PR #1, opened 2016-03-08
Modifications to Support Globus CLI

sirosen:
To briefly enumerate the changes:

  - Add `requests` to `install_requires` and switch to setuptools.
    Trying to be permissive here and allowing versions as old as 2.0.0. We'll see if we need to narrow this when we add tests.
  - Add default `globus.cfg` to package_data (it was missing)
  - Allow arbitrary kwargs to pass through BaseClient requests wrappers.
    This was a convenience for some of the other work, but I also think it's a good behavior to have as we add clients.
  - Instead of `get_goauth_token` being a weird one-off method on TransferClients, moved that to a new NexusClient which only has that method.

@bd4 Could you look this over, make sure there aren't any glaring mistakes? I was able to use the new NexusClient with a version of globuscli that I'm working on, so I'm pretty confident.

### DIFF
--- a/globus_sdk/nexus.py
+++ b/globus_sdk/nexus.py
@@ -1,0 +1,26 @@
+from __future__ import print_function
+
+from globus_sdk.base import BaseClient
+
+
+class NexusClient(BaseClient):
+    def __init__(self, environment="default"):
+        BaseClient.__init__(self, "nexus", environment)
+
+    def get_goauth_token(self, username, password):
+        """
+        Legacy method for getting a GOAuth token from nexus using
+        globusid username and password. Clients using this should
+        be prepared to migrate to Globus Auth.
+
+        Note that these tokens have a long lifetime and should be saved
+        and re-used.
+        """
+        r = self.get('/goauth/token?grant_type=client_credentials',
+                     auth=(username, password))
+        try:
+            return r.json['access_token']
+        except KeyError:
+            raise GlobusError(r)
+        else:
+            raise
--- a/globus_sdk/__init__.py
+++ b/globus_sdk/__init__.py
@@ -1,3 +1,4 @@
 
 from globus_sdk.base import GlobusError, GlobusResponse
 from globus_sdk.transfer import TransferClient
+from globus_sdk.nexus import NexusClient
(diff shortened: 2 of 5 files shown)


globus_sdk/nexus.py
```diff
@@ -0,0 +1,26 @@
+from __future__ import print_function
+
+from globus_sdk.base import BaseClient
+
+
+class NexusClient(BaseClient):
+    def __init__(self, environment="default"):
+        BaseClient.__init__(self, "nexus", environment)
+
+    def get_goauth_token(self, username, password):
```
bjmc: This should raise a `DeprecationWarning`

globus_sdk/nexus.py
```diff
@@ -0,0 +1,26 @@
...
+        """
+        Legacy method for getting a GOAuth token from nexus using
+        globusid username and password. Clients using this should
+        be prepared to migrate to Globus Auth.
+
+        Note that these tokens have a long lifetime and should be saved
+        and re-used.
+        """
+        r = self.get('/goauth/token?grant_type=client_credentials',
```
bjmc: :-1: single-letter variable names.

bd4: I disagree, I think for short bits of code, single var names are fine and in some ways easier to read.

globus_sdk/nexus.py
bd4: Agreed but we should wait until 2-legged or possibly API keys/app passwords are available as a replacement. Otherwise we'll get a lot of questions about it, and have not alternative to provide them.

sirosen: I was actually thinking of having the `NexusClient` initializer fire a `PendingDeprecationWarning` warning.
I think that's a lot cleaner and more clearly expresses the true state of the entire Nexus API to any would-be developers.

Especially since these tokens are needed for pretty much any scripting against our APIs at present.

I'll open an issue to discuss.

globus_sdk/nexus.py
bd4: That being said, one char instance variables and param names can be confusing, I'm going to fix those in the base class.

Issue #4, opened 2016-03-09
Support Loading Token when Instantiating a Client


sirosen:
After some thought about this, I think this should be an SDK responsibility.
I think that we should have an option somewhere in `~/.globus.cfg` for this, but also maybe an option in the client initializers.

Here's the issue: if I write code that talks to multiple APIs, I'll end up having a bunch of cases in which I instantiate the clients myself.
There are a couple of tempting options for how to pass an auth token to all of them uniformly, but they all seem worse than having this in the SDK itself.

People will write (with varying degrees of correctness) something like

```python
def make_clients(token):
    ac, nc, tc = AuthClient(), NexusClient(), TransferClient()
    for c in (ac, nc, tc):
        c.set_auth_token(token)
    return ac, nc, tc
```

because so many calls must be authenticated.

I'd rather see `set_auth_token` be invoked if you pass `auth_token=...` to the constructor.
However, better yet, I'd like for the SDK itself to support loading the auth token from a file, like `~/.globus.cfg`.

We could tell people to define something like
```ini
[default]
auth_token = ABC123

[environment beta]
auth_token = DEF456
```

That would make the simplest possible config file
```ini
auth_token = GHI789
```

I think that's appealingly simple.

I'm fighting hard to make sure that the CLI isn't just a re-implementation of curl, and I think handling the token outside of the SDK pushes me to make some compromises, like adding `--headers` to handle this.

@bd4 What do you think of this?
I think the ideal is that you can put it in a config file or an environment variable and things just work.
It only gets tricky when you want to support multiple tokens for multiple environments (which, of course, is something of an internal must).

bjmc:
Access tokens are going to be short-lived, so putting them in config files and updated them sounds like a sisyphean endeavor to me.

Maybe it would make sense for the SDK to take a refresh token in a config file and automatically handle periodically exchanging it for fresh access tokens for you.

sirosen:
Today, GOAuth tokens are not short lived, and don't have refresh tokens.
I'm interested in building something that works today, but will serve as a basis for loading whatever the *real* credentials will be.
This is something of a stopgap measure, but we can convert it to load a refresh token or maybe even 2-legged oauth credentials.
I was going to build a wrapper that does this a bit for the CLI because the SDK doesn't have a good credential loading mechanism, but then I thought it would be better to implement here.

I understand the desire to not build throwaway features for interacting with Nexus, but I also want to be pragmatic about this.
[...]

PR #5, opened 2016-03-10
auth_token Config Value and Config Extensions

sirosen:
Adds a config value of
```ini
[general]
auth_token = xyz987
```
which gets loaded on `BaseClient` initialization. That value can, at present, be a GOAuth token, and everything works there.

I also made some more extensive changes to the config module to support two use cases which we discussed or mentioned in #4.
The first is an override to the ordinary `ConfigParser` behavior when it sees a file with no section heading before its first value, as in
```ini
auth_token = abc123
```
or
```ini
auth_token = def456
[general]
...
```
This makes `ConfigParser` fail hard with an error because it's insisting on every value being in a section, but I think this is unnecessarily harsh. More notes in the commit message, but I added a class that overrides this behavior to just insert `[general]\n` at the start of any files that trigger that exception.
`ConfigParser` already handles multiple instances of the same section (and loads them in order, overriding values), so there's no real danger here, and it makes smaller and simpler configs possible.
I especially like the possibility of
```shell
echo "auth_token = $(globus nexus get-goauth-token -F text)" > ~/.globus.cfg
```
or similar as a future recommendation, and this only works so cleanly if we remove the section header constraint.

The other change is less controversial -- the addition of an optional load-from-environment behavior.
It's controlled by a kwarg to `config.py`'s `_get()` with a default value of `False`.
Environment values are always named as `GLOBUS_{uppercase_option_name}`, with their config section omitted.
I think that leads to the easiest to understand behavior, without having to stick `GENERAL_` in every single name.
Environment variables have (as you would expect) higher precedence than any config files.

@bd4, I want to make sure I'm not going overboard with the modification of `ConfigParser`. I like it (a lot, actually), but I recognize the flaws and arguments against it. I would very much appreciate your input on it.

### DIFF
--- a/globus_sdk/base.py
+++ b/globus_sdk/base.py
@@ -1,5 +1,6 @@
 import urllib
 import json
+import warnings
 
 import requests
 
@@ -21,6 +22,18 @@ class BaseClient(object):
             self.base_url = slash_join(self.base_url, base_path)
         self._session = requests.Session()
         self._headers = dict(Accepts="application/json")
+
+        # potentially add an Authorization header, if a token is specified
+        auth_token = config.get_auth_token(environment)
+        if auth_token:
+            warnings.warn(
+                ('Providing raw Auth Tokens is not recommended, and is slated '
+                 'for deprecation. If you use this feature, be ready to '
+                 'transition to using a new authentication mechanism after we '
+                 'announce its availability.'),
+                PendingDeprecationWarning)
+            self.set_auth_token(auth_token)
+
         # TODO: get this from config file, default True
         self._verify = True
         if "ec2" in self.base_url:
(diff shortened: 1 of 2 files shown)


sirosen:
After an inspection of the GT env vars, and the sprawling landscape that presents, I agree that `GLOBUS_SDK_` is acceptable.
I wish we could have the `GLOBUS_` namespace to ourselves, but, alas, it was not meant to be.

sirosen:
@bd4 Per our discussion, I've rebuilt the `GlobusConfigParser` to wrap `SafeConfigParser`, rather than being part of the `ConfigParser` class hierarchy.

I'm pretty happy with what I have so far -- I don't think it's too complicated, but I think it's sufficiently much better than the base `ConfigParser` behaviors to serve our needs.

For now I don't think that we need a top-level `get()` which wraps up
```python
p = _get_parser()
p.get(*args, **kwargs)
```
but that would be an easy way to extend this in the future.

PR #6, opened 2016-03-11
improve error handling, make extensible

bd4:


globus_sdk/base.py
```diff
@@ -7,6 +7,46 @@
 from globus_sdk import config
 
 
+class GlobusError(Exception):
+    def __init__(self, r, *args, **kw):
```
sirosen: `r` :cry:

globus_sdk/base.py
```diff
@@ -7,6 +7,46 @@
...
+        self._underlying_response = r
+        self.http_status = r.status_code
+        if "application/json" in r.headers["Content-Type"]:
```
sirosen: Question: should we even be checking `Content-Type` here?
A JSON response could be returned without this header, or with a wrong header (apparently).

Should we just *always* try to load from JSON, catch failed parsing, and fallback to text?

globus_sdk/transfer.py
```diff
@@ -1,9 +1,25 @@
 from __future__ import print_function
 
-from globus_sdk.base import BaseClient
+from globus_sdk.base import BaseClient, GlobusError
+
+
+class TransferError(GlobusError):
```
sirosen: Add some docstring or comments that explain the desired `request_id` behavior.

PR #7, opened 2016-03-12
Refactor Endpoint Search, build generic PaginatedResource Transfer tooling

sirosen:
Adds a few things.

- The `merge_params` helper

I put a lot of doc into this, even though it's a really small function because I think it's important to understand that it's necessary to do things this way.
Not only is it easier to write, but it's also safer than doing it manually each time.
I thought a bit about adding a `@with_merged_params` decorator which would magically stuff all parameters into a single dict to pass to an underlying function, but this was hairy, complicated, and less clear.
I also think a decorator poses some correctness problems because it's less clear how to use in unusual cases -- and it's never very clear how much a decorator is modifying its underlying function.

I also considered a decorator we could slap on like
```python
@requires_params('filter_fulltext')
def endpoint_search(...):
   ...
```
but then someone needs to read and understand that decorator, as opposed to just seeing a positional argument.

- Update `TransferClient.endpoint_search` to have `filter_fulltext` and `filter_scope` as explicit params

Rather than having these buried inside of `**params`, endpoint search now has a method signature that looks like this:
```python
endpoint_search(self, filter_fulltext, filter_scope=None, **params)
```

That makes it very clear that we expect anyone invoking search to pass a fulltext query for us to use. I don't think we need to support, via the SDK, search calls without queries -- they can always pass `None`, if they want to get cute, and the above `merge_params` will filter it out.

- A pagination iterator for `endpoint_search`

I don't know that this is necessarily the final form that we want for this to take, but pagination is hard to do right.
I think that the correct and natural way to handle paginated calls in the SDK is to build iterators.
Clients using those iterators can pass any parameters that they want to, but we will overwrite them if they pertain to pagination (`limit`, `offset`).

This works really nicely for both simple and sophisticated clients.

I was on the fence about setting limit inside of the iterator, but if a client wants < 100 search results, they can just use the non-iterating version of that call.
Additionally, in the case of python at least, an iterator actually gives its caller a chance to control the behavior as results come in -- if you want to stop iteration as the caller, it's very simple and immediately stops the API calls.

One interesting possibility for this and future iterators it to accept a `limit` parameter which does not get passed to the underlying API call, but instead counts the number of results returned and stops after that many.
I don't know if that's needful though, so I'm leaving it out.

Another option worth mentioning is to have all or some of our helpers all accept `paginate=True` and set `paginate=False` by default. 
If `paginate=True`, the function acts as an iterator, and otherwise it directly returns the result.
In a dream world, that could even be a simple decorator, but our APIs aren't uniform enough to do that.

For now, this is a simple version of pagination for endpoint search that definitely works and works well -- I was able to adapt the CLI to use it in two or three lines of change.

### DIFF
--- a/globus_sdk/transfer.py
+++ b/globus_sdk/transfer.py
@@ -24,21 +24,21 @@ class PaginatedResource(object):
       the returned JSON document
     """
 
-    def __init__(self, page_size, max_total_results):
+    def __init__(self, max_results_per_call, max_total_results):
         """
         PaginatedResource takes two arguments
-        page_size is the (max) size of a page to fetch from the API.
+        max_results_per_call is the max size of a page to fetch from the API.
         max_total_results is a pagination limit for total results to return
         from the API.
 
         When using this decorator, it is a good idea to name the arguments,
         even though they are positionals, for greater clarity inline, as in
 
-        >>> @PaginatedResource(page_size=25, max_total_results=250)
+        >>> @PaginatedResource(max_results_per_call=25, max_total_results=250)
         >>> def call_that_supports_paging(...):
         >>>     ...
         """
-        self.page_size = page_size
+        self.max_results_per_call = max_results_per_call
         self.max_total_results = max_total_results
 
     def _get_paging_kwargs(self, **kwargs):
@@ -50,18 +50,16 @@ class PaginatedResource(object):
         disrupt the flow of original positional arguments to the wrapped
         function, `func`
         """
-        # limit for num results, with a default given by the decorator
-        # definition
-        # paginated calls should always define a `limit` kwarg, but just in
-        # case someone forgets to do so, we can fall back on the page size
-        limit = kwargs.get('limit', self.page_size)
+        # paginated calls should always define a `num_results` kwarg, but just
+        # in case someone forgets to do so, we can fall back on the max allowed
+        num_results = kwargs.get('num_results', self.max_results_per_call)
 
         # offset should rarely be set, but we need to handle the case in
         # which a caller wants to start their results at a given offset
         # manually
         offset = kwargs.get('offset', 0)
 
-        return limit, offset
+        return num_results, offset
 
     def __call__(self, func):
         """
@@ -80,28 +78,28 @@ class PaginatedResource(object):
             interface.
             """
             # keyword args used by the paging itself
-            limit, offset = self._get_paging_kwargs(**kwargs)
+            num_results, offset = self._get_paging_kwargs(**kwargs)
 
-            # now, cap the limit per request to the page size
-            per_call_limit = min(limit, self.page_size)
+            # now, cap the limit per request to the max per request size
+            limit = min(num_results, self.max_results_per_call)
 
-            # check the limit to see if it exceeds the maximum total
-            # number of results allowed by the API
-            if limit > self.max_total_results:
+            # check the requested num results to see if it exceeds the maximum
+            # total number of results allowed by the API
+            if num_results > self.max_total_results:
                 raise exc.PaginationOverrunError((
                     'Paginated call would exceed API limit. Pass a smaller '
-                    'limit parameter -- the maximum for this call is {}')
+                    'num_results parameter -- the maximum for this call is {}')
                     .format(self.max_total_results))
 
             has_next_page = True
             while has_next_page:
                 # if we're about to request more results than the user asked
                 # for, limit ourselves on the last paginated call to the API
-                if offset + per_call_limit > limit:
-                    per_call_limit = limit - offset
+                if offset + limit > num_results:
+                    limit = num_results - offset
 
                 kwargs['offset'] = offset
-                kwargs['limit'] = per_call_limit
+                kwargs['limit'] = limit
 
                 res = func(*args, **kwargs).json_body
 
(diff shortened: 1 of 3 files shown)


bd4:
I really don't like the idea of using iterators for pagination. Paging through everything in out db is not something I want to encourage, especially for endpoint search. That's why there is a hard limit at 999.

bd4:
I guess the bottom line is that I want clients to think carefully about how they are paginating, and if we make it too easy/hidden, they won't. Most use cases should be handled by getting a single reasonably sized page, and for the 1% where you really want to go through as much as you are allowed to, they should be thinking carefully about what the best page size is for the application.

sirosen:
@bd4 There's a significant distinction I want to draw here -- do you object to iterators for paging as a design choice in a python application, or as a component of the SDK?
I assume you're fine with iterators as a way of implementing a paging API call -- I took the idea from boto3, and in practice I've found that it works great. Their implementation, of course, is a bit more sophisticated: http://boto3.readthedocs.org/en/latest/_modules/boto3/resources/collection.html

So, if every reasonable python client will implement pagination the same way, using iterators, why shouldn't it just be part of the SDK itself?

In fact, your argument that making it too easy for people will encourage them to abuse the API is the _opposite_ of what I believe to be the real concern here.
I would be much more worried that someone implementing pagination without a clear understanding of our API will do something like this:

```python
def bad_paginating_search(client, query):
    has_next_page = True
    offset = 0
    limit = 999
    while has_next_page:
        res = client.endpoint_search(query, offset=offset)
        # uh-oh! this looks very wrong!
        yield res.json_body['DATA'][0]
        offset += 1
        has_next_page = res['has_next_page'] or offset > limit
```

That's super-abusive of the API, obviously wrong and slow, and, I would say, *very likely* to be written by someone careless. After all, it has the same effect, logic-wise, as the correct version I included in this PR -- it just does it in a definitively worse way.

If your real worry is that making pagination easy will encourage everyone to pull down 1.5 MB of search results (the approximate size of 999 entries, when I did it) for every search, I would say that means we need a pagination limit coming out of the gates.
So, every paginator has a `limit` kwarg, distinct from the page size limit, which restricts the total number of entries it retrieves by default.

That way, you don't need to worry about people writing

```python
def all_eps(client):
    return list(client.paginated_endpoint_search(''))

for ep in all_eps:
    ...
```

because a naive user of the paginated search feature (or paginated task listing) will only see 25 or 100 results; whatever we set as the default.

An alternate form of that is the `paginate=True` / `paginate=False` setting on the `TransferClient.endpoint_search` method itself.
If we do that, then no one will turn pagination on until they actually need it, and when they do, they just update their calls to an existing function whose name and signature they already know.
If one or two clients turns on pagination when they don't actually need it, I think it puts us in a much preferable position than one or two clients implementing it incorrectly, and calling `/endpoint_search` 999 times in a row.

I actually feel very strongly that the SDK should handle this for you, so this might be something we discuss more.

bjmc:
For my two cents, I agree with Stephen that an iterator seems like a nice Pythonic way to return paginated results, but I agree with Bryce that it should be something users need to explicitly ask for (a keyword argument defaulting `multiple_pages=False`, as suggested above, sounds good to me)

bd4:
Ok, I think part of what I'm reacting to is how we seem to be making the SDK more and more full featured, which makes me worry about whether we can deliver on the docs. The nice thing about the bare bones client is that the API docs are basically all you need. For the iterator, we need a description and a code example.

Writing good documentation actually takes way more time than writing code. So I guess I'm open to going down this path, as long as we do it with our eyes open to what we are signing up for.

For this specific feature, non-paginated by default, with an explicit limit param to emphasize that you can't always page through everything, seems reasonable. I'd like to see well written end-developer docs for this before I merge though.

sirosen:
I'll make an updated version of this which has `limit` param, and raises an exception if the full result set exceeds the API's pagination limit.
I think that's the most sensible behavior.

I'm very much onboard with the idea that this requires good documentation, and that documentation takes more time than writing. However, I don't trust arbitrary developers to do this correctly at all.

So far, pagination is the only case that's come up where I think the SDK *must* offer a higher-level helper. We can see if there are more such cases as well spec things out and evolve the body of SDK-dependent code.

sirosen:
I have a new version with lots of doc.

The decorator business is full of comments because I think it's potentially confusing, but they right way to do it.
Really, it's a matter of the decorator syntax requiring a lot of levels of nesting even for a relatively simple closure with some added arguments.
Plus, the argument parsing in python kind of explodes on this use case.

@bd4 What do you think of this, and its accompanying documentation? It will have to be sphinx-ified at some point, of course.

globus_sdk/transfer.py
```diff
@@ -1,9 +1,159 @@
...
+            limit = kwargs.get('limit', default_limit)
+
+            # offset should rarely be set, but we need to handle the case in
+            # which a caller wants to start their results at a given offset
+            # manually
+            offset = kwargs.get('offset', 0)
+
+            return paginated, limit, offset
+
+        def wrapped_func(*args, **kwargs):
```
bjmc: Consider using [`functools.wraps`](https://docs.python.org/2/library/functools.html#functools.wraps)

sirosen: You mean for function metadata preservation, right?
I don't think `functools.wraps` handles any of the messy problems in the decorator pattern argument handling.

Definitely a good best-practices thing to do. I'm working on a paginated task listing method as well, to make sure that all of this works well. I'll fix it there.

bjmc: > You mean for function metadata preservation, right?

Yeah. It makes it less confusing if somebody is inspecting that method in a debugger or an interactive shell.

sirosen:
I think there's been enough discussion of this.
I'm going to go ahead and merge -- we talked about the ugliness of the decorator in #8.

PR #8, opened 2016-03-14
Add paginated task_list Transfer helper

sirosen:
Adds a notion of "paging styles" to PaginatedResource, because Transfer is actually not uniform about paging.
Also adds handling for `max_total_results=None`, for the case in which there is no limit on the API.

### DIFF
--- a/globus_sdk/transfer.py
+++ b/globus_sdk/transfer.py
@@ -1,6 +1,7 @@
 from __future__ import print_function
 
 import functools
+import inspect
 
 from globus_sdk.base import BaseClient, merge_params
 from globus_sdk import exc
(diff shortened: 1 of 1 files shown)


sirosen:
@bd4 Could you look at this again?

I found a weird bug regarding correct handling of keyword argument default values which was previously being masked by having a default for the requested `num_results`.
The only solution I could find was to use `inspect` to piece together the keyword argument default values, if present.

I'm satisfied with this (although it's a little ugly), but I'm concerned that others might find it objectionable.

bd4:
The inspect thing is ugly, but I do like hiding all the messy details of three different styles of paging. Go ahead and merge.

PR #10, opened 2016-03-18
Python3 compatibility

jaswilli:
- Use "six" to import version agnostic packages

### DIFF
--- a/globus_sdk/auth.py
+++ b/globus_sdk/auth.py
@@ -1,6 +1,6 @@
 from __future__ import print_function
 
-import urllib
+from six.moves.urllib.parse import urlencode
 
 from globus_sdk import config
 from globus_sdk.base import BaseClient, merge_params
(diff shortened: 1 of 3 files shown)


bjmc:
:+1: :100:

bjmc:
Also consider https://tox.readthedocs.org/en/latest/ to test this with different Python versions automatically.

bd4:
I'd prefer using conditional imports in the few cases where we run into this, vs introducing a new dependency. Maybe I'm being naive about how often it will come up, could be persuaded otherwise, but that's my 2 cents.

sirosen:
I'm torn on this. Conditional imports are easier for most cases.
However, `six` is required if you want to use any language features that have undergone significant changes. My favorite example is metaclasses, but I there are a few others as well.

It's a staple for any complex codebase that supports 2 and 3 to use six, but I don't know that we're complicated enough to actually demand it.

That said, it's a pretty stable library, heavily used, and well supported. I have no real qualms about merging this, and wouldn't consider it dependency bloat -- it's not just an import renaming bundle, but a real library with real functionality that we might want to leverage.

jaswilli:
I don't have strong feelings about any particular solution. I just ran into trouble today working with python3 so here we are...

bd4:
The fact that 3 raises a DuplicateSectionError and 2 doesn't is one example of where a simple conditional import isn't enough. I guess I'm fine with using `six` if it's well supported. I'll merge it.

PR #12, opened 2016-03-20
Auth tokens

jaswilli:
* I missed a compatibility import on the previous PR. I also had modified `globus.cfg` to get it working while I was playing around so missed that the file handle needs to be reset. :blush: 
* Accept Globus Auth bearer tokens in the client constructors.

#### Context
Using a config file and/or environment variables for tokens is not practical for a globus-sdk consumer that may be serving multiple clients, or cycling through tokens, etc.

This changes the client constructors to taken an optional `auth_token` argument that will then be used in the `Authorization` header.

```python
from globus_sdk import TransferClient

TOKEN = 'AQBW8J40AAAAAAACzrTL00Q4hThVgOXDo-Kz1xoErMDvtfV-2sOw8P-VNMSwxO6A4DGjx5nyBs9ohLrta_dZ'

#go ep1
EP = 'ddb59aef-6d04-11e5-ba46-22000b92c6ec'

t = TransferClient(auth_token=TOKEN)

r = t.operation_ls(EP)
```

### DIFF
--- a/globus_sdk/auth.py
+++ b/globus_sdk/auth.py
@@ -7,8 +7,9 @@ from globus_sdk.base import BaseClient, merge_params
 
 
 class AuthClient(BaseClient):
-    def __init__(self, environment=config.get_default_environ()):
-        BaseClient.__init__(self, "auth", environment)
+    def __init__(self, environment=config.get_default_environ(),
+                 auth_token=None):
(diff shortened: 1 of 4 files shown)


sirosen:
:cry: `DuplicateSectionError`.
Looks like a good fix.

We discussed whether or not to have the client accept auth tokens in the constructor, and I think we agreed that it would be useful, but didn't act on it.

PR #13, opened 2016-04-01
Add transfer submission helpers

sirosen:
Uses the pattern of make_<methodname>_* to clarify that these are
helpers associated strongly with the submit_transfer method. It's a bit
verbose, but `make_submit_transfer_data` is unlikely to collide with
other functionality in the future and cause confusion.

Includes handling of the submission ID, since this is a logical place for it.

A decision that I will grant is questionable: I'm requiring the label here.
I thought that would be a good way of strongly encouraging the use of the label field to disambiguate transfers started by a program (vs. a human user in the web UI).

I'd like this to serve as a model for some similar things (delete tasks, and maybe some of the ACL management), so we should take our time and make sure we're happy with the result here.

### DIFF
--- a/globus_sdk/transfer/client.py
+++ b/globus_sdk/transfer/client.py
@@ -345,9 +345,15 @@ class TransferClient(BaseClient):
         sync_level can be 1, 2, or 3, but it can also be
         "exists", "mtime", or "checksum" if you want greater clarity in
         client code.
+
+        Includes fetching the submission ID as part of document generation. The
+        submission ID can be pulled out of here to inspect, but the document can
+        be used as-is multiple times over to retry a potential submission
+        failure (so there shouldn't be any need to inspect it).
         """
         datadoc = {
             'DATA_TYPE': 'transfer',
+            'submission_id': self.get_submission_id().data['value'],
             'label': label,
             'source_endpoint': source_endpoint,
             'destination_endpoint': dest_endpoint,
--- a/globus_sdk/transfer/client.py
+++ b/globus_sdk/transfer/client.py
@@ -335,7 +335,7 @@ class TransferClient(BaseClient):
         return self.post('/delete', data)
 
     def make_submit_transfer_data(self, source_endpoint, dest_endpoint,
-                                  transfer_items, label, sync_level=None,
+                                  transfer_items, label=None, sync_level=None,
                                   **kwargs):
         """
         Build a full document for submitting a Transfer.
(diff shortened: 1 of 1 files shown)


globus_sdk/transfer/client.py
```diff
@@ -334,6 +334,67 @@ def submit_delete(self, data):
         """
         return self.post('/delete', data)
 
+    def make_submit_transfer_data(self, source_endpoint, dest_endpoint,
+                                  transfer_items, label, sync_level=None,
+                                  **kwargs):
```
jaswilli: Perhaps this should be `label=None`? If `label` is required without a default there might be a tendency to pass in an empty string here, and unless it's changed recently Transfer will reject an empty string `label`.

sirosen: My feeling was that "Label is optional for the API, but let's make it mandatory for the SDK".

@jaswilli Do you really think people will pass the empty string? Or are you more of the camp that the label should be optional, as it is for the API?
I don't feel terribly strongly, but I think any SDK user should be setting the label to identify the program that's starting transfers (because the tasks are still owned by the user, and there's no other way to track that).

bd4: I think label should be optional in the sdk. If label is None, the webapp displays the source/dest endpoint names, which is probably good enough for a lot of common use cases.

jaswilli: I don't know. If it's required in the SDK I can see people saying, "I don't want a label, `label` is a string, therefore I'm going to pass in `''`."

In our web app we actually check for an empty string and convert it to `null` before sending the request off to the API (because if you're taking input from a user it's really easy to end up with a "no input" -> empty string -> transfer api error situation).

So unless there's a compelling reason to make `label` required it seems like it might cause hassle for no reason.

sirosen:
Given that I'm the only one who saw "require label" as a good thing, and I'm not really committed to that idea, I am convinced.
I've fixed it, should be good for any further review.

If you want to see what this looks like in action, here's a sample from the current CLI version:

```python
    if args.batch is not None:
        transfer_items = [
            client.make_submit_transfer_item(item['source'], item['dest'])
            for item in args.batch['DATA']
            ]
    else:
        transfer_items = [client.make_submit_transfer_item(args.source_path,
                                                           args.dest_path)]

    datadoc = client.make_submit_transfer_data(
        args.source_endpoint, args.dest_endpoint, transfer_items,
        label='globus-cli transfer', sync_level="mtime")

    res = client.submit_transfer(datadoc)
```

globus_sdk/transfer/client.py
jimPruyne: Regardless of whether we allow label to be None, if as @jaswilli says that the API will complain on the empty string, we should catch that in the SDK and report that as an error/exception rather than waiting for the API to return with an exception.

bd4: I disagree very strongly with trying to do an extra level of verification in the sdk. The REST APIs already should be doing verification and returning useful errors. Maintaining yet another level of verification is a waste of resources. In a user interface, it may be necessary, but I don't see how it is in an sdk.

PR #16, opened 2016-04-05
Handle per-service tokens, kill NexusClient

sirosen:
Now that tokens.globus.org is a thing (for now, as a stopgap measure),
we can kill all mention of the NexusClient. Because of a mismatch in our
understanding of how this was going to work, some of our prior naming
was not so great. A little bit of name shuffling was necessary to
accomodate.
Makes BaseClient more abstract by putting a `NotImplemented` method into
its initializer -- so it won't be possible to instantiate a straight
BaseClient anymore (not that you'd really want to). If you want a client
to an API that we don't have baked into the SDK, you have to subclass
BaseClient and define a method for loading its token as a string. Still
extensible, but less monkey-patch-able.

Open question: should BaseClient be a true ABC, with the token loading as an abstract method?


bjmc:
Did I miss something? What is `tokens.globus.org` (other than a thing)?

mattias-lidman:
Same here. I can kinda deduce from going to tokens.globus.org what the idea is, and at first blush the idea seems not insane, but this strikes me as the sort of thing that should've been brought up for discussion.

globus_sdk/config.py
```diff
@@ -154,16 +154,27 @@ def _bool_cast(value):
     raise ValueError("Invalid config bool")
 
 
+def _get_token_type(tok_type, environment):
```
bd4: Nitpick: I think this should be named _get_token(token_type, environment). I was expecting it to return a token type, not a token of the specified type.

sirosen:
Bryce signed off on this in its current state over Slack. Merging.

PR #20, opened 2016-04-12
Transfer Client Paging triggers errors "eagerly" at iterator instantiation time

sirosen:
The idea here is to force any paged calls to fetch the first page of results immediately, when the iterable object is instantiated.
That way, when you make a paged call, you get an error at the invocation site, not at the iteration site.
Protects us partially against people doing something like

```
tc = TransferClient()
try:
    eps = tc.endpoint_search(...)
except GlobusAPIError as e:
    ...

for ep in eps:
    ...
```

Importantly, we behave *much* better with structures like this:

```
try:
    eps = tc.endpoint_search(...)
except GlobusAPIError as e:
    ...
else:
    print('Successfully got endpoint search results!')
    ... # some code that assumes we'll be able to iterate
    for ep in eps:
        ...
```

During iteration, a call to the API could still fail, but a variety of cases, like a malformed call to endpoint_search, will be caught immediately.
Excepting credential expiration during iteration on a paged call, or a network failure, it's unlikely that the first call will succeed, but later calls will fail.
To do this, the `@PaginatedResource` decorator had to go -- binding of arguments at call time vs. definition time and other issues abounded.
Instead, we keep that name, but `PaginatedResource` is now a class, instantiated with a function or instance method to wrap, which implements a custom `__iter__` to have an iterable interface.
This has the added benefit of enhancing code clarity, since the return type of paged functions when decorated no longer mismatches with the return type of the undecorated functions.

This PR also enforces the "eager" error semantics on non-paged calls which were doing deferred API calls because they used the `yield` iterator function pattern.
To resolve, just wraps their results in list comprehensions -- still quite tidy, and less likely to confuse people down the line.

### DIFF
--- a/globus_sdk/transfer/client.py
+++ b/globus_sdk/transfer/client.py
@@ -147,7 +147,7 @@ class TransferClient(BaseClient):
         merge_params(params, filter_scope=filter_scope,
                      filter_fulltext=filter_fulltext)
         return PaginatedResource(
-            self.get, ["endpoint_search"], {'params': params},
+            self.get, "endpoint_search", {'params': params},
             num_results=num_results, max_results_per_call=100,
             max_total_results=1000)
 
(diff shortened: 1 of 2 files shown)


sirosen:
Approved by Bryce, merging

Issue #23, opened 2016-04-12
Improving response classes


bd4:
I find myself constantly forgetting to add the `.data`, and it seems superfluous. I think it's especially confusing for the iter responses. I propose that we add a `__getitem__` to `GlobusResponse` that operates on the `.data`, so you can do `response['path']` instead of `response.data['path']`, and make it have a default `__next__` that directly exposes `response['DATA']`. That makes it more convenient, while still giving us a central response class that we can tweak in the future if needed.

With the current response formats, it feels like Java, not like Python.

bd4:
I'd also like to kill `.json_body`, it's confusing having two ways to access the same thing. Maybe we should rename `.text_body` to just text, it was designed to be symmetric with the params to `.get(...)` and friends, but I don't think that actually buys anything.

bd4:
The `DATA` iter thing is very specific to the Transfer client, so that should be in a subclass, but `__getitem__` may be general enough to live in the base. If the response is not JSON, it could raise `TypeError` or similar.

sirosen:
I approve of this message in principle, but I think we need to be careful to not tread on the toes of a non-Transfer API that gives non-JSON responses.
Transfer is very good about JSON everywhere, all the time, but it's not just a matter of being well behaved: other APIs may have use cases that warrant a different format.

So, my thoughts on how to do this:

- Keep the `data` property -- I like having something I can turn to and say "If a response type wants to define what all of its contents look like as a python object, do so here and only here."
- `GlobusResponse` gets a `__getitem__` method which tries to do a `__getitem__` on whatever its `data` property is. That way, non-`TransferResponse` and non-`GlobusHTTPResponse` classes can make themselves a `__getitem__` by simply ensuring that it works on `.data`. Likely that will typically a dict. Otherwise, we get a lot of same-y `__getitem__` definitions in a bunch of classes, and it gets ugly.
- If `TransferResponse` has iterable responses, that's fine, but I think it's very ugly to throw a `KeyError` if something doesn't have the `DATA` field. Either use a specialized subclass, or make iterability an option to the `TransferResponse` constructor. Attempting to iterate non-iterable responses should be a `TypeError` (just put that in `__next__`, I think?)
- Paging calls will not suddenly start returning a `TransferResponse` object -- I don't see a way of having a generic `TransferResponse` be responsible for paging without contorting things significantly. We might make `PaginatedResource` a `TransferResponse` subclass and maybe rename it.

More questions:
[...]

PR #24, opened 2016-04-13
Improve GlobusResponses to have native __getitem__ and optional __iter__

sirosen:
This handles #23 

Short summary:

- HTTP methods on the `BaseClient` now accept `response_class` as a keyword arg
- `GlobusResponse` has `__getitem__`, which does a `__getitem__` on the `data` property
- Drop `GlobusHTTPResponse.json_body`
- Rename `GlobusHTTPResponse.text_body` to `GlobusHTTPResponse.text`
- Iterable responses from `TransferClient` calls have `response_class=IterableTransferResponse`, which defines `__iter__` on the `"DATA"` array
- `PaginatedResource` for the `TransferClient` now inherits directly from `GlobusResponse`, so that we're always returning a `GlobusResponse`, no matter what

### DIFF
--- a/globus_sdk/base.py
+++ b/globus_sdk/base.py
@@ -35,7 +35,7 @@ class BaseClient(object):
 
     # Can be overridden by subclasses, but must be a subclass of GlobusError
     error_class = exc.GlobusAPIError
-    response_class = GlobusHTTPResponse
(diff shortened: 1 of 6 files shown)


bd4:
Looks great, merging.

PR #27, opened 2016-04-15
add OO transfer data helper

bd4:
Not ready to merge, just putting this up for comments. I think the OO version, where add_item is a method on the object itself, is much cleaner.

I'm not sure about having the public api as a separate class that takes a `TransferClient` in it's constructor, vs modifying `make_submit_transfer_data` so it returns the `TransferData` instance and killing the item helper.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -35,6 +35,10 @@ Transfer Client
    :show-inheritance:
    :exclude-members: error_class, response_class
 
+.. autoclass:: globus_sdk.TransferData
+   :members:
+   :show-inheritance:
+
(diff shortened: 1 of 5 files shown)


sirosen:
I like `TransferData` and `DeleteData` as separate classes model-wise, even if you have to pass them a client object to use.
Ultimately, this will likely be a little bit clumsy no matter what, so, accepting that, this seems like the least-clumsy approach.

We should remove the old helpers before GlobusWorld -- should that be part of this PR?

bd4:
I'm going to leave the old helpers there for now, give everyone a day to migrate.

PR #28, opened 2016-04-20
Documentation Cleanup Pre-GlobusWorld 2016

sirosen:
@bd4 If you don't get to this, I'll merge it early tomorrow morning, before the conference starts.

I want us to have cleaned up docs before people start seeing the site.
Things in here that I think are important:
- `AuthClient` method docs link to locations on `docs.globus.org`
- `TransferClient` has `:rtype:` set on every method -- you'll know for sure if its iterable or not from the docs
- Old Transfer helpers are clearly deprecated and refer to new OOP helpers. They also raise a deprecation warning
- Fix to Basic Usage example
- Get Deprecation info off of index page (it's not that level of importance, distracts from real content)
- Add an example for using `TransferData`

There are also a few little things here and there.

Can we call this our "final" review of docs, and we'll leave it frozen after this?
We should also push to pypi when we're done, so that we're sure that everything is synced up.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -33,11 +33,21 @@ Transfer Client
    :member-order: bysource
    :inherited-members:
    :show-inheritance:
-   :exclude-members: error_class, response_class
+   :exclude-members: error_class, default_response_class
+
+Helper Objects
+~~~~~~~~~~~~~~
 
 .. autoclass:: globus_sdk.TransferData
    :members:
    :show-inheritance:
+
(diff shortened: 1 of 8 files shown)


bd4:
I want to avoid having doc for task submission in two places, so I added few things from the new doc to the `submit_transfer` example, and removed the example in `TransferData`. I didn't copy over the once-and-only-once submit example; we can resurrect it later, I'd like to think about it more first and I don't think it's necessary at this stage.

PR #35, opened 2016-06-18
Added mgmt_task_list function to query all tasks for a given endpoint

fred-lefebvre:
The mgmt_task_list queries the Globus REST API for 'task_list' under 'endpoint_manager'

### DIFF
--- a/globus_sdk/transfer/client.py
+++ b/globus_sdk/transfer/client.py
@@ -134,7 +134,7 @@ class TransferClient(BaseClient):
         path = self.qjoin_path("endpoint", endpoint_id)
         return self.delete(path)
 
-    def mgmt_monitored_endpoints(self, **params):
+    def management_monitored_endpoints(self, **params):
         """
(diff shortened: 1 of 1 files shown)


sirosen:
I think we want to add a bit more doc to these, and to some similar methods (like `task_list`), but we definitely shouldn't block merging this on that.

I tend to agree with @bd4 about names. I know it feels like bikeshedding a bit, but once we choose a name it's hard to change it.
@fred-lefebvre As suggested, can we change the names to `endpoint_manager_*` ? With that minor tweak, I'll happily merge this.

(Aside: serious :+1: for figuring out the non-trivial `PaginatedResource` stuff. Hopefully the docs were good enough. :smile: )

fred-lefebvre:
@sirosen , I really don't mind the name change.  I just want the functionality :P  Do you want me to resubmit or you will do the change yourself ?

sirosen:
Looks good. Merging.

PR #39, opened 2016-07-20
DOC - minor fix

NelleV:
This patch fixes a minor problem in the documentation for downloading the latest git repository:
It is ``git clone`` and not ``git checkout`` (unlike for svn where svn checkout downloads the latest version of the repository.


bjmc:
Good catch! Thank you!

PR #52, opened 2016-09-16
Complete Version 0.4.0

sirosen:
This includes all of the various items in the 0.4.0 milestone.
To enumerate them here:

- Add `TransferClient.task_wait`
- Update docstring style to better handle long descriptions for parameters
- Add "Authorization Code" (3-legged OAuth) Flow
- Add Client Credentials & Dependent Token Grant support
- Remove some deprecated functions and rename some exceptions which conflict with builtins

Additionally, adds support for token revocation, fixes some doc issues, adds an optional dependency on `python-jose` for parsing the OIDC `id_token`.

@jaswilli If you want, it might make sense to review this in tandem with the changes to the sample data portal. I'll be opening up that PR momentarily.


sirosen:
@jaswilli This is good to merge as far as I'm concerned. When you're comfortable with it, feel free to merge.
I'll tag as `0.4.0` and upload to pypi whenever.

jaswilli:
This all looks good to me. I manually tested the native app flow and then used the sample data portal to test the three-legged auth code flow, dependent token grant, token introspect, and client credentials grant.

There was a minor issue with the base64 encoding on python3 but I fixed that and was able to run the portal all the way through on both 2.7 and 3.

I added a couple things we might want to consider before officially releasing `0.4`: https://github.com/globus/globus-sdk-python/projects/1

PR #53, opened 2016-09-19
Fix mock import on python3

jaswilli:


jaswilli:
Yeah. I don't think any of the tests are currently passing but this is at least a step in the right direction.

PR #54, opened 2016-09-19
Add .editorconfig

jaswilli:


jaswilli:
> Set indent_style = 2 on *.json, not *.yml -- we don't have any yaml in the repo, and I doubt we will anytime soon

I was thinking we might want to consider setting this project up on Travis so that there would be public visibility into the test runs...

I'll mark `.json` as two spaces if that's your preference.

> Set values for *.rst files, since that's what the docs/ dir contains and what I generally try to use for a changelog and readme on python projects (so that PyPi will be happy with it)

I looked briefly to see if there were special formatting requirements for `.rst` files but didn't really see any that weren't covered by the `[*]`. Tell me what you'd like specifically and I'll add a section.

sirosen:
Er, actually `*.json` doesn't matter -- I forgot that there are no JSON files in the repo either.

As for the `*.rst`, I figured that if you're going to call out `*.md` you should call out `*.rst` as well. If you think it's well covered by the `[*]` rule, then feel free to leave it.

We should talk more about Travis.
I think that's a good idea, personally, and it handles a lot of the multi-python issues for us, but we have to contend with the credentials that we'll eventually need for test runs against the service (or a *lot* of mocking).

sirosen:
Also, nitpicks satisfied -- approved in the current state.

jaswilli:
> As for the *.rst, I figured that if you're going to call out *.md you should call out *.rst as well. If you think it's well covered by the [*] rule, then feel free to leave it.

Markdown sometimes needs trailing whitespace, so that feature needs to be turned off. I couldn't find anything in rst about trailing whitespace being significant.

sirosen:
To the best of my knowledge, and judging from my past experience with it, ReST never needs trailing whitespace *unless* you are writing a code sample that must include trailing whitespace.
I don't know of any reason why we would need that.

PR #57, opened 2016-09-27
Remove unused method parameter

jaswilli:
* Removes unused parameter `client_id` from `NativeAppAuthClient.oauth2_start_flow_native_app()`.


sirosen:
Good catch. Merged.

PR #58, opened 2016-09-28
Turn expires_in into a concrete expires at time.

jaswilli:
- An OAuth2 token response includes an `expires_in`
  field which is the number of seconds a token is valid.
  Since there's no reference point to base the `expires_in`
  offset against, it's only useful immediately upon receipt.
  In order to provide an accurate view of when a token expires,
  we turn this value into a POSIX timestamp.

The net result of this is the token payload, as presented by the SDK now has an `expires_at_seconds` property instead of `expires_in`.

### DIFF
--- a/globus_sdk/auth/token_response.py
+++ b/globus_sdk/auth/token_response.py
@@ -1,5 +1,6 @@
 import json
 import requests
+import time
 
 from globus_sdk.response import GlobusHTTPResponse
 from globus_sdk.exc import GlobusOptionalDependencyError
@@ -12,12 +13,14 @@ def _convert_token_info_dict(source_dict):
         - "refresh_token"
         - "token_type"
     """
+    expires_in = source_dict.get('expires_in', 0)
+
     return {
         'scope': source_dict['scope'],
         'access_token': source_dict['access_token'],
         'refresh_token': source_dict.get('refresh_token'),
         'token_type': source_dict.get('token_type'),
-        'expires_in': source_dict['expires_in']
+        'expires_at_seconds': int(time.time() + expires_in)
     }
 
 
@@ -28,8 +31,15 @@ class OAuthTokenResponse(GlobusHTTPResponse):
     """
     def __init__(self, *args, **kwargs):
         GlobusHTTPResponse.__init__(self, *args, **kwargs)
-        # dict used to avoid re-computing self.by_resource_server
-        self._by_resource_server = None
+
+        # call the helper at the top level
+        self._by_resource_server = {
+            self.resource_server: _convert_token_info_dict(self)}
+        # call the helper on everything in 'other_tokens'
+        self._by_resource_server.update(dict(
+            (unprocessed_item['resource_server'],
+             _convert_token_info_dict(unprocessed_item))
+            for unprocessed_item in self.other_tokens))
 
     @property
     def by_resource_server(self):
@@ -41,26 +51,23 @@ class OAuthTokenResponse(GlobusHTTPResponse):
         valid, this representation is typically more desirable for applications
         doing inspection of access tokens and refresh tokens.
         """
-        # memoize the results of this property computation in
-        # '_by_resource_server'
-        if self._by_resource_server is None:
-            # call the helper at the top level
-            self._by_resource_server = {
-                self.resource_server: _convert_token_info_dict(self)}
-            # call the helper on everything in 'other_tokens'
-            self._by_resource_server.update(dict(
-                (unprocessed_item['resource_server'],
-                 _convert_token_info_dict(unprocessed_item))
-                for unprocessed_item in self.other_tokens))
-
         return self._by_resource_server
+
+    @property
+    def expires_at_seconds(self):
+        """
+        A POSIX timestamp representation of the expiration
+        time for the top-level token in the response.
+        """
+        top_level_token = self._by_resource_server[self.resource_server]
+        return top_level_token['expires_at_seconds']
 
     @property
     def expires_in(self):
         """
         The ``expires_in`` value for the top-level token in the response.
         """
-        return self['expires_in']
+        return int(self.expires_at_seconds - time.time())
 
     @property
     def access_token(self):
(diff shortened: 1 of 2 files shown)


jaswilli:
Ignore the AppVeyor build failure. I was looking at setting it up for Windows testing and didn't realize it had actually turned on the integration.

globus_sdk/auth/token_response.py
```diff
@@ -56,11 +60,19 @@ def by_resource_server(self):
         return self._by_resource_server
 
     @property
+    def expires_at(self):
+        """
+        A POSIX timestamp representation of the expiration
+        time for the top-level token in the response.
+        """
+        return self['expires_at']
```
sirosen: This doesn't seem right. Lookups on `self` hit `self.data`, the parsed API response (inherited from `GlobusResponse`), which won't have an `expires_at` key in it.

What you really want is, I think, `self.by_resource_server[self.resource_server]['expires_at_seconds']`, which is ugly as sin, and maybe suggests that we shouldn't bother or need to restructure this.

I was iffy when I added these convenience properties for the top-level token, maybe we should just drop them?

globus_sdk/auth/token_response.py
```diff
@@ -56,11 +60,19 @@ def by_resource_server(self):
...
+
+    @property
     def expires_in(self):
         """
         The ``expires_in`` value for the top-level token in the response.
         """
-        return self['expires_in']
+        return int(self['expires_at'] - time.time())
```
sirosen: As with `self.expires_at`, I think this is problematic. Maybe drop this one too? Maybe drop all of the top-level token properties?

globus_sdk/authorizers/refresh_token.py
```diff
@@ -103,7 +89,7 @@ def _get_new_access_token(self):
         """
         token_response = self.auth_client.oauth2_refresh_token(
             self.refresh_token)
-        self._set_expiration_time(token_response.expires_in)
+        self._set_expiration_time(token_response.expires_at)
```
sirosen: Maybe this is the counterargument on the top-level token response properties. We know that there's only one token in the response, and this reads rather nicely.

If we change that around a bunch, we need to remember to fix these as well.

globus_sdk/auth/token_response.py
```diff
@@ -30,6 +33,7 @@ def __init__(self, *args, **kwargs):
         GlobusHTTPResponse.__init__(self, *args, **kwargs)
         # dict used to avoid re-computing self.by_resource_server
         self._by_resource_server = None
+        self._by_resource_server = self.by_resource_server
```
sirosen: This works, but I think it's confusing to read.

Can we just move the body of the `by_resource_server` property here, into `__init__`, and have the property just return the underlying value?
Or even just make `self.by_resource_server` an attribute on the object, and drop the `property` decorator altogether? (We'd need to move its description elsewhere in the docs, of course.)

sirosen: I think we also need to do this in the Dependent Token response class, to avoid a delay between fetching it and computing the `expires_at_seconds` times.

globus_sdk/authorizers/refresh_token.py
```diff
@@ -1,19 +1,12 @@
...
-# this is a fixed number of seconds to be pulled down from the expiration
-# timer, so that when the expires_in value is very small, the percent error
-# doesn't go to 0 and make us run the timer all the way down
-# so, we want (NOW - expires_in - (expires_in*PCT_ERROR) - FIXED_ERROR)
-# ultimately
-_EXPIRATION_FIXED_ERROR = 5
+
+# Provides a buffer for token expiration time to account for
+# possible delays or clock skew.
+EXPIRES_ADJUST_SECONDS = 60
```
sirosen: I think we spoke about this, but I just wanted to confirm that I'm cool with this change.
If it ends up not being enough, we can deal with it then.

mattias-lidman: We're not worried about (stable) clock skew. That's solved by returning `expires_in` rather than `expires_at`; Auth can think it's 1970 for all you care. What we are worried about is:

- Clock drift, but this is probably mostly a theoretical worry with modern hardware over the 48 hour lifetime of a token.
- Network delays. Both:
  - Time from token minting in Auth until the response is processed in the client.
  - Time from "is expiry time good?" check in the client until "is expiry time good?" check in Auth.

The two network delays can potentially add up to more than 60 seconds. The only way to get a truly robust value for  `EXPIRES_ADJUST_SECONDS` is to force a specific `timeout` for both these calls and setting the adjustment parameter to `timeout*2 + fudge_factor_for_potential_clock_drift_or_whatever_just_in_case`.

Now, 60 seconds is clearly sufficient in the vast majority of cases, but someone will run into this, potentially many someones if we have problems on the Auth side.

(We can still run into someone actually changing the clock of a node by some significant factor, but there's no programmatic way of guarding against that so we don't worry about it.)

globus_sdk/auth/token_response.py
jaswilli: Sorry, you're right. I was going to do it in the dependent token class but forgot.

globus_sdk/auth/token_response.py
jaswilli: Oops--you're right.

I agree that the top-level properties are sort of strange. But as you mention below, they do come in useful for quickly grabbing something like `expires_in`.

Although as discussed yesterday, if and when we end up in a world where not all tokens have the same `expires_in` value even that small benefit goes away.

jaswilli:
@sirosen This has been updated with the changes we discussed (I wish there was a way to mark a code review as "update" or something).

sirosen:
Closed/merged.

@jaswilli Note that I haven't bumped `globus_sdk/version.py` yet, so you don't need to `python setup.py develop` on your working virtualenvs.
I'll do that with a doc rebuild.

jaswilli:
👍  Doing a release would be great so that I can use `expires_at_seconds` in the workshop examples.

PR #59, opened 2016-09-29
Fixup endpoint_activate call

### LINKED ISSUE #56: TransferClient.endpoint_activate isn't including its body in the POST
This is almost definitely a bug, but I don't have time at the moment to test a fix (or I would just patch it right now).

The `self.post` call in `endpoint_activate` is missing the body.

jaswilli:
Closes #56


sirosen:
Another embarrassment down. Clearly I never checked what I wrote for the first version of this.

https://www.youtube.com/watch?v=rY0WxgSXdEE

PR #60, opened 2016-09-30
Send client_id if available when revoking token

jaswilli:
*If an `AuthClient` can't authenticate itself (e.g. a public/native app client), send its `client_id` in the post body.

Globus Auth requires either the request either be authenticated or that `client_id` is available in the body of the request.


sirosen:
Thanks for the fixup to use the type of `self.authorizer`.

In case you run into other things, I'm going to hold off on `0.4.2` for a while.
Let me know when you need it.

PR #66, opened 2016-10-18
Correct mapping for transfer submission sync_level

corpulentcoffee:
Fixes `exists` to be sync level `0` and adds `size` for sync level `1`.

https://docs.globus.org/api/transfer/task_submit#transfer_specific_fields


sirosen:
Looks good. As a separate issue, should we also log the level at which transfers are running? I think it would be a good `INFO` log message to have, so that if you are suspicious that your sync level is being passed through correctly, you can check.

PR #67, opened 2016-10-18
Add a specialized ActivationRequirementsResponse type

sirosen:
In two commits for ease of review (separates doc changes).
Also fixes a small bug in which the `TransferClient.endpoint_get_activation_requirements` method was calling autoactivate. Semantically, they're very similar, so it's not a big deal.

Supports the following features on a response named `r`:
- `r.supports_auto_activation` as a boolean property
- `r.active_until` as a method that takes either a relative time in seconds or a POSIX timestamp
- `r.supports_web_activation` sees if there's a known type of web activation supported

I'm a little edgy about that last one but wanted to capture the distinction between web activation and magical hosted CLI activation. Should we not try to do that?

It turns out that `r.active_until` is kind of important because it ensures that you're checking things in the right order. That is, `r["expires_in"] == -1` takes precedence over `r["activated"] == False` takes precedence over actual time comparison. It's weird and potentially confusing that endpoints with `expires_in=-1` also have `activated=False`, so hiding that is a Good Thing.

This response type is only used on explicit calls to get activation requirements and not on autoactivate and activate calls because those claim that they might not return an activation_requirements document. However, the API docs don't describe the mentioned `result` document type, so it's not obvious whether or not that's really true. Should this be applied to those calls as well? Is that safe? A quick check reveals that for most endpoints, this appears to be a specialization of the ActivationRequirements response.


sirosen:
Oh yeah, also `globus_sdk.transfer.response` got converted from a module to a package. This response type needed a bit of room to breathe.

jaswilli:
> I'm a little edgy about that last one but wanted to capture the distinction between web activation and magical hosted CLI activation. Should we not try to do that?

I need to be educated on this--are there endpoints that can _only_ be activated via the hosted CLI? My assumption would have been that `r.supports_web_activation` would always be true.

sirosen:
The only ones that I know break that rule are S3 endpoints. For those you *must* use the SSH CLI and use the activation command there.
There might be some other obscure type of endpoints for which that's true -- I think it's been used for experimental endpoint types and features in the past.

Of course, that won't be possible at all in a future world.

jaswilli:
Aren't S3 endpoints getting a new DSI in which they'll start behaving more like traditional endpoints?

Anyway, I'm not sure `supports_web_activation` is going to be terribly useful but I suppose there's no harm in having it either.

sirosen:
Yeah, but we don't know how long the S3-DSI work will take -- likely on the order of many months.

Let's leave it in, if you don't see any harm in it.
If there are future more sophisticated activation types that can't be done through the web (can't imagine what those would be...), we can use it to send that info along.

PR #69, opened 2016-10-19
Update TransferClient to accept UUID for endpoints

sirosen:
Anywhere that we were assuming something to be an endpoint_id as a string, we now merely assume that it has a `__str__` which produces an endpoint ID.

This captures any cases where the SDK is given a uuid.UUID, is fully backwards compatible, and supports potential future types. For example, you could have your own `GlobusEndpoint` type, and as long as its `__str__` is the endpoint ID, it's safe to pass to SDK methods.

This is needed for globus/globus-cli#86

### DIFF
--- a/globus_sdk/base.py
+++ b/globus_sdk/base.py
@@ -2,7 +2,7 @@ import json
 import logging
 
 import requests
-
+import six
 from six.moves.urllib.parse import quote
 
 from globus_sdk import config, exc
@@ -419,3 +419,16 @@ def merge_params(base_params, **more_params):
     for param in more_params:
         if more_params[param] is not None:
             base_params[param] = more_params[param]
+
+
+def safe_stringify(value):
+    """
+    Convert bytes by decoding, strings verbatim, and __str__ on anything
+    else
+    """
+    if isinstance(value, bytes):
+        return value.decode('utf-8')
+    elif isinstance(value, six.string_types):
+        return value
+    else:
+        return str(value)
--- a/globus_sdk/transfer/client.py
+++ b/globus_sdk/transfer/client.py
@@ -2,7 +2,7 @@ import logging
 import time
 
 from globus_sdk import exc, config
-from globus_sdk.base import BaseClient, merge_params
+from globus_sdk.base import BaseClient, merge_params, safe_stringify
 from globus_sdk.authorizers import (
     AccessTokenAuthorizer, RefreshTokenAuthorizer)
 from globus_sdk.transfer.response import (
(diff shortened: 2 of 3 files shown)


corpulentcoffee:
Do you think anyone would ever be holding an endpoint ID in a non-unicode (i.e. byte) string in Python 3?

Technically, this works in the SDK on Python 3 on `master`:
```
>>> transfer_client.endpoint_autoactivate(b'ddb59aef-6d04-11e5-ba46-22000b92c6ec')
TransferResponse({'request_id': 'KqpJeDFVI', 'oauth_server': None, 'length': 0, 'code': 'AutoActivated.GlobusOnlineCredential', 'endpoint': 'go#ep1', 'DATA_TYPE': 'activation_result', 'DATA': [], 'expire_time': '2016-10-20 19:43:18+00:00', 'message': 'Endpoint activated successfully using Globus Online credentials.', 'subject': '/C=US/O=Globus Consortium/OU=Globus Connect User/CN=shifflett', 'resource': '/endpoint/ddb59aef-6d04-11e5-ba46-22000b92c6ec/autoactivate', 'expires_in': -1})
```

With this PR, this ceases to work:
```
>>> transfer_client.endpoint_autoactivate(b'ddb59aef-6d04-11e5-ba46-22000b92c6ec')
  File "/home/mayor/Repos/globus-sdk/globus_sdk/transfer/client.py", line 310, in endpoint_autoactivate
    return self.post(path, params=params)
  File "/home/mayor/Repos/globus-sdk/globus_sdk/base.py", line 190, in post
    retry_401=retry_401)
  File "/home/mayor/Repos/globus-sdk/globus_sdk/base.py", line 364, in _request
    raise self.error_class(r)
globus_sdk.exc.TransferAPIError: (400, 'BadRequest', 'Invalid endpoint name u"shifflett#b\'ddb59aef-6d04-11e5-ba46-22000b92c6ec\'": Invalid characters', 'qh59roPFt')
```

This is because of the `str()` representation of a bytestring on Python 3:
```
~$ python2 -c "print(str(b'ddb59aef-6d04-11e5-ba46-22000b92c6ec'))"
ddb59aef-6d04-11e5-ba46-22000b92c6ec

~$ python3 !*
python3 -c "print(str(b'ddb59aef-6d04-11e5-ba46-22000b92c6ec'))"
b'ddb59aef-6d04-11e5-ba46-22000b92c6ec'
```

sirosen:
I think it's more important to support `uuid.UUID`s than it is to support bytestrings, but if we can have our cake and eat it too, let's do that.

The main issue, as I see it, is that in py2, bytestrings were "a kind of string, and in py3, that's no longer the case.
So we can't just look at `isinstance(endpoint_id, six.string_types)` and walk along -- bytestring will fall into the same category as UUIDs that way.
What is the correct check for us to be doing here? (I'm looking into this now.)

sirosen:
The logic for this is all in `globus_sdk.base.safe_stringify` -- that's used everywhere.

I didn't apply it to the preexisting calls to `str(server_id)` which I see in a few places.
Those were/are there to support passing a raw `int` as the `server_id`, and I don't think they need the special treatment. Maybe they should get it, for uniformity?

PR #70, opened 2016-10-19
Support for `include` param on token introspect

### LINKED ISSUE #64: Token Introspect Doesn't Support "include" parameter
I ran into this while trying to use the introspect call.

[Per the docs](https://docs.globus.org/api/auth/reference/#token_introspection_post_v2_oauth2_token_introspect), you have the include the additional parameter: `{'include': 'identities_set'}` to get a full identity set as part of the introspect response.

For now I've got my code using `AuthClient.post` directly, so this is easy enough to work around, but we should add support for that parameter out of the box.
Since that's the only supported value, and that most introspect callers will want it, I would even argue that it should be set by default.

sirosen:
Done by the same basic model used throughout TransferClient, with
`merge_params()`.
Must pass `identities_set` as a string, doesn't make it a boolean toggle
or anything fancy.
Resolves #64.


mattias-lidman:
Obligatory warning that the future of `include=identities_set` is somewhat up for debate. Is there a pressing need to include it in the SDK?

sirosen:
The need comes from the fact that trying to build an "identity set aware" application without this is impossible (and that's what Jim and I are trying to do, which is how #64 originally surfaced).

The only sensible alternative is to support more arbitrary parameter passthroughs, which is what Bryce liked for the `TransferClient` for a lot of its future-compatibility properties. In this particular case, I don't see that as strongly beneficial. As a bonus, if we do drop `include` as a supported parameter, the `TypeError` people will get if they upgrade the SDK communicates that unambiguously, whereas a service error might not.

I'm aware that its future is up for debate, but
- there is no timeline for its removal
- it's on our site as a documented feature
- even if it is removed, that will be a slow process with lots of prior warning
- it is arguably one of the most important things that Auth can do

PR #74, opened 2016-10-19
Fix typo

jaswilli:
(I fixed the docs by hand since re-making them locally caused a lot of unrelated changes. I'm guessing I probably use a different version of sphinx?)


sirosen:
This is fine, I'll merge.

We should check that we're synced on sphinx versions, but we also haven't been rigorous about doing a sphinx rebuild after every change. It wouldn't be surprising if sphinx sees a lot of updates.
I've been waiting for the dust to settle a little, and then I'll fix the doc on token revoke and clean up.

PR #75, opened 2016-10-20
Fix logging for transfer data sync_level

corpulentcoffee:
Two problems with the existing implementation:

- Python 3 needs `next(generator)` rather than `generator.next()`
- if the SDK caller uses a `sync_level` that isn't in the dict, the use
  of a fallback argument in `sync_dict.get(sync_level, sync_level)` will
  make back to the user's input, which means that the use of the
  generator here will raise a `StopIteration`

Really, we should be able to just avoid the use of the generator
altogether by not clobbering the original `sync_level` that the caller
used and outputting it with what we map it to (`self['sync_level']`).

### DIFF
--- a/globus_sdk/transfer/data.py
+++ b/globus_sdk/transfer/data.py
@@ -62,13 +62,9 @@ class TransferData(dict):
         # garbage overnight
         if sync_level is not None:
             sync_dict = {"exists": 0, "size": 1, "mtime": 2, "checksum": 3}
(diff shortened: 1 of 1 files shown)


sirosen:
I like this fix. I didn't like the generator as I wrote it, but didn't think of handling it this way.
The important thing is that if someone calls with `sync_level="checksum"`, they see `checksum` in the logs. Preserving the original argument is the easiest path to that.

PR #76, opened 2016-10-20
create acl doc fix

jackkordas:
I assume the html documentation is auto generated and will be updated as well.


sirosen:
It is (sphinx).

I'm holding off on an SDK release until we have a little more work done in the CLI, but this will be captured with the doc update in the next release.

Issue #63, opened 2016-10-03
Error in docs: Revoke Token is more finicky than indicated


sirosen:
The docstring for `AuthClient.oauth2_revoke_token` indicates, in its examples section, that you can do revocations with bad creds or wrong creds.
I think this gives a bad impression, since apparently the endpoint is more restrictive than that.

Although I maintain that this should be fixed to be highly permissive server-side, we should meanwhile update the docstring to be accurate.

I'm not bothering about this for `0.4.2`, however.

PR #77, opened 2016-10-20
Finalize v0.4.3

sirosen:
Includes the fix for #63, a version bump, and a rebuild of the docs.

There was more that I wanted to get into this version, but it also has been pushed back further than I anticipated.

Is there anything more that we demand for v0.4.3?


corpulentcoffee:
This PR isn't off the very current tip of `master`, so it might be missing the doc change in #76...?

sirosen:
@corpulentcoffee I've rebased to capture those changes, and also added a Makefile which does many useful things. Importantly, there's `make help`, which I stole from Sphinx as an idea, and very much like.

PR #78, opened 2016-10-21
Make transfer options explicit and add logging

jaswilli:
* Adds named parameters for the transfer task options added in https://github.com/globus/globus-cli/pull/94


sirosen:
Somehow I missed the update to add kwargs logging, as discussed. Merging now -- that was my only hold on this.

PR #81, opened 2016-10-25
Avoid fetching a submission ID when unnecessary

### LINKED ISSUE #80: TransferData and DeleteData should not call transfer_client.get_submission_id() in their __init__()s if kwargs already has a submission_id
If the caller provides a `submission_id`, this is a wasted call, adds unnecessary delay, and prevents dry-run testings without a valid token.

Additionally, if the caller explicitly passes a `submission_id` kwarg, but its value is `None`, then the behavior should be the same as if they passed no `submission_id` at all (i.e. a `get_submission_id()` call should indeed be made and the `self.update(kwargs)` call that happens at the end of initialization should not clobber it).

We should probably just explicitly name `submission_id` in the `__init__()` arguments list with a default of `None`, and explicitly do the right thing in the function body before the `self.update()` call happens (similar to `sync_level` and `label`).

corpulentcoffee:
When constructing a `TransferData` or `DeleteData` instance, if the caller supplies us with a `submission_id` that they already obtained, do not fetch another one.

Additionally, allow callers to pass `submission_id=None` explicitly to signal that they need a submission ID.

Closes #80.

---

```
(.venv3) ~/Repos/globus-sdk$ python
Python 3.5.2 (default, Jun 28 2016, 08:46:01) 
[GCC 6.1.1 20160602] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from globus_sdk import TransferClient, TransferData
>>> 
>>> tc = TransferClient()
>>> tc.get_submission_id = lambda: print("\nget_submission_id() was called!\n") or \
...                                {'value': '99999999-9999-9999-9999-999999999999'}
>>> 
>>> TransferData(transfer_client=tc,
...              submission_id='c0ffeec0ff-eec0-ffee-c0ff-eec0ffeec0ff',
...              source_endpoint='00000000-0000-0000-0000-000000000000',
...              destination_endpoint='00000000-0000-0000-0000-000000000000')
{'destination_endpoint': '00000000-0000-0000-0000-000000000000', 'DATA_TYPE': 'transfer', 'submission_id': 'c0ffeec0ff-eec0-ffee-c0ff-eec0ffeec0ff', 'verify_checksum': False, 'preserve_timestamp': False, 'encrypt_data': False, 'DATA': [], 'source_endpoint': '00000000-0000-0000-0000-000000000000'}
>>> 
>>> TransferData(transfer_client=tc,
...              source_endpoint='00000000-0000-0000-0000-000000000000',
...              destination_endpoint='00000000-0000-0000-0000-000000000000')

get_submission_id() was called!

{'destination_endpoint': '00000000-0000-0000-0000-000000000000', 'DATA_TYPE': 'transfer', 'submission_id': '99999999-9999-9999-9999-999999999999', 'verify_checksum': False, 'preserve_timestamp': False, 'encrypt_data': False, 'DATA': [], 'source_endpoint': '00000000-0000-0000-0000-000000000000'}
>>> 
>>> TransferData(transfer_client=tc,
...              submission_id=None,
...              source_endpoint='00000000-0000-0000-0000-000000000000',
...              destination_endpoint='00000000-0000-0000-0000-000000000000')

get_submission_id() was called!

{'destination_endpoint': '00000000-0000-0000-0000-000000000000', 'DATA_TYPE': 'transfer', 'submission_id': '99999999-9999-9999-9999-999999999999', 'verify_checksum': False, 'preserve_timestamp': False, 'encrypt_data': False, 'DATA': [], 'source_endpoint': '00000000-0000-0000-0000-000000000000'}
>>>

### DIFF
--- a/globus_sdk/transfer/data.py
+++ b/globus_sdk/transfer/data.py
@@ -25,7 +25,7 @@ class TransferData(dict):
     ``"exists"``, ``"size"``, ``"mtime"``, or ``"checksum"`` if you want
     greater clarity in client code.
 
-    Includes fetching the submission ID as part of document generation. The
+    If ``submission_id`` isn't passed, one will be fetched automatically. The
     submission ID can be pulled out of here to inspect, but the document
     can be used as-is multiple times over to retry a potential submission
     failure (so there shouldn't be any need to inspect it).
@@ -35,13 +35,15 @@ class TransferData(dict):
     documentation for example usage.
     """
     def __init__(self, transfer_client, source_endpoint, destination_endpoint,
-                 label=None, sync_level=None, verify_checksum=False,
-                 preserve_timestamp=False, encrypt_data=False, **kwargs):
+                 label=None, submission_id=None, sync_level=None,
+                 verify_checksum=False, preserve_timestamp=False,
+                 encrypt_data=False, **kwargs):
         source_endpoint = safe_stringify(source_endpoint)
         destination_endpoint = safe_stringify(destination_endpoint)
         logger.info("Creating a new TransferData object")
         self["DATA_TYPE"] = "transfer"
-        self["submission_id"] = transfer_client.get_submission_id()["value"]
+        self["submission_id"] = submission_id or \
+            transfer_client.get_submission_id()["value"]
         logger.info("TransferData.submission_id = {}"
                     .format(self["submission_id"]))
         self["source_endpoint"] = source_endpoint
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: globus_sdk/transfer/data.py [hunk 1]
```diff
@@ -25,7 +25,7 @@ class TransferData(dict):
     ``"exists"``, ``"size"``, ``"mtime"``, or ``"checksum"`` if you want
     greater clarity in client code.
 
-    Includes fetching the submission ID as part of document generation. The
+    If `submission_id` isn't passed, one will be fetched automatically. The
```
sirosen: I hate to nitpick this, but ReST requires double-ticks for this.
Applies to `DeleteData` below as well.

Let me know when that's fixed and I'll merge.

corpulentcoffee: Amended, thanks.

PR #82, opened 2016-10-26
Handle missing Content-Type in headers

jimPruyne:
In the case where Content-Type is completely missing from response
headers, the test for Content-Type value is failing. Check for presence
of Content-Type in the headers before checking for the value in the
Content-Type header.

I've actually seen this behavior (at least in an error condition) on a response from a server.

### DIFF
--- a/globus_sdk/exc.py
+++ b/globus_sdk/exc.py
@@ -22,10 +22,12 @@ class GlobusAPIError(GlobusError):
                    useful to developers, but there may be cases where it's
                    suitable for display to end users.
     """
+
     def __init__(self, r, *args, **kw):
         self._underlying_response = r
         self.http_status = r.status_code
-        if "application/json" in r.headers["Content-Type"]:
+        if "Content-Type" in r.headers and (
+                "application/json" in r.headers["Content-Type"]):
             logger.debug(('Content-Type on error is application/json. '
                           'Doing error load from JSON'))
             try:


jaswilli:
Do you know which call (or Auth operation) generates the response without `Content-Type`? It might also be worthwhile to fix this upstream.

jimPruyne:
I'm getting a 500 from Nexus (I need groups for search authorization stuff, so building it in in an sdk compatible way). I assume I'm doing something really awful in formatting my request (though not sure what I'm doing that could be so bad).

If anyone wants to debug nexus-side, I can send out info when I see that (though the request I've made is likely to be embarrassing assuming I can re-produce...)

jaswilli:
So you're using the SDK to make calls against nexus using `BaseClient`'s `.get`/`.post` methods?

I'm less interested in trying to fix nexus 😄

sirosen:
@jimPruyne I hope you aren't trying to instantiate a `BaseClient` directly? That will probably be more difficult. I just want to clarify/confirm that you are defining a new `BaseClient > NexusClient` class.

You can file a bug against Nexus if you want, but probably won't get any response, realistically.

jimPruyne:
That is the plan. I'll not bug you with nexus. Perhaps @mattias-lidman can let us know if he wants to investigate further.

jimPruyne:
No, not a direct BaseClient. I started from the NexusClient that you'd done and pointed out and have seemingly brought it up to date with current sdk behavior, but that's a separate discussion. I think this one-liner PR while perhaps minor does actually fix a bizarre case that should be handled.

jaswilli:
> I think this one-liner PR while perhaps minor does actually fix a bizarre case that should be handled.

It does. Sending `Content-Type` is not mandatory so this is a good change, I think.

I just assumed you were talking about Auth so that's why I said we should (in addition to this) fix it upstream.

jimPruyne:
yep, understood wrt Auth.

globus_sdk/exc.py
```diff
@@ -90,6 +92,7 @@ class TransferAPIError(GlobusAPIError):
     :ivar request_id: Unique identifier for the request, which should be
                       provided when contacting support@globus.org.
     """
+
```
sirosen: I know this is super-picky, but could you scrub these extra newlines and squash that with the current commit? Do that and I'll merge.

I want to avoid the noisy diff and inconsistent formatting with all of the other files.
Ultimately, I'd end up tearing back through here and fixing these lines at a later date, so it's just better to fix them now.

jimPruyne:
@sirosen After fun with rebase, I think I satisfied your squash request. Turns out that the extra nl showed up in another place so I got to go through it twice.

In general, I guess my use of flake8 not so compatible with everyone else's.

sirosen:
Were those newlines added by some version of autopep8 or similar?
`flake8`, at least as specified in the `test-requirements.txt`, doesn't care about whether or not there are newlines between a class docstring and the first method.

Since the class docstrings actually document `__init__` (one of several sphinx styles which I chose early on for us), I like them to have no whitespace.

jimPruyne:
was most likely autopep8 as that's in my worflow.

PR #85, opened 2016-10-27
Complete version 0.4.4

sirosen:
This is a little noisy because of the documentation update, but it only does a couple of things:
- Update version to `0.4.4`
- Fix a bug in `ClientLogAdapter` which shows up in python2.7

This will also release the content type fix that Jim wanted, and the mistreatment of submission_ids that Dave fixed.

@jaswilli @corpulentcoffee Any other blockers before putting this out there?


sirosen:
Merged my own branch with the blessing of my peers.

PR #89, opened 2016-11-02
Add an always_activated property to the activation response

### LINKED ISSUE #86: Add `ActivationRequirementsResponse.always_activated`
This should just be a property which returns the results of `self["expires_in"] == -1`.

Otherwise, we'll get more confusion like the discussion in globus/globus-cli#119

jaswilli:
* Adds an `always_activated` property that will be `True` if the endpoint is always activated (e.g., shared or GCP endpoints).

Closes #86


sirosen:
@jaswilli Sorry you had to wait so long for a merge. I completely missed this.

jaswilli:
No worries, if it was urgent I would have mentioned it to you.

PR #93, opened 2016-11-09
Add task_successful_transfers to TransferClient

### LINKED ISSUE #87: Add `TransferClient.task_successful_transfers()`
This should be a simple get call to the successful_transfers resource on the given task.

Docs:
https://docs.globus.org/api/transfer/task/#get_task_successful_transfers

sirosen:
Also minor fix to docstring on pause_info.
Resolves #87

### DIFF
--- a/globus_sdk/transfer/paging.py
+++ b/globus_sdk/transfer/paging.py
@@ -19,9 +19,14 @@ class PaginatedResource(GlobusResponse, six.Iterator):
     - They support `limit` and `offset` query params, with `limit` being a
       count of elements to return, and offset being an offset into the result
       set (as opposed to a page number), 0-based
+      OR
+      They support a `marker` query param, which is an opaque value
     - They return a JSON result with `has_next_page` as a boolean key,
       indicating whether or not there are more results available -- even if the
       hard limit for the API forbids requesting these results
+      OR
+      They return a JSON result with `marker` as a key indicating whether or
+      not there are more results available
     - Individual results are JSON objects inside of an array named `DATA` in
       the returned JSON document
 
@@ -40,6 +45,9 @@ class PaginatedResource(GlobusResponse, six.Iterator):
     # pages have a 'has_next_page', but use 'last_key' rather than 'offset' +
     # 'limit'
     PAGING_STYLE_LAST_KEY = 2
+    # pages have a 'marker' attribute, which refers to the "next page" of
+    # results, but which is opaque
+    PAGING_STYLE_MARKER = 3
 
     # bind an object at class def time to act as a sentinel value for iteration
     # Basically grabbing something that can't be duplicated by iteration
@@ -82,6 +90,14 @@ class PaginatedResource(GlobusResponse, six.Iterator):
         self.max_total_results = max_total_results
         self.offset = offset
         self.paging_style = paging_style
+
+        # counter for how many results we've gotten thusfar, used to cap paging
+        # in non-offset based styles
+        self.num_results_fetched = 0
+
+        # potentially necessary params during paging
+        self.limit = None
+        self.next_marker = None
 
         # check the requested num results to see if it exceeds the maximum
         # total number of results allowed by the API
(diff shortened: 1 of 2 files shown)


jaswilli:
I've been playing with this call on a local branch. It's a paged result set, using a form of paging that's different than the other transfer calls (a "marker" param that's meaning is opaque).

I suppose we could just ignore that for now and put the burden of dealing with that on the caller.

sirosen:
Ugh. I hadn't realized.
Probably just scrap this branch, in that case.

We shouldn't put the paging burden on the caller if we can help it -- does this require that `PaginatedResource` gets another paging style to support?

jaswilli:
I've been debating about whether or not to try and adapt your pagination class to this additional style or just doing a one-off `SuccessfulTransfersResponse` class to handle it.

`/successful_transfers` supports `limit` and `marker` so I think the calling style of `PaginatedResource` _might_ be a little awkward but since it's not something for public consumption maybe that'll be okay.

sirosen:
The whole `PaginatedResource` class is pretty awkward, TBH.
But it makes sure we share important things like making sure we fetch the first page of results immediately on the first call.

I'd be more in favor of an added "paging style" that's at least as awkward as the already-present 3 different ways of doing it.
Adding another response class seems like it would require replicating a lot of the work already in the `PaginatedResource`.

An alternative, if it's more attractive, is to refactor that a bit to separate the implementation of paging styles from the `PaginatedResource` interface.
Perhaps via simple inheritance, or perhaps a kind of encapsulation... whatever seems best.

sirosen:
I added marker style paging and adjusted the implementation to use it.

Unfortunately, I've not yet had a chance to test the new behavior and make sure it works.
I have confirmed that these changes do not break the existing paging styles.

globus_sdk/transfer/paging.py
```diff
@@ -163,22 +179,68 @@ def iterable_func(self):
...
+            Check that the API says there are more results available.
+
+            Additionally, update the PaginatedResource.maker or
+            PaginatedResource.offset based on the response
+            """
+            # if the paging style is MARKER, look at the marker
+            if self.paging_style == self.PAGING_STYLE_MARKER:
+                # marker may be 0, null, or absent if no more results
+                # API docs aren't 100% clear -- looks like 0 is what we should
+                # expect, but we'll also accept null or absent to be safe
```
jaswilli: I believe how this works (when I was playing around with this I looked at the Transfer code) is that `marker` is always an opaque value that marks where in set of available data you are (by default it will be `0` if you have all the results in this response, or if this is the beginning of the set of results). It shouldn't ever be `null` but there's no harm in handling that just to be safe.

If there are more pages of results, that will be indicated by the `next_marker` property in the response, which will be set to a value that the caller should use for the next requests. `next_marker` will be `null` if there are no more results available.

So basically, you make your first call as `/task/:id/successful_transfers` and you'll get back a payload that will have a `marker` set to some value that you could send back with `?marker=<val>` if you want to re-fetch the exact same set of results again. If the task has no more results, `next_marker` will be `null`. If there are more results, `next_marker` will be an opaque value that you can fetch with `?marker=<next_marker>`.

globus_sdk/transfer/paging.py
```diff
@@ -163,22 +179,68 @@ def iterable_func(self):
 
         This method is the real workhorse of this entire module.
         """
-        # now, cap the limit per request to the max per request size
-        limit = min(self.num_results, self.max_results_per_call)
+        if not self.client_kwargs['params']:
+            self.client_kwargs['params'] = {}
+
+        # to start with, cap the limit per request to the max per request size
+        self.limit = min(self.num_results, self.max_results_per_call)
```
jaswilli: Transfer uses a default limit of `100` for `/successful_transfers` so we could consider using that for marker style paging as an optimization that should not put undue burden on Transfer.

globus_sdk/transfer/paging.py
```diff
@@ -19,9 +19,14 @@ class PaginatedResource(GlobusResponse, six.Iterator):
...
       count of elements to return, and offset being an offset into the result
       set (as opposed to a page number), 0-based
+      OR
+      They support a `marker` query param, which is an opaque value
     - They return a JSON result with `has_next_page` as a boolean key,
       indicating whether or not there are more results available -- even if the
       hard limit for the API forbids requesting these results
+      OR
+      They return a JSON result with `marker` as a key indicating whether or
+      not there are more results available
```
jaswilli: See below where I (attempt) to explain how this paging style works.

globus_sdk/transfer/paging.py
sirosen: You are 100% correct. And yet, I remember this working... Perhaps I got confused in long lists of files and just saw the same results over and over. Manual validation at its finest.

I should be fetchng `next_marker` -- for this pattern of purely paging results, there's no reason to keep around the value of `marker`. I'll fix and re-test, more carefully this time.

globus_sdk/transfer/paging.py
sirosen: Does Transfer actually support the `?limit=100` param though? The docs for this call show `?marker=...` as the only supported parameter.

jaswilli: Yeah, it's not documented but it should support `limit`.

I don't think it's super important that we special case this, I just thought that for cases where the result set is huge it might be nice to have a default of `100` instead of `10`.

sirosen:
I've fixed the `next_marker` vs. `marker` confusion of mine and found out what the flaw was in my testing that made me incorrectly believe that I had this working.


After weighing the options a little, I don't think we should use `limit` at all. It's not in the docs, and Transfer may consider it a "bug" and remove it down the line. Maybe unlikely, but if we bake assumptions into the SDK those propagate into a potentially large number of clients, so not using it is safer and more conservative.


@jaswilli Should be ready for (hopefully final) review.

PR #94, opened 2016-11-09
Let AuthClient.get_identities accept UUID objects and iterables of strings or UUIDs

### LINKED ISSUE #79: Allow UUIDs to be passed to AuthClient.get_identities
For uniformity with the TransferClient accepting a UUID type for endpoint IDs, single-identity calls to `get_identities` should be allowed to pass a UUID.

We can also expand the set of types it takes. For example, iterables of strings or of UUIDs are things which we can handle if desired.
Mixed type iterables are kind of nasty, but it doesn't hurt us much to support them if we want (in fact, it might be harder to ensure that we _aren't_ getting them).

sirosen:
Uses the same safe_stringify method as the Transfer client to allow `ids` to be given a single UUID object without choking.
Resolves #79

### DIFF
--- a/globus_sdk/auth/client_types/base.py
+++ b/globus_sdk/auth/client_types/base.py
@@ -1,5 +1,7 @@
 from __future__ import print_function
 
+import six
+import collections
 import logging
 
 from globus_sdk import config
@@ -67,7 +69,8 @@ class AuthClient(BaseClient):
         arguments, looks up identity information for the set of identities
         provided.
         ``<U>`` and ``<I>`` in this case are comma-delimited strings listing
-        multiple Identity Usernames or Identity IDs.
+        multiple Identity Usernames or Identity IDs, or iterables of strings,
+        each of which is an Identity Username or Identity ID.
 
         Available with any authentication/client type.
 
@@ -85,7 +88,7 @@ class AuthClient(BaseClient):
            u'status': u'unused',
            u'username': u'globus@globus.org'}]}
         >>> ac.get_identities(
-        >>>     identities=",".join(
+        >>>     ids=",".join(
         >>>         ("46bd0f56-e24f-11e5-a510-131bef46955c",
         >>>          "168edc3d-c6ba-478c-9cf8-541ff5ebdc1c"))
         ...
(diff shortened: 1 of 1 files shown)


globus_sdk/auth/client_types/base.py
```diff
@@ -106,10 +106,22 @@ def get_identities(self, **params):
...
         self.logger.info('Looking up Globus Auth Identities')
+
+        # if either of these params has a truthy value, stringify it safely,
+        # letting us consume args whose `__str__` methods produce "the right
+        # thing"
+        # most notably, lets `ids` take a single UUID object safely
+        if usernames:
+            params['usernames'] = safe_stringify(usernames)
+        if ids:
+            params['ids'] = safe_stringify(usernames)
```
corpulentcoffee: Wrong source variable?

globus_sdk/auth/client_types/base.py
```diff
@@ -106,10 +106,22 @@ def get_identities(self, **params):
...
+
         self.logger.debug('params={}'.format(params))
+
         if 'usernames' in params and 'identities' in params:
             self.logger.warn(('get_identities call with both usernames and '
                               'identities set! Expected to result in errors'))
+
```
corpulentcoffee: Is `'identities' in params` correct, or should it be `'id' in params`?

sirosen:
Fixed those typos. I'm surprised that my testing didn't catch them?
I didn't check on passing both though -- that bug with `identities` vs `ids` is pretty old, but it's just supposed to trigger a warning, which is why we haven't caught it.

corpulentcoffee:
Eh, okay, so I think if you do this, you might need to handle the caller passing a list.

On master, passing a list results in Requests serializing the call like `usernames=a&usernames=b`, which doesn't work on Globus Auth, but is harmless in that it only returns one result back.

With this change, passing a list results in it being stringified like `['username1@example.com', 'username2@example.com', 'username3@example.com']`, which Requests happily passes along as-is, which causes Globus Auth to create garbage identities (literally usernames containing single quotes and brackets).

corpulentcoffee:
(And yes, I know the docstring has helpful examples that show joining against a sequence before calling `get_identities()`, but we know that people will occasionally not read that and only see the plural nouns in the method arguments.)

sirosen:
I think you're probably right. Letting this method consume iterables has been a long time coming. I'll try to knock this out today.

sirosen:
My current path on this is to use `six.string_types` and `collections.Iterable` to handle this:

```python
>>> isinstance('abc', six.string_types)
True
>>> isinstance('abc', collections.Iterable)
True
>>> isinstance(['abc'], six.string_types)
False
>>> isinstance(['abc'], collections.Iterable)
True
```

So, first check for string type, then check for Iterable, then failover to `safe_stringify`. That should handle most reasonable things.

sirosen:
@corpulentcoffee I think this is good now?

It works on UUIDs, Strings, Iterables of Strings, Iterables of UUIDs, and of course anything that looks like UUID and has a `__str__` which produces something we can use.

corpulentcoffee:
Thanks!

Issue #99, opened 2016-11-30
Authorization Code Grant flow logs message about Native App


sirosen:
This line: https://github.com/globus/globus-sdk-python/blob/master/globus_sdk/auth/oauth2_authorization_code.py#L120

Just needs a new message.

jaswilli:
Sorry, I didn't notice Stephen had opened an issue about this.

PR #101, opened 2016-11-30
Fix logging output

jaswilli:
* Authorization Code flow was accidentally logging a message about Native Apps.


corpulentcoffee:
Fixes #99

PR #102, opened 2016-12-01
Fix merge_params() usage

jaswilli:


sirosen:
Wow. That was an embarrassing oversight.

Thanks for finding this!
We should discuss doing a release of `0.4.6` to get this out

PR #108, opened 2016-12-21
Update Native App Tutorial to list redirect URI

sirosen:
The necessary Redirect URI value to take advantage of our default
behaviors is now listed at the start of the older tutorial doc.
This should be reworked into the tutorial section in the future, but is
a good fixup for now.


corpulentcoffee:
We should revisit this at some point and talk about maybe making `https://auth.globus.org/v2/web/auth-code` automatic for all native apps -- either implicitly whitelisted (like `localhost`) or added automatically when interacting with the developer pages.

PR #112, opened 2017-01-10
Allow "unlimited" results for paginated TransferClient methods

sirosen:
All `TransferClient` methods which are paginated under the hood accept a `num_results` parameter.
We now allow setting that parameter to `None` to indicate that as many results as possible should be fetched.

Updates docstrings of the various methods which accept `num_results` to ensure that this behavior is clear.

Since Jason requested this, I'm assigning it to him.


jaswilli:
Forgot about this one. I'll test it out soon 👍

PR #119, opened 2017-01-20
adding update create and delete unit tests for transfer client

aaschaer:
Submitting the first few unit tests for the transfer client to make sure I'm on the right track.
Feedback is much appreciated!

### DIFF
--- a/tests/unit/test_transfer_client.py
+++ b/tests/unit/test_transfer_client.py
@@ -1,31 +1,19 @@
 import globus_sdk
 
-from tests.framework import (
-    CapturedIOTestCase, get_client_data,
-    GO_EP1_ID, GO_EP2_ID, SDKTESTER1A_NATIVE1_RT)
+from tests.framework import (CapturedIOTestCase, get_client_data,
+                             GO_EP1_ID, GO_EP2_ID,
+                             GO_USER_ID, SDK_USER_ID,
+                             SDKTESTER1A_NATIVE1_RT)
 from globus_sdk.exc import TransferAPIError
+from globus_sdk.transfer.paging import PaginatedResource
 
 
 class TransferClientTests(CapturedIOTestCase):
 
-    def clean(self):
+    @classmethod
+    def setUpClass(self):
         """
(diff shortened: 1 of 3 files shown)


tests/unit/test_transfer_client.py
```diff
@@ -3,22 +3,56 @@
...
+
+        r = self.tc.create_endpoint({"display_name": "SDK Test Endpoint"})
+        self.test_ep_id = r["id"]
+
+    def tearDown(self):
+        """
+        calls clean to remove any data created in setUp or individual tests
+        """
+        self.clean()
+        # TODO: should SDK Tester log out here?
```
sirosen: Given that we're using a hardcoded token right now, there isn't much log out behavior to speak of.
But yes, when we start doing login flows as part of testing, revoking those creds will have to be part of it too.

tests/unit/test_transfer_client.py
```diff
@@ -3,22 +3,56 @@
...
             client_id=get_client_data()["native_app_client1"]["id"])
         authorizer = globus_sdk.RefreshTokenAuthorizer(
             SDKTESTER1A_NATIVE1_RT, ac)
-        tc = globus_sdk.TransferClient(authorizer=authorizer)
+        self.tc = globus_sdk.TransferClient(authorizer=authorizer)
+
+        self.clean()
+
+        r = self.tc.create_endpoint({"display_name": "SDK Test Endpoint"})
+        self.test_ep_id = r["id"]
```
sirosen: We can roll with this for now, but come to think of it, we should think about what parts of this should be in `setUp` and which should maybe be in `setUpClass`.

I think having endpoint cleanup and creation as part of `setUp` is probably right, and maybe the `TransferClient` should be instantiated in `setUpClass`.

aaschaer:
Made the requested changes and added tests up through the acl functions.

Most TODOs require having constant endpoints with different characteristics than go#ep1 and go#ep2, I'll continue testing as much as I can off those two endpoints for now.

PR #120, opened 2017-01-27
Add remaining transfer client tests and move shared endpoint tests to seperate class

aaschaer:
The inheritance feels a little clunky, but inheriting TransferClientTests directly causes all its tests to be run in duplicate, and having the separate classes makes the tests run about a minute faster, so I think its worth it.

A few issues I'm worried about.
If go#ep1 or go#ep2 were momentarily unavailable, should the tests catch this and restart?
If two users ran the testing suite at the same time, it would likely cause io errors and failed tests as they attempted to make/delete the same files, is there a good way to deal with this? Maybe make a test dir on go#ep1 with a unique name per run of the suite?

I appreciate any and all feedback.

### DIFF
--- a/tests/unit/test_transfer_client.py
+++ b/tests/unit/test_transfer_client.py
@@ -8,7 +8,10 @@ from globus_sdk.exc import TransferAPIError
 from globus_sdk.transfer.paging import PaginatedResource
 
 
-class TransferClientTests(CapturedIOTestCase):
+# class that has setUp and tearDown for all transfer client testing classes
+class BaseTransferClientTests(CapturedIOTestCase):
+
+    __test__ = False  # prevents base class from trying to run tests
 
     @classmethod
     def setUpClass(self):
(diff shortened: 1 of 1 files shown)


sirosen:
> The inheritance feels a little clunky

Although I understand what you mean, I actually think it's okay. Once you accept that "test classes own test functions", it becomes natural that you cannot inherit from them.
Using an auxiliary class to encapsulate shared setup/teardown is better than anything else I can come up with.

> If go#ep1 or go#ep2 were momentarily unavailable, should the tests catch this and restart?

This should probably not be the responsibility of the testsuite, so we're okay in the current state. If there were a problem with the service, it could cause the tests to fail in all sorts of undetectable/non-recoverable ways.

> If two users ran the testing suite at the same time, it would likely cause io errors and failed tests as they attempted to make/delete the same files, is there a good way to deal with this? Maybe make a test dir on go#ep1 with a unique name per run of the suite?

The problem actually runs deeper, as one test user could delete endpoints that another one is using.
I turned off concurrent builds on Travis when I noticed that with your initial tests.

Isolated dirs on the tutorial endpoints is probably a good gesture towards concurrency safety, but, personally, I'd rather have more extensive unsafe tests than less extensive, but safe, tests.

@jaswilli, @corpulentcoffee, what do you guys think of this last issue? How do you handle this problem in the webapp testsuite?
Non-concurrency safe tests on an open source repo bothers me a bit more than it would on something closed (where access to the tests is controlled).
Could we maybe make test cleanup less aggressive to handle this?

Current inclination: merge and resolve concurrency issues later.

corpulentcoffee:
> @jaswilli <https://github.com/jaswilli>, @corpulentcoffee
<https://github.com/corpulentcoffee>, what do you guys think of this last
issue? How do you handle this problem in the webapp testsuite?

This isn't a totally solved problem for us either, as we have been leaning
on running the tests without concurrency.

Our newer tests, however, generally setup and tear down the specific things
they need (e.g. groups, shares, directories) for the purposes of that test,
and give those things random names to avoid conflicts with another test
runner doing the same test. When tests do fail, that can leave behind junk
on the environment, which I've been handling by just periodically reaping
anything that is older than 24 hours.

jaswilli:
👍on random names for things you're creating. A prefix+random string goes a long way towards eliminating the concern over concurrent test runs colliding.

sirosen:
We agreed in Slack -- merge this and open an issue to make it more concurrency-safe.

PR #121, opened 2017-01-27
fix and add API links to transfer client doc strings

aaschaer:


sirosen:
Awesome! Thanks for adding these.

Issue #122, opened 2017-01-31
Make testsuite concurrency-safe


sirosen:
This is a known problem discussed partly in #120 
Many thanks to @aaschaer for his work on the testsuite (!), but he noted that there are issues with concurrency safety.

We should rectify this.

A couple of things we can do to start:
- [x] Normal teardown should only delete assets created during *this* test run
- [x] Tests should track all assets that are created and delete them all on teardown
- [x] Complete cleanup (e.g. delete all endpoints belonging to user) should be a manual task runnable from the repo, just like generating a new refresh token
- [x] Tests which use shared resources (e.x. Tutorial Endpoints) should insert random, unique strings into any identifiers used

This will get us into a better state, and we can attack any remaining issues as they arise.

sirosen:
Resolved by #125

PR #125, opened 2017-02-09
Concurrent tests

aaschaer:
Implements test-suite concurrency safety measures described in issue #122

Tests ran without issue on trials with 8 concurrent "make test"s running, and the only error that occurred with more running was an occasional GlobusConnectionError: ('Connection aborted.', BadStatusLine("''",)), which may simply have been due to the larger sample size rather than a concurrency problem.

Should the test suite bother with Connection Errors such as these?

### DIFF
--- a/tests/manual_tools/clean_sdk_test_assets.py
+++ b/tests/manual_tools/clean_sdk_test_assets.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python
+
+import globus_sdk
+
+
+# removes all endpoints, bookmarks, and files/folders on tutorial enpdoints
+# intended to be used to clean the testing areas for the SDK Tester user
+def clean():
+
+    # constants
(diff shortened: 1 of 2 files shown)


sirosen:
I think we should improve robustness against network errors, but let's just open an issue to track that. Unless we start seeing them very frequently, it's an acceptable known flaw in the tests.

I like the asset cleanup dict. It's tidy, and anything unusual could be wrapped in a one-off function we put in there, of necessary. It also leaves us free to annotate -- I see you did this with an extra "name" attribute in one case.

It might be safer to use `uuid` over random bits to guarantee uniqueness, but the chance of collision is so low that I can't imagine it being a real problem.

I'm going to merge, but I noticed that you'll need the test user's password to test out the manual cleanup. You'll also need it to test login flows, so maybe it just goes directly into the repo... I'm still not sure how best to handle it.

PR #126, opened 2017-02-10
Remove deprecated behavior loading access tokens from config `[general]` section

sirosen:
This is the last change that I wanted to slip into `0.5.0`
It's breaking for some people, probably, but we've wanted to drop this behavior pretty much since `GlobusAuthorizer` and its subclasses were added.

This includes some amendment to docs to purge all remaining reference to these tokens.

@jaswilli, @corpulentcoffee Does this seem too aggressive to you? Or is it probably okay?


jaswilli:
Seems reasonable to me. I can't imagine anyone would be using this "feature" at this point, and if they are, when it breaks and they realize they don't need to go to the tokens site every 48hrs they'll be super happy.

PR #128, opened 2017-02-14
TransferData, DeleteData, and PaginatedResource unit tests

aaschaer:
First batch of unit tests. Figure I should split these up into multiple pull requests as I go to make reviewing easier.

Question: when testing PaginatedResource I didn't test the Marker paging style, as it is opaque and seems to be pretty much unused in the SDK. Is that okay?

Any and all feedback is much appreciated.

### DIFF
--- a/tests/unit/test_data.py
+++ b/tests/unit/test_data.py
@@ -1,0 +1,119 @@
+import globus_sdk
+
+from tests.framework import (CapturedIOTestCase, get_client_data,
+                             SDKTESTER1A_NATIVE1_RT,
+                             GO_EP1_ID, GO_EP2_ID)
+
+
+class DataTests(CapturedIOTestCase):
+
+    def setUp(self):
+        """
+        Creates two TransferData objects and two Delete objects for testing,
+        one with defaults and one with custom paramaters
+
+        """
+        super(DataTests, self).setUp()
+
+        ac = globus_sdk.NativeAppAuthClient(
+            client_id=get_client_data()["native_app_client1"]["id"])
+        authorizer = globus_sdk.RefreshTokenAuthorizer(
+            SDKTESTER1A_NATIVE1_RT, ac)
+        self.tc = globus_sdk.TransferClient(authorizer=authorizer)
+
+        # constants
+        self.label = "label"
+        self.params = {"param1": "value1", "param2": "value2"}
+
+        # default transfer data
+        self.tdata = globus_sdk.TransferData(self.tc, GO_EP1_ID, GO_EP2_ID)
+        # transfer data with paramaters
+        self.param_tdata = globus_sdk.TransferData(self.tc,
+                                                   GO_EP1_ID, GO_EP2_ID,
+                                                   label=self.label,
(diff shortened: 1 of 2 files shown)


sirosen:
The marker style paging is only used for one resource right now, but it doesn't fit very well with the other kinds of paging. I think leaving it out is fine.

I'm doing a review pass now. I think there are going to be one or two minor edits I'll request before merge.

tests/unit/test_data.py
```diff
@@ -0,0 +1,119 @@
...
+        Verifies TransferData field initialization
+        """
+
+        # default init
+        self.assertEqual(self.tdata["DATA_TYPE"], "transfer")
+        self.assertEqual(self.tdata["source_endpoint"], GO_EP1_ID)
+        self.assertEqual(self.tdata["destination_endpoint"], GO_EP2_ID)
+        self.assertIn("submission_id", self.tdata)
+        self.assertIn("DATA", self.tdata)
+        self.assertEqual(len(self.tdata["DATA"]), 0)
```
sirosen: Doing initialization steps in `setUp()` and then testing the results in your test functions is a little unusual.
It's not that bad, but it will pose a problem if we ever want to test out various initialization behaviors.

My preference would be to move the creation of `TransferData` and `DeleteData` to helpers, with the ability to pass overrides for various specific values. i.e. allow callers to `self.make_tdata()` or `self.make_tdata(sync_level="checksum")`

It's not necessary for the current set of tests, but it may be in the future.

tests/unit/test_paging.py
```diff
@@ -0,0 +1,98 @@
+import requests
+import json
+
+from tests.framework import (CapturedIOTestCase)
+from globus_sdk.transfer.paging import PaginatedResource
+from globus_sdk.transfer.response import IterableTransferResponse
+
+
+class PagingSimulator():
```
sirosen: Should be `class PagingSimulator(object):`
Otherwise, this will be an old-style class. Unlikely to be a problem, but good to fix for cleanliness.

tests/unit/test_paging.py
```diff
@@ -0,0 +1,98 @@
...
+        data["DATA"] = []
+        for i in range(offset, min(self.n, offset + limit)):
+            data["DATA"].append({"value": i})
+        # fill has_next_page field
+        data["has_next_page"] = (offset + limit) < self.n
+
+        # make the simulated response
+        response = requests.Response()
+        response._content = bytes(json.dumps(data))
+        response.headers["Content-Type"] = "text/plain"
```
sirosen: I assume something blew up with `Content-Type` unset? If we're going to set it, we should use `application/json`, as that is the one the service will send.

aaschaer:
Made requested changes.

When I started making the helper method for make_tdata and make_ddata, I realized they were such thin wrappers over the object constructors that it would be easier just to use the object constructors, so I just made testing initialization values not dependent on setup.

sirosen:
I think that's a fair place to end up. Initializing the objects as part of the tests is reasonable.

All looks good now. Merging.

Issue #105, opened 2016-12-01
Test suite expansion


corpulentcoffee:
Conversation topics to get us going:

- [x] what additional nose tests need adding to globus_sdk/tests?
- [x] for testing methods that are actually supposed to call live services, do we want to use real test tokens, or do we want to fake the services, or both?
- [ ] expand to AppVeyor testing for Windows? add OS X on Travis?

corpulentcoffee:
12/1 we discussed possibly registering some test clients of different types and using live refresh tokens.

sirosen:
01/18 we agreed to create a 3-legged oauth app, a native app, a single test user, fetch refresh tokens for each app, and use this as a starting point.

bjmc:
:+1: :100:

aaschaer:
Listing tests for globus_sdk/tests on #127, please let me know if anything is missing.

sirosen:
#131 added AppVeyor.

@corpulentcoffee Do we want to add OSX on Travis? I don't really see the harm in it, so why not, right?

corpulentcoffee:
If it works and is reliable, sure.

sirosen:
> If it works and is reliable, sure.

Nope. It doesn't work. We couldn't even get one OSX build through Travis's enormously overburdened queue for that.

I'm going to close this, as that resolves the question of whether or not to add OSX via Travis.

corpulentcoffee:
Yes, the basis of my comment was hearing Jason mention that it was slow/unreliable.

PR #132, opened 2017-02-17
Improve builds on AppVeyor; add pypy builds to Travis

sirosen:
As soon as I added the Appveyor build, I was bothered by how slow it is. Too slow!
I wanted to make it faster, so I looked into parallelization features.
Good thing the tests are now concurrency-safe! They have to be if two independent CI servers will run them.
Unfortunately, Appveyor doesn't do parallel builds on free repos.

Two main angles of attack: do parallel builds on Travis, reduce the number of versions we cover on Appveyor.

With Travis, because we can do parallel builds, we use the build matrix feature. While we're add it, add pypy builds and allow failure to get some extra coverage.

In the case of Appveyor, remove all of the Python3 on 32-bit builds. We'll keep the 2.7 on 32-bit build so that we hit a 32 bit machine at some point, but if you're running py3 on a non-64-bit box, that's pretty unusual. I think this is okay to drop.
Additionally, simplify the Appveyor config per direction given by packaging.python.org

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,16 @@ python:
   - "3.4"
   - "3.5"
   - "3.6"
+  # allow failures on CPython dev and pypy
+  # we want to be warned about these, but they aen't critical
   - "3.6-dev"
+  - "pypy"
+  - "pypy3"
 matrix:
   allow_failures:
     - python: "3.6-dev"
+    - python: "pypy"
+    - python: "pypy3"
 install:
   - pip install .
   - pip install -r test-requirements.txt
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 sudo: false
-dist: trusty
 language: python
 python:
   - "2.7"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,46 +2,24 @@ environment:
   matrix:
     # Appveyor may upgrade python to new point releases
     # See: http://www.appveyor.com/docs/installed-software#python
-
     - PYTHON: "C:\\Python27"
-      PYTHON_VERSION: "2.7.x"
-      PYTHON_ARCH: "32"
-
     - PYTHON: "C:\\Python27-x64"
-      PYTHON_VERSION: "2.7.x"
-      PYTHON_ARCH: "64"
-
-    - PYTHON: "C:\\Python33"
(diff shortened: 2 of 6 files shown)


sirosen:
It seems I misread the docs on Travis. It also only allows one concurrent job on free repos. :(

sirosen:
Okay, there's still refactoring work here that's useful -- the Appveyor config is way less messy now.
Salvaging that.

I thought that if we specified our pythons in the `.travis.yml` `matrix.includes` directive, we'd get parallelism. Seems that's not the case.
From that part, I'm salvaging the addition of `pypy` and `pypy3` as `allow_failures` targets.

Builds appear to be failing because of `task_wait` taking too long. Presumably this is due to concurrent test runs creating tasks which queue simultaneously, and block on the 3-task-per-user limit.
There might also be faulting tasks for the SDK test user which are blocking up the queue.

sirosen:
The name of my branch for this work is becoming wrong at best, and a grim parody of life at worst.
Builds will probably be overall slower after I'm done here, as we'll be testing OSX on Travis to complete #105 , doubling the size of the Travis build matrix.

sirosen:
I've reverted the addition of OSX builds.
Not even the first OSX build made it through the Travis queue in over an hour.

sirosen:
Looks like the older AppVeyor config wasn't correctly invoking the variable python interpreters.
Fixing it reveals some py3/Windows issues, which I'll try to clean up here.

sirosen:
I needed the pull request to drive the CI builds.
They're now passing and working as expected, I think, so this is ready for review.

jaswilli:
> It seems I misread the docs on Travis. It also only allows one concurrent job on free repos. :(

I think you can have concurrent jobs. I think they just changed the default to be `1` at some point. You can go into settings and bump it up.

.travis.yml
```diff
@@ -1,16 +1,21 @@
 sudo: false
-dist: trusty
```
jaswilli: Is there a reason we want to go back to early 90s ubuntu?

sirosen: :(

`pypy` installation fails on trusty, at least as of right now. I'd like to have `pypy` and `pypy3` in the allowed failures builds for Travis. If it works on cpython and "another python", it probably works everywhere.

jaswilli: 😭

PR #134, opened 2017-02-17
Clean .globus files older than an hour. Also a few minor python 3 compatability changes

aaschaer:
Took a little more fiddling with timezones than I was expecting, but seems to be working after testing on some shared dirs I made before lunch.

Also I confirmed on the globus transfer web app that deleting a shared endpoint does not clean up ~/.globus/sharing. Should this issue be tracked somewhere?

### DIFF
--- a/tests/unit/test_transfer_client.py
+++ b/tests/unit/test_transfer_client.py
@@ -24,9 +24,7 @@ def setUpModule():
     path = "~/.globus/sharing/"
     hour_ago = datetime.utcnow() - timedelta(hours=1)
     filter_string = "last_modified:," + hour_ago.strftime("%Y-%m-%d %H:%M:%S")
-
-    ls_params = {"path": path, "filter": filter_string}
-    old_files = tc.operation_ls(GO_EP1_ID, **ls_params)
+    old_files = tc.operation_ls(GO_EP1_ID, path=path, filter=filter_string)
 
     kwargs = {"notify_on_succeeded": False, "notify_on_fail": False}
     ddata = globus_sdk.DeleteData(tc, GO_EP1_ID, **kwargs)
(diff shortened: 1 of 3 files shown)


tests/unit/test_transfer_client.py
```diff
@@ -29,6 +59,7 @@ def setUp(self):
         Creates a list for tracking cleanup of assets created during testing
         Sets up a test endpoint
         """
+        super(BaseTransferClientTests, self).setUp()
```
sirosen: Good catch! We're not really relying on the IO capturing for much of anything, but this way we won't see warnings or other noise.

tests/unit/test_transfer_client.py
```diff
@@ -416,7 +448,7 @@ def test_get_endpoint_role(self):
 
         # get role id from role list, assumes admin id is first
         list_doc = self.tc.endpoint_role_list(self.test_ep_id)
-        role_id = iter(list_doc["DATA"]).next()["id"]
+        role_id = list_doc["DATA"][0]["id"]
```
sirosen: Not only is this more compatible, but I also think more readable.

jaswilli:
FYI, I restarted the build on this after turning on concurrent jobs and there was a failure--not sure if it's material to this PR or just a transient oops.

https://travis-ci.org/globus/globus-sdk-python/jobs/202781697

sirosen:
I'm restarting the builds to see if they pass on a retry. I think that the timing on the task wait test is too tight.
I'm going to up it by a factor of 10 in an upcoming PR.

sirosen:
Oh, #132 fixed some of these same python3 issues, so we now have a conflict.

I'll just resolve on merge, assuming the build passes.

PR #135, opened 2017-02-20
Add ValueError catching to exc

aaschaer:
The exceptions were breaking when testing on malformed json. I feel like catching these provides a little more information should something get really messed up, but if we want to break explicitly when json cant even be parsed then just ignore this PR.


sirosen:
Yeah, I'm 100% onboard with this change.

The expectation we have set is that any API error will be propagated client-side as a `GlobusAPIError`. Anytime that that class breaks, we're raising the wrong type of exception.

PR #136, opened 2017-02-20
Unit tests for base.py, config.py, exc.py, and response.py

aaschaer:
I ran test_base_client.py with 16 concurrent test suites running without any errors or assets getting past cleanup, so it should still be safe to have multiple builds running.

test_config.py is an expansion of test_config_loading.py and should probably replace it at some point?

For testing exc.py I only tested exceptions that were more than a wrapper around hard-coded text as I assume those don't need tests?

### DIFF
--- a/tests/files/sample_configs/get_test.cfg
+++ b/tests/files/sample_configs/get_test.cfg
@@ -1,0 +1,10 @@
+[general]
+option = general_value
+
+[section]
+option = section_value
+
+[environment section]
+option = environment_value
+
+[nonexistant]
--- a/tests/files/sample_configs/ssl_test.cfg
+++ b/tests/files/sample_configs/ssl_test.cfg
@@ -1,0 +1,12 @@
+[general]
+
+[environment true]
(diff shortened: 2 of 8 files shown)


sirosen:
My only reservation, and it is a mild one, is that we might make `BaseClient` an abstract class in the future.
In that case, we could rewrite these tests to implement a very simple subclass of it and work with that.

That's not really necessary right now though.

Waiting for tests to pass, then merging.

sirosen:
Ah, failures seem to revolve around cleanup hitting an unexpected condition:
```
TransferAPIError: (404, u'ClientError.NotFound', u"Directory '/~/.globus/sharing/' not found on endpoint 'ddb59aef-6d04-11e5-ba46-22000b92c6ec'", u'HXJF3lVcU')
```

Want to just fix that here? It would be ideal to do a separate PR and rebase this off of the fix, but it's not necessary if you want to move faster.

aaschaer:
Ah, I'll fix that here real quick.

PR #137, opened 2017-02-22
AuthClient unit tests, and changes needed for them.

aaschaer:
I made some changes to facilitate AuthClient unit tests, and wanted to make sure they were reasonable.

I added an additional_params argument to AuthClient.oauth2_refresh_token() so I could pass the client_id needed for native authorization. I figure this shouldn't hurt anything, especially since AuthClient.oauth2_revoke_token() already does this.

I added openid, profile, and email to the requested scopes for the Native app grant and requested new auth and transfer refresh tokens that are now in constants.

It felt wrong to have a long list of user data values in constants for validating auth responses, so I made a get_user_data() function similar to get_client_data()

### DIFF
--- a/globus_sdk/auth/client_types/base.py
+++ b/globus_sdk/auth/client_types/base.py
@@ -187,7 +187,7 @@ class AuthClient(BaseClient):
         return self.current_oauth2_flow_manager.exchange_code_for_tokens(
             auth_code)
 
-    def oauth2_refresh_token(self, refresh_token):
+    def oauth2_refresh_token(self, refresh_token, additional_params=None):
         r"""
         Exchange a refresh token for a :class:`OAuthTokenResponse
         <globus_sdk.auth.token_response.OAuthTokenResponse>`, containing
@@ -213,6 +213,8 @@ class AuthClient(BaseClient):
         form_data = {'refresh_token': refresh_token,
                      'grant_type': 'refresh_token'}
 
+        if additional_params:
+            form_data.update(additional_params)
         return self.oauth2_token(form_data)
 
     def oauth2_revoke_token(self, token, additional_params=None):
--- a/tests/unit/test_auth_client.py
+++ b/tests/unit/test_auth_client.py
@@ -7,7 +7,6 @@ import globus_sdk
 from tests.framework import (CapturedIOTestCase,
                              get_client_data, get_user_data,
                              SDKTESTER1A_NATIVE1_AUTH_RT)
-from globus_sdk.auth import GlobusNativeAppFlowManager
 from globus_sdk.exc import GlobusAPIError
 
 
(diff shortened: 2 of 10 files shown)


globus_sdk/auth/client_types/base.py
```diff
@@ -187,7 +187,7 @@ def oauth2_exchange_code_for_tokens(self, auth_code):
         return self.current_oauth2_flow_manager.exchange_code_for_tokens(
             auth_code)
 
-    def oauth2_refresh_token(self, refresh_token):
+    def oauth2_refresh_token(self, refresh_token, additional_params=None):
```
sirosen: I will never object to extension points like this, as they're exposing underlying API functionality that already exists.
So, this is good and we should keep it.

My one question is: why do you need it? I know you said that you needed to send the `client_id` for the Native App, but the `NativeAppAuthClient` class provides that customization already. Is there a reason you need that capability here as well?

aaschaer:
Ah yes, I was being so careful not to miss any of the files with minor changes I forgot the actual unit tests. Whoops xD

As for why I needed the additional_params, I wanted to be able to test the base AuthClient without being dependent on NativeAppAuthClient, especially since NativeAppAuthClient overrides some of AuthClient's functions.

Also, I noticed that invalid_client errors from Auth are not being parsed properly, I'm guessing this is because they don't have the error in an "errors" list: {"error":"invalid_client"}, does the error need to be reformatted, or do we need to change the way we parse errors?

aaschaer:
Looks like there's a python3 compatibility error and or I made too many assumptions about urls. I'm surprised the travis build passed though?

sirosen:
Wingdings OS never fails to amaze. :stuck_out_tongue_closed_eyes: 
We also had a `flake8` failure at some point in the past that only showed up on appveyor, so I'm growing more thankful for/resentful of that integration with each passing day.

As for the `{"error": "invalid_client"}` errors, I think we need to do a better job handling those, but we can tackle that as a separate issue.
Most of the assumptions built into error handling today are centered around the Transfer API, so we should expect gaps as we test more of our Auth interaction.

tests/unit/test_auth_client.py
```diff
@@ -0,0 +1,230 @@
...
+        Instantiates an AuthClient with an access_token authorizer
+        """
+        super(AuthClientTests, self).setUp()
+
+        self.access_token = self.test_oauth2_token()
+        self.ac = globus_sdk.AuthClient(
+            authorizer=globus_sdk.AccessTokenAuthorizer(self.access_token),
+            client_id=get_client_data()["native_app_client1"]["id"])
+
+    def test_get_identities(self):
```
sirosen: This test is quite long, and I think it would be beneficial to break it up a little bit.
e.x. If `go@globusid.org` has changed and returns different data, we should still correctly test that sending `usernames` and `ids` results in an error.

We don't necessarily need to be very strict about where/how we break things down. Under the most rigorous regimen, we'd have a distinct test for each behavior of each call, with each set of parameters we give it -- seems a bit like overkill, IMHO.

However, I think it's useful, especially with a test like this which is vulnerable to breakage due to external changes, to separate testing of error conditions from working features.

I'd also be curious about @corpulentcoffee's and @jaswilli's opinions on this matter -- where and how they like to have unit tests broken apart.

aaschaer:
Made the requested changes. Will put integration tests in another PR later.
I'll go ahead and make issues to track the auth Error handling issues that have come up.

sirosen:
For some reason, GitHub marked my comment on `test_oauth2_exchange_code_for_tokens` as being out of date, even though that content didn't change.

Could you update that method to use the same mocking approach? I like the results very much.

PR #141, opened 2017-02-24
Unit tests for Native clients, Confidential clients, and flow managers.

aaschaer:
I tested as much as I could without having the clients tests import the flow managers and vice versa, Integration tests will cover the interactions not covered here.

I'll make an issue tracking the eventual need for a Resource Server to test oauth2_get_dependent_tokens.


sirosen:
I'm not sure about the checks on the verifier/challenge for Native Apps -- those tests might prove fragile in the long run.
However, I'd rather have them here, with the risk that we might have to remove them, than have nothing at all.

PR #143, opened 2017-02-25
Add prefill_named_grant support

### LINKED ISSUE #130: Add `prefill_named_grant` support to NativeApp grant flow
Per globus/globus-cli#176 , there's an option to prefill the named grant label, `prefill_named_grant`. It's probably a string.

We need to support this so that the CLI and similar applications can use it.

Not considering this blocked on documentation.

jaswilli:
* Adds support for using the Native App's `prefill_named_grant` query string parameter when setting up an authorization flow.

Example:
```python
native_app.oauth2_start_flow(
    requested_scopes=scopes,
    redirect_uri=redirect_uri,
    prefill_named_grant='host.example.com')
```

Result:
screen shot 2017-02-24 at 9 58 34 pm

Closes #130

### DIFF
--- a/globus_sdk/auth/client_types/native_client.py
+++ b/globus_sdk/auth/client_types/native_client.py
@@ -46,7 +46,8 @@ class NativeAppAuthClient(AuthClient):
 
     def oauth2_start_flow(
             self, requested_scopes=None, redirect_uri=None,
-            state='_default', verifier=None, refresh_tokens=False):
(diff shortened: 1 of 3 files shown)


sirosen:
Love it. Let's rebuild docs and release 0.5.1 once the tests pass and I can merge it.

I thought you might not include the test for it, so I had just written something near identical to include. So wonderful to have a test suite into which we can slot these little things.

PR #145, opened 2017-02-25
Make auto-provisioning identities opt-in

jaswilli:
* Make `get_identities()` **not** auto-provision identities by default.

I think this is the desired behavior for the vast majority of users so I propose we make auto-provisioning of identities opt-in.

It would surprise me if any SDK users are currently relying on the auto-provisioning behavior but just in case we should highlight this change in the release notes.

### DIFF
--- a/globus_sdk/auth/client_types/base.py
+++ b/globus_sdk/auth/client_types/base.py
@@ -51,7 +51,8 @@ class AuthClient(BaseClient):
 
         BaseClient.__init__(self, "auth", authorizer=authorizer, **kwargs)
 
-    def get_identities(self, usernames=None, ids=None, **params):
+    def get_identities(self, usernames=None, ids=None,
+                       provision=False, **params):
         r"""
         GET /v2/api/identities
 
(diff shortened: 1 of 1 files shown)


sirosen:
What about Nexus? It takes user input at least when we do searches. Will it be okay with the new behavior? i.e. Does it pass `provision=True` correctly?

This smells to me like a backwards-incompatible change on a practical level, even if the docs are phrased loosely enough to allow for it, because people may be doing single-identity lookups and expecting 1 result (rather than 0).

As a user of Auth, this feels like a good change to make -- given that both options now exist, it's a more sane default. I'm just advocating to take this one slow, and send out a warning 2-4 weeks before changing it.

jaswilli:
> What about Nexus? It takes user input at least when we do searches. Will it be okay with the new behavior? i.e. Does it pass provision=True correctly?

I'm not sure how the SDK relates to Nexus, can you explain? Are you using this in particular way for your private fork with groups support?

> send out a warning 2-4 weeks before changing it.

Timing doesn't really matter to me. We can call it a breaking change and ship this with 1.0.

I'd definitely like to turn off auto-provisioning off in the context of the CLI sooner rather than later but I believe we could get `provision='false'` working via the params/kwargs.

sirosen:
We should think about the CLI behavior once it starts using newer versions of the SDK. I doubt it will matter for most clients.

PR #146, opened 2017-02-27
Integration tests between auth clients and flow managers

aaschaer:
Integration tests for AuthClient, ConfidentialAppAuthClient, and NativeAppAuthClient with GlobusAuthorizationCodeFlowManager and GlobusNativeAppFlowManager.

Testing code exchange only tests that auth returns 401's for now. When we handle those auth errors better we should check that they are invalid_grant errors and not invalid_client errors, but unless we automate SDK tester login or somehow get a code that wont expire I don't think we can automate testing a successful code exchange.

### DIFF
--- a/tests/integration/test_auth_client_flow.py
+++ b/tests/integration/test_auth_client_flow.py
@@ -1,0 +1,133 @@
+from six.moves.urllib.parse import quote_plus
+
+import globus_sdk
+from globus_sdk.auth.oauth2_constants import DEFAULT_REQUESTED_SCOPES
+from globus_sdk.auth.oauth2_native_app import make_native_app_challenge
+from tests.framework import CapturedIOTestCase, get_client_data
+from globus_sdk.exc import GlobusAPIError
+
+
+class AuthClientIntegrationTests(CapturedIOTestCase):
+
+    def test_oauth2_get_authorize_url_native(self):
+        """
+        Starts an auth flow with a NativeAppFlowManager, gets the authorize url
+        validates expected results with both default and specified parameters.
+        """
+        ac = globus_sdk.AuthClient(
+            client_id=get_client_data()["native_app_client1"]["id"])
+
+        # default parameters for starting auth flow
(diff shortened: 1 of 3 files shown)


tests/integration/test_auth_client_flow.py
```diff
@@ -0,0 +1,133 @@
...
+                         "code_challenge_method=" + "S256",
+                         "access_type=" + "online"]
+        for val in expected_vals:
+            self.assertIn(val, url_res)
+
+        # starting flow with specified paramaters
+        flow_manager = globus_sdk.auth.GlobusNativeAppFlowManager(
+            ac, requested_scopes="scopes", redirect_uri="uri",
+            state="state", verifier="verifier", refresh_tokens=True)
+        ac.current_oauth2_flow_manager = flow_manager
```
sirosen: I don't _object_ to us testing this, but this here is one of the side effects of the fact that the base `AuthClient` is never expected to have a flow manager. I wouldn't bother stripping out these tests, but you could have skipped the tests on `AuthClient` and it would have done no harm.

tests/integration/test_auth_client_flow.py
```diff
@@ -0,0 +1,133 @@
...
+        """
+        ac = globus_sdk.AuthClient(
+            client_id=get_client_data()["native_app_client1"]["id"])
+        flow_manager = globus_sdk.auth.GlobusNativeAppFlowManager(ac)
+        ac.current_oauth2_flow_manager = flow_manager
+
+        with self.assertRaises(GlobusAPIError) as apiErr:
+            ac.oauth2_exchange_code_for_tokens("invalid_code")
+        self.assertEqual(apiErr.exception.http_status, 401)
+        self.assertEqual(apiErr.exception.code, "Error")
```
sirosen: I know you expressed some concern over the fact that we can only test for errors on these, not a real code exchange.
This is fine for now.

If we want to test a successful code exchange, we need to bite off the work of automating a login flow in the browser. We decided to skip on that for now, so this is the best we can do for today.

PR #149, opened 2017-03-01
Authorizer unit and integration tests

aaschaer:
So it turns out the refresh_token returned in a refresh token grant is identical to the one sent, so I wasn't able to test revoking a refresh token in the integration tests like we wanted to, unless there is another way to get new refresh tokens from an existing one?

### DIFF
--- a/tests/integration/test_refresh_token_auth_client.py
+++ b/tests/integration/test_refresh_token_auth_client.py
@@ -1,0 +1,110 @@
+try:
+    import mock
+except ImportError:
+    from unittest import mock
+
+import globus_sdk
+from tests.framework import (CapturedIOTestCase,
+                             get_client_data, GO_EP1_ID,
+                             SDKTESTER1A_NATIVE1_TRANSFER_RT)
+from globus_sdk.exc import GlobusAPIError
+
+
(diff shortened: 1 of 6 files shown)


sirosen:
No, there is not. I guess I wasn't clear -- I wanted to test revoking a refresh token, but we can't without the browser pieces. So, that needs to be left out.

Ultimately, it's okay, but in a perfect world we'd have all of our toys. :smilie:

I'll review this now.

tests/integration/test_refresh_token_auth_client.py
```diff
@@ -0,0 +1,110 @@
...
+        continues to function.
+        """
+        # TODO: Get a new refresh_token to revoke here instead
+        self.authorizer.refresh_token = None
+
+        # make request, confirm successful
+        get_res = self.tc.get_endpoint(GO_EP1_ID)
+        self.assertEqual(get_res["id"], GO_EP1_ID)
+        # confirm no new token was gotten
+        self.assertEqual(self.access_token, self.authorizer.access_token)
```
sirosen: I think we should just remove this test for now? I wouldn't bank on the behavior going forward when the is no refresh token attached.

PR #150, opened 2017-03-02
token_response unit tests, activation_response unit tests, and framework changes for them

aaschaer:
I added a JOSE_FLAG in the code to only run logic dependent on python-jose if it was able to be imported, but this makes some tests pass trivially without warning that they are skipping logic based on dependencies, so I'm not completely happy with it.

There is a chance that some of these tests will break when the id_token expires in 2 days, but I will need to update these tests then anyways.


aaschaer:
The id_token tests did in-fact break once the token expired, so the only thing we can test on an expired id_token is that decoding it sees that its expired, which should suffice.

However, it seems that these tests are being skipped on unsuccessful python-jose imports. Travis claims that it successfully pip installs python-jose, and its working locally for me so I'm not quite sure what's wrong?

tests/unit/test_activation_response.py
```diff
@@ -0,0 +1,98 @@
...
+        response._content = six.b(json.dumps(data))
+        return ActivationRequirementsResponse(response)
+
+    def test_expires_at(self):
+        """
+        Confirms expires_at is set properly by __init__
+        """
+        for seconds in [0, 10, 100, 1000, -10]:  # should -10 work here?
+            response = self.make_response(expires_in=seconds)
+            self.assertEqual(response.expires_at, int(time.time()) + seconds)
```
sirosen: This is super nitpicky, but can we change to account for the rare possibility that we're very close to a second boundary when this runs?
e.g.
```python
now = int(time.time()) + seconds
self.assertIn(response.expires_at, (now - 1, now, now + 1))
```

sirosen: Also, I don't know that the API will every actually respond with `-10`, but there's nothing saying that it can't. We should continue to include it, IMO.

tests/unit/test_token_response.py
```diff
@@ -0,0 +1,236 @@
...
+            "refresh_token": "RT3", "other_tokens": [], "token_type": "3",
+            "id_token": SDKTESTER1A_NATIVE1_ID_TOKEN,
+            "access_token": "invalid_access_token"}
+        self.top_token["other_tokens"] = [self.other_token1, self.other_token2]
+
+        # create the response
+        http_response = requests.Response()
+        http_response._content = six.b(json.dumps(self.top_token))
+        http_response.headers["Content-Type"] = "application/json"
+        self.response = OAuthTokenResponse(http_response)
```
sirosen: This is not a "right now" issue, but maybe it would be nice to have a helper in `tests.framework` which converts a dict to a response, given a response class. Just thought of it seeing this a couple of times in this PR.

tests/unit/test_token_response.py
```diff
@@ -0,0 +1,236 @@
...
+        """
+        top_convert = _convert_token_info_dict(self.top_token)
+
+        # expected results for top token data
+        for key in ["scope", "access_token", "refresh_token", "token_type"]:
+            self.assertEqual(self.top_token[key], top_convert[key])
+
+        # sanity check expires at, assuming test will run well within 1 second
+        self.assertEqual(int(time.time()) + self.top_token["expires_in"],
+                         top_convert["expires_at_seconds"])
```
sirosen: As in the other case, let's account for the second-boundary case with `assertIn(..., (now - 1, now, now + 1))`

The assumption that we run within a second is okay, but this protects against rare, but confusing, spurious failures.
We could get by with `(now - 1, now)`, but it seems safer to include `now + 1` just to protect against some future refactor breaking this.

sirosen: There are a few more of these below. I'm not going to drop the same comment on them, but let's update 'em all.

tests/unit/test_token_response.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def test_expires_in(self):
+        """
+        Gets two measurements of response's expires_in attribute,
+        Sanity checks that expires_in decreases as time goes on
+        """
+        first_measure = self.response.expires_in
+        time.sleep(1)
+        second_measure = self.response.expires_in
+        self.assertTrue(first_measure < self.top_token["expires_in"])
```
sirosen: Maybe I've missed something. Why should this be true? Wouldn't you expect them to be equal most of the time, and rarely different?
If this is right, please add a comment explaining.

sirosen:
I pushed a commit to your branch which should fix the installation of `python-jose` on Travis.

aaschaer:
Made requested changes.

And I think test-requirements is showing up despite there not being changes because I at first added python-jose to there directly and then removed it when appveyor was having trouble importing pycrypto. If its a problem I could probably squash some commits.

sirosen:
I'd appreciate an interactive rebase to clean out things that didn't actually make it in.
It would also be good to not have a merge commit from a remote branch into the local copy -- that's something that always makes me cranky.

That cleanup is all that's left though -- in terms of actual content this seems good.

PR #154, opened 2017-03-08
Improve Native App PKCE support

jaswilli:
* Generate `code_verifier` and `code_challenge` strings that are RFC 7636 compliant:
  * 43-128 characters in length.
  * Contain only the following characters: `a-z`, `A-Z`, `-`, `_`, `.`, `~`.

Needs to wait until Auth handles code verifiers without the padding.


jaswilli:
The Globus Auth changes have been released to production so this can be merged/released.

Issue #152, opened 2017-03-07
Improve `requests` error wrapping to provide a more controlled surface for our package


sirosen:
We've seen `requests.exceptions.ConnectionError` come back to use with a body of `('Connection aborted.', BadStatusLine("''",))`. We then present that as a `GlobusConnectionError` to better encapsulate our usage of `requests` as an HTTP layer.

This is a confusing error from `httplib` being wrapped by `requests` being wrapped by us. It likely relates to a malformed response. The discussion on a [related, underlying issue](kennethreitz/requests#2364) in `requests` itself suggested the possibility of a proxy chopping off parts of encrypted data, and breaking decryption.

To make matters worse, on newer Pythons, `httplib` will potentially change, changing `requests` and therefore us. This is bad: people may try to rely on our error messages, and we should have more control.

So, let's should do something like the following:
- Let `GlobusNetworkError`s take a custom message and a contained exception
- Keep the underlying exception distinct -- don't pass its args back to our own exception class
  - Do we need to pass some of its args through?
  - Preserve the original traceback?

We may want to move the `RequestsException` handler out of our underlying caller and into a helper, just to avoid bulking up `BaseClient` too much.

sirosen:
Closed by #155

PR #155, opened 2017-03-08
Improve requests error wrapping

aaschaer:
Addresses #152 

I gave all the exceptions very simple messages for now. We could expand on the conversion logic to look at the underlying exceptions and give more helpful messages based on those data if wanted.

### DIFF
--- a/globus_sdk/exc.py
+++ b/globus_sdk/exc.py
@@ -1,5 +1,6 @@
 import logging
 import textwrap
+import requests
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


globus_sdk/exc.py
```diff
@@ -141,11 +142,17 @@ class InvalidDocumentBodyError(GlobusError):
     """
 
 
-# Thin wrappers around requests exceptions, so the SDK is API independent.
+# Wrappers around requests exceptions, so the SDK is API independent.
```
sirosen: Haha! Indeed, this comment change summarizes the whole PR! :rofl:

sirosen:
Somehow, an attribute which I thought was promised to us by Python is missing in Windows versions of Python3? Not all `Exception` classes have a `message` attribute?

This is totally weird to me, and I'm not sure what we need to do to make sure we don't trip over it.

aaschaer:
It seems python3 removed Exception storing an explicit message field and just uses the arguments passed to it to form a string. I'll adjust the code to reflect this.

However, I am once again dubious that Travis is actually running our tests with python3

sirosen:
Oh dear, I believe that you are correct...
Taken from one of the Py3 builds:
```
Creating /home/travis/build/globus/globus-sdk-python/.venv/lib/python2.7/site-packages/globus-sdk.egg-link (link to .)
```

Let's resolve in a follow-up issue.

PR #156, opened 2017-03-08
Add support for setting python version for venv

sirosen:
Fixes Travis testing always using Python2 (whoops!).

Makes it easier to test against all of your available pythons.
And also have Travis set it explicitly to `TRAVIS_PYTHON_VERSION`.


sirosen:
I had to rework this, as the version number handling only got me so far. For whatever reason, setting up a py3.6 venv blew up with this config. Rather than trying too much harder, took a direct approach of linking `.venv` to the location of the pre-built Travis venv. Keep the version stuff in case it's handy -- it mostly worked.

PR #161, opened 2017-03-22
Make jwk_data get respect ssl_verify=no

### LINKED ISSUE #160: ssl_verify=no is ignored on ID Token parsing, maybe elsewhere too
The whole purpose of the `ssl_verify` config setting is really to support internal use cases in which we're running against a non-production server and don't want to jump through the hoops to get `requests` to play nice and support SNI.
I don't think there's much other purpose to it.

However, we fail to use it when making an explicit `requests.get` call in `decode_id_token`.

There may be other locations where we make explicit requests. I'll check for that now, but this is the only one which I know is causing issues.

sirosen:
Take the value off of the given AuthClient object.
Closes #160 

It turns out this was the only call where `verify` wasn't set.

I think I might be the only person who cares about this?

### DIFF
--- a/globus_sdk/auth/token_response.py
+++ b/globus_sdk/auth/token_response.py
@@ -116,7 +116,8 @@ class OAuthTokenResponse(GlobusHTTPResponse):
         logger.debug('Fetch JWK Data: Start')
         oidc_conf = auth_client.get('/.well-known/openid-configuration')
         jwks_uri = oidc_conf['jwks_uri']
-        jwk_data = requests.get(jwks_uri).json()
+        # use the auth_client's decision on ssl_verify=yes/no
+        jwk_data = requests.get(jwks_uri, verify=auth_client._verify).json()
         logger.debug('Fetch JWK Data: Complete')
 
         return jwt.decode(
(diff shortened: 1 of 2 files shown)


jaswilli:
> I think I might be the only person who cares about this?

I'm curious where you're actually running into an error. All our Auth environments should pass verification, no?

This also seems to have a legitimate test failure.

sirosen:
I'll take a look at the test failure tomorrow.

I see this on Python2, as an SNI issue. I can either turn off verification or fix the actual problem. While working up the fix, I tried turning off verification. I'm still working on getting my SNI change in place, but figured that this would be good to take care of as well.

sirosen:
The AppVeyor build passed because it doesn't run the `python-jose` dependent tests.

Amended my commit to include a very minor tweak to add the `_verify` property to the mocked `AuthClient` used in those tests. The inscrutable failure messages came of that value being a `mock.Mock` instead of a bool or string.

sirosen:
@jaswilli, can I ask that we push this through?

I can't get [this internal issue](https://github.com/globusonline/globus-search/pull/222) fixed -- I'm having trouble even figuring out what's wrong -- so this is actually blocking me from assembling some working components.

Issue #158, opened 2017-03-17
Remove top-level attributes of token response structure for access_token, etc.?


sirosen:
I was helping some OSG folks get started with the SDK, and the extra properties on the `OAuthTokenResponse` structure which reflect the top level token were a point of confusion. I even realized that I could not adequately answer how that token gets chosen/decided by Auth (first requested scope? I think?).

Should just remove `OAuthTokenResponse.access_token`, and similar properties, and force everyone to just use `OAuthTokenResponse.by_resource_server`?

@jaswilli, @corpulentcoffee Any strong feelings on the matter?

PR #162, opened 2017-03-23
Remove confusing top-level attrs of Token Response

sirosen:
After some thought about #158, I've become convinced that these attributes should be removed.
The silent "promotion" of scopes by Auth, the lack of any public specification as to the method of determination for that top-level token, and the inability of users to change what that token is by any means mean that these properties offer very little value and a lot of confusion.

`OAuthTokenResponse` top-level properties for accessing the top-level token aren't the desired/canonical/safe way to access tokens. We're telling people to always use `by_resource_server`, so we should cut down the interface for this class to that and only that.

Minor change included: make the dependent token response a subclass of the typical token response, only overriding a helper used by `__init__`.

This will require that we bump to `0.7.0` in the next release.

### DIFF
--- a/globus_sdk/auth/token_response.py
+++ b/globus_sdk/auth/token_response.py
@@ -34,15 +34,17 @@ class OAuthTokenResponse(GlobusHTTPResponse):
     """
     def __init__(self, *args, **kwargs):
         GlobusHTTPResponse.__init__(self, *args, **kwargs)
+        self._init_rs_dict()
 
+    def _init_rs_dict(self):
         # call the helper at the top level
         self._by_resource_server = {
-            self.resource_server: _convert_token_info_dict(self)}
+            self['resource_server']: _convert_token_info_dict(self)}
         # call the helper on everything in 'other_tokens'
         self._by_resource_server.update(dict(
             (unprocessed_item['resource_server'],
              _convert_token_info_dict(unprocessed_item))
-            for unprocessed_item in self.other_tokens))
+            for unprocessed_item in self['other_tokens']))
 
     @property
     def by_resource_server(self):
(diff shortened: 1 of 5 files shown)


bjmc:
We've always said the "top-levelness" is a meaningless artifact of wedging in extra tokens while still complying with the spec. :+1: for this change. The "top level" tokens are not special or different in any way.

sirosen:
I accidentally closed this by cleaning up all of the branches on my fork which were merged with it -- obviously including the branch itself. Whoops.

Sorry for the noise.

globus_sdk/auth/token_response.py
```diff
@@ -56,50 +58,6 @@ def by_resource_server(self):
...
-        """
-        return self['resource_server']
-
-    @property
-    def other_tokens(self):
-        """
-        The array of tokens other than the top-level one.
-        """
-        return self['other_tokens']
-
```
jaswilli: We definitely have some internal usage of these properties that will need to be refactored if they get dropped.

sirosen:
The update should pass tests.

The internal usage seems to be isolated to the `RefreshTokenAuthorizer`. That makes sense since the refresh tokens are a specific scenario in which we _know_ that there is one and only one token coming back on the response.

Outside of that, all of the changes necessary were made in the testsuite. Unfortunately, several tests simply had to be dropped because they were inspecting the exact attributes being removed here.

As an unfortunate side-effect, the `RefreshTokenAuthorizer` now needs to extract a single token -- with an unknown resource server name -- from a token response. The only way to do that, under this model, is to take `response.by_resource_server.values()[0]`. We happen to know that to be safe, but I wrapped it in a safety check just so that we raise a cleaner error if it ever gets broken.

PR #163, opened 2017-03-29
Change "identities_set" to "identity_set" for token introspection

mattias-lidman:
Per https://github.com/globusonline/globus-auth/pull/931/. Auth supports both, and eventually we'd like to deprecate the awkward `identities_set`. Do not merge until https://github.com/globusonline/globus-auth/pull/931 in deployed everywhere.


sirosen:
Looks like API support for the new name is out. Merging.

Issue #123, opened 2017-01-31
Implement Client Credentials Authorizer


sirosen:
I was working on an example of the Client Credentials grant when I realized that this is a possibility.

The `RefreshTokenAuthorizer` tracks expiration time and automatically requests new tokens for the user.
However, there are no automated utilities for managing tokens fetched via Client Credentials grants.

It occurred to me that we could define a `ClientCredentialsAuthorizer` built from a `ConfidentialAppAuthClient` and a set of scopes.
The scopes are necessary in order to specify what resource server (and scopes against it) are used -- the authorizer can fetch and store access tokens via the client credentials grant, always choosing the top-level token.
We could even have it error-out if `other_tokens` are returned (i.e. bad scopes supplied).

Does this have value? Would it help client authors?

stumartin:
Decide this is a "nice to have" for 1.0

aaschaer:
closed by #164

PR #164, opened 2017-03-30
Client credentials authorizer

aaschaer:
First pass at the ClientCredentialsAuthorizer requested in #123.

The vast majority of the logic for the ClientCredentialsAuthorizer was identical to the RefreshTokenAuthorizer, so I moved it into an abstract RenewingAuthorizer that both now implement.

One design question I ran into was how to handle scopes that cross multiple resource servers?
Right now I say its required that all scopes be for the same resource server and raise a GlobusError when we get back other_tokens. But this prevents using the DEFAULT_REQUESTED_SCOPES and requires a decent amount of oauth2 knowledge for using the authorizer.

Some potential solutions?:
Require a resource_server parameter along with the scopes, so we can use .by_resource_server like we suggest our users do. Could make this optional and default to auth or transfer.
Give a warning if other_tokens are found, but still keep using the top level token, and default to DEFAULT_REQUESTED_SCOPES.
Error explicitly if other_tokens are found, but make a new default set of scopes for this authorizer that only uses one resource server.

Unless of course its possible to have more than one Bearer token in the Authorization header, in which case I can just use all of the tokens, but I don't think the oauth2 spec allows that? (and if so could someone point me to the spec for those headers?)

Also, I would like to remove the option to pass an initial access token to these authorizers, unless there is a good reason to keep it that I'm missing? My issue is that if a user passes a valid initial access token and an invalid refresh token/confidential client, errors will only come up when the authorizer tries to get a new access token, which feels like it could get complaints of "why did it suddenly stop working".


sirosen:
This has conflicts now that #162 is merged. I also think there are some (unfortunate) reasons why you'll need to make some logical tweaks around the handling of the tokens you get back.

IIRC, Auth is upgrading all requests to include OIDC scopes, 'profile', 'email', 'openid', even if they aren't requested.
That means that even if someone passes `scopes='...transfer.api...'`, they'll get a two-token response. And, though I don't think this is a public promise, the auth.globus.org token will be the "top level" one.

These behaviors are part of the rationale for #162. There are some scenarios (e.g. refreshing a refresh token) when only one token _is_ returned. The results are inconsistent and confusing without internal knowledge of Auth, so the only safe thing to do is to use the `by_resource_server` representation and assume that there might be multiple tokens in every response.

> One design question I ran into was how to handle scopes that cross multiple resource servers?
Right now I say its required that all scopes be for the same resource server and raise a GlobusError when we get back other_tokens. But this prevents using the DEFAULT_REQUESTED_SCOPES and requires a decent amount of oauth2 knowledge for using the authorizer.

Others may differ, but I think this is the right basic approach. However, given that your scope set may be changed by the server, the error needs to be "when the provided scopes span multiple tokens in the response". It's a more laborious check than just `other_tokens` being nonempty.

The fact that it requires some more knowledge to use is okay. I think trying to make it too automagic may ultimately hurt the user in this case -- they'll need to gain some basic familiarity at some point.

> Require a resource_server parameter along with the scopes, so we can use .by_resource_server like we suggest our users do. Could make this optional and default to auth or transfer.

I'd be okay with this, but I don't think it's as good as deducing the RS from the scopes. If someone is passing scopes to this authorizer which span multiple resource servers, they've probably misunderstood something. I don't know that helping them select back into that data will actually help, and it will be an irritant for folks who _do_ know what they're doing.

> Unless of course its possible to have more than one Bearer token in the Authorization header, in which case I can just use all of the tokens, but I don't think the oauth2 spec allows that? (and if so could someone point me to the spec for those headers?)

It is not possible for spec-compliant servers. The whole idea of having multiple tokens is to not let one server see credentials intended for other resource servers -- doing this would violate the spirit of multiple tokens.

If you really want the spec, [RFC 6749](https://tools.ietf.org/html/rfc6749) is one of my _least favorite_ documents.
You should also feel free to ask the Auth devs questions like this: Auth is an opinionated OAuth2 service, so many things which are left discretionary or open by the spec are actually subject to choices that they've made.

> Also, I would like to remove the option to pass an initial access token to these authorizers, unless there is a good reason to keep it that I'm missing? My issue is that if a user passes a valid initial access token and an invalid refresh token/confidential client, errors will only come up when the authorizer tries to get a new access token, which feels like it could get complaints of "why did it suddenly stop working".

I don't mind dropping it for the `ClientCredentialsAuthorizer`, but it should stay for the `RefreshTokenAuthorizer`. It makes more sense for the latter than the former.

When you request refresh tokens, you get access tokens back alongside them. If the user can't pass in the initial access token, we're forcing an extra API round-trip, and making Auth mint an extra token.
Personally, I don't consider this a big deal, but we discussed it when we first added `RefreshTokenAuthorizer`s, and the consensus was that it was important to support and recommend a behavior which is the preferred OAuth2 client behavior.

That said, there isn't the same constraint on `ClientCredentialsAuthorizer`. There's no reason I can think of to force someone to do a client credentials grant by hand just to pass the result into the Authorizer.

bjmc:
> IIRC, Auth is upgrading all requests to include OIDC scopes, 'profile', 'email', 'openid', even if they aren't requested.

I think there's been some miscommunication. We're not going to be returning an access token (and id_token) for the auth resource server unless a client requests them. There was discussion that we might want to include the `preferred_username` claim in the id token, even if a client didn't ask for the `profile` scope (because that information is already available through the identities API without asking for any scopes).

We're not going to be giving you an extra token unless you want it. The `id_token` you asked for might potentially contain more details.

bjmc:
>  I don't think this is a public promise, the auth.globus.org token will be the "top level" one.

This is true. If an `auth.globus.org` is returned, it will always be the 'top level' one. This was done because the `id_token` contains an `at_hash` claim that can be used to validate the `access_token` returned alongside it. It would have been weird if the `id_token` at the top level contained a hash for one of the other access tokens buried in `other_tokens` and you had to find the matching one. It also would potentially have broken third-party clients expecting to easily validate against the top level token.

mattias-lidman:
>  There was discussion that we might want to include the preferred_username claim in the id token, even if a client didn't ask for the profile scope (because that information is already available through the identities API without asking for any scopes).

And to clarify further:
- this would be in the case you requested the `openid` scope, which gives you the `sub` field, from which you can deduce the username using public APIs. We're never going to start magically issue access/id_tokens for the Auth RS if none were requested.
- we decided not to do this right now. We may or may not do it at some point in the future.

sirosen:
> We're not going to be giving you an extra token unless you want it. The `id_token` you asked for might potentially contain more details.

Okay, sorry about that. It seems I misremembered [this internal issue about adding scopes to the legacy client support](https://github.com/globusonline/globus-auth/pull/869).

So, @aaschaer, I think that means that the nicest way to check for one-and-only-one token is just to check that `len(tokens.by_resource_server) == 1`.

I stand by the decision to do #162 , as the rules on what the top level token is are confusing. It's the Auth token if there is one, but otherwise it might be determined by the order of your requested scopes.
It seems to me that our general attitude is that all clients should be examining the resource server associated with each token, and that failing to do so is asking for trouble. Is that right? If that's the case, the SDK should have those "good behaviors" which we want for everyone.

bjmc:
The legacy support is a convoluted hack and the sooner we retire it, the better. It should have zero impact on the modern SDK clients.

Absolutely agree with #162. Don't rely on the top-levelness. My suggestion would be to pretend you've received a flat list: `tokens = response.pop('other_tokens') + [response]` and then go from there. If we could get away with returning a list without breaking 3rd party clients and spec compliance, we would.

mattias-lidman:
> Okay, sorry about that. It seems I misremembered this internal issue about adding scopes to the legacy client support.

Yeah, nothing about legacy anything should be used to draw conclusions about non-legacy anything.

> It seems to me that our general attitude is that all clients should be examining the resource server associated with each token, and that failing to do so is asking for trouble. Is that right?

Yes. We've always said that (multi-RS cognizant) clients should have no expectations at all about "top-level" vs. "other tokens", since the only reason we even have "other tokens" instead of just making it a flat list is for spec compliance, and the special rule that if there's an Auth token it will be the top-level follows from that, and @bjmc explained.

aaschaer:
As suggested authorizers that renew access tokens now check if tokens by_resource_server returns exactly one token and raise ValueErrors if this is untrue. I left this logic in the individual authorizers rather than the RenewingAuthorizer to give back more helpful error messages.

I still dislike that a valid access token and an invalid method of getting new ones will postpone errors till the original token expires, but if saving API calls is important for the RefreshTokenAuthorizer I'd prefer to be consistent in letting users pass in initial tokens.

globus_sdk/authorizers/client_credentials.py
```diff
@@ -0,0 +1,82 @@
...
+    Access Tokens using a ConfidentialAppAuthClient and a set of scopes to
+    fetch a new Access Token when the old one expires.
+
+    Example usage looks something like this:
+
+    >>> import globus_sdk
+    >>> confidential_client = globus_sdk.ConfidentiallAppAuthClient(
+        client_id=..., client_secret=...)
+    >>> cc_authorizer = globus_sdk.ClientCredentialsAuthorizer(
+    >>>     confidential_client)
```
sirosen: This needs to include the scopes requested.

globus_sdk/authorizers/renewing.py
```diff
@@ -0,0 +1,139 @@
...
+logger = logging.getLogger(__name__)
+# Provides a buffer for token expiration time to account for
+# possible delays or clock skew.
+EXPIRES_ADJUST_SECONDS = 60
+
+
+@six.add_metaclass(abc.ABCMeta)
+class RenewingAuthorizer(GlobusAuthorizer):
+    """
+    A ``RenwingAuthorizer`` is an abstract superclass to any authorizer
```
sirosen: nitpick: typo in `RenewingAuthorizer` name

tests/unit/test_refresh_token_authorizer.py
```diff
@@ -16,188 +14,58 @@ def setUp(self):
...
-        header_dict = {}
-        self.authorizer.expires_at = None
-
-        self.authorizer.set_authorization_header(header_dict)
-        # confirm value
-        self.assertEqual(header_dict["Authorization"],
-                         "Bearer " + self.rs_data['access_token'])
 
-    def test_handle_missing_authorization(self):
+    def test_multiple_resource_servers(self):
```
sirosen: I definitely should have added this test when changing this behavior.
Thanks for having my back! :+1:

sirosen:
Er, minor addendum to that review:
Could you also update `docs/authorization.rst` to add an autodoc directives for the two new classes?

Put the `RenewingAuthorizer` up with the base class, since it's abstract, please.

PR #175, opened 2017-04-07
Add AuthAPIError exception class

### LINKED ISSUE #138: invalid_client errors not being json parsed
The following error: {"error": "invalid_client"} is defaulting to text and logging a complaint about malformed json.

Not a huge issue, since defaulting to text doesn’t loose any information in this case, but we might want to edit GlobusAPIError._load_from_json() to handle these in the future.

sirosen:
Handles the case of Auth errors with a single top-level "error" key more gracefully. Also gives Auth errors their own class.
Although not all of Auth is an API, the parts used by the SDK are, so the class name is okay.

Resolves #138 

Because we're under the gun for GW on Tuesday, I'm not trying to think through tests for this right now. I know it's bad not to have them, but limited time -> limited quality.

### DIFF
--- a/globus_sdk/exc.py
+++ b/globus_sdk/exc.py
@@ -1,5 +1,6 @@
 import logging
 import textwrap
+import six
 import requests
 
 logger = logging.getLogger(__name__)
@@ -131,6 +132,30 @@ class TransferAPIError(GlobusAPIError):
         self.code = data["code"]
         self.message = data["message"]
         self.request_id = data["request_id"]
+
+
+class AuthAPIError(GlobusAPIError):
+    """
+    Error class for the API components of Globus Auth.
+
+    Customizes JSON parsing.
+    """
+    def _load_from_json(self, data):
+        """
+        Load error data from a JSON document.
+
+        Sets `code` statically because Auth doesn't have API error codes, and
+        looks for a top-level "error" attribute.
+        """
+        self.code = "Error"
+        if "message" in data:
+            self.message = data["message"]
+        elif "detail" in data:
+            self.message = data["detail"]
+        elif "error" in data and isinstance(data["error"], six.string_types):
+            self.message = data["error"]
+        else:
+            self.message = "no_extractable_message"
 
 
 class InvalidDocumentBodyError(GlobusError):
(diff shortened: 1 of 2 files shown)


globus_sdk/exc.py
```diff
@@ -133,6 +134,30 @@ def _load_from_json(self, data):
...
+    Customizes JSON parsing.
+    """
+    def _load_from_json(self, data):
+        """
+        Load error data from a JSON document.
+
+        Sets `code` statically because Auth doesn't have API error codes, and
+        looks for a top-level "error" attribute.
+        """
+        self.code = "Error"
```
aaschaer: Looks like the tests are failing on this static setting of self.code.

I would expect something like the following to pass tests, but if the tests themselves should be changed I can do that.

`if "code" in data:`
    `self.code = data["code"]`
`else:`
    `self.code = "Error"`

sirosen: Hang on, how was code being set before? Maybe the prior failover is better?

sirosen: Judging by the tests, this was being set to the Requests `reason` field? I don't see that done explicitly though, and it makes me nervous.
@aaschaer, would you be able to update the tests? That would be great.

aaschaer: Yeah, I'll update them and make a PR to your repo.

bjmc: We do have error codes: https://github.com/globusonline/globus-auth/blob/integration/globus/auth/error/codes.py

sirosen: Ah, that explains it! I saw things that looked like the `reason`, because that's how a lot of the codes read.

@aaschaer, please scratch that work. I'll fix my PR to not assume that `code` is absent, as you suggested originally.

aaschaer:
Changing all instances of GlobusAPIError to AuthAPIError in the Auth unit tests broke nothing, so I'm fine with merging this as is. I can add individual unit tests as well later.

globus_sdk/exc.py
```diff
@@ -133,6 +134,40 @@ def _load_from_json(self, data):
...
+    """
+    Error class for the API components of Globus Auth.
+
+    Customizes JSON parsing.
+    """
+    def _load_from_json(self, data):
+        """
+        Load error data from a JSON document.
+
+        Sets `code` statically because Auth doesn't have API error codes, and
```
bjmc: Minor: This comment is outdated/inaccurate.

PR #185, opened 2017-04-13
Support iterable requested_scopes everywhere

### LINKED ISSUE #184: Make requested_scopes accept (in all contexts) an iterable of strings
Right now, we're requiring that people give us a space-delimited string.

Many people have pointed out that it's awkward / uncomfortable.
Wouldn't it be great if we could take tuples and lists? Yes, yes it would.

This shouldn't be too hard.

sirosen:
You can pass an iterable like a list of strings and we'll join with spaces. Applies to both forms of 3-legged oauth flow, and to client credentials grants.
Includes several tests to ensure that it's all working, and a round of documentation updates to make sure people know it's there.

Closes #184

### DIFF
--- a/globus_sdk/auth/client_types/confidential_client.py
+++ b/globus_sdk/auth/client_types/confidential_client.py
@@ -86,12 +86,43 @@ class ConfidentialAppAuthClient(AuthClient):
             self, redirect_uri, requested_scopes=None,
             state='_default', refresh_tokens=False):
         """
-        Starts an Authorization Code OAuth2 flow by instantiating a
+        Starts or resumes an Authorization Code OAuth2 flow.
+
+        Under the hood, this is done by instantiating a
         :class:`GlobusAuthorizationCodeFlowManager
         <globus_sdk.auth.GlobusAuthorizationCodeFlowManager>`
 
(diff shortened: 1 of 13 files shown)


globus_sdk/auth/client_types/confidential_client.py
```diff
@@ -72,6 +74,9 @@ def oauth2_client_credentials_tokens(self, requested_scopes=None):
         """
         self.logger.info('Fetching token(s) using client credentials')
         requested_scopes = requested_scopes or DEFAULT_REQUESTED_SCOPES
+        # convert scopes iterable to string immediately on load
+        if not isinstance(requested_scopes, six.string_types):
```
bjmc: Consider `hasattr(requested_scopes, '__iter__')` for duck-typing, as opposed to `isinstance`-checking.

corpulentcoffee: A string is iterable too, unfortunately.

bjmc: Under Python2, strings are iterable but they don't have the `__iter__` method. Apparently that's changed in Python3.

PR #186, opened 2017-04-13
AuthAPIError tests

aaschaer:
Updated some existing tests to use the AuthAPIError class, and added a few unit tests to test_exc.py


sirosen:
More tests. Music to my ears. Merging.

PR #190, opened 2017-04-17
Fix docs references to oauth2_start_flow_*

sirosen:
Dan pointed out that these docs are using an out-of-date name for this
method from pre-1.0 release versions.


aaschaer:
Looks good to me. Merging.

PR #191, opened 2017-04-18
Add endpoint_manager_task_successful_transfers, and unit tests

### LINKED ISSUE #189: Add endpoint_manager Task Successful Transfers method to TransferClient
We have a method for Task Successful Transfers ( #93 ), but we need one for the `endpoint_manager` API.

Transfer API docs for this:
https://docs.globus.org/api/transfer/advanced_endpoint_management/#get_task_successful_transfers_as_admin

aaschaer:
Resolves #189 

Adds endpoint_manager_task_successful_transfers() to transfer client
Updates the testing framework to add user info and a transfer refresh token for sdktester2b
And adds unit tests for the three endpoint_manager_* methods in transfer client

### DIFF
--- a/tests/unit/test_transfer_client.py
+++ b/tests/unit/test_transfer_client.py
@@ -1,3 +1,4 @@
+import re
 from random import getrandbits
 from datetime import datetime, timedelta
 
@@ -5,7 +6,8 @@ import globus_sdk
 from tests.framework import (CapturedIOTestCase,
                              get_client_data, get_user_data,
                              GO_EP1_ID, GO_EP2_ID,
-                             SDKTESTER1A_NATIVE1_TRANSFER_RT)
+                             SDKTESTER1A_NATIVE1_TRANSFER_RT,
+                             SDKTESTER2B_NATIVE1_TRANSFER_RT)
 from globus_sdk.exc import TransferAPIError
 from globus_sdk.transfer.paging import PaginatedResource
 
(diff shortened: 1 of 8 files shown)


tests/unit/test_transfer_client.py
```diff
@@ -1412,3 +1411,172 @@ def test_delete_endpoint_acl_rule(self):
...
+            task_id)
+
+        # confirm results
+        self.assertIsInstance(success_doc, PaginatedResource)
+        count = 0
+        for transfer in success_doc:
+
+            self.assertEqual(transfer["DATA_TYPE"], "successful_transfer")
+            self.assertIsNotNone(re.match(dest_path + "file[1-3].txt",
+                                 transfer["destination_path"]))
```
sirosen: This will be my one and only nitpick with this PR: `flake8` doesn't seem to fail on this, but it should.

The indentation doesn't match because `transfer["destination_path"]` is an argument to `re.match`.

sirosen:
I have one arguably unimportant cosmetic quibble, but the rest of this checks out okay.

One concern is that we might get caught looping on EP deletions with the way you're handling it. I'm not sure that this is actually possible, and don't care that much if it is. Worst case, the tests lock up until terminated manually. That's not actually that bad.

aaschaer:
Fixed indentation and squashed the cleanup commit.

Issue #198, opened 2017-04-19
Add endpoint manager endpoint acl list method to transfer client


aaschaer:
[API docs](https://docs.globus.org/api/transfer/advanced_endpoint_management/#get_endpoint_access_list_as_admin)

Path: GET /endpoint_manager/endpoint/<endpoint_id>/access_list

Proposed client method name: endpoint_manager_acl_list

aaschaer:
Closed via #199

Issue #193, opened 2017-04-19
Add endpoint_manager get endpoint to transfer client


aaschaer:
[API Docs](https://docs.globus.org/api/transfer/advanced_endpoint_management/#mc_get_endpoint)

Path: GET /endpoint_manager/endpoint/<endpoint_id>

Proposed client method name: endpoint_manager_get_endpoint()

aaschaer:
Closed via #199

PR #199, opened 2017-04-19
Add endpoint_manager/endpoint methods

### LINKED ISSUE #192: Add endpoint_manager hosted endpoint list to transfer client
[API Docs](https://docs.globus.org/api/transfer/advanced_endpoint_management/#get_hosted_endpoint_list>)

Path: GET /endpoint_manager/endpoint/<endpoint_id>/hosted_endpoint_list

Proposed client method name: endpoint_manager_hosted_endpoint_list()

aaschaer:
Resolves #192 #193 and #198 

One naming question was weather to use `endpoint_manager_endpoint_acl_list` or `endpoint_manager_acl_list`. I went with the latter since the former seemed redundant, but then it doesn't match the non-managed `endpoint_acl_list` method?

### DIFF
--- a/globus_sdk/transfer/client.py
+++ b/globus_sdk/transfer/client.py
@@ -1396,15 +1396,90 @@ class TransferClient(BaseClient):
 
     def endpoint_manager_monitored_endpoints(self, **params):
         """
+        Get endpoints the current user is a monitor or manager on.
+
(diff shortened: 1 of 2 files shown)


sirosen:
I think your choice of naming is fine.

PR #203, opened 2017-04-20
Add endpoint_manager pause_rule methods and unit tests

### LINKED ISSUE #197: Add endpoint_manager pause rule methods to transfer client
Path: GET /endpoint_manager/pause_rule_list [?filter_endpoint=ENDPOINT_ID]
[API Docs](https://docs.globus.org/api/transfer/advanced_endpoint_management/#get_pause_rules)
Proposed method name: endpoint_manager_pause_rule_list

Path: POST /endpoint_manager/pause_rule
[API Docs](https://docs.globus.org/api/transfer/advanced_endpoint_management/#create_pause_rule)
Proposed method name: endpoint_manager_create_pause_rule

Path: GET /endpoint_manager/pause_rule/<pause_rule_id>
[API Docs](https://docs.globus.org/api/transfer/advanced_endpoint_management/#get_pause_rule)
Proposed method name: endpoint_manager_get_pause_rule

Path: PUT /endpoint_manager/pause_rule/<pause_rule_id>
[API Docs](https://docs.globus.org/api/transfer/advanced_endpoint_management/#update_pause_rule)
Proposed method name: endpoint_manager_update_pause_rule

Path: DELETE /endpoint_manager/pause_rule/<pause_rule_id>
[API Docs](https://docs.globus.org/api/transfer/advanced_endpoint_management/#delete_pause_rule)
Proposed method name: endpoint_manager_delete_pause_rule

aaschaer:
Resolves #197 

Once this is merged I'll go back and test things that were blocked on pausing.

### DIFF
--- a/globus_sdk/transfer/client.py
+++ b/globus_sdk/transfer/client.py
@@ -1817,8 +1817,7 @@ class TransferClient(BaseClient):
     # endpoint manager pause rule methods
     #
 
-    def endpoint_manager_pause_rule_list(self, filter_endpoint=None,
-                                         filter_host_endpoint=None, **params):
+    def endpoint_manager_pause_rule_list(self, filter_endpoint=None, **params):
         """
         Get a list of pause rules on endpoints that the current user has the
         activity monitor effective role on.
@@ -1851,8 +1850,7 @@ class TransferClient(BaseClient):
         self.logger.info(("TransferClient.endpoint_manager_"
                           "pause_rule_list(...)"))
         path = self.qjoin_path("endpoint_manager", "pause_rule_list")
-        merge_params(params, filter_endpoint=filter_endpoint,
-                     filter_host_endpoint=filter_host_endpoint)
+        merge_params(params, filter_endpoint=filter_endpoint)
         return self.get(path, params=params,
                         response_class=IterableTransferResponse)
 
--- a/globus_sdk/transfer/client.py
+++ b/globus_sdk/transfer/client.py
@@ -1812,3 +1812,175 @@ class TransferClient(BaseClient):
         }
         path = self.qjoin_path("endpoint_manager", "admin_resume")
         return self.post(path, json_body=json_body, params=params)
+
+    #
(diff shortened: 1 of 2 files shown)


sirosen:
The TransferClient tests are getting awfully long -- we might want to try to segment those tests into a few files/modules just for ease of management/perusal.

This looks good, assuming that tests pass.

aaschaer:
Ah, I forgot to merge params in pause_rule_list.

I'm slightly disappointed in myself for this, but also proud of myself for writing tests that caught my mistake.

aaschaer:
2 of the failures were a race condition I assumed would never happen in an unrelated test and a connection error.

The third however is still violating my assertion that only one pause rule will be created per shared endpoint in the pause rule tests.
Either the filter parameter still isn't working, or something has gone terribly wrong. I'll see if I can figure out which.

aaschaer:
Ah, I think I was somehow still looking at the old failures (maybe caused by my rebase?) since now there is only one failure.

I'll test some more locally, but I think the problem is fixed now?

aaschaer:
Upon closer inspection, it seems the filter_host_endpoint param is either being ignored by the API, or not behaving as expected (according to the docs I should have been getting back an empty list). 

I removed the named parameter, but if it is behaving properly for actual host endpoint managers it can still be passed through **params.

I'll add this to the list of small bugs / documentation inconsistencies with the APIs I've run into testing endpoint_manager calls. Where should I make issues to track those?

sirosen:
> I'll add this to the list of small bugs / documentation inconsistencies with the APIs I've run into testing endpoint_manager calls. Where should I make issues to track those?

I'd say that you should file these directly against Transfer, and that way we can figure out if they need to be fixed there (underlying bug) or just in the docs (misdocumented feature / outdated docs). If we go straight to the docs, it may get fixed to reflect the actual state of the system even if we don't really want it that way.

Please do so in the language of "GET against /foo/bar", rather than the language of "SDK call TransferClient.foo(bar)".

PR #208, opened 2017-05-08
Better handle UTF-8 inupts

### LINKED ISSUE #207: non ASCII compatible UTF-8 inputs break on python 2.7
When looking into / writing tests for https://github.com/globus/globus-cli/issues/65, I ran into some `UnicodeEncodeError: 'ascii' codec can't encode characters` errors from the SDK when using python 2.7

It seems that python 2.7 breaks when it tries to use the default ascii codec to deal with the type inconsistency between string literals being byte strings and `safe_stringify` always returning unicode.

The simplest solution I could think of was to import `unicode_literals` from `__future__` in modules that use `safe_stringify` to make the necessary internal strings be unicode regardless of python version, but was wondering if there was a better solution? I think `safe_stringify` was originally intended to output byte strings for python 2, which might also solve this problem?

aaschaer:
Fixes #207 and adds integration tests that test operations using non url-safe ascii names, non-ascii utf-8 names, and non-utf-8 latin-1 names.

It seems latin-1 works fine as long as it doesn't need to be decoded from a byte string internally. I attempted to do what the APIs do and fall back to latin-1 if decoding with utf-8 fails, but this caused errors in the json library and got a 500 back from transfer, so I figured it best to only support decoding utf-8 for now.

Should we raise some kind of `GlobusError` when `safe_stringify` fails on an input, or would this mask the original error too much?

### DIFF
--- a/globus_sdk/auth/client_types/base.py
+++ b/globus_sdk/auth/client_types/base.py
@@ -1,4 +1,5 @@
 from __future__ import print_function
+from __future__ import unicode_literals
 
 import six
 import collections
--- a/globus_sdk/base.py
+++ b/globus_sdk/base.py
@@ -1,3 +1,4 @@
+from __future__ import unicode_literals
 import json
 import logging
 
--- a/globus_sdk/transfer/client.py
+++ b/globus_sdk/transfer/client.py
@@ -1,3 +1,4 @@
+from __future__ import unicode_literals
 import logging
 import time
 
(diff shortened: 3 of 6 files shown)


sirosen:
Just spoke with Aaron to clarify what's going on here. To summarize: calling `<bytestr>.format(<unicode>)` blows up, make our own format strings unicodes to fix.

> Should we raise some kind of `GlobusError` when `safe_stringify` fails on an input, or would this mask the original error too much?

I wouldn't bother for now. If someone passes a non-string thing to a function that, by contract, consumes strings, blowing up with a `ValueError` or somesuch seems reasonable. And, as you indicate, it could mask the actual cause.

Merging.

PR #209, opened 2017-05-09
Upgrade pip in Makefile

jaswilli:
As seen with the release of `requests` 2.14, old versions of pip can cause problems. Since we're still using one of Ubuntu 12.04 containers on Travis we get a _really_ old version of pip.

**update:** and `setuptools`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,13 +7,13 @@ python:
   - "3.5"
   - "3.6"
   # allow failures on CPython dev and pypy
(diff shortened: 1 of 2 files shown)


jaswilli:
New plan based on all the various constraints we've run into:

* New make target, `travis` which will not use a virtualenv.
* Install the latest pip and setuptools, except for in the `travis` target, where we install the latest version less than 30, since we need to support the version of pypy3 that Travis makes available.

PR #214, opened 2017-06-02
Recursive operation_ls

aaschaer:
Needed for https://github.com/globus/globus-cli/issues/46. Unless we want to just wait for endpoint level support, in which case I'll just add this to the CLI.

Uses BFS, rate-limiting, and dosen't buffer the whole response in memory.

I decided the simplest way to handle non-buffered output was to make a `RecursiveLsResponse` response class that inherits from `PaginatedResource` so we can return an iterator.

Since this gets rid of access to top level fields and changed the response class I made this a separate `recursive_operation_ls` method in the transfer client rather than requiring `recursive=True` for the normal `operation_ls` method. I also re-added the top level path field to each item.

The `filter_after_first` paramater is to help the CLI match the behavior of a batch ls while using filter to simulate globbing.

### DIFF
--- a/globus_sdk/transfer/client.py
+++ b/globus_sdk/transfer/client.py
@@ -7,7 +7,8 @@ from globus_sdk.base import BaseClient, merge_params, safe_stringify
 from globus_sdk.authorizers import (
     AccessTokenAuthorizer, RefreshTokenAuthorizer, ClientCredentialsAuthorizer)
 from globus_sdk.transfer.response import (
-    TransferResponse, IterableTransferResponse, ActivationRequirementsResponse)
+    TransferResponse, IterableTransferResponse, ActivationRequirementsResponse,
+    RecursiveLsResponse)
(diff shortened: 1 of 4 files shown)


sirosen:
@aaschaer, I feel a little bit uncomfortable about this going into the SDK.

The implementation is really solid, and looks to me totally satisfactory. However, when endpoints start to natively support recursive `ls`, we'll have two implementations to deal with. There's a significant potential for confusion, especially since they'll almost certainly have slightly different signatures and even potentially different supported behaviors.

However, given that this response type and implementation seem complete, how would you feel about adding the operation, as implemented here, to the `RetryingTransferClient`? That seems like an easy way to port this into the CLI and keep it well-isolated in the `globus_cli.services` subpackage.
The existing `globus ls -r` is a bit scary, as discussed elsewhere, so making it safer seems like a worthwhile expenditure of effort.

aaschaer:
Sounds good to me, I'll go ahead and close this PR then.

PR #217, opened 2017-07-05
Add explicit documentation for PaginatedResources

### LINKED ISSUE #215: How should we improve PaginatedResource documentation to clarify that second iteration is empty? Are new/additional behaviors desirable or required?
In various forms, we've seen confusion over the fact that a PaginatedResource does not record the results of paging and yields the results as it gets them.
Usually, this is manifest in the form of someone trying to walk a PaginatedResource multiple times, and being surprised that the second iteration is empty.

So, first and foremost:
How do we make this clearer in the docs?

Secondly, which, if any, of the following behaviors would help resolve the issue?
1. Have paginated calls accept `store_results=True`, enabling the use of a list to record the results as they are iterated
2. Add a method, `PaginatedResource.reset()` which restarts paging, and which can document that it's discouraged but useful for fetching fresh results
3. On repeat iterations (i.e. after the first `StopIteration`), have `PaginatedResource` _raise an error_ which explains how to do multiple iterations correctly (either listify, or make a fresh call)
4. On repeat iterations, have `PaginatedResource` restart paging, even though it's inefficient (optionally log a warning)

EDIT: updated to numbered list to make things easier to discuss.

sirosen:
Previously, the docs just said "an iterable of GlobusResponses", which led to confusion by consumers between `IterableTransferResponse` and `PaginatedResource`.

Now, call out the use of `PaginatedResource` explicitly, and document the fact that it acts as a typical iterator only allowing callers to walk its contents once.

Closes #215 

Thoughts on how this is done? Directing people to the `PaginatedResource` doc feels a bit like throwing them to the wolves -- it's very much an internal class and documented as such -- but it's maybe better than doing nothing?

### DIFF
--- a/globus_sdk/transfer/paging.py
+++ b/globus_sdk/transfer/paging.py
@@ -16,24 +16,27 @@ class PaginatedResource(GlobusResponse, six.Iterator):
 
     This is a class and not a function because it needs to enforce distinct
     actions between initialization and iteration. If defined as a generator
-    function with the `yield` syntax, python won't let us distinguish between
+    function with the ``yield`` syntax, python won't let us distinguish between
     creating the iterable object and iterating it.
     To eagerly trigger errors from the first call, we need to wrap it up in a
     class.
 
     Expectations about Paginated Transfer API Resources:
-    - They support `limit` and `offset` query params, with `limit` being a
-      count of elements to return, and offset being an offset into the result
+
+    - They support ``limit`` and ``offset`` query params, with ``limit`` being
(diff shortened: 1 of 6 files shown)


globus_sdk/transfer/paging.py
```diff
@@ -16,24 +16,27 @@ class PaginatedResource(GlobusResponse, six.Iterator):
```
aaschaer: Somewhere in this helptext I think we should re-emphasize that the iterator can only be walked once and that each iteration makes a network-call.

PR #218, opened 2017-07-05
Add symlink support to sdk

### LINKED ISSUE #216: Symlink Support
Here is my proposal for how we should add Transfer's upcoming support of symlinks to the SDK:

**Task submission:**

Add `recursive_symlink="ignore"` to `TransferData.__init__`'s named arguments, then add it as one of the `TransferData`'s query paramaters.

SDK will not enforce that the value is in "ignore", "keep", or "copy" to stay in our model of giving back the errors Transfer gives us.

Add `add_symlink_item(self, source_path, destination_path)` to `transferData`'s methods, akin to `TransferDat.add_item` that adds a `transfer_symlink_item` to the DATA field of the submission data.

(could also add `symlink=False` to the existing `add_item` method, but a seperate method avoids the issue of `recursive` and `symlink` both being True.)

No code changes to `DeleteData` `TransferClient.submit_transfer` or `TransferClientsubmit_delete` should be needed.

**Task Monitoring:**

No code changes should be needed for task monitoring as we just return a `TransferResponse` that will now happen to have a few more data fields.

**Symink Creation:**

add an `operation_symlink(self, endpoint_id, symlink_target, path, **params)` method to `TransferClient`.

That makes a POST request to /operation/endpoint/<endpoint_id>/symlink with the given request parameters and returns a `TransferResponse`.

aaschaer:
Resolves #216 and is implemented as described in the issue.

This was passing tests against sandbox, but I would feel a bit safer waiting to merge this until there is a tutorial endpoint that supports symlinks on production so I can add my unit tests to this PR and make sure everything is working as expected. (That said, the CLI supporting symlinks is blocked on this being released so we might want to go ahead and get this out for that reason?)

### DIFF
--- a/globus_sdk/transfer/client.py
+++ b/globus_sdk/transfer/client.py
@@ -980,9 +980,43 @@ class TransferClient(BaseClient):
         resource_path = self.qjoin_path("operation/endpoint", endpoint_id,
                                         "rename")
         json_body = {
-            'DATA_TYPE': 'mkdir',
+            'DATA_TYPE': 'rename',
             'old_path': oldpath,
             'new_path': newpath
+        }
+        return self.post(resource_path, json_body=json_body, params=params)
+
+    def operation_symlink(self, endpoint_id, symlink_target, path, **params):
+        """
+        ``POST /operation/endpoint/<endpoint_id>/symlink``
+
+        :rtype: :class:`TransferResponse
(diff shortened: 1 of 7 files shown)


sirosen:
I made a minor clarifying adjustment to the docstring, which is the only issue I found with this work. Merging (even though builds haven't passed yet).

PR #223, opened 2017-07-25
Add example to docs of retrying task submission

sirosen:
Add a case to the `advanced_transfer` examples, in which we define a fairly generic retry handler, with some healthy logging, and handle `NetworkError, GlobusAPIError` well. Shows an application of that
handler to the TransferData structure submitted via TransferClient.submit_transfer. The whole example aims for minimal exposition.
It will read comfortably to anyone familiar with the idea of defining functions on-the-fly, but very novice developers will likely struggle with it (which is why it's "advanced").

### DIFF
--- a/docs/examples/advanced_transfer.rst
+++ b/docs/examples/advanced_transfer.rst
@@ -48,3 +48,122 @@ as in
     submission_data = globus_sdk.TransferData(
         transfer_client, source_endpoint_uuid, dest_endpoint_uuid,
         deadline=str(future_1minute))
+
+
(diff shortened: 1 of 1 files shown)


aaschaer:
Looks good to me.

I was able to get expected results with the sample code, and the explanations all made sense to me, although I probably would have had  `retry_globus_function` take `*args` and `**kwargs` to pass to `func` rather than explain why we needed to create a function with no arguments to pass to the retry handler.

I'll go ahead and merge without waiting for AppVoyer since this is all doc changes.

PR #229, opened 2017-10-03
Update doc string for add_endpoint_acl_rule

aaschaer:
Updating doc string as requested by a user on developer discuss.

Original request:

> Would you please update the globus sdk documentation to clarify that the path specified when creating an acl on a shared endpoint is relative to the host_path of that endpoint?


sirosen:
I didn't see the request, but this change is reasonable enough. Merging.

PR #231, opened 2017-10-04
Expand doc for TransferData to detail parameters

sirosen:
Specifies parameter types, defaults, and meaning. Full understanding of behaviors will probably always require reading the API docs, but this should eliminate confusion about things like Endpoint Documents (dicts) and Endpoint IDs (strings).

I want to do another round of this to document `DeleteData` once we're happy with the results here.

### DIFF
--- a/globus_sdk/transfer/data.py
+++ b/globus_sdk/transfer/data.py
@@ -1,6 +1,6 @@
 """
 Data helper classes for constructing Transfer API documents. All classes should
-extend ``dict``, so they can be passed seemlesly to
+extend ``dict``, so they can be passed seamlessly to
 :class:`TransferClient <globus_sdk.TransferClient>` methods without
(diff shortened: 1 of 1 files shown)


globus_sdk/transfer/data.py
```diff
@@ -1,6 +1,6 @@
 """
 Data helper classes for constructing Transfer API documents. All classes should
-extend ``dict``, so they can be passed seemlesly to
+extend ``dict``, so they can be passed seamlesly to
```
jaswilli: needs more `s`

sirosen: Haha, yes, typo fixing should actually fix the typos...

sirosen:
Amended to fix a typo which Jason caught and remove a trailing space which made linting fail.

PR #233, opened 2017-10-10
Expand docstring for DeleteData and fix typos

sirosen:
Add a docstring for DeleteData which explains the various parameters (like TransferData). Fixes several minor typos in TransferClient as well.


aaschaer:
Looks good to me. I might suggest giving an example ISO-8601 timestamp string, but I feel like users who aren't familiar with those will probably stick with datetime objects anyways.

sirosen:
Added some ISO-8601 examples. I think it's a good call to do so, even if most people use datetime.

PR #234, opened 2017-10-12
Retry NetworkErrors in testsuite

sirosen:
It's quite common to get spurious failures in the testsuite because one GET somewhere along the line times out. These should all be retried, but we don't want to put painstaking retry logic into each testcase.

Add a decorator which sets up retry behavior to the tests.framework package, and decorate every network-sensitive test with it. It is possible to write a test such that the retry will fail, potentially worse than it would have failed without the retry, but (1) that's not the common case and (2) you can use `retry_errors` on a locally defined function to restrict its behavior if you're writing a sensitive test.

Although this is pretty much universally applied, it should be a per-test decision, so keep the decorator pattern very simple to shuffle around and re-apply (at the cost of some redundancy).

This worked for me locally, but I'm concerned that across pythons and platforms we might see some weirdness.

### DIFF
--- a/tests/framework/__init__.py
+++ b/tests/framework/__init__.py
@@ -1,7 +1,8 @@
 from tests.framework.capturedio_testcase import CapturedIOTestCase
 from tests.framework.transfer_client_testcase import TransferClientTestCase
 from tests.framework.tools import (get_fixture_file_dir,
-                                   get_client_data, get_user_data)
+                                   get_client_data, get_user_data,
+                                   retry_errors)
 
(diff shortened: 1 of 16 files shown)


aaschaer:
Might have been nice if this were part of `CapturedIOTestCase` to prevent needing a decorator for every single test function that makes network calls, but the only way I can think of doing that off the top of my head would be to override `CapturedIOTestCase`'s `__getattribute__` with a retry wrapper, and that seems a little janky.

Since you already went to the effort of adding the decorators and we really need some way to deal with Networ Errors, this looks good to merge to me.

PR #241, opened 2017-11-01
Allow passing a base_url to BaseClient objects

aaschaer:
Minor change motivated by globus-dns wanting to pass the auth_url it already has in its own config instead of taking another step to create a redundant /etc/globus.cfg file.


sirosen:
Look good. Assuming that this passes, I'll merge.
@aaschaer, want to look at #239, and I can push out a release containing this fix once both merge?

Issue #252, opened 2017-12-02
decoding jwt requires cryptography


vsoch:
The `id_token` that can be decoded with jwt is great for getting user info without needing to do a second call to user info, but when doing the decode it is missing an algorithm:

```
jwk_as_pem = jwt.algorithms.RSAAlgorithm.from_jwk(
uwsgi_1   | AttributeError: module 'jwt.algorithms' has no attribute 'RSAAlgorithm'
``` 

I was able to fix this by installing cryptography. Should this be a dependency for the globus-sdk-python application or third parties that use it?

vsoch:
Oh and there is no huge issue / anything blocking what I'm working on, I've added the dependency to my application, but wanted to discuss which strategy made most sense. Thank you!

bjmc:
The `jwt` features are [listed as a optional dependency](https://github.com/globus/globus-sdk-python/blob/master/setup.py#L32), because some people might want to use the SDK without them (particularly if they're in an environment where `cryptography` is difficult to build)

If you install the SDK using `pip install globus-sdk[jwt]` it should pull in `cryptography` for you as a dependency automatically.

Maybe we need to do a better job of highlighting this in the documentation.

bjmc:
Also, thanks for your report! It's great getting feedback from developers on ways we can improve and make things easier.

vsoch:
Oh fantastic! I'll use the `[jwt]` module addin instead of installing crytography separately. And +1 to tweaking the docs. They are really well written/detailed, and a few helper "entrypoints" might lead to the right place.

jaswilli:
Hi @vsoch,

Also, the SDK provides a convenience method for handling the the `id_token` so you don't need to mess with the JWT library directly.

Once you have the `cryptography` dependency installed (as discussed, `globus_sdk[jwt]`) you can call the `decode_id_token` function against a token response and get the contents of the `id_token`, e.g.:

```python
import globus_sdk

# ... set up and make a token request that includes an id_token

id_token = token_response.decode_id_token(globus_sdk.AuthClient())
```

[Link to the docs](http://globus-sdk-python.readthedocs.io/en/stable/optional_dependencies/#oidc-id-tokens)

vsoch:
Thanks! Yeah that one I figured out, very useful!

https://github.com/singularityhub/sregistry/blob/add/globus/shub/plugins/globus/utils.py#L89

PR #255, opened 2017-12-04
Add reference from Responses to originating Client

sirosen:
Motivated largely by #252 , where an otherwise clean explanation of `decode_id_token` seemed encumbered a little bit by the ugly detail of having to pass an `AuthClient`.
This has also always been a bit of a wart on the SDK, and maybe it's time that we cleaned it up a bit.

Give every GlobusResponse object a ref to the client object which created it. Use this newfound power to simplify the OAuthTokenResponse.decode_id_token method to not require that you pass the client object back into it (which is confusing for both naive and experienced users).
This may have implications for other response classes in the future.

Updates in only a few places: the response definitions which don't consume `*args, **kwargs`, and the base client _request method (which instantiates the response objects).

This is carefully compatible with current usage, but deprecates passing `auth_client` into `decode_id_token` and starts logging a warning under that usage.

Also improves docstring for `decode_id_token`.

### DIFF
--- a/globus_sdk/auth/token_response.py
+++ b/globus_sdk/auth/token_response.py
@@ -58,9 +58,19 @@ class OAuthTokenResponse(GlobusHTTPResponse):
         """
         return self._by_resource_server
 
-    def decode_id_token(self, auth_client):
+    def decode_id_token(self, auth_client=None):
         """
         A parsed ID Token (OIDC) as a dict.
+
+        Requires the globus-sdk[jwt] optional dependency. For more details, see
+        the :ref:`Optional Dependency Docs <optional_dependencies>`.
+
+        **Parameters**
+
+            ``auth_client`` (:class:`AuthClient <globus_sdk.AuthClient>`)
+              Deprecated parameter for providing the AuthClient used to request
+              this token back to the OAuthTokenResponse. The SDK now tracks
+              this internally, so it is no longer necessary.
         """
         logger.info('Decoding ID Token "{}"'.format(self['id_token']))
         try:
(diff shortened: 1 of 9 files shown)


sirosen:
All of the tests which directly instantiate response objects need to update, which is where this is going wrong. I'll amend to fix.

globus_sdk/auth/token_response.py
```diff
@@ -58,9 +58,19 @@ def by_resource_server(self):
         """
         return self._by_resource_server
 
-    def decode_id_token(self, auth_client):
+    def decode_id_token(self, auth_client=None):
         """
         A parsed ID Token (OIDC) as a dict.
+
+        Requires the globus-sdk[jwt] optional dependency. For more details, see
+        the :ref:`Optional Dependency Docs <optional_dependencies>`.
```
jaswilli: We can drop this note now.

globus_sdk/response.py
```diff
@@ -15,13 +15,16 @@ class GlobusResponse(object):
...
+    return to a caller. Most basic actions on a GlobusResponse are
+    just ways of interacting with this data.
     """
-    def __init__(self, data):
-        """
-        ``GlobusResponse`` objects *always* wrap some kind of data to
-        return to a caller. Most basic actions on a GlobusResponse are
-        just ways of interacting with this data.
-        """
+    def __init__(self, client, data):
```
jaswilli: I think this is a breaking change unless we're going to try and assert that `GlobusResponse` is somehow private.

globus_sdk/response.py
```diff
@@ -79,10 +82,10 @@ class GlobusHTTPResponse(GlobusResponse):
     :ivar http_status: HTTP status code returned by the server (int)
     :ivar content_type: Content-Type header returned by the server (str)
     """
-    def __init__(self, http_response):
+    def __init__(self, client, http_response):
```
jaswilli: Same as `GlobusResponse` above...

jaswilli:
I agree with this in principal and would like to see it happen but I think we should try to get it done without breaking changes if at all possible. 👍

globus_sdk/response.py
sirosen: I thought about this -- nobody other than us is probably using this initializer, but it *is* a backwards incompatible way to add the correct.
Would you be satisfied by making`client` an optional argument?

corpulentcoffee: I had the same thought as @jaswilli when I originally saw this; just structuring this as `self, data, client=None` would avoid any possibility of drama and I think that you wouldn't have needed so many changes in the test suite.

sirosen:
@corpulentcoffee, your note about the changes to the testsuite was borne out by changing this to that pattern. In retrospect, I should have just done this from the start.

There is some irony in my taking care not to break compatibility for one method (`decode_id_token`) and not having taken care not to break it for the initializer of one of the most fundamental classes.

PR #257, opened 2017-12-04
Require cryptography and pyjwt on all installs, deprecate [jwt] target

sirosen:
Cryptography and pyjwt become required (via `pyjwt[crypto]`), and this removes py3.3 from supported versions (because it breaks on Windows, or at least AppVeyor).

Since this makes a previously supported py3 version unsupported, add a warning at install time as a courtesy.

The `[jwt]` target needs to stay as an empty target so that SDK installs which specify it don't break.

The big tada moment for this is that the tests pass on Windows, so we can probably roll forward.

Per discussions about this:
IF and ONLY IF people run into problems installing `cryptography` on platforms and versions we want to target, we can pursue the addition of an env var which controls setup.py and removes `cryptography` and `pyjwt` from install_requires.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,3 +1,4 @@
+import warnings
 import os.path
 import sys
 from setuptools import setup, find_packages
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -9,6 +10,13 @@
 # globus-sdk does not support python versions older than 2.7 #
 ##############################################################""")
 
+# warn on older/untested python3s
+# it's not disallowed, but it could be an issue for some people
+if sys.version_info > (3,) and sys.version_info < (3, 4):
+    warnings.warn(
+        'Installing globus-sdk on Python 3 versions older than 3.4 '
+        'may result in degraded functionality or even errors.')
```
jaswilli: 😢

aaschaer: Part of me wants this warning to be a little louder, but I guess people stuck on EOL python already have enough warnings yelling at them.

PR #259, opened 2017-12-08
Add a (BETA) SearchClient class which provides an interface to the Search API

sirosen:
Even though Search is a production service, creating an index is not publicly exposed. As such, it's hard for people in our team to function independently with respect to it and gain familiarity.
We're considering basic review to be the best we can hope for at present.

@corpulentcoffee has kindly offered to review this for obvious errors.
If you would like me to create an index for you to toy with, we can do that too -- it's very little effort.

This contains the following components
- SearchClient class
- SearchQuery object which helps you build structured queries to send to `SearchClient.post_search()`
- SearchAPIError class for API errors
- Docs for the above

Because the docs are still in progress towards docs.globus.org (but not there yet!) there are no `External Documentation` linkages in the docstrings.

### DIFF
--- a/globus_sdk/__init__.py
+++ b/globus_sdk/__init__.py
@@ -1,39 +1,46 @@
 import logging
 
 
+from globus_sdk.version import __version__
+
 from globus_sdk.response import GlobusResponse, GlobusHTTPResponse
-from globus_sdk.transfer import TransferClient
-from globus_sdk.transfer.data import TransferData, DeleteData
-from globus_sdk.auth import (
-    AuthClient, NativeAppAuthClient, ConfidentialAppAuthClient)
+
+from globus_sdk.exc import (
+    GlobusError, GlobusAPIError, TransferAPIError, SearchAPIError,
+    NetworkError, GlobusConnectionError, GlobusTimeoutError)
 
 from globus_sdk.authorizers import (
     NullAuthorizer, BasicAuthorizer, AccessTokenAuthorizer,
     RefreshTokenAuthorizer, ClientCredentialsAuthorizer)
 
-from globus_sdk.exc import GlobusError, GlobusAPIError, TransferAPIError
-from globus_sdk.exc import (
-    NetworkError, GlobusConnectionError, GlobusTimeoutError)
+from globus_sdk.auth import (
+    AuthClient, NativeAppAuthClient, ConfidentialAppAuthClient)
 
-from globus_sdk.version import __version__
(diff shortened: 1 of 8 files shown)


globus_sdk/globus.cfg
```diff
@@ -2,30 +2,36 @@
...
 auth_service = https://auth.preview.globus.org/
 nexus_service = https://nexus.api.preview.globus.org/
+search_service = https://search.api.preview.globus.org/
 transfer_service = https://transfer.api.preview.globus.org/
 
 [environment sandbox]
 auth_service = https://auth.sandbox.globuscs.info/
-nexus_service = https://graph.api.go.sandbox.globuscs.info/
+nexus_service = https://graph.api.sandbox.globuscs.info/
+search_service = https://search.api.sandbox.globuscs.infog/
```
jaswilli: It looks like all these search URLs have a typo? ("g")

Also, I think we need to switch all these `graph` entries to `nexus`.

jaswilli:
Aside from `globus.cfg` this LGTM 👍

sirosen:
I amended to fix the config (no need to persist typo-ed hostnames in the history).
Test failures are unrelated to these changes -- I'll rerun them later today if I remember.

globus_sdk/__init__.py
```diff
@@ -1,39 +1,46 @@
...
+__all__ = (
+    "__version__",
 
-__all__ = ["GlobusResponse", "GlobusHTTPResponse",
+    "GlobusResponse", "GlobusHTTPResponse",
 
-           "TransferClient", "TransferData", "DeleteData",
+    "GlobusError", "GlobusAPIError", "TransferAPIError", "SearchAPIError",
+    "NetworkError", "GlobusConnectionError", "GlobusTimeoutError",
+    "GlobusOptionalDependencyError",
```
corpulentcoffee: `GlobusOptionalDependencyError` isn't actually availed from `globus_sdk.exc` above.

```
>>> from globus_sdk import *
AttributeError: module 'globus_sdk' has no attribute 'GlobusOptionalDependencyError'
```

(This looks like this was broken before you rearranged this though.)

corpulentcoffee:
Everything else seemed OK to me. I don't care if we fix the `GlobusOptionalDependencyError` in this PR or after.

sirosen:
I'm fixing the import error in #260. I just pushed a relevant change with a couple of tests and the requisite line removal.

PR #264, opened 2017-12-13
Improve handling of connection timeouts

sirosen:
Add a default connection timeout of 60s (settable via env var or config). Default is wait indefinitely, which could hang clients... bad.

Also, improve error propagation so that callers can distinguish connection timeout (waiting for TCP connection establishment) from read timeout (waiting for data back from server). This is just a matter of wrapping some requests errors better.

### DIFF
--- a/globus_sdk/base.py
+++ b/globus_sdk/base.py
@@ -45,6 +45,8 @@ class BaseClient(object):
 
        ``http_timeout`` (*float*)
          Number of seconds to wait on HTTP connections. Default is 60.
+         A value of -1 indicates that no timeout should be used (requests can
+         hang indefinitely).
 
     All other parameters are for internal use and should be ignored.
     """
@@ -114,6 +116,9 @@ class BaseClient(object):
         if http_timeout is None:
             http_timeout = config.get_http_timeout(environment)
         self._http_timeout = http_timeout
+        # handle -1 by passing None to requests
+        if self._http_timeout == -1:
+            self._http_timeout = None
 
         # set application name if given
         self.app_name = None
--- a/globus_sdk/base.py
+++ b/globus_sdk/base.py
@@ -43,6 +43,9 @@ class BaseClient(object):
           User-Agent string, and may be useful when debugging issues with the
           Globus Team
 
+       ``http_timeout`` (*float*)
+         Number of seconds to wait on HTTP connections. Default is 60.
+
     All other parameters are for internal use and should be ignored.
     """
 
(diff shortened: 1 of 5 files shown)


jaswilli:
Having a global timeout settable via config is fine but I was thinking we'd add a `timeout` keyword arg to the base client so that timeouts could be specified on a per-client basis. It would also allow the advanced usage of specifying both connection timeout and a read timeout by passing in a tuple.

Thoughts?

sirosen:
Yeah, I don't much care how it's exposed to be settable, so long as there's a default of 30 or 60s to keep it from hanging indefinitely.

I think having both ways of doing it is bad, so I'll just move it into the initializer for clients.

globus_sdk/config.py
```diff
@@ -143,6 +143,17 @@ def get_service_url(environment, service):
     return url
 
 
+def get_http_timeout(environment):
+    p = _get_parser()
+    value = p.get("http_timeout", environment=environment,
+                  failover_to_general=True, check_env=True,
+                  type_cast=int)
```
jaswilli: Not a huge deal but if we have `http_timeout` documented as a `float` maybe we should be casting to that here instead of `int`?

globus_sdk/base.py
```diff
@@ -55,7 +58,9 @@ class BaseClient(object):
     BASE_USER_AGENT = 'globus-sdk-py-{0}'.format(__version__)
 
     def __init__(self, service, environment=None, base_url=None,
-                 base_path=None, authorizer=None, app_name=None):
+                 base_path=None, authorizer=None, app_name=None,
+                 http_timeout=None,
```
jaswilli: I think our usage of `None` is going to be problematic in that we've lost the ability to tell Requests to not timeout. Requests uses `timeout=None` but we'll convert `None` to `60`.

globus_sdk/config.py
sirosen: Good catch. Will fix.

globus_sdk/base.py
sirosen: Is that... actually a problem? I mean, will anyone actually want to set no-timeout?

I see a couple of okay workarounds:
 - We could translate `-1` to no timeout
 - We could make `http_timeout` part of `**kwargs`, and treat its absence and its presence (with a value of `None`) differently

Do you like either of these? (I have no strong feelings)

jaswilli: It _shouldn't_ be a problem but the world is a crazy place and someone, maybe even someone we know, might end up with a service that takes more than 60 seconds to respond...

`-1` for no timeout seems intuitive and is easy to document. I'd probably go with that.

sirosen:
@jaswilli, assuming that the fixes look okay to you, would you mind merging this and #268 ?
I can then do a release of v1.4.1

PR #265, opened 2017-12-14
Add Content-Type to SearchClient headers

sirosen:
The Search API is a stickler for correct Content-Type on JSON payloads.
Set this header all the time when talking to the API to keep things happy.


jaswilli:
Could we just handle this universally?: https://github.com/jaswilli/globus-sdk-python/commit/e1f7c4ce37c432bc64a97dfa615843e43e78a9e7

sirosen:
@jaswilli I like that much better, actually.
Want to put it in a PR and close this?

PR #274, opened 2018-01-23
Bug fixes for search functionality

NickolausDS:
A bug in CRUD for both Search Subjects and Entries caused parameters
to be lost before the request was sent, resulting in errors from
Globus Search.

### DIFF
--- a/globus_sdk/search/client.py
+++ b/globus_sdk/search/client.py
@@ -267,7 +267,7 @@ class SearchClient(BaseClient):
         in the API documentation for details.
         """
         index_id = safe_stringify(index_id)
-        params = merge_params(params, subject=subject)
+        merge_params(params, subject=subject)
 
         self.logger.info("SearchClient.get_subject({}, {}, ...)"
                          .format(index_id, subject))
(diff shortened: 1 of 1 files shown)


sirosen:
`merge_params` modifies a dict in place and returns `None`. You don't want to use `x = merge_params(x, ...)` -- the current usage is correct.

See the `merge_params` docstring and contents (which are quite small) for detail:
https://github.com/globus/globus-sdk-python/blob/1672d3ca318f60091a70ceb2ac730ae7657ccb19/globus_sdk/base.py#L404-L444

I'm going to close this out, since we don't want this patch, but please open an issue with the actual bug details so that we can dig into this.

NickolausDS:
https://github.com/globus/globus-sdk-python/issues/275

sirosen:
Oh, wow! Someone just pointed out to me that I got this totally wrong.
I thought you were changing good usage to bad, but you were changing bad usage to good!

sirosen:
Yes, this looks correct, now that I take a closer look at it.
Sorry for the runaround.

There aren't any tests for this, so I'll go ahead and merge without waiting for the testsuite to run.

NickolausDS:
Thanks!

PR #276, opened 2018-01-25
First draft of LocalGlobusConnectPersonal

sirosen:
This is a new class which describes the local personal endpoint. It has one property: endpoint_id, which gets the endpoint ID as a string.

@jaswilli mentioned maybe not including a deleter for `LocalGlobusConnectPersonal.endpoint_id` -- I don't have much of an opinion on this.
I could also remove it from the docstring for the `endpoint_id` property, to keep the docs simpler and shorter, but I'd leave the actual deleter in place, if we think it might be confusing to novice users.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -27,6 +27,7 @@ Table of Contents
    clients
    responses
    exceptions
+   local_endpoints
    oauth
    authorization
    config
--- a/docs/local_endpoints.rst
+++ b/docs/local_endpoints.rst
@@ -1,0 +1,27 @@
+.. _local_endpoints:
+
+Local Endpoints
+===============
+
+Unlike SDK functionality for accessing Globus APIs, the locally available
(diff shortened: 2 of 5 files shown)


jaswilli:
Yeah, I just don't get the use case around wanting to delete an endpoint id vs. just instantiating a new object.

That said, I don't much care either way about the actual presence of the deleter it was mostly just a feeling that all the documentation and example code around using it that seemed like it would muddy the waters.

sirosen:
In hindsight, reinstalling GCP while you have a `LocalGlobusConnectPersonal` in memory is pretty niche -- I've cut down the docstring content on that subject to a single phrase, not even a full usage example.

You could definitely create a new object instead. I don't have a use-case for using `del` vs just creating a new `LocalGlobusConnectPersonal`; including the deleter just seemed natural to me, I guess. _shrugs_

PR #279, opened 2018-02-16
Misc doc updates

jaswilli:


docs/tutorial.rst
```diff
@@ -27,15 +27,15 @@ In order to complete an OAuth2 flow to get tokens, you must have a client or
...
+"Register your app with Globus."
 You will be prompted to login -- do so with the account you wish to use as your
 app's administrator.
 
 When prompted, create a Project named "SDK Tutorial Project".
 Projects let you share the administrative burden of a collection of apps, but
 we won't be sharing the SDK Tutorial Project.
 
-In the "Manage Project" menu for "SDK Tutorial Project", select "Add new App".
+In the "Manage Project" menu for "SDK Tutorial Project", select "Add new app".
```
corpulentcoffee: I think it's just an "Add..." menu now (the GCS resource server registration changeset switched it away from "Manage Project").

jaswilli: I didn't notice that one, nice catch.

PR #281, opened 2018-02-23
Add GlobusSDKUsageError & replace most ValueErrors

### LINKED ISSUE #139: change AuthClient errors from ValueErrors to ?
The ValueErrors thrown by AuthClient's oauth2_get_authorize_url() and oauth2_exchange_code_for_tokens() when called out of order should probably be some kind of GlobusError?

sirosen:
The GlobusSDKUsageError is just `class GlobusSDKUsageError(GlobusError, ValueError): pass`

It replaces many uses of ValueError to indicate that an SDK method has been misused. This acts as an additional hint to users that they haven't read docs carefully enough, and makes us more conformant to our stated policy of only raising GlobusErrors unless otherwise noted.

That latter part (being consistent with our docs) is the one I care more about.
I left ValueError in place where it "felt appropriate", so feel free to call me out on that if any of it looks wrong.

Simple example:
```python
globus_sdk.ConfidentialAppAuthClient(
    'a', 'b', authorizer=globus_sdk.AccessTokenAuthorizer('foo'))
```
now raises GlobusSDKUsageError, not ValueError, for passing authorizer to a confidential client.

Resolves #139

### DIFF
--- a/globus_sdk/__init__.py
+++ b/globus_sdk/__init__.py
@@ -6,7 +6,8 @@ from globus_sdk.version import __version__
 from globus_sdk.response import GlobusResponse, GlobusHTTPResponse
 
 from globus_sdk.exc import (
-    GlobusError, GlobusAPIError, TransferAPIError, SearchAPIError,
+    GlobusError, GlobusSDKUsageError,
+    GlobusAPIError, TransferAPIError, SearchAPIError,
     NetworkError, GlobusConnectionError, GlobusTimeoutError,
     GlobusConnectionTimeoutError)
 
@@ -30,7 +31,8 @@ __all__ = (
 
     "GlobusResponse", "GlobusHTTPResponse",
 
-    "GlobusError", "GlobusAPIError", "TransferAPIError", "SearchAPIError",
+    "GlobusError", "GlobusSDKUsageError",
+    "GlobusAPIError", "TransferAPIError", "SearchAPIError",
     "NetworkError", "GlobusConnectionError", "GlobusTimeoutError",
     "GlobusConnectionTimeoutError",
 
--- a/globus_sdk/local_endpoint/personal.py
+++ b/globus_sdk/local_endpoint/personal.py
@@ -1,5 +1,7 @@
 import os
 
+from globus_sdk.exc import GlobusSDKUsageError
+
 
 def _on_windows():
     """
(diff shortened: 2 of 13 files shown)


globus_sdk/__init__.py
```diff
@@ -30,7 +31,8 @@
 
     "GlobusResponse", "GlobusHTTPResponse",
 
-    "GlobusError", "GlobusAPIError", "TransferAPIError", "SearchAPIError",
+    "GlobusError", "GlobusSDKUsageError",
```
corpulentcoffee: Do you need to introduce this new error class in `docs/exceptions.rst` as well, including maybe rephrasing the "malformed calls to Globus SDK methods may raise standard python exceptions" bit?

globus_sdk/local_endpoint/personal.py
```diff
@@ -49,7 +51,7 @@ def endpoint_id(self):
                 if _on_windows():
                     appdata = os.getenv("LOCALAPPDATA")
                     if appdata is None:
-                        raise ValueError(
+                        raise GlobusSDKUsageError(
```
corpulentcoffee: Is the absence of `LOCALAPPDATA` on a Windows system the fault of usage, or is it the fault of a broken Windows installation? I don't think it matters, and we should probably leave it alone as it would be a breaking change, but maybe an `OSError` would have been better here.

globus_sdk/__init__.py
sirosen: Yeah, I think a doc update would be appropriate. Will add.

corpulentcoffee:
> It's not backwards-compatible in the strictest, most paranoid, sense, but it's probably fine for all or almost all real usage.

Agreed; just a quick mention in release notes is fine.

globus_sdk/local_endpoint/personal.py
sirosen: I was considering it a usage error from the start because it's not okay to try to use this method after clearing `os.environ` or invoking python with a really messed up env in the first place.
It's not possible for us to know why it wasn't set. So I just assume that it's the caller's fault at some level! :smile: 

I doubt this will ever get tripped, but #robustnessprinciple

Anyway, it sounds like you're okay with this error as it is in this PR.

sirosen:
Rebased to include the `ValueError` in `BaseClient.__init__` + a doc update.

PR #283, opened 2018-03-23
make work with python 2.6.9

rhaas80:
This makes the module usable with python 2.6.9 as found eg on Blue Waters. 

There are basically two changes:

1. I need to use `{0}` and `{1}` etc. in all string.format() calls and cannot use just `{}`
2. `logging.NullHandler` does not yet exist so I have to provide a substitute

### DIFF
--- a/docs/examples/advanced_transfer.rst
+++ b/docs/examples/advanced_transfer.rst
@@ -103,7 +103,7 @@ retry logic from the core task submission logic.
                     must be run inside an exception handler
             """
             if retries < 1:
-                logger.error('Retried {} too many times.'
+                logger.error('Retried {0} too many times.'
                              .format(func_name))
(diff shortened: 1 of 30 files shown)


jaswilli:
Hi @rhaas80,

I appreciate the effort but we're not able to commit to supporting python 2.6 given that it reached end-of-life almost 5 years ago.

rhaas80:
Ah I see. Fine with me. You may consider to remove the comment (in the original master) in https://github.com/globus/globus-sdk-python/blob/a0be7d6c12dc24f9b8b4b982da0395ed56e82b84/globus_sdk/__init__.py#L53-L55 then which is the one that gave me the impression (hope actually) that 2.6.9 might still be supportable.

PR #300, opened 2018-07-02
Adjust travis config to test against python 3.7

jaswilli:
* Adds a python 3.7 build.
* Adds python 3.8-dev.
* Use the `matrix` feature to select the ubuntu distro per-build so we can get all our desired pythons.


sirosen:
This looks fancy.
Awaiting full build output, but assuming it passes I'll merge.

If `xenial` on Travis doesn't add some of these (3.4, pypy), I'm okay with dropping them from our tests when `trusty` EOLs next year.

Do we need to do a similar update on AppVeyor?

jaswilli:
> Do we need to do a similar update on AppVeyor?

Maybe? When I looked they were still showing `3.6.5` as the latest release.

PR #307, opened 2018-08-24
Fix handling of `environment="production"`

### LINKED ISSUE #306: Creating a client with `environment="production"` does not properly translate as it would with the environment variable
The default config section has the "production" environment values.
We have handling code for this in `globus_sdk.config_get_default_environ`, but that only runs if `environment` isn't passed to a client.
As a result, config lookups are done against the `production` section and fail.

There are several ways of handling this, but one option which is probably pretty foolproof is to update this:
https://github.com/globus/globus-sdk-python/blob/d761ba9a104ba2d8a70d2e064d12ea95101596df/globus_sdk/base.py#L85-L86

Instead of doing that, try
```python
environment = config.get_default_environ(environment)
```
and update `get_default_environ` to take an input.

sirosen:
When passed explicitly, this value bypassed the check we do while looking up the environment variable, and then results in failed config lookups.
Force the given `environment` through that config lookup code which translates `production` to `default` and logs the use of non-production environments.

Minor note: the logging used to say `GLOBUS_SDK_ENVIRONMENT=...` with a strong implication that it matched the env var. Since this is no longer always taken from there, I've modified it to just say `globus_environment=...`

closes #306

### DIFF
--- a/globus_sdk/config.py
+++ b/globus_sdk/config.py
@@ -175,17 +175,25 @@ def _bool_cast(value):
     raise ValueError("Invalid config bool")
 
 
-def get_default_environ():
+def get_globus_environ(inputenv=None):
     """
-    Get the default environment to look for in the config, as a string.
+    Get the environment to look for in the config, as a string.
+
     Typically just "default", but it can be overridden with
     `GLOBUS_SDK_ENVIRONMENT` in the shell environment. In that case, any client
     which does not explicitly specify its environment will use this value.
+
+    :param inputenv: An environment which was passed, e.g. to a client
+                     instantiation
     """
-    env = os.environ.get('GLOBUS_SDK_ENVIRONMENT', 'default')
+    if inputenv is None:
+        env = os.environ.get('GLOBUS_SDK_ENVIRONMENT', 'default')
+    else:
+        env = inputenv
+
     if env == 'production':
         env = 'default'
     if env != 'default':
         logger.info(('On lookup, non-default environment: '
-                     'GLOBUS_SDK_ENVIRONMENT={}'.format(env)))
+                     'globus_environment={}'.format(env)))
     return env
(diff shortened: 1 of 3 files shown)


tests/unit/test_config.py
```diff
@@ -157,10 +157,10 @@ def test_get_default_environ(self):
         if "GLOBUS_SDK_ENVIRONMENT" in os.environ:
             prev_setting = os.environ["GLOBUS_SDK_ENVIRONMENT"]
             del os.environ["GLOBUS_SDK_ENVIRONMENT"]
-        self.assertEqual(globus_sdk.config.get_default_environ(), "default")
+        self.assertEqual(globus_sdk.config.get_globus_environ(), "default")
         # otherwise environ value
         os.environ["GLOBUS_SDK_ENVIRONMENT"] = "beta"
-        self.assertEqual(globus_sdk.config.get_default_environ(), "beta")
+        self.assertEqual(globus_sdk.config.get_globus_environ(), "beta")
```
aaschaer: Wouldn't hurt to add a test for the changed behavior here

`os.environ["GLOBUS_SDK_ENVIRONMENT"] = "production"`
`self.assertEqual(globus_sdk.config.get_globus_environ(), "default")`

globus_sdk/config.py
```diff
@@ -175,17 +175,25 @@ def _bool_cast(value):
...
+    :param inputenv: An environment which was passed, e.g. to a client
+                     instantiation
     """
-    env = os.environ.get('GLOBUS_SDK_ENVIRONMENT', 'default')
+    if inputenv is None:
+        env = os.environ.get('GLOBUS_SDK_ENVIRONMENT', 'default')
+    else:
+        env = inputenv
+
     if env == 'production':
```
aaschaer: Oh also, would we also want to catch "prod" in addition to "production"?

PR #309, opened 2018-08-28
Make RenewingAuthorizer expiration check public

sirosen:
Simply rename:
```
-    def _check_expiration_time(self):
+    def check_expiration_time(self):
```
so that it's official.

Also improve documentation around Renewing Authorizer and its subclasses. In order to do this without writing some complex sphinx extension code or other shenanigans, I accepted a bit of repetition. It makes the docs a tiny bit harder to maintain in exchange for making the docs a lot easier to maintain.


aaschaer:
Looks good to me! Just waiting for tests to pass.

I guess anyone who was using `_check_expiration_time` before was doing so at their own risk since it was a private method?

sirosen:
Yes, that's my perspective on it. It was an `_...` method, which is as private as python gets.

I want to get this shipped before the CLI, so that we can use the new name.

PR #308, opened 2018-08-28
Dependent token enhancement

mikkel14:
Tiny changes to address #278. I just added in the `additional_params` parameter, similar to other functions in the module.

PR #305, opened 2018-08-14
Replace token suffixes in debug logging with SHA-256 hash value

mikkel14:
This is my first time working with this repository and submitting a pull request, so feel free to tear things apart. This is addressing #299 

I added `access_token_hash` data member to the `AccessTokenAuthorizer` and `RenewingAuthorizer` classes to hold a SHA-256 hash value. Additionally, I added a utility function for calculating the hash value if that functionality is needed elsewhere.

Again, feel free to suggest things to tweak and move around, tests to add, etc.

PR #310, opened 2018-08-28
Update to v1.6.0

sirosen:
Minor version increase as this will add the `.patch` method to all `BaseClient` subclasses.

Release notes will contain:

- Correct handling of `environment="production"` as an argument to client construction (#307)
- `RenewingAuthorizer` and its subclasses now expose the `check_expiration_time` method (#309)
- Allow parameters to be passed to customize the request body of `ConfidentialAppAuthClient.oauth2_get_dependent_tokens` (#308)
- Use sha256 hashes of tokens (instead of last 5 chars) in debug logging (#305)
- Add the `patch()` method to `BaseClient` and its subclasses, sending an `HTTP PATCH` request (#302)
- Officially add support for python 3.7 (#300)
- Make pickling SDK objects safer (but still not officially supported!) (#284)
- Malformed SDK usage may now raise `GlobusSDKUsageError` instead of `ValueError`. `GlobusSDKUsageError` inherits from `ValueError` (#281)
- Numerous documentation improvements (#279)(#294)(#296)(#297)


aaschaer:
Error was a 500 filtering up from groups, looks good to me

PR #311, opened 2018-08-29
fix acl for sdktester2b in test suite

aaschaer:
Fixes the test suite error raised by transfer now checking ACLs before task submission

We could add this on creation of the shared endpoint, but unless someone feels strongly about that I'd rather not add 19 extra round trips to Transfer


sirosen:
> I'd rather not add 19 extra round trips to Transfer
Agreed. The testsuite is slow enough as it is.

(Aside: I'm more and more seriously considering the option of going back and building mocks over all of the APIs for our testing.)

When tests pass, I'll merge.

sirosen:
No, I've been doing some work on it but wanted to wait until the current work in progress gets closed out before merging. I'll retrigger runs of the failed builds, but perhaps this can just merge given that one Travis and one AppVeyor build each passed?

Issue #313, opened 2018-09-03
Provide wheels


nsoranzo:
It would be nice to have wheels on PyPI (see https://pythonwheels.com/ ), instead (or in addition) to eggs.

PR #314, opened 2018-09-04
Build wheels instead of eggs

nsoranzo:
Fix https://github.com/globus/globus-sdk-python/issues/313 .


sirosen:
As I mentioned on #313, this may not be quite trivial. I didn't elaborate, but I mostly meant that WRT testing the resulting wheel to make sure it works on all of our target platforms. In the past, even universal wheels have had issues with UCS-2 and UCS-4 python2 issues -- I don't know if those have all been sorted out or not by modern versions of `wheel`.

Thanks for doing the legwork -- it looks right to me -- but I do want to take a day or two to make sure it works correctly before changing the format.

nsoranzo:
@sirosen Thanks for the quick reply, most high profile Python packages use wheels nowadays (see e.g. https://pypi.org/project/numpy/#files ) and dropped eggs long ago. For universal wheels (like for this project) there is absolutely no issue.

sirosen:
I didn't see anything in the `wheel` issue history suggesting that the USC-{2,4} issues are all wrapped up, but I did find [this relevant note](https://www.python.org/dev/peps/pep-0513/#ucs-2-vs-ucs-4-builds) in the `manylinux` tag spec.

It sounds like `pip` and `wheel` have internally resolved this issue.
I'm building a USC-2 python2 to verify this right now, and I'll be able to check installation wheels built with USC-4 python2 and with python3.

As soon as I get through that testing, assuming everything works I'm happy to accept this change.

upload-requirements.txt
```diff
@@ -1 +1 @@
-twine==1.9.1
+twine>=1.9.1
```
sirosen: Could you amend to revert this change?
Aggressively upgrading `twine` has actually bitten us in the past. It's pegged at a specific version intentionally (and yes, I know that's a touch weird).

Otherwise, I'll revert in the merge.

nsoranzo: I can do that, what about setting it to `==1.11.0` ? It's been released in March and quite stable, with important updates (see https://twine.readthedocs.io/en/latest/changelog.html ).

sirosen:
Okay, I've confirmed that py2 usc-4 universal wheel builds work with py2 usc-2 , and that upgrades and downgrades between wheels and eggs work correctly for us.

Thanks for the work! I'm going to go ahead and merge without the tests, since they don't cover our packaging anyway.
Wheel packaging will roll out in our next release -- I do not plan to backport this to `1.6.0` since it would violate our rule of having the tags mark exactly what was released.

nsoranzo:
Thanks for testing and merging!

> Wheel packaging will roll out in our next release -- I do not plan to backport this to 1.6.0 since it would violate our rule of having the tags mark exactly what was released.

No problem!

PR #315, opened 2018-09-04
Convert testsuite to pytest, mock all API interactions

sirosen:
This is a work in progress. I'll rename when it's complete, but it's not ready yet, at least.

It consists of two primary changes:
1. Convert the testsuite to `pytest` from `nose2`
2. Mock all API interactions using the `httpretty` lib, never directly communicate with the APIs

The first commit in this PR removes the nose2 testsuite, and I'm porting its functionality (or at least, what I can of it) to pytest bit by bit.

I've also changed the way that development and testing requirements are installed by adding an extra meant for internal use, `globus_sdk[development]`. So you can `pip install -e '.[development]'` and get everything you need for docs, testing, and package publishing. That makes things a bit more uniform across appveyor, travis, and the various make targets.

---

Planned tests to add:
- [ ] Check loading GCP ID from file with the local_endpoint module
  - [ ] Make sure this covers testing the windows variant (previously broken, #321)
- [ ] Add unicode to a few response objects, make sure it comes through translations okay
- [ ] Full successful login flow
- [ ] Functional tests for RenewingAuthorizer usage
  - [ ] RefreshToken and ClientCredentials
  - [ ] Check 401s on first response and second response
  - [ ] Check with a client method with `retry_401s=False`
  - [ ] RenewingAuthorizer callout may fail (e.g. network error)
  - [ ] Ensure `on_refresh` callback is invoked
  - [ ] Propagate errors from malformed `on_refresh` callback
    - [ ] Does the Authorizer state change? Documenting this behavior would be a good add


sirosen:
Trying to debug the py3 pickling errors in this, it seems a potential cause is [this python bug about weakrefs and proxy objects](https://bugs.python.org/issue14577).
I'm not sure if I believe that we're doing anything sophisticated enough that we should get tripped up by this though... Probably needs more investigation.

sirosen:
I fixed the pickle test issue -- not in a necessarily beautiful way, but it works.

I know we're busy right now, and this is not priority, but I would greatly appreciate any time people can spare me to look at the change as it stands today. I don't intend to do any more work on it unless we decide to go forward with this conversion, as I risk wasting effort.

tests/unit/authorizers/test_basic_authorizer.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+    Confirms that BasicAuthorizer doesn't handle missing authorization
+    """
+    assert not authorizer.handle_missing_authorization()
+
+
+@pytest.mark.parametrize(
+    'username, password',
+    [("user", u"テスト"), (u"дум", 'pass'), (u"テスト", u"дум")])
+def test_unicode_handling(username, password):
```
aaschaer: I like testing non ascii characters here, is there a good way to make sure we're testing it across the whole SDK?

tests/functional/auth/test_auth_client_flow.py
```diff
@@ -0,0 +1,145 @@
...
+
+import globus_sdk
+from globus_sdk.auth.oauth2_constants import DEFAULT_REQUESTED_SCOPES
+from globus_sdk.auth.oauth2_native_app import make_native_app_challenge
+
+from tests.common import register_api_route
+
+
+CLIENT_ID = 'd0f1d9b0-bd81-4108-be74-ea981664453a'
+INVALID_GRANT_RESPONSE_BODY = '{"error":"invalid_grant"}'
```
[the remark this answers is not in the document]
sirosen: Yes, this is one of the things which we can do if we aren't talking to the live APIs. It just seemed awfully laborious to do if we aren't yet fully convinced that we want to make these changes.

sirosen:
I've retitled from `[WIP]`. This doesn't have all of the testing that I want to add, but I think what's here is good enough for us to actually accept as a replacement testsuite.
Especially with coverage reporting, we can work on getting everything included as time is available.

After a brief chat in Slack, I'm removing Dave from the reviewers. He said that Aaron's sign off is good enough for him, but it doesn't seem right to ask him to approve of the changes in GitHub based on that.

I'm going to continue to annoy Jason for "proper" review, at least of the framework for the tests, if not of all of the individual test contents.

tests/unit/authorizers/test_basic_authorizer.py
aaschaer: I think sprinkling some non ascii unicode values into the fixtures is probably sufficient, I mostly want to know that in our handling of both python 2 and 3 we aren't choking on utf8 conversions somewhere in the SDK itself.

sirosen:
I don't want this PR to block and wait on a bunch of new tests that we want to add, but I also feel I should keep a list of useful/important ideas. I've started listing new test ideas at the bottom of the PR description.

Issue #292, opened 2018-05-08
Document SDK client classes as unsafe for multiprocessing and not guaranteed to be threadsafe


sirosen:
This came up in the context of a Search user using a process pool to index data in parallel.

Each SDK client object contains a requests Session.
Sessions hold urllib3 connection pools.
Connection pools are documented as threadsafe, but [are known to be unsafe prefork](https://github.com/urllib3/urllib3/issues/850), so creating clients before a fork and using them afterwards is unsafe.

There may be other methods of SDK clients which are not threadsafe and I don't want to tie us in knots worrying about this.

SDK clients should be considered unsafe to create prefork and thread-unsafe.
We should recommend creating your client objects after forking and once-per-process.

Technically, someone could have a use-case for driving multiple client instances off of a single authorizer and want us to improve this to function out of the box, but today it _won't_ and I don't want people getting the wrong idea. If someone has a strong enough driving need, we can consider making them threadsafe at that time.

mikkel14:
Would a note added to `docs/clients.rst` be acceptable for this? I wrote up a small paragraph. I could add a little warning to `globus_sdk.base.BaseClient` docstring if desired, since all the other clients inherit from that. Although maybe this is best kept in the full SDK documentation only.

sirosen:
@mikkel14, I totally missed this comment (just noticed it during a review). Sorry about that!

I think a note in the narrative docs, in `docs/clients.rst` would be quite nice.
If you open up a pull request, I'd be more than happy to review.

PR #319, opened 2018-10-23
Add note on multithreading and multiprocessing applications

mikkel14:
Stab at addressing #292


sirosen:
This looks good to me. Merging.

PR #328, opened 2018-10-30
Add Windows builds to Travis

jaswilli:
Adds Windows builds for python 2.7 and python 3.7.

Note: leaving AppVeyor in place for now while we see how building with Windows on Travis goes.


sirosen:
I will wait for the build results, but I just realized that we can ditch the deeply crazy build.cmd script if this works. Suddenly I go from "meh" to super-excited about this.

Issue #282, opened 2018-03-11
Feature Request: add AuthCode local webbrowser handler to SDK


sirosen:
We've had requests come in to add the local `webbrowser` integration used by the CLI to the SDK.
To start, the feature request:

> Allow SDK users to emulate the login behavior of `globus login`

That probably breaks into components:
- A `GlobusSDKAuthCodeHandlerServer` which can be used as a context manager to cover both the local HTTP server and `webbrowser` invocation
- Handling for custom HTML templates
- Handling for custom scopes and other params on the oauth2 flow
- Well-defined exceptions for different failure modes (instead of catching `Exception`)

When you start to add in all of the window-dressing features (like detecting remote connections), it will inevitably bloat a bit more.

---

After spending a couple of hours trying to figure out how to slot it in, I'm convinced.
It doesn't belong in the SDK.

This is only being requested as an SDK feature "by process of elimination" -- people want to emulate `globus login`, and they're not sure where else to request such a feature.
When you look at what we offer today, we have the APIs, Local Globus Connect Personal, and various objects to support these. Authorizers, Responses, and Exceptions are all the most minimal, necessary tooling for managing these things.

The CLI is a python product which will do all kinds of things which users want to emulate -- but not all of them need to be here in the SDK. If such things need to be shared and exposed to 3rd party developers (and I'm not convinced that there is any such need!), then let's make another lib rather than distorting this one and losing sight of its goals and the level of quality and support which it offers.
Maybe it will make its way in, against my protestations, but it would be a mistake that sets a really bad precedent.

Issue #285, opened 2018-04-09
Add documentation example which demonstrates proper use of `on_refresh`


sirosen:
Variants of the Flask example, the tutorial native app client, and the simple client creds example demonstrating proper use of `on_refresh` for `RefreshTokenAuthorizer` and `ClientCredentialsAuthorizer` would be nice references to have.

Issue #256, opened 2017-12-04
Reorganize tutorial into sections, cover many more topics


sirosen:
The Tutorial in the SDK, which is where I (and hopefully others) send people to get started, is missing a lot of detail and important information. It's focused on "quickstarting" people with a Native App so that they can really start using the Transfer API.

The Tutorial should really have the following sections
- Create a Native App Client
  - Create AccessTokenAuthorizer, TransferClient, do `endpoint_search()`
- Create a Confidential App Client
  - Create AccessTokenAuthorizer, TransferClient, do `endpoint_search()`
- Look at token response content
  - Show refresh tokens with one of the client types
  - Demonstrate `decode_id_token`
  - Validate tokens, revoke, and validate again
- Feed refresh tokens to RefreshTokenAuthorizer, TransferClient, do `endpoint_search()`
  - Revoke AccessToken, do `endpoint_search()` again
- Show how to use `on_refresh` with renewing authorizers to save most recent token in a file
  - Make sure to set umask to get 0600 perms
- Refer to the Examples pages for Flask 3-legged, ClientCredentialsAuthorizer
- TransferClient instantiation with arbitrarily chosen Authorizer
  - Demonstrate switching from TransferClient with RefreshTokenAuthorizer to ClientCredentialsAuthorizer
- TransferClient detailed usage
  - Iterate on `ls` results, `task_list` results
  - Create bookmark, resolve bookmark, `ls` on bookmark + relative path
  - Submit Transfer Task, get explicit Submission ID and submit Delete Task
  - Link to retry example, relative deadline example
  - Double-iteration on `endpoint_search()` yields empty iterator -- paging APIs and implications
- Create AuthClient using Auth tokens
  - Explain `internal_auth_client` vs `auth_client` (or however these are named...)
  - Call get Identities API

@jaswilli, @aaschaer, @corpulentcoffee, is there additional content that needs to be in an expansive tutorial which tries to cover most of the "core" SDK functionality?

sirosen:
It would be great to have someone maintaining an intro guide, but I have enough excitement in my life without it. This is stale; closing.

Issue #254, opened 2017-12-04
Add Transfer API Endpoint Search Example to docs


sirosen:
Based off of #253, I think it would be nice to have another advanced TransferClient usage example focusing on `endpoint_search()`

I don't want to be re-documenting the API behaviors, but rather showing how you can use the iterable results to get more sophisticated behavior.
Something along these lines:
- Create three endpoints with content in their Keywords, Organization, and Department fields
- Perform searches which capture cross-sections of these endpoints
- Show a search which captures two of them "accidentally" (e.g. Organization for EP1 and Keywords for EP2 both have the same content)
- Show how to check fields for stricter matching client-side to fix the prior search (capture EPs with overlapping fields, then explicitly check that Organization=X)

PR #332, opened 2018-11-02
Added features targeted for developers writing custom clients

NickolausDS:
These changes make it easier for developers to create their own
clients by adding simple built-in token storage, and a customized
auth flow with a built-in local server. A built-in client_id was
also added to reduce the barrier of entry required for a native auth
flow. Developers writing simple scripts can simply do:

    from globus_sdk import native_auth
    my_tokens = native_auth()

No functionality here is meant to replace existing SDK functionality.
Instead, the changes are intended to address commonly copied and
re-implemented code.

Adding @rpwagner and @ranantha by request. 

To keep things as simple, only `native_auth` was added to `__init__.py`. 

This does address #282, but these changes don't fully address address
the requirements there. Mainly, local_server functionality isn't exposed 
for general use and templates can't be customized for branding.

I'll make notes about changes here I think warrant extra discussion.

### DIFF
--- a/globus_sdk/auth/oauth2_native_app_shortcut.py
+++ b/globus_sdk/auth/oauth2_native_app_shortcut.py
@@ -1,0 +1,183 @@
+import logging
+import os
+import webbrowser
+from socket import gethostname
+
+from six.moves import input
+
+from globus_sdk.auth.client_types.native_client import NativeAppAuthClient
+from globus_sdk.auth.oauth2_constants import DEFAULT_REQUESTED_SCOPES
+from globus_sdk.exc import ConfigError
+from globus_sdk.utils.local_server import is_remote_session, start_local_server
+from globus_sdk.utils.safeio import safe_print
+from globus_sdk.utils.token_storage import clear_tokens, load_tokens, save_tokens
+
+logger = logging.getLogger(__name__)
+
+AUTH_CODE_REDIRECT = "https://auth.globus.org/v2/web/auth-code"
+
+NATIVE_AUTH_DEFAULTS = {
+    "config_filename": os.path.expanduser("~/.globus-native-apps.cfg"),
+    "config_section": None,  # Defaults to client_id if not set
+    "client_id": "0af96eea-fec8-4d6e-aad2-c87feed8151c",
+    "requested_scopes": DEFAULT_REQUESTED_SCOPES,
+    "refresh_tokens": False,
+    "prefill_named_grant": gethostname(),
+    "additional_auth_params": {},
+    "save_tokens": False,
+    "check_tokens_expired": True,
+    "force_login": False,
+    "no_local_server": False,
+    "no_browser": False,
+    "server_hostname": "127.0.0.1",
+    "server_port": 8890,
+    "redirect_uri": "http://localhost:8890/",
+}
+
+
+def native_auth(**kwargs):
+    """
+    Provides a simple shortcut for doing a native auth flow for most use-cases
+    by setting common defaults for frequently used fields. Although a default
+    client id is provided, production apps should define their own at
+    https://developers.globus.org. See `NativeAppAuthClient` for constructing
+    a more fine-tuned native auth flow. Returns tokens organized by resource
+    server.
+
+    **Native App Parameters**
+        ``client_id`` (*string*)
+          Client App id registered at https://developers.globus.org. Defaults
+          to a built-in one for testing.
+
+        ``requested_scopes`` (*iterable* or *string*)
+          The scopes on the token(s) being requested, as a space-separated
+          string or iterable of strings. Defaults to ``openid profile email
+          urn:globus:auth:scope:transfer.api.globus.org:all``
+
+        ``redirect_uri`` (*string*)
+          The page that users should be directed to after authenticating at
+          the authorize URL. Defaults to
+          'https://auth.globus.org/v2/web/auth-code', which displays the
+          resulting ``auth_code`` for users to copy-paste back into your
+          application (and thereby be passed back to the
+          ``GlobusNativeAppFlowManager``)
+
+        ``refresh_tokens`` (*bool*)
+          When True, request refresh tokens in addition to access tokens
+
+        ``prefill_named_grant`` (*string*)
+          Optionally prefill the named grant label on the consent page
+
+        ``additional_auth_params`` (*dict*)
+          Set ``additional_parameters`` in
+          NativeAppAuthClient.oauth2_get_authorize_url()
+
+    **Login Parameters**
+        ``save_tokens`` (*bool*)
+          Save user tokens to disk and reload them on repeated calls.
+          Defaults to False.
+
+        ``check_tokens_expired`` (*bool*)
+          Check if loaded access tokens have expired since the last login.
+          You should set this to False if using Refresh Tokens.
+          Defaults to True.
+
+        ``force_login`` (*bool*)
+          Do not attempt to load save tokens, and complete a new auth flow
+          instead. Defaults to False.
+
+        ``no_local_server`` (*bool*)
+          Do not start a local server for fetching the auth_code. Setting
+          this to false will require the user to copy paste a code into
+          the console. Defaults to False.
+
+        ``no_browser`` (*bool*)
+          Do not automatically attempt to open a browser for the auth flow.
+          Defaults to False.
+
+        ``server_hostname`` (*string*)
+          Hostname for the local server to use. No effect if
+          ``no_local_server`` is set. MUST be specified in ``redirect_uri``.
+          Defaults to 127.0.0.1.
+
+        ``server_port`` (*string*)
+          Port for the local server to use. No effect if ``no_local_server``
+          is set. MUST be specified in ``redirect_uri``. Defaults to 8890.
+
+    **Configfile Parameters**
+        ``config_filename`` (*string*)
+          Filename to use for reading and writing values.
+
+        ``config_section`` (*string*)
+          Section within the config file to store information (like tokens).
+
+    **Examples**
+
+    ``native_auth()``
+
+    Or to save tokens: ``native_auth(save_tokens=True)``
+    """
+    unaccepted = [k for k in kwargs.keys() if k not in NATIVE_AUTH_DEFAULTS.keys()]
+    if any(unaccepted):
+        raise ValueError("Invalid args: {}".format(unaccepted))
+
+    opts = {k: kwargs.get(k, v) for k, v in NATIVE_AUTH_DEFAULTS.items()}
+
+    # Default to the auth-code page redirect if the user is copy-pasting
+    if (
+        opts["no_local_server"] is True
+        and opts["redirect_uri"] == NATIVE_AUTH_DEFAULTS["redirect_uri"]
+    ):
+        opts["redirect_uri"] = AUTH_CODE_REDIRECT
+
+    config_section = opts["config_section"] or opts["client_id"]
+
+    if opts["force_login"] is False:
+        try:
+            return load_tokens(
+                config_section, opts["requested_scopes"], opts["check_tokens_expired"]
+            )
+        except ConfigError as ce:
+            logger.debug(
+                "Loading Tokens Failed, doing auth flow instead. "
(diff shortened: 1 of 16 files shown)


globus_sdk/auth/oauth2_native_app_shortcut.py
```diff
@@ -0,0 +1,182 @@
...
+
+        ``config_section`` (*string*)
+          Section within the config file to store information (like tokens).
+
+    **Examples**
+
+    ``native_auth()``
+
+    Or to save tokens: ``native_auth(save_tokens=True)``
+    """
```
NickolausDS: The native_auth function was written from the perspective of making the native auth flow as simple as humanly possible. At its most basic, import a thing and then call it. Having this be a simple function call makes it very easy to include in a simple script or serve as a basic client, and this will be especially useful fo folks creating their own custom resource servers and need a small native app to test. It’s intended to serve a very narrow but very common use case, and wasn’t intended to support all features of the NativeAppAuthClient.

globus_sdk/auth/oauth2_native_app_shortcut.py
```diff
@@ -0,0 +1,182 @@
...
+          Defaults to False.
+
+        ``server_hostname`` (*string*)
+          Hostname for the local server to use. No effect if
+          ``no_local_server`` is set. MUST be specified in ``redirect_uri``.
+          Defaults to 127.0.0.1.
+
+        ``server_port`` (*string*)
+          Port for the local server to use. No effect if ``no_local_server``
+          is set. MUST be specified in ``redirect_uri``. Defaults to 8890.
```
NickolausDS: In retrospect, I didn't realize auth now supported port numbers not explicitly accounted for in the `redirect_urls` for the registered app. With that, `server_hostname`, and `server_port` options here should probably be removed.

globus_sdk/utils/safeio.py
```diff
@@ -0,0 +1,34 @@
...
+
+def get_safe_io():
+    global _safe_io
+    if _safe_io is None:
+        _safe_io = SafeIO()
+    return _safe_io
+
+
+def safe_print(message, *args, **kwargs):
+    get_safe_io().write(message, *args, **kwargs)
```
NickolausDS: When a user specifies the option for `no_local_server=True`, or `no_browser=True`, we need to print to a console so the user can copy/paste the auth code. In most or probably all cases, we _could_ just print to stderr and be fine. But libraries like the `globus-cli` use `@click.echo` and this was intended as a way to respect printing output for libraries like those.

globus_sdk/utils/token_storage.py
```diff
@@ -0,0 +1,217 @@
...
+    "expires_at_seconds",
+    "resource_server",
+]
+REQUIRED_KEYS = ["scope", "access_token", "expires_at_seconds", "resource_server"]
+CONFIG_TOKEN_GROUPS = "token_groups"
+
+logger = logging.getLogger(__name__)
+
+
+def save_tokens(tokens, config_section=None):
```
NickolausDS: `save_tokens` is intended to be used for arbitrary token storage, specifically by people registering their own scopes and wanting to easily save them for future use. This is most useful for people writing very simple clients or scripts. When somebody wants to write a custom client and save their own specific (non-token) related config-values, they should use their own config and not rely on the globus_sdk. A compromise there is to not expose this directly, and only allow it to be used indirectly by other SDK functionality.

jimPruyne:
I have not reviewed code or read all of Nick's updates, but I did want to register a +1 on the concept. Helpers for creating new Native App clients supporting token storage, etc. would be something I'd have used a couple times by now.

sirosen:
I find it difficult to put all of my thoughts in order on this. This is the short version:

This PR significantly expands the surface of the SDK and would constitute potentially significant maintenance burden. I'm really concerned about merging it in. My gut tells me that the right thing to do is to build it as a distinct python package entirely -- _not_ part of the SDK.

That was the view I expressed in #282 as well.

I'm closing because I think the appropriate next step is for this to get its legs as a distinct, standalone lib (on pypi or just in a GitHub repo). Then we can talk about what use-cases it does or does not satisfy, change the interface in significant and breaking ways version-to-version -- a luxury we are not afforded here -- and potentially work our way up to something we'd be comfortable moving into the SDK.

This is the longer version:

Optics aside, I think that's the best path for supporting a collection of utilities which touch a number of things which we have very carefully avoided to date.

That also sets the stage for discussions about porting individual, well-isolated pieces into the SDK, and converting this second-order library to be a consumer of these additions. Those discussions can take place about individual components, and would be much more productive than trying to take this in all at once.

There are important considerations missing from this work so far. Most prominently, this only appears to implement a small part of what you need to actually manage user logins.
You need login _and logout_ at a minimum. Session boost if you want to use HA endpoints. And you need to think about how the interface will look when consumed by multiple distinct applications. As written, two naive callers of this new function can see badly inconsistent behavior, as they "thrash" the set of scopes in use, and only function when they were the last application used for login.

This work needs to be expanded to handle more sophisticated -- but realistic -- cases before we can seriously consider it. And I'd like to get all of the maintainers involved in discussing what exactly it is we're trying to build, and why, before a bunch of code is written and we might have the unpleasant, but self-preservatory, duty of rejecting it.

Given that I haven't had time in the past _calendar year_ (and this is not to my credit) to do [major and important](https://github.com/globus/globus-sdk-python/issues/256) [work on documentation](https://github.com/globus/globus-sdk-python/issues/254) for [features we already have](https://github.com/globus/globus-sdk-python/issues/285), I find it difficult to hop onboard with significant expansions to our feature-set beyond without some strong justification.

We need to be really careful what we sign up for here in part because the SDK has a limited dependency load and high portability as one of its core goals. People use the SDK to _implement resource servers_, not just to implement clients.

Trying to be portable, lightweight, only-what-you-need, is in tension with trying to build something "batteries included". Batteries-included things are great, but they need to be clearly delineated and separated from the core, underlying components out of which they are built.
If this does move into the SDK, I'd want us to discuss putting it in a distinct subpackage as a "porcelain" layer.

Also, trying to treat good documentation as an "after the fact" effort is more or less doomed, in my experience. We need to be working on quality documentation as we write it, or it will never get written.

NickolausDS:
Hmm, the things that strike me as most evident is not having a larger discussion beforehand about what problems we need to solve and what would be the best way to go about solving them. 

Firstly, in response to this:

> This PR significantly expands the surface of the SDK and would constitute potentially significant maintenance burden. 

This is definitely a fair point and probably my biggest overall worry when submitting it. These changes do address some current limitations in the SDK, especially with regard to creating new standalone clients, like the Globus CLI. However, Internally the changes do expand the SDK past what it was originally intended to do (I'm thinking specifically of the config changes). 

Moving back to a discussion about the use-cases, there are two major ones:

The first is developers that want to build their own resource servers. Building the resource server is straight forward, but what if the developer needs a command line client to talk to it? The SDK contains features for doing a native app flow with a registered Globus App, but it lacks token storage and using a local server with the native auth flow. Those aren't critical features but severely dampen the user experience if they're not present, so as a result in practice the features are simply copied from the Globus CLI.

The second use-case is developers that don't really need a full-blown SDK but need to write a quick python script to use globus services. This covers both folks that only want to use core Globus services such as transfer, but is also a necessity for anyone creating their own resource server (You do need to test it at some point, after all). The same two features from the first use case are needed here, but the cost of having them for a simple script are even higher. 

Given those two use cases, as you suggest, a good solution would be implementing them in a separate package that users could import if they needed it. The only downside I can see is the need to support a separate package. Apart from that, I would argue these changes don't add any external dependencies and wouldn't impact developers that don't use them in any significant way. But again, as you suggest, putting them in a separate package then transferring them over if they're successful seems a great way to test their necessity before supporting them as part of the SDK.

Some other points:

> There are important considerations missing from this work so far. Most prominently, this only appears to implement a small part of what you need to actually manage user logins.
You need login and logout at a minimum. Session boost if you want to use HA endpoints. And you need to think about how the interface will look when consumed by multiple distinct applications. As written, two naive callers of this new function can see badly inconsistent behavior, as they "thrash" the set of scopes in use, and only function when they were the last application used for login.

This does handle login, logout, specifying custom auth params, revoking tokens, logging in with multiple applications, and handling the edge-case where a user calls this multiple times in the same application with different requested scopes, which it handles by revoking the old ones and clears them out before new ones are saved. A huge criticism of the code is it doesn't expose this functionality for developers, specifically loading or revoking tokens. However, I think there's a lot of discussion there on what we want to expose to developers to allow them to do, and at what point we would want to prevent developers from attempting to save things through the SDK vs forcing them to create their own config files. 

> This work needs to be expanded to handle more sophisticated -- but realistic -- cases before we can seriously consider it. And I'd like to get all of the maintainers involved in discussing what exactly it is we're trying to build, and why, before a bunch of code is written and we might have the unpleasant, but self-preservatory, duty of rejecting it.

This code was written in response to the most common use cases people have for doing a native auth flow, and specifically code that is commonly copied out of the Globus cli. But you're right, the scope of these changes was too large not to have had a discussion about the specific use cases. And in addition, finding best solution in solving them while keeping in mind the cost it will take to maintain them in the future.

sirosen:
We can discuss more via other fora, but...

> The SDK contains features for doing a native app flow with a registered Globus App, but it lacks token storage and using a local server with the native auth flow.

This is intentional.
I'm well aware that people want these, and _many other_ features from the SDK.

In fact, #282 was opened in part so that I could clearly lay out the rationale for not including these features in the SDK and making sure that anyone who wanted to add them would at least start by creating their separate, standalone project.
It even includes some notes on a good potential design for such a feature.

> The second use-case is developers that don't really need a full-blown SDK but need to write a quick python script to use globus services.

I think what you actually mean is "plastering over the cracks left by Globus Auth not supporting personal access tokens or developer API keys".
I am quite unhappy myself with the amount of nonsense we have to put new developers through just to get started, but I'm not yet convinced that this is the right place to solve it.

I'm _definitely_ not convinced that enough thought has gone into the design of an interface that we can expose to users in our next SDK release without shooting ourselves in the foot.

> The only downside I can see is the need to support a separate package.

When you write something, you have to support it. Otherwise, it's just more abandonware added to the web.

We've been very choosy about what features we include here, being sure to only include those that we're willing to help users troubleshoot on _any_ supported platform, including pain points like old versions of Windows running old versions of python2.7.

> This does handle login, logout, ...

Only one of these things is exported at the top level. And there's no narrative documentation to suggest or guide users through proper usage. On my first read, I saw `native_auth` (which, by the way, is very ambiguous in its meaning) as the only top-level export. So the only public-facing interface appears to be login.

Are users meant to import from `globus_sdk.utils.local_server`? Why or why not?

Consider: what does minimal, production-ready usage of your code look like? Does the documentation guide users to success?

A single function with a lot of params may be a very simple interface, but it's not a always very _good_ one.

Some of what I'm missing with this is:
- What is the interface that you're trying to design for users?
- Will it work when used correctly and fail when used incorrectly?
- Will users be able to understand what it's doing and know what side-effects it will have?
- Which parts of the problem are _you_ solving, and which parts of the problem are left for _users_ to solve?

---

At a minimum, this change seems to be adding
- The features documented as best served in an external lib in #282
- A built-in client ID; I'm not sure this is a good idea
- Read-write to a shared ini file for all client code
- A wrapper around `print`?
- A set of top-level functions, meant to be imported from `globus_sdk` and from `globus_sdk.utils.token_storage`

What people want more than anything:
- Some implementation of #282
- A pre-written `on_refresh` callback for serializing tokens to a file

```python
def gen_on_refresh(resource_server, filename):
    def on_refresh(token_response):
        tok = token_response.by_resource_server[resource_server]
        with open(filename, 'w') as f:
            f.write(json.dumps(tok))
    return on_refresh
```

🎉 But that's not really good enough for everyone. There are a million and one variations on this theme depending on what users need out of it.

We very much on purpose made _where and how you store_ tokens an external problem to be solved by the user, and _how you get tokens_ to be an internal problem solved by the SDK.
That was something we learned by looking at and using `oauth2client`, which does not distinguish the responsibilities for these two activities. That decision weighs in at around 2000 LOC (for the in-memory implementation alone!) + a clunky and uncomfortable interface.

PR #339, opened 2019-01-18
allow arbitrary params to be passed to add_item

aaschaer:
Upcoming external checksum support in transfer will allow passing an `external_checksum` and a `checksum_algorithm` per transfer data item.

I figured it would best to follow the model for other transfer API calls and allow passing arbitrary params, but if preferred I can just add the new values by name.

### DIFF
--- a/globus_sdk/transfer/data.py
+++ b/globus_sdk/transfer/data.py
@@ -165,7 +165,8 @@ class TransferData(dict):
                 )
             )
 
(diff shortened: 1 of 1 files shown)


sirosen:
Once Transfer does a release with the new functionality, I think adding it explicitly here so that it can be documented nicely would be a good (but non-essential) step.
However, I think adding `**params` is valuable regardless, because it makes the SDK methods more "future proof".

PR #341, opened 2019-02-04
Shared endpoint list

### LINKED ISSUE #318: Add support for get shared endpoint list
GET /endpoint/<endpoint_xid>/shared_endpoint_list

Preliminary doc is here:
https://github.com/globusonline/koa/blob/master/doc/manual/endpoint.adoc#get_shared_endpoint_list2

Will require a new pagination method.

Karl has also pointed out that pagination is more likely to hit rate limiting errors, but maybe that should be a separate issue since I feel like it makes the most sense to address rate limiting in the BaseClient?

aaschaer:
Resolves #318 

Turns out this API endpoint does more than just use a new pagination method, it also doesn't use `DATA` as its key for iterable data.

I decided to handle this by adding a specifiable `iter_key` to the `IterableTransferResponse` class which required passing a lot of kwargs around.

If this is too much of a mess, I would not be opposed to a more straightforward solution.

### DIFF
--- a/globus_sdk/base.py
+++ b/globus_sdk/base.py
@@ -198,7 +198,15 @@ class BaseClient(object):
     def qjoin_path(self, *parts):
         return "/" + "/".join(quote(part) for part in parts)
 
-    def get(self, path, params=None, headers=None, response_class=None, retry_401=True):
+    def get(
+        self,
+        path,
+        params=None,
+        headers=None,
+        response_class=None,
+        response_kwargs=None,
+        retry_401=True,
+    ):
         """
         Make a GET request to the specified path.
 
@@ -217,6 +225,10 @@ class BaseClient(object):
               Class for response object, overrides the client's
               ``default_response_class``
 
+            ``response_kwargs`` (*dict*)
+              Kwargs that will be passed to the response_class's __init__
+              method
+
             ``retry_401`` (*bool*)
               Retry on 401 responses with fresh Authorization if
               ``self.authorizer`` supports it
@@ -231,6 +243,7 @@ class BaseClient(object):
             params=params,
             headers=headers,
             response_class=response_class,
+            response_kwargs=response_kwargs,
             retry_401=retry_401,
         )
 
@@ -242,6 +255,7 @@ class BaseClient(object):
         headers=None,
         text_body=None,
         response_class=None,
+        response_kwargs=None,
         retry_401=True,
     ):
         """
@@ -269,6 +283,10 @@ class BaseClient(object):
               Class for response object, overrides the client's
               ``default_response_class``
 
+            ``response_kwargs`` (*dict*)
+              Kwargs that will be passed to the response_class's __init__
+              method
+
             ``retry_401`` (*bool*)
               Retry on 401 responses with fresh Authorization if
               ``self.authorizer`` supports it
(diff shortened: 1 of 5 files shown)


sirosen:
> I would not be opposed to a more straightforward solution.

I understand/share this instinct that "there must be a simpler way!" but I'm not sure there really is. There is an alternative which I'll expand on below which might be easier to reason about, and if it appeals to you as well, we can do that.

The only thing I really don't like about this is that we end up with `response_class_kwargs` in the top-level `BaseClient` methods, whereas in the past we could simply delegate any/all customization to the response_class.

One thing that occurs to me is that we could make `PaginatedResource` take a `response_class` and let `IterableTransferResponse` merely be the default.

We could then pass something like this in `TransferClient.shared_endpoint_list`:
```python
return PaginatedResource(
    self.get,
    path,
    {"params": params},
    response_class=SharedEndpointsResponse,
    paging_style=PaginatedResource.PAGING_STYLE_TOKEN,
)
```

with

```python
class SharedEnpointsResponse(IterableTransferResponse):
    ...  # non-default iter_key gets set in here
```
and then pass that through to `PaginatedResource`.

While this might turn out to be longer than the proposed change, I think it would save us logic elsewhere (e.g. in `BaseClient._request`).

Let me know if you see this as better, or if perhaps it sparks some other idea for a clean approach.

Everything other than the management of `response_class_kwargs` seems fine to me.

aaschaer:
I do like not having to mess with base.py too much.

But I'm starting to think the cleanest option might be to change the Transfer API itself.

Both to use the consistent DATA key and to return more than just the id of each endpoint, since as it is we would have to make more round trips back to Transfer if we want the CLI's output to match between `shared-endpoint-list` and `my-shared-endpoint-list`

sirosen:
If changing Transfer is easy enough, I'm onboard with that. Staying consistent in using `DATA` as the key would be nice -- I suspect the webapp and other clients would benefit from that as well.

I'm going to mark this `blocked: implementation` for now, until we have whatever changes are wanted in Transfer preview or test. Just ping me and untag -- or close and open a replacement PR -- when you want me to look again.

jaswilli:
I’ve been meaning to respond to this to get clarification and it seems like the above answers my question which was, “transfer has an api that returns a non-standard response?”

That seems strange and unless there’s a reason I think we should change it to conform to the standard envelope response.

I’m also a little surprised that yet another paging style has been introduced in transfer. Why wasn’t one of the existing methods sufficient?

sirosen:
> I’m also a little surprised that yet another paging style has been introduced in transfer. Why wasn’t one of the existing methods sufficient?

You mean: Haven't we implemented all theoretically possible paging styles at this point? 😂 

I wasn't going to bring this up, on the assumption that the Transfer piece was locked in at this point. If it's possible to change, it looks to me like this is another variant on the `marker` style and it would be nice to just call it `marker` so that we can use that in this case too.

However, if there are reasons that it's important/necessary for Transfer to call this something different, it doesn't seem like a big deal to me to add the option to `PaginatedResource` (which, let's be honest, is already pretty wild anyway).

sirosen:
I tried a rebase to see how this could clean up with `SharedEndpointListResponse` inheriting from `IterableTransferResponse` and that part is fine.

But pagination is super-hairy and I don't think we're going anywhere on this at present. I could see us picking this effort up again if there are appropriate changes in the backend, but as it stands I would not like this to be left open indefinitely. Feel free to reopen or ping me if there's some change in the situation.

PR #344, opened 2019-02-21
Release v1.7.1

jaswilli:


sirosen:
Thanks AppVeyor. I think it's now actually gone, but of course dirties this PR.
I looked at `git diff 1.7.0..master` just to be sure, and this is going to pass and I see no reason to wait for Travis.

I'll run a release after merging (as a reminder of who does releases, since _I_ have definitely forgotten in the past).

PR #346, opened 2019-03-14
Property to reveal more results available

jaswilli:
### Context

Discussions have taken place about the Transfer API endpoint search, the number of results returned and how a user might be able to know the total number of results or some other way to know they aren't seeing everything that's available.

### Proposal
This is one proposed solution which adds a property to the PagedResponse class that will become true if the underlying API indicates there are more results available than the user requested (i.e., by way of a limit parameter).

Example:

```
import globus_sdk

tc = globus_sdk.TransferClient(
    authorizer=globus_sdk.AccessTokenAuthorizer(access_token=<token>)
)

res = tc.endpoint_search(filter_fulltext='globus', limit=5)
endpoints = list(res)

assert res.limit_less_than_available_results is True
```

Which then could be leveraged in the CLI to inform the user that they're not seeing all available results.

### DIFF
--- a/globus_sdk/transfer/paging.py
+++ b/globus_sdk/transfer/paging.py
@@ -129,6 +129,9 @@ class PaginatedResource(GlobusResponse, six.Iterator):
                 path,
             )
         )
+
+        self._limit_less_than_available_results = False
+
         self.max_results_per_call = max_results_per_call
         self.max_total_results = max_total_results
         self.offset = offset
@@ -179,6 +182,19 @@ class PaginatedResource(GlobusResponse, six.Iterator):
             # express this internally as "generator is null" -- just need some
             # way of making sure that it's clear
             self.generator = None
+
+    @property
+    def limit_less_than_available_results(self):
+        """
+        Indicates that the Transfer API had more results
+        available than were requested by way of the `limit` parameter.
+
+        Note: this will always be false until the iterator containing the
+        results has been exhausted.
+
+        :rtype: bool
+        """
+        return self._limit_less_than_available_results
 
     @property
     def data(self):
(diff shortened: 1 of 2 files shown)


sirosen:
I have to look at this in more detail, hopefully tomorrow, but I'm 👍 on exposing this as a property.

> Which then could be leveraged in the CLI to inform the user that they're not seeing all available results.

Perhaps not for right now / in this thread, but I'm curious how you're thinking this would work. A warning on stderr, perhaps? Ping me out of band if you want to discuss in depth.

globus_sdk/transfer/paging.py
```diff
@@ -180,6 +183,19 @@ def __init__(
...
             self.generator = None
 
+    @property
+    def limit_less_than_available_results(self):
+        """
+        Indicates that the Transfer API had more results
+        available than were requested by way of the `limit` parameter.
+
+        Note: this will always be false until the iterator containing the
+        results has been exhausted.
```
sirosen: The phrasing you've used is accurate, but I can't shake the feeling that this is confusing.

The issue is that the iterator can be exhausted _without_ setting this to True if `num_results=None`, but if the API is limiting our results, you still got "less than available".

But I don't see a way to make it better than what you've done.

jaswilli: Yeah, I'm not invested in this name in any way. I went through a couple and landed on this for lack of anything better.

sirosen: Not a fan of `limit_param_was_obeyed_and_thats_why_you_got_fewer_results_than_available_but_if_false_could_indicate_that_other_limiting_factor_was_at_play`?
That's my first instinct, obviously. Explicit is better than implicit.

sirosen:
After thinking about this over the weekend, I can come up with equivalently confusing names or _very_ long names, so I'm going with this and merging.

PR #351, opened 2019-03-30
Add changelog doc, tweak contrib doc very mildly

sirosen:
Not a big deal, but I sometimes notice the absence of a singular changelog doc. Most prominently, there's no trivially easy view of "what hasn't been released yet". Adding a changelog and having a typical "Unreleased" section makes this nicer.

Add a changelog generated from GitHub release data. (Use md format so that we can just `#NUM` for issue links in GitHub).

Also mention the Unreleased section of the changelog in the contrib doc.

I used this script to generate it:
```
$ cat assemble_changelog.py
#!/usr/bin/env python

import requests

releases = requests.get(
    "https://api.github.com/repos/globus/globus-sdk-python/releases",
    headers={"Accept": "application/vnd.github.v3+json"},
).json()

print("CHANGELOG")
print("=========")
print("")
print("Unreleased")
print("----------")
for release in releases:
    print("")
    print(release["name"])
    print("-" * len(release["name"]))
    print("")
    print(release["body"].replace("\r\n", "\n"))
```
and then did some very mild reformatting to be a little more stylistically consistent.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,0 +1,229 @@
+CHANGELOG
+=========
+
+Unreleased
+----------
+
+* Add a property to paginated responses to indicate that there are more results available than the user requested (`#346`_)
+* Add CHANGELOG file to the repo; not just GitHub releases
+
+.. _#346: https://github.com/globus/globus-sdk-python/pull/346
+
+v1.7.1
+------
+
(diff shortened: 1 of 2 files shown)


sirosen:
Oy, forgot that GitHub autolinking of issue numbers doesn't happen in repo files, only issues and so forth. Will amend to use adoc and plain-old links.
Also, I should point at this from the README.

PR #365, opened 2019-08-26
add external_checksum and checksum_algorithm

aaschaer:
Adding these as explicit named arguments along with doc for how they are used by Transfer.

First PR since the linting and pytest changes, but I think `make test` did the right thing?


aaschaer:
chery-picked the commit and the indenting seems fixed, thank you!

PR #367, opened 2020-01-03
Add globus_sdk.IdentityMap for Auth ID lookups

### LINKED ISSUE #366: Feature: Add LazyIdentityMap, based on globus-cli's LazyIdentityMap
I've found cases outside of the globus-cli where this pattern would be useful.
I also believe it would be healthy to rebuild this tool with some better testing (and less time pressure). I don't intend to work on this starting today, but I'm putting it in the tracker (a) as a wish-list item and (b) in case someone wants to take a crack at contributing it.

[`LazyIdentityMap` is defined here](https://github.com/globus/globus-cli/blob/acf3f09672c7d6eda2ca37e424a12aa824cec555/globus_cli/services/auth.py#L80-L115) and is specifically meant to speed up rendering of results to the text formatted output. It's batching lookups in groups of 100, and doesn't order the lookups or anything. It's little more than a mild abstraction over a dict.

Here's what I would do if we decided to make it an SDK-level tool:
1. Change the parameters a bit:
  - Accept IDs or Usernames as inputs
  - Take the AuthClient object as an input
  - Allow tuning of the batch size
2. Map to the full identity object in the response body, not just one field within it
3. When a value is looked up, only make one request to populate that part of the map (don't eagerly look up everything)
4. Allow more IDs to be added to an existing map (mostly because this is easy and I see no reason not to allow it)
5. Support lookups with `__getitem__`, possibly implement the [full `Mapping` protocol](https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping) by supporting `__iter__` and `__len__`
6. Test, test, test. Because the SDK testsuite is based on fully mocking the socket layer, we can test things in interesting ways. But an additional important test is to modify the globus-cli to use the SDK implementation and ensure that it works (e.g. run _its_ testsuite, run some acl listing commands and other relevant exercises).

sirosen:
The IdentityMap is mostly a port of the LazyIdentityMap found within the Globus CLI. It has several notable differences, however:

- it accepts usernames or identity IDs and internally differentiates them
- it is lazier than the LazyIdentityMap because it only looks up one batch at a time (maybe the CLI implementation should have been called EagerIdentityMap)
- it has an external interface more clearly defined for consumers to interact with it (because that's the point of doing this work, basically)
- it will "prioritize" a given key by forcing it into a record batch for lookup
- it returns full record dicts, not just usernames

In addition to the IdentityMap itself, this includes tests which exercise almost all of the IdentityMap behaviors.
The IdentityMap is primarily a wrapper over an internal `_cache` dictionary which maps both usernames and IDs, undifferentiated, to records. You can `del` items from the cache to force fresh lookups if you believe data has gone stale, and `IdentityMap.get` has the same signature as `dict.get`.

It's important to note that IdentityMap does not implement the full `Mapping` protocol (defined in modern pythons by the collections.abc module). The issue for us in this case is that certain operations like `__contains__` are very unclear. For example, if the map has a key added to it but it hasn't looked it up yet, does it "contain" that key? If you were checking to see if a key had already been added the
answer is "yes". But if you wanted to know if `idmap[key]` will KeyError, the answer is "maybe/unknown". The only way to answer "contains" if you want to know that is to actually do a lookup, and having `in` checks side-effecting in python seems bad.

Therefore, `in` is intentionally left undefined. We can add new methods, similar to `has_key(key, resolved/unresolved=True/False)` in the future if users come to us with a desire for this information and can clarify their use-cases.
For the most part, `idmap.get(key) is not None` should suffice if anyone ever cares to check.

I worked through tests mostly with the goal of complete line coverage, but there are some untested behaviors I can target if it seems important. 

Absent a more obvious place to put this in docs, I've included it on the Auth clients page (similar to the `TransferData` and `DeleteData` objects).

Resolves #366

### DIFF
--- a/globus_sdk/auth/identity_map.py
+++ b/globus_sdk/auth/identity_map.py
@@ -1,11 +1,16 @@
-import re
-
-# what qualifies as a valid Identity Username?
-_USERNAME_PATTERN = re.compile(r"^[a-zA-Z0-9]+.*@[a-zA-z0-9-]+\..*[a-zA-Z]+$")
+import uuid
 
 
 def is_username(val):
-    return _USERNAME_PATTERN.match(val) is not None
+    """
+    If the value parses as a UUID, then it's an ID, not a username.
+    If it does not parse as such, then it must be a username.
+    """
+    try:
+        uuid.UUID(val)
+        return False
+    except ValueError:
+        return True
 
 
 def split_ids_and_usernames(identity_ids):
@@ -49,7 +54,7 @@ class IdentityMap(object):
     - seed the ``IdentityMap`` with IDs and Usernames via :py:meth:`~IdentityMap.add` (you
       can also do this during initialization)
 
-    - retreive identity IDs or Usernames from the map
+    - retrieve identity IDs or Usernames from the map
 
     Because the map can be populated with a collection of identity IDs and Usernames
     prior to lookups being performed, it can improve the efficiency of these operations
@@ -71,7 +76,7 @@ class IdentityMap(object):
     Correct usage looks something like so::
 
         ac = globus_sdk.AuthClient(...)
-        idmap = globus_sdk.ext.IdentityMap(
+        idmap = globus_sdk.IdentityMap(
             ac, ["foo@globusid.org", "bar@uchicago.edu"]
         )
         idmap.add("baz@xsede.org")
--- a/globus_sdk/auth/identity_map.py
+++ b/globus_sdk/auth/identity_map.py
@@ -1,0 +1,222 @@
+import re
+
+# what qualifies as a valid Identity Username?
+_USERNAME_PATTERN = re.compile(r"^[a-zA-Z0-9]+.*@[a-zA-z0-9-]+\..*[a-zA-Z]+$")
+
+
+def is_username(val):
+    return _USERNAME_PATTERN.match(val) is not None
+
+
+def split_ids_and_usernames(identity_ids):
+    ids = set()
+    usernames = set()
+
+    for val in identity_ids:
+        if is_username(val):
+            usernames.add(val)
+        else:
+            ids.add(val)
+
+    return ids, usernames
+
+
+class IdentityMap(object):
+    r"""
+    There's a common pattern of having a large batch of Globus Auth Identities which you
+    want to inspect. For example, you may have a list of identity IDs fetched from
+    Access Control Lists on Globus Endpoints. In order to display these identities to an
+    end user, you may want to resolve them to usernames.
+
+    However, naively looking up the identities one-by-one is very inefficient. It's best
+    to do batched lookups with multiple identities at once. In these cases, an
+    ``IdentityMap`` can be used to do those batched lookups for you.
+
+    An ``IdentityMap`` is a mapping-like type which converts Identity IDs and Identity
+    Names to Identity records (dictionaries) using the Globus Auth API.
+
+    .. note::
+
+        ``IdentityMap`` objects are not full Mappings in the same sense as python dicts
+        and similar objects. By design, they only implement a small part of the Mapping
+        protocol.
+
+    The basic usage pattern is
+
+    - create an ``IdentityMap`` with an AuthClient which will be used to call out to
+      Globus Auth
+
+    - seed the ``IdentityMap`` with IDs and Usernames via :py:meth:`~IdentityMap.add` (you
+      can also do this during initialization)
+
+    - retreive identity IDs or Usernames from the map
+
+    Because the map can be populated with a collection of identity IDs and Usernames
+    prior to lookups being performed, it can improve the efficiency of these operations
+    up to 100x over individual lookups.
+
+    If you attempt to retrieve an identity which has not been previously added to the
+    map, it will be immediately added. But adding many identities beforehand will
+    improve performance.
+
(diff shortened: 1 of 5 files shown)


globus_sdk/auth/identity_map.py
```diff
@@ -0,0 +1,222 @@
+import re
+
+# what qualifies as a valid Identity Username?
+_USERNAME_PATTERN = re.compile(r"^[a-zA-Z0-9]+.*@[a-zA-z0-9-]+\..*[a-zA-Z]+$")
```
jaswilli: I'm wondering if we shouldn't do this check from the other direction, i.e., check if the input is a UUID. That regex clearly matches things that aren't valid identity usernames and it's also complex enough that it conveys a false sense of security. I also worry that it might get reused somewhere else...

Anyway, I think what we really care about here is whether or not the input is a UUID so we know to treat it as an identity id--everything else gets dumped into the username bucket.

globus_sdk/auth/identity_map.py
```diff
@@ -0,0 +1,222 @@
...
+
+    The basic usage pattern is
+
+    - create an ``IdentityMap`` with an AuthClient which will be used to call out to
+      Globus Auth
+
+    - seed the ``IdentityMap`` with IDs and Usernames via :py:meth:`~IdentityMap.add` (you
+      can also do this during initialization)
+
+    - retreive identity IDs or Usernames from the map
```
jaswilli: retrieve

globus_sdk/auth/identity_map.py
```diff
@@ -0,0 +1,222 @@
...
+
+    .. warning::
+
+        If an Identity is not found in Globus Auth, it will trigger a KeyError when
+        looked up. Your code must be ready to handle KeyErrors when doing a lookup.
+
+    Correct usage looks something like so::
+
+        ac = globus_sdk.AuthClient(...)
+        idmap = globus_sdk.ext.IdentityMap(
```
jaswilli: This isn't in an `ext` namespace anymore, right?

globus_sdk/auth/identity_map.py
```diff
@@ -0,0 +1,222 @@
...
+        self.auth_client = auth_client
+        self.id_batch_size = id_batch_size or self._default_id_batch_size
+
+        # uniquify, copy, and split into IDs vs usernames
+        self.unresolved_ids, self.unresolved_usernames = split_ids_and_usernames(
+            [] if identity_ids is None else identity_ids
+        )
+
+        # the cache is a dict mapping IDs and Usernames
+        self._cache = {}
```
jaswilli: Should we consider sharing the cache between instances? I know it likely won't ever apply for CLI usage but since this is in the SDK now...

globus_sdk/auth/identity_map.py
sirosen: Yeah, that seems very reasonable.
I think I pulled this regex out of globus-cli source (I'll dig it up later), so it's already "getting reused somewhere else".

I'll switch to a check that does `try: uuid.UUID(value) ; except ValueError: ...`

globus_sdk/auth/identity_map.py
sirosen: Good catch, thank you. Yes, per our brief chat on that subject, no more `globus_sdk.ext` subpackage. I'll grep for other cases of it.

globus_sdk/auth/identity_map.py
sirosen: I think the way I would approach that, if anyone ever wants it, is to let you pass a `cache` into the constructor, and default to constructing a new dict.
So we could, someday, have
```python
cache = {}
idmap1 = globus_sdk.IdentityMap(authclient, cache=cache)
idmap2 = globus_sdk.IdentityMap(authclient, cache=cache)
```

The advantage of that is that it lets us support other mappings, like

```python
cache = build_memcache_backed_id_cache()
idmap = globus_sdk.IdentityMap(authclient, cache=cache)
```

but I thought that was all too fancy / unnecessary for v1. Would you like me to take a crack at it, and see if it's simple enough to include?

jaswilli: I was thinking of a middle ground of simply making the cache a module level variable because I think it's unlikely that the casual user of this SDK is going to make advanced use of being able to inject their own custom cache.

sirosen: I really don't like the idea of anything which is going to persist over the lifetime of the interpreter.

If someone builds a long-running application and doesn't know that there's a persistent cache underneath, they could get stale results and know that it's happening. It also means that memory usage for the cache will grow over time.

Maybe it's best to just leave this as-is?

sirosen:
I've done some cleanup per review, and also because I realized that I had used httpretty in a slightly confusing way.

I also went ahead and implemented the "pass a custom cache" variant, but kept it in a separate branch ( https://github.com/sirosen/globus-sdk-python/commit/bbf7983ccd94ee70af2471e35070add0abc16768 ). We can pull it in, or just discard. (Or, if you strongly disagree and want a default shared cache, we can discuss further, of course.)

globus_sdk/auth/identity_map.py
jaswilli: I'm fine with leaving it as is. If anyone else ends up using this and wants to optimize the caching behavior I suppose they're free to implement a singleton on top of this to get it done.

PR #370, opened 2020-03-27
Doc typos

pn2200:
Fix a few typos in example documentation.


sirosen:
Thank you for these fixes! The `scopes` vs `scope` one is particularly nefarious -- I'll try to audit our repos for other cases of it, in case it's out there.

PR #372, opened 2020-07-21
Add pre-commit config, convert to tox

sirosen:
Replace makefile logic for managing venvs with tox usage and add pre-commit config.
Also, trim down the sphinx config so that linting can pass reasonably on it.

Update contrib doc to match. Separate linting and testing requirements from the recommendations (editor config, pre-commit). Note that testing requires tox.

Removes some extra targets from tox and Travis (like `lowestdeps` and multi-platform linting).

---

At this point, I've been using pre-commit happily in other projects and repos, and I think it's time to overhaul and modernize this config. It's also nice to use tox a bit more and clear out some of the somewhat-painful stuff in the makefile.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,28 @@
+repos:
+- repo: https://github.com/pre-commit/pre-commit-hooks.git
+  rev: v0.9.1
+  hooks:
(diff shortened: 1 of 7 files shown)


jaswilli:
This all seems pretty reasonable.

PR #373, opened 2020-07-22
Cleanup linting and dev tool versions

sirosen:
- synchronize pre-commit config with setup.py (oops)
- move doc build to modern things (sphinx-material)
- fix classifiers to reflect what we really test
- apply rtd config and update travis to do a doc build (so you can't merge with docs failing)

This is the cleanup we talked about, mostly from noticing that I got sloppy when adding pre-commit config.
I'm intentionally holding back `black` because of a bug in the latest version which I've noticed causing issues in other projects, so I've commented it in both places.

A lot of tools are moving to python 3.6+, and I'm not really fighting it. But I've put version specifiers in because we still have our 2.7 build (for now, at least).

Note that `sphinx-material` handles `:param <paramname>:` and `:type <paramname>:` in docstrings in a not-totally-trashcan-on-fire way (vs the guzzle theme), so I've looked at updating the docs to use that. I might want to do a follow-up PR to do the conversion. I think it will have to be by hand though, since I can't think of an efficient way to script it.


setup.py
```diff
@@ -49,11 +49,14 @@
             "tox>=3.5.3,<4.0",
             # linting
             "flake8>=3.0,<4.0",
-            "isort>=4.3,<5.0",
+            "isort>=5.1.4,<6.0",
```
jaswilli, line 52: I think this is going to kill 2.7

sirosen: I'm being punished for not running `tox` before opening this PR. Yeah, I'll add the version specifier here too... 😢

sirosen:
Ah, I forgot about the warning, though you mentioned it earlier. Yes, I'll tweak that when I amend to fix the warning for 3.4

setup.py
```diff
@@ -49,11 +49,14 @@
             "tox>=3.5.3,<4.0",
             # linting
             "flake8>=3.0,<4.0",
-            "isort>=4.3,<5.0",
+            'isort>=5.1.4,<6.0;python_version>="3.6"',
```
jaswilli, line 52: If we're going to make the switch to isort 5 we should be able to trim our config down to:

```
[isort]
profile = black
known_first_party = tests, globus_sdk
```

sirosen: I didn't know about `profile = black`. Thank goodness.
`flake8` rejected the idea of such a mode.

Let's save that for follow-up, if that's okay with you.

sirosen:
Okay, I got tripped up by the fact that `flake8-bugbear` dropped 3.5 support, but this is all cleaned up now. (Yeesh!)

PR #383, opened 2020-12-02
Update testing matrix and switch from httpretty to responses

sirosen:
`responses` is simpler for us to use than `httpretty`, and because we wrapped our primary usage of `httpretty` (hooray!) in the past, the replacement is basically painless.
Some bits here and there need to be twiddled, of course.

I gave up on tracking down exactly why the tests have started failing again, but httpretty was clearly somehow involved/at-fault.

`responses` isn't at 1.0, but it appears to be stable. I think it's doing the thing that m2crypto used to do, in which 1.0 is never released so that backwards compatibility and semver is never something they have to think about.

In addition to dropping 3.5 and adding 3.9 to testing, this makes the matrix a bit smaller.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -7,7 +7,14 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ["2.7", "3.5", "3.6", "3.7", "3.8"]
+        python-version: ["2.7", "3.9"]
+        include:
+          - os: ubuntu-latest
+            python-version: 3.6
+          - os: ubuntu-latest
+            python-version: 3.7
+          - os: ubuntu-latest
+            python-version: 3.8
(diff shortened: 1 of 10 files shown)


jaswilli (commented):
I never really scrutinized the usage of `httppretty` so I have no real opinion/a lateral move to a different library is fine with me.

I'm curious about the switch away from `six.b()` in some of the tests though. Is that because `httppretty` was using explicit bytestrings and `responses` just uses native strings?

Otherwise seems good to me.

sirosen:
> Is that because `httppretty` was using explicit bytestrings and `responses` just uses native strings?

Yeah, it's tied to what the "request" object is in each library. In `httpretty` it's their own specialized mock, and since it lives down at the socket layer everything is done in bytes. In `responses`, you're actually handling a `requests.PreparedRequest`, which (apparently, I just learned this yesterday) operates in unicode strings.

I didn't expect to need those changes at the outset, but when things failed because of it I was not that surprised.

PR #386, opened 2020-12-04
Update dev dependencies

jaswilli:
Bumps a few of the development dependencies for no reason other than to keep current.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,15 +12,15 @@ repos:
       types: [python]
       language_version: python3
 - repo: https://gitlab.com/pycqa/flake8
-  rev: 3.8.3
+  rev: 3.8.4
   hooks:
     - id: flake8
       name: "Lint python files"
       types: [python]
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -5,25 +5,22 @@ repos:
     - id: check-merge-conflict
     - id: trailing-whitespace
 - repo: https://github.com/python/black
-  # refrain from using 19.10b0 or later until
-  #   https://github.com/psf/black/issues/1288
-  # is fixed
-  rev: 19.3b0
+  rev: 20.8b1
```
jaswilli, line 8: https://github.com/psf/black/issues/1288 was resolved in the 20.8b0 [release](https://github.com/psf/black/issues/1288)

sirosen: It took a long time but I've been happy with the behavior in the `>=20.8b0` versions.
We should check CLI for this as well. I bet it's still on 19.3b0

PR #387, opened 2020-12-15
Add a pyinstaller hook and simple test

### LINKED ISSUE #384: Support installation with pyinstaller
## User Story / Use-case

If you want to build a cross-platform application which bundles python, pyinstaller is one of the ways, if not _the_ de-facto standard way, to do it.
In order to do so, pyinstaller has to bundle all of the dependency packages for the application, including the SDK, requests, six, etc.

## Problem

Right now, an installation with pyinstaller fails because pyinstaller doesn't respect normal python package metadata. It doesn't use setuptools and it won't read `setup.py`, `setup.cfg` or `MANIFEST.in`.
Instead, pyinstaller has [hooks](https://pyinstaller.readthedocs.io/en/stable/hooks.html) and pre-defines a whole host of them for "common" python packages. We're obviously not in that list, so `globus_sdk/globus.cfg` is not loaded by pyinstaller and you get a borked install.

We've had this come up in user support, but we also have run into it internally where we use pyinstaller.

Given that we provide a very boring (metadata-wise) package by the most standard of means, it's kind of absurd that we're not supported out of the box. But whining about it won't fix things, and pyinstaller has drawn a clear line in the sand that they are not going to add support for any python build tools. (Given the pyproject.toml and "pluggable build system" changes, I'm not even sure I blame them...)

## Implementation Options

1. For completeness: we could get rid of `globus.cfg` and have no non-python package data. I don't like this option at all for a whole host of reasons. Let's not.

2. Get a hook built into pyinstaller. (Won't happen; they'd be sure to reject it.)

3. pyinstaller has a contrib package for adding hooks: https://github.com/pyinstaller/pyinstaller-hooks-contrib
We could try to get a hook added there.

4. We could write and add a hook to `globus_sdk`.

I think 3 and 4 are the only viable ones in this list, and I like 4 best because it works with pyinstaller out of the box and we're not beholden to anyone else who works on that contrib repo. The contrib hooks are really for cases like `boto3` where a hook is needed and someone who is _not_ affiliated with the project wants to add it.

My inclination right now is to write and add a hook, add a CI job which builds and runs a very minimal application with pyinstaller, and call it a day.

sirosen:
closes #384

This adds a pyinstaller hook and a simple test to the testsuite which tries to do a pyinstaller build of a miniscule (two line) app.

The pyinstaller hook is quite simple -- it just ensures that non-python files are included in the SDK package. The hook itself needs a setuptools entry point which identifies the hook location (one of the only bits of setuptools config pyinstaller respects).

In order for the testing to work, a few steps are needed:
- add pyinstaller to 'development' requirements for py3+
- add the test to the testsuite
- skip the test if the environment doesn't support a pyinstaller build

Skipping in the right contexts is the trickiest part. It seems that pyinstaller needs a shared-library python build. That's the default for a lot of system packaging, but it's not guaranteed. Furthermore, pyenv builds won't include shared libraries by default, but will use static linking instead. Precisely because this is so finicky, we want to allow tests without it. That way, developer laptop testing can work even if the python env isn't "just so".

GitHub Actions should build pythons appropriately to run the pyinstaller test, so we have a good guarantee of things working.

We don't test the pyinstaller hook for py2 -- if it doesn't work for that version, that's fine (wontfix).

---

The one thing I wasn't sure about was where to put internal doc on this subject. Basically, I've imitated the sample repo from pyinstaller ( https://github.com/pyinstaller/hooksample ) and tried not to think too hard about this. But if we need more internal doc, maybe I could add `globus_sdk/_pyinstaller/README`? I'm just not certain it's necessary.

### DIFF
--- a/globus_sdk/_pyinstaller/__init__.py
+++ b/globus_sdk/_pyinstaller/__init__.py
@@ -1,0 +1,5 @@
+import os
+
+
+def get_hook_dirs():
+    return [os.path.dirname(__file__)]
--- a/globus_sdk/_pyinstaller/hook-globus_sdk.py
+++ b/globus_sdk/_pyinstaller/hook-globus_sdk.py
@@ -1,0 +1,18 @@
+#
+# this is the pyinstaller hook which describes how to collect data to freeze the
+# 'globus_sdk' correctly
+#
+# pyinstaller hooks are modules which define a number of known variables, which
+# pyinstaller then loads and uses
+#
+#
+# This module is based on the PyInstaller Hook Sample
+# see:  https://github.com/pyinstaller/hooksample
+#
+
+from PyInstaller.utils.hooks import collect_data_files
+
+# "datas" is a list of files which need to be included as data files
+# collect_data_files will crawl the source tree and grab up *all* non-python files
+# exclude any files in the pyinstaller dir
+datas = collect_data_files("globus_sdk", excludes=["_pyinstaller"])
--- a/setup.py
+++ b/setup.py
@@ -61,6 +61,8 @@ setup(
             'mock==2.0.0;python_version<"3.6"',
             # mocking HTTP responses
             "responses==0.12.1",
+            # pyinstaller is needed in order to test the pyinstaller hook
+            'pyinstaller;python_version>="3.6"',
             # builds + uploads to pypi
             'twine==3.2.0;python_version>="3.6"',
             'wheel==0.34.2;python_version>="3.6"',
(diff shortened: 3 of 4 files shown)


jaswilli:
Without having any experience with pyinstaller this all seems pretty reasonable. 

I can:
* Produce a broken pyinstaller bundle by using the sdk as a dependency in a test script
* See that I can build a working bundle by swapping in this branch.

Good enough for me.

PR #389, opened 2020-12-18
Push release tag to upstream during release

jaswilli:
This adds an attempt to push the release tag to the caller's upstream during `make release`.


sirosen:
Cool. 👍 

As long as it has the `-git` (I hate `make` sometimes...) there's no harm in trying.

Reminds me that I wanted to look into the GitHub Actions pypi publisher action, which could possibly be driven off of the tag being pushed, to simplify our work here a bit.

PR #393, opened 2021-01-27
Skip and quota

aaschaer:
Adds named arguments for skip_source_errors and fail_on_quota_errors to TransferData and methods for the new GET /task/<task_id>/skipped_errors and GET /endpoint_manager/task/<task_id>/skipped_errors to the TransferClient

### DIFF
--- a/globus_sdk/transfer/client.py
+++ b/globus_sdk/transfer/client.py
@@ -1419,8 +1419,8 @@ class TransferClient(BaseClient):
         .. note::
 
             This does not include files that were checked but skipped as part of a sync
-            transfer, only files that were actually transferred, and does not include
-            any directories.
+            transfer or skipped due to skip_source_errors, only files that were
+            actually transferred, and does not include any directories.
 
         ``GET /task/<task_id>/successful_transfers``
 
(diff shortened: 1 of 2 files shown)


globus_sdk/transfer/client.py
```diff
@@ -1772,6 +1821,52 @@ def endpoint_manager_task_successful_transfers(
...
+        <https://docs.globus.org/api/transfer/advanced_endpoint_management/#get_task_skipped_errors_as_admin>`_
+        in the REST documentation for details.
+        """
+        task_id = safe_stringify(task_id)
+        self.logger.info(
+            (
+                "TransferClient.endpoint_manager_task_"
+                "skipped_errors({}, ...)".format(task_id)
+            )
+        )
```
sirosen, line 1856: This doesn't really matter, but may I suggest
```python
        self.logger.info(
            "TransferClient."
            "endpoint_manager_task_skipped_errors({}, ...)".format(task_id)
        )
```
as an alternative? The long-line rule may demand that we break things up, but I think the above is more readable because it avoids breaking a "long word".

globus_sdk/transfer/client.py
```diff
@@ -1419,8 +1419,8 @@ def task_successful_transfers(self, task_id, num_results=100, **params):
         .. note::
 
             This does not include files that were checked but skipped as part of a sync
-            transfer, only files that were actually transferred, and does not include
-            any directories.
+            transfer or skipped due to skip_source_errors, only files that were
+            actually transferred, and does not include any directories.
```
sirosen, line 1423: This sentence is now very long; I think we should break it up.

I suggest:
> Only files that were actually transferred are included.
> This does not include files that were checked but skipped as part of a sync
> transfer, files which were skipped due to ``skip_source_errors``, or
> directory names.

But if you have an alternate phrasing you prefer, go for it.

sirosen (commented):
The logical lines of code are definitely fine/correct (although I want to see and read the new API docs before approving, just to triple-check everything).

I have a couple of bits of feedback on inline docs.

globus_sdk/transfer/data.py
```diff
@@ -134,6 +134,8 @@ def __init__(
         preserve_timestamp=False,
         encrypt_data=False,
         deadline=None,
+        skip_source_errors=False,
+        fail_on_quota_errors=False,
```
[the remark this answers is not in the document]
aaschaer: Ah missed those, added!

globus_sdk/transfer/client.py
aaschaer: Agreed - I think this might have been auto-formatted since I tend to err in the direction of too long a line ^^"

globus_sdk/transfer/client.py
aaschaer: Thank you! I slightly adjusted the phrasing to put directories first since that seems the most common case.

sirosen:
I made a very minor fix to `:param ...` in the TransferData docstring, but other than that, all was spot on.
Thanks for handling the fixes, and 👍 to your rephrase of the successfully transferred files explanation.

PR #395, opened 2021-01-29
Fixup variable reference

jaswilli:


sirosen:
Gah! No! Unbelievable! make is my favorite program.

PR #396, opened 2021-01-30
Drop python2

sirosen:
- Remove it from package data and the build matrix
- Apply `pyupgrade` and invoke that in cycles until it, `black`, and a handful of manual edits for `flake8` converge
- Remove `six`
- blacken-docs and look for py2-compatible bits of doc (turned out just to be a hunt for `raw_input`)

There are so many files changed by pyupgrade, it's a bit dizzying. I'm going to use GitHub to self-review and try to double-check everything, but obviously this deserves a careful read.

We have the `1.x-line` branch if we ever need to go back. This is for `main`.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ["2.7", "3.9"]
+        python-version: ["3.9"]
         include:
           - os: ubuntu-latest
             python-version: 3.6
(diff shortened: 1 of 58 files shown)


jaswilli (approved):
I read through this and tested it against the sample data portal workflows without issue. I'm satisfied.

I'm not going to merge it because it sounded like you might still be looking at it.

sirosen:
Thanks for the speedy review!
I plan to read this over again tomorrow or Monday. I'll merge it when I'm satisfied that I've caught everything I reasonably can.

sirosen (commented):
I used the GitHub web review tool to go file-by-file and it was actually a pretty pleasant experience. I noted a couple of string formatting oddities which I'll come back to in a (tiny) follow-up PR.

PR #398, opened 2021-02-01
Remove six usage

jaswilli:


sirosen:
Thanks for the catch. Although not a problem today, I bet it would have broken when `cryptography` v3.4 lands and drops py2 support.

PR #403, opened 2021-02-23
Refactor decode_id_token (decompose it)

### LINKED ISSUE #298: Make it possible to cache JWK (external to SDK) and use it for decode_id_token
Low priority.

I realized while answering a listhost thread that even though `OAuthTokenResponse.decode_id_token` could be optimized (in theory) to use a cached copy of the JWK and therefore no network callout, we don't leave that avenue open.

Caching should not be done in the SDK for a couple of reasons:
- Unnecessary complexity and bloat
- Multithreaded and multiprocess SDK usage don't share client objects

We could simply expose some methods for (1) getting the JWK and (2) using that JWK with `decode_id_token`.
Following the model we used for TransferData submission_ids, it could be:
- `AuthClient.get_jwk()`
- `AuthClient.decode_id_token(jwk=string)`
with the default behavior for `decode_id_token()` obviously being what it is today.

sirosen:
decode_id_token() previously did two round-trips to Auth + a decoding step on the JWK prior to handling the ID token. This is prohibitively expensive if anyone were to try to build a service using decode_id_token. Instead, you would have to rewrite the function yourself.

So this is now broken down into two fetch methods, one of which can optionally do a decoding step, and their results *may* be passed to decode_id_token. As a result, loading data over the network can always be taken care of separately from parsing an id token, and parsing a JWK to a public key object can also be included in those steps.

AuthClient.get_openid_configuration is a new method for fetching OIDC config.
AuthClient.get_jwk is a new method for getting, and optionally parsing the JWK.

Unfortunately, the JWK URI provided by OIDC config is a full string including `https://`, not just the path component. We could strip the leading `https://auth.globus.org` from it, but I chose to allow `BaseClient._request` to handle requests to full URLs. It's a single line of dispatch there vs parsing and unparsing the JWK URI.

get_jwk can take OIDC config as a parameter, or fetch it automatically.

decode_id_token can take both the OIDC config and the JWK as paramters, or fetch and use both automatically (and it will pass OIDC config to the JWK to avoid a double-fetch of that data).

If an application wants to cache the parsed JWK and OIDC config, that is possible, and they should simply be passed back to the decode_id_token function. Caching the parsed JWK would require defining some serialization format, but that's not the SDK's problem to solve. And we could make things more variable by letting parsing of the JWK to a public key be separated out even further, but that does not seem to be necessary.

closes #298

---

I'm submitting this as a draft so that I can ask if this is the right interface to provide. I haven't added new tests yet but I think this may actually simplify testing of `decode_id_token`. At the very least, I think it doesn't make things worse, although maybe that full-URI handling in BaseClient is a mistake.

I'd like to split this up because right now I don't think `decode_id_token` is very useful.

### DIFF
--- a/globus_sdk/auth/token_response.py
+++ b/globus_sdk/auth/token_response.py
@@ -3,7 +3,6 @@ import logging
 import time
 
 import jwt
-import requests
 
 from globus_sdk.response import GlobusHTTPResponse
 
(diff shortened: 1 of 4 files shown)


globus_sdk/auth/token_response.py
```diff
@@ -162,32 +161,40 @@ def by_scopes(self):
...
-        )
+        if not openid_configuration:
+            logger.debug("No OIDC Config provided, autofetching...")
+            openid_configuration = auth_client.get_openid_configuration()
+
+        if not jwk:
+            logger.debug("No JWK provided, autofetching + decoding...")
+            jwk = auth_client.get_jwk(
+                openid_configuration=openid_configuration, as_pem=True
+            )
```
jaswilli, line 186: `.decode_id_token(jwk=jwk_data)` is valid but also sort of strange since it'll pointless fetch the openid config but not use it.

Considering our openid provider uses a jwks that doesn't change I wonder if it doesn't make sense to make `decode_id_token` take the jwks as an optional argument and then either fetch both the config/jwks or nothing? Do you think there's a scenario in which an SDK consumer would try to optimize by fetching only the openid config but not the jwks?

sirosen: I did consider making a named tuple or something like `JWTRequirements.openid_configuration` and `JWTRequirements.jwk`, and only allowing
- `AuthClient.get_openid_configuration`
- `AuthClient.get_jwt_requirements`
- `OAuthTokenResponse.decode_id_token(jwt_requirements=...)`

Or we could make it an error to pass the JWK without passing openid configuration (but not vice versa).

If you want to do anything about this, I think it's one of those two? I'm a bigger fan of the usage error than a new type that wraps the pair of data.

sirosen: I went ahead and added a commit which makes it an error to pass `jwk` without `openid_configuration`.

If this looks like a reasonable direction, I'd like to move ahead and build out tests for it and un-mark it as a draft when I have time. Let me know if the general bent of this seems okay. (We can work out details later.)

aaschaer (approved):
Looks good to me, I like that we have a real token to test this on.

Should make caching the config and jwk auth responses while working with the sdk easier

PR #405, opened 2021-03-10
Import and adapt globus-sdk-tokenstorage

sirosen:
For the most part, this is

    globus_sdk_tokenstorage -> globus_sdk.tokenstorage

But there are some notable caveats. One is that the StorageAdapter and FileAdapter are public parts of the module and explicitly set their metaclass to "abc.ABCMeta" (which marks them as abstract). With that change in hand we can drop "Abstract" from their names for an overall tidier appearance.

A StorageAdapter is literally any object implementing `on_refresh` and `store` methods. And the default implementation of `on_refresh` just calls `store`. FileAdapter just extends this with some notion that the storage is in a file somewhere.

The other main thing that is new here is that fairly comprehensive type annotations are now in place, at least enough to note the argument and return types of the new interfaces.
I've tried to strike a balance on the annotations between flexibility (e.g. mapping vs dict) and simplicity. So maybe some methods would accept types other than `str`, but they're all noted as `str` and that's what should be used.

The tests are pretty much a verbatim port of the globus-sdk-tokenstorage tests.

One potential issue is that FileAdapter still only sets umask appropriately for a unix environment. I would be happy to work on changes to add windows compatibility, but for now there is no attempt at doing so. (globus-sdk-tokenstorage does not solve this problem either)

Also notable: SQLiteAdapter inherits from FileAdapter (this was not the case in the standalone lib) so that we can control the mode of the created db via `FileAdapter.user_only_umask()`.

<details><summary>it may be too large to be useful, but here's a diff between the original implementation and this sdk-ized version</summary>

```diff

diff -r sdk/globus_sdk/tokenstorage/__init__.py tokenstore/src/globus_sdk_tokenstorage/__init__.py
1,3c1,3
< from globus_sdk.tokenstorage.base import FileAdapter, StorageAdapter
< from globus_sdk.tokenstorage.file_adapters import SimpleJSONFileAdapter
< from globus_sdk.tokenstorage.sqlite_adapter import SQLiteAdapter
---
> from .file_adapters import SimpleJSONFileAdapter
> from .sqlite_adapter import SQLiteAdapter
> from .version import __version__
5c5
< __all__ = ("SimpleJSONFileAdapter", "SQLiteAdapter", "StorageAdapter", "FileAdapter")
---
> __all__ = ("__version__", "SimpleJSONFileAdapter", "SQLiteAdapter")
Only in tokenstore/src/globus_sdk_tokenstorage/: abstract_base.py
Only in sdk/globus_sdk/tokenstorage: base.py
diff -r sdk/globus_sdk/tokenstorage/file_adapters.py tokenstore/src/globus_sdk_tokenstorage/file_adapters.py
2c2
< import typing
---
> import os
4,5c4,5
< from globus_sdk.tokenstorage.base import FileAdapter
< from globus_sdk.version import __version__
---
> from .abstract_base import AbstractStorageAdapter
> from .version import __version__
7,8d6
< if typing.TYPE_CHECKING:  # conditional imports reduce the risk of dependency cycles
<     from globus_sdk.auth import OAuthTokenResponse
9a8,22
> class AbstractFileAdapter(AbstractStorageAdapter):
>     """
>     File adapters are for single-user cases, where we can assume that there's a
>     simple file-per-user and users are only ever attempting to read their own
>     files.
>     """
>
>     def file_exists(self):
>         """
>         Check if the file used by this file storage adapter exists.
>         """
>         return os.path.exists(self.filename)
>
>     def read_as_dict(self):
>         raise NotImplementedError
11c24,25
< class SimpleJSONFileAdapter(FileAdapter):
---
>
> class SimpleJSONFileAdapter(AbstractFileAdapter):
23,28c37,52
<     def __init__(
<         self,
<         filename: str,
<         resource_server: typing.Optional[str] = None,
<         scopes: typing.Optional[str] = None,
<     ):
---
>     def __init__(self, filename, resource_server=None, scopes=None):
>         self.filename = filename
>         self._set_resource_server_or_scopes(
>             resource_server=resource_server, scopes=scopes
>         )
>
>     def _set_resource_server_or_scopes(self, resource_server=None, scopes=None):
>         """
>         :param resource_server: the resource server name for tokens to look up
>                                 in a token response object
>         :param scopes: a list of scope names for tokens to look up in a token
>                        response object
>
>         Set the scopes or resource server used to look up tokens in a response object.
>         Callers must pass exactly one of ``resource_server`` and ``scopes``.
>         """
33,34d56
<
<         self.filename = filename
38,40c60
<     def _lookup_data_from_response(
<         self, token_response: "OAuthTokenResponse"
<     ) -> typing.Dict:
---
>     def _lookup_data_from_response(self, token_response):
55c75
<     def store(self, token_response: "OAuthTokenResponse"):
---
>     def store(self, token_response):
72c92,93
<         with self.user_only_umask():
---
>         old_umask = os.umask(0o177)
>         try:
75c96
<             to_write["globus-sdk.version"] = __version__
---
>             to_write["globus-sdk-tokenstorage.version"] = __version__
77a99,101
>         finally:
>             # reset umask
>             os.umask(old_umask)
79c103
<     def read_as_dict(self) -> typing.Dict:
---
>     def read_as_dict(self):
88,92c112,113
<         with open(self.filename) as f:
<             val = json.load(f)
<         if not isinstance(val, dict):
<             raise ValueError("reading from json file got non-dict data")
<         return val
---
>         with open(self.filename, "r") as f:
>             return json.load(f)
diff -r sdk/globus_sdk/tokenstorage/sqlite_adapter.py tokenstore/src/globus_sdk_tokenstorage/sqlite_adapter.py
1a2
> import os
3d3
< import typing
5,6c5,6
< from globus_sdk.tokenstorage.base import FileAdapter
< from globus_sdk.version import __version__
---
> from .abstract_base import AbstractStorageAdapter
> from .version import __version__
8,9d7
< if typing.TYPE_CHECKING:  # conditional imports reduce the risk of dependency cycles
<     from globus_sdk.auth import OAuthTokenResponse
11,12c9
<
< class SQLiteAdapter(FileAdapter):
---
> class SQLiteAdapter(AbstractStorageAdapter):
34,35c31,32
<     def __init__(self, dbname: str, namespace: str = "DEFAULT"):
<         self.filename = self.dbname = dbname
---
>     def __init__(self, dbname, namespace="DEFAULT"):
>         self.dbname = dbname
39,41d35
<     def _is_memory_db(self):
<         return self.dbname == ":memory:"
<
43,48c37,38
<         init_tables = self._is_memory_db() or not self.file_exists()
<         if init_tables and not self._is_memory_db():  # real file needs to be created
<             with self.user_only_umask():
<                 conn = sqlite3.connect(self.dbname)
<         else:
<             conn = sqlite3.connect(self.dbname)
---
>         init_tables = self.dbname == ":memory:" or not os.path.exists(self.dbname)
>         conn = sqlite3.connect(self.dbname)
72,73c62
<             # also mark the "database schema version" in case we ever need to handle
<             # graceful upgrades
---
>             # also mark the "database schema version", useful for upgrades
79,80c68,69
<                     ("globus-sdk.version", __version__),
<                     ("globus-sdk.database_schema_version", "1"),
---
>                     ("globus-sdk-tokenstorage.version", __version__),
>                     ("globus-sdk-tokenstorage.database_schema_version", "1"),
86c75
<     def store_config(self, config_name: str, config_dict: typing.Mapping) -> None:
---
>     def store_config(self, config_name, config_dict):
106c95
<     def read_config(self, config_name: str) -> typing.Optional[typing.Dict]:
---
>     def read_config(self, config_name):
123,126c112
<         val = json.loads(config_data_json)
<         if not isinstance(val, dict):
<             raise ValueError("reading config data and got non-dict result")
<         return val
---
>         return json.loads(config_data_json)
128c114
<     def remove_config(self, config_name: str) -> bool:
---
>     def remove_config(self, config_name):
143c129
<         return typing.cast(bool, rowcount != 0)
---
>         return rowcount != 0
145c131
<     def store(self, token_response: "OAuthTokenResponse") -> None:
---
>     def store(self, token_response):
171c157
<     def read_as_dict(self) -> typing.Dict[str, typing.Any]:
---
>     def read_as_dict(self):
189c175
<     def remove_tokens_for_resource_server(self, resource_server: str) -> bool:
---
>     def remove_tokens_for_resource_server(self, resource_server):
205c191
<         return typing.cast(bool, rowcount != 0)
---
>         return rowcount != 0
Only in tokenstore/src/globus_sdk_tokenstorage/: version.py
```
</details>

### DIFF
--- a/src/globus_sdk/tokenstorage/sqlite_adapter.py
+++ b/src/globus_sdk/tokenstorage/sqlite_adapter.py
@@ -1,0 +1,203 @@
+import json
+import sqlite3
+import typing
+
+from globus_sdk.auth import OAuthTokenResponse
+from globus_sdk.tokenstorage.base import FileAdapter
+from globus_sdk.version import __version__
+
+
+class SQLiteAdapter(FileAdapter):
+    """
+    :param dbname: the name of the DB file to write to and read from
+    :param namespace: A "namespace" to use within the database. All operations will
+                      be performed indexed under this string, so that multiple distinct
+                      sets of tokens may be stored in the database. You might use
+                      usernames as the namespace to implement a multi-user system, or
+                      profile names to allow multiple Globus accounts to be used by a
+                      single user.
+
+    A storage adapter for storing tokens in sqlite databases.
+
+    SQLite adapters are for more complex cases, where there may be multiple users or
+    "profiles" in play, and additionally a dynamic set of resource servers which need to
+    be stored in an extensible way.
+
+    The ``namespace`` is a user-supplied way of partitioning data, and any token
+    responses passed to the storage adapter are broken apart and stored indexed by
+    *resource_server*. If you have a more complex use-case in which this scheme will be
+    insufficient, you should encode that in your choice of ``namespace`` values.
+    """
+
+    def __init__(self, dbname: str, namespace: str = "DEFAULT"):
+        self.filename = self.dbname = dbname
+        self.namespace = namespace
+        self._connection = self._init_and_connect()
+
+    def _is_memory_db(self):
+        return self.dbname == ":memory:"
+
+    def _init_and_connect(self):
+        init_tables = self._is_memory_db() or not self.file_exists()
+        if init_tables and not self._is_memory_db():  # real file needs to be created
+            with self.user_only_umask():
+                conn = sqlite3.connect(self.dbname)
+        else:
+            conn = sqlite3.connect(self.dbname)
+        if init_tables:
+            conn.executescript(
+                """
+CREATE TABLE config_storage (
+    namespace VARCHAR NOT NULL,
+    config_name VARCHAR NOT NULL,
+    config_data_json VARCHAR NOT NULL,
+    PRIMARY KEY (namespace, config_name)
+);
+CREATE TABLE token_storage (
+    namespace VARCHAR NOT NULL,
+    resource_server VARCHAR NOT NULL,
+    token_data_json VARCHAR NOT NULL,
+    PRIMARY KEY (namespace, resource_server)
+);
+CREATE TABLE sdk_storage_adapter_internal (
+    attribute VARCHAR NOT NULL,
+    value VARCHAR NOT NULL,
+    PRIMARY KEY (attribute)
+);
+            """
+            )
+            # mark the version which was used to create the DB
+            # also mark the "database schema version" in case we ever need to handle
+            # graceful upgrades
+            conn.executemany(
+                """
+INSERT INTO sdk_storage_adapter_internal(attribute, value) VALUES (?, ?);
+            """,
+                [
+                    ("globus-sdk.version", __version__),
+                    ("globus-sdk.database_schema_version", "1"),
+                ],
+            )
+            conn.commit()
+        return conn
+
+    def store_config(self, config_name: str, config_dict: typing.Mapping) -> None:
+        """
+        :param config_name: A string name for the configuration value
+        :param config_dict: A dict of config which will be stored serialized as JSON
+
+        Store a config dict under the current namespace in the config table.
+        Allows arbitrary configuration data to be namespaced under the namespace, so
+        that application config may be associated with the stored tokens.
+
+        Uses sqlite "REPLACE" to perform the operation.
+        """
+        self._connection.execute(
+            """
+REPLACE INTO config_storage(namespace, config_name, config_data_json)
+VALUES (?, ?, ?)
+            """,
+            (self.namespace, config_name, json.dumps(config_dict)),
+        )
+        self._connection.commit()
+
+    def read_config(self, config_name: str) -> typing.Optional[typing.Dict]:
+        """
+        :param config_name: A string name for the configuration value
+
+        Load a config dict under the current namespace in the config table.
+        If no value is found, returns None
+        """
+        row = self._connection.execute(
+            """
+SELECT config_data_json FROM config_storage WHERE namespace=? AND config_name=?
+        """,
+            (self.namespace, config_name),
+        ).fetchone()
+
+        if row is None:
+            return None
+        config_data_json = row[0]
+        val = json.loads(config_data_json)
+        if not isinstance(val, dict):
+            raise ValueError("reading config data and got non-dict result")
+        return val
+
+    def remove_config(self, config_name: str) -> bool:
+        """
+        :param config_name: A string name for the configuration value
+
+        Delete a previously stored configuration value.
+
+        Returns True if data was deleted, False if none was found to delete.
+        """
+        rowcount = self._connection.execute(
+            """
+DELETE FROM config_storage WHERE namespace=? AND config_name=?;
+            """,
+            (self.namespace, config_name),
+        ).rowcount
+        self._connection.commit()
+        return typing.cast(bool, rowcount != 0)
+
+    def store(self, token_response: "OAuthTokenResponse") -> None:
+        """
+        :param token_response: a globus_sdk.OAuthTokenResponse object containing token
+                               data to store
+
+        By default, ``self.on_refresh`` is just an alias for this function.
+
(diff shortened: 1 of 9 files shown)


sirosen (commented):
Sorry it took so long for me to get back to you.

I'd like to avoid adding too many new things as part of this initial porting effort, but we can definitely do more as follow-up work, maybe even prior to releasing v3.
The major thing I don't want to do is add methods for doing token revocation because it grows the scope of this beyond just storing and retrieving data.

But this layer could at least help with logout, with respect to the stored data. Some interface like `pop()` would probably be good. e.g. `adapter.pop(resource_server)` gets and removes all the data for that resource server.

I'll take another look at the JSON file storage adapter. Right now, it doesn't work cleanly when there are multiple tokens (multiple resource servers), and I'm not sure that there's a good reason for that.

sirosen:
I've just done a pretty significant rewrite of the JSON file storage adapter and added `get_token_data` as I proposed above.

Things still on the list as definite TODOs:
- add a method for cleaning up the file for the file storage adapters (`delete_file()`?)
- add a method for deleting namespaces from the DB (`delete_namespace(str)`?)

However, I think we should do these as follow-up work so that we can look more clearly at those parts as they're added. So in my mind, this is ready for another round of review.

sirosen:
@aaschaer, how do you feel about this work at present?

I'd like to pull together the several threads of development I currently have going on the SDK, and start to look at a CLI branch that either works against `main` or uses a `3.0.0a1` release. This is a key component -- I'd like CLI 3.0 to migrate from `~/.globus.cfg` to `~/.globus/cli.db` (or something similar).

sirosen:
@aaschaer, I just rebased in order to handle recent changes, so I squashed away the `if TYPE_CHECKING: ...` changes entirely.

src/globus_sdk/tokenstorage/sqlite_adapter.py
```diff
@@ -0,0 +1,217 @@
...
+from typing import Any, Dict, Mapping, Optional, cast
+
+from globus_sdk.auth import OAuthTokenResponse
+from globus_sdk.tokenstorage.base import FileAdapter
+from globus_sdk.version import __version__
+
+
+class SQLiteAdapter(FileAdapter):
+    """
+    :param dbname: the name of the DB file to write to and read from
```
aaschaer, line 12: Would be good to mention that passing dbname as :memory: has special behavior here.

I almost would rather we didn't rely on magic string behavior, and instead had some in_memory boolean param to pass but we're inheriting the :memory: usage from sqlite so its probably fine.

aaschaer (approved):
Looks good to me for the most part! Nothing that should block merging.

Left one comment on improving docs on how to use :memory: for db.

Also I noticed in general we aren't wrapping json.loads in try-excepts. Is it okay if users hit a JSONDecodeError on bad data? Their storage is probably pretty broken at that point, but we could maybe at least say what we were trying to decode that broke?

src/globus_sdk/tokenstorage/sqlite_adapter.py
sirosen: I'll do a follow-up PR to add that to the doc here.

Given that `:memory:` is specially treated by sqlite itself -- not merely the python sqlite bindings -- I think it's least confusing to stick with this behavior, at least for anyone familiar with sqlite. If I were designing an interface for something backed by a file or memory in a vacuum, without the sqlite context, I'd also probably choose something different.

sirosen:
> Is it okay if users hit a JSONDecodeError on bad data? Their storage is probably pretty broken at that point, but we could maybe at least say what we were trying to decode that broke?

We could try to produce a more useful error in this case, but I'd want to think about it more and not just tack it on to the existing PR. The possibility of errors when loading is real -- even if it requires that someone or something is mucking up the data -- and merits some more thought.

I'm going to proceed with a merge as-is, and we can follow-up.

PR #406, opened 2021-03-16
Fix doc on external_checksum

sirosen:
In response to confusion arising around external_checksum, this fixes the doc to state that the checksum is verified after data transfer. We also clarify that the task will fail if verification fails.

---

I'm sure it's still somewhat confusing/ambiguous for some readers, but this at least fixes the part that was inaccurate.
If you have ideas for further refining the wording, I'm game for additional changes.


aaschaer (approved):
There are times when the checksum is checked before transfer (notably sync tasks) but since it will always be checked after the transfer I think this is easier to understand. I'll go ahead and merge.

PR #407, opened 2021-03-16
Remove 'allowed_authorizer_types' restriction

sirosen:
Instead, the clients themselves should allow any authorizer (including, for example, custom ones constructed by the user) to be passed.

Inevitably, *someone* will pass a basic authorizer to a TransferClient. Then it won't work, they'll get errors back from Transfer, and they'll sort it out. I don't think I still buy the reasoning behind this restriction, especially since it runs counter to the surrounding design (that a client is agnostic towards the inner workings of its authorizer).

In the very worst case, if we do want to keep this, I would invert it and make it a `forbidden_authorizers` list, so we can specify

    TransferClient.forbidden_authorizers = [
        NullAuthorizer, BasicAuthorizer
    ]

and call it a day.

### DIFF
--- a/globus_sdk/auth/client_types/confidential_client.py
+++ b/globus_sdk/auth/client_types/confidential_client.py
@@ -31,11 +31,6 @@ class ConfidentialAppAuthClient(AuthClient):
     .. automethodlist:: globus_sdk.ConfidentialAppAuthClient
     """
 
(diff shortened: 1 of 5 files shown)


jaswilli:
Agreed. I don't think the extra code makes much of a difference in the feedback loop to justify its existence.

PR #410, opened 2021-05-15
Strip down the contrib doc and add badges

sirosen:
A lot of advisory guidance in the contrib doc is not essential and takes up space/time/attention. e.g. It is easiest to handle PRs which merge cleanly, and commit messages ought to use the imperative mood. However, these issues are not dealbreakers -- we can afford to rebase PRs or put up with awkwardly phrased commit messages.

Rather than populating the contrib doc with best practices and advisory info, cut everything down to just the necessary tox commands.

To get info about code style out into the world, setup badges for black and isort.

---

I started this because I wanted to revise the guidance on relative vs absolute imports. But it looks like we removed that at some point. By the time I realized, I had a branch open and had started doing cleanup, so I just finished what I was doing.

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -3,85 +3,51 @@ Contributing to the Globus SDK
 
 First off, thank you so much for taking the time to contribute! :+1:
 
-Reporting Bugs & Requesting Features
-------------------------------------
+Bugs & Feature Requests
+-----------------------
 
-  - Check if there's a matching
-      https://github.com/globus/globus-sdk-python/issues[issue]
-      before opening a new issue or pull request
-  - When possible, provide a code sample to reproduce bugs
+Should be reported as
+https://github.com/globus/globus-sdk-python/issues[GitHub Issues]
+
+For a good bug report:
+
+  - Check if there's a matching issue before opening a new issue
+  - Provide a code sample to reproduce bugs
 
 Linting & Testing
 -----------------
 
 Testing the SDK requires https://tox.readthedocs.io/en/latest/[tox].
 
-Once you have `tox` installed, you can run testing or linting on the repo
-with `tox` or `tox -e lint`. These are wrapped for convenience as `make test`
-and `make lint` as well.
+Run tests with
 
-All code must pass `make lint test`, which runs linting and tests.
+    tox
 
-Optional, but recommended linting setup
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+And linting with
+
+    tox -e lint
+
+Optional, but recommended, linting setup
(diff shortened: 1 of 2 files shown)


rudyardrichter (commented):
For interested parties, especially internal, I imagine the added detail and `Expectations for Pull Requests` would be useful and worthwhile. I understand wanting to keep this document as concise as possible, but would you consider moving the deleted portions to a new section of "here's stuff that isn't vital to get started, but a reference one might point at for style/conventions/etc."?

sirosen:
I can add a "code guidelines" section at the end of the doc, but most of the current items aren't useful. Several are either obvious to professional developers, or else things that I don't want to hand-hold novices through doing. e.g.

> - *Squash intermediate and fixup commits*. We recommend running `git rebase --interactive` prior to submitting a pull request.

This is kind of useful in that it recommends a specific command for doing cleanup. But either you're comfortable enough with git to do this -- in which case there's no need for SDK documentation to explain it because you know about rebasing -- or you aren't -- in which case it might be more expedient to squash merge that person's contributions.

sirosen:
I added a small "Code Guidelines" section. I could write a great deal more: a whole book chapter on "composability over inheritance", for example. But I tried to stay very concise because the long-winded nature of this doc is exactly what I'm fighting. We can reintroduce other items from the prior doc if I cut something which is really useful.

rudyardrichter:
> I could write a great deal more: ...

Anything veering towards "philosophy" or deeper topics definitely makes sense to not include and is likely to come up in reviews anyways. ("Think very hard before adding a new dependency" is a good example here.)

However practices that aren't necessarily universal like

> Squash intermediate and fixup commits.

I think are worth writing down once, somewhere. I've definitely seen repositories where force-pushing (even for said rebases) is disallowed, or you're supposed to do it only after a review, or whatever. The point being that if someone isn't following a desired behavior which may have multiple common approaches, you can just point at this and say "please follow those guidelines."

rudyardrichter:
Anyways, I definitely like the conciseness of the new version, and having somewhere to put `Code Guidelines`.

sirosen:
I used to care more about asking people to rebase before GitHub had a squash-merge button. A dirty git history was more of an imposition on maintainers, since cleaning up the history was manual.

I do personally prefer a clean git history. That leaves reviewers with the option of preserving the commit history. If someone comes in which a branch that read `wip foo; finish adding foo; fixup; fixup; fixup`, you really have no choice but to squash it. But I'm not sure the repo needs to express an opinion on this topic -- it's fine either way so long as the quality of the actual work is good.

sirosen:
@rudyardrichter, we may or may not disagree slightly over details here. Are you satisfied with where the doc is in the current state of this PR?

@aaschaer, although your feedback on this is welcome, it's not strictly required. Let me know if you want me to hold off on merging this, but otherwise I'll consider Rudy and I reaching agreement to be sufficient.

PR #411, opened 2021-05-18
Fix task list docs

### LINKED ISSUE #408: TransferClient.endpoint_manager_task_list has invalid usage as an example
Endpoint manager task listing requires that you supply at least one filter. However, it looks like the [doc example](https://github.com/globus/globus-sdk-python/blob/4e55f8c61f6c384d0b9b3102994cddab57b929d4/globus_sdk/transfer/client.py#L1589-L1632) was written naively, or it was insufficiently tested. The example does not have a filter in place.

Also, the doc doesn't cover the supported filters, which we should probably copy into the SDK doc so that anyone looking at those docs can see how to use the API without linking out to the Transfer docs.

The Transfer reference is here:
https://docs.globus.org/api/transfer/advanced_endpoint_management/#query_parameters

rudyardrichter:
- Include a table of filters that can be applied to `TransferClient.endpoint_manager_task_list`, taken from the Transfer docs.
    - (I didn't include `filter_username` which is marked as deprecated.)
- Update examples for `TransferClient.endpoint_manager_task_list` to correctly include at least one filter.

Resolves #408.

### DIFF
--- a/globus_sdk/transfer/client.py
+++ b/globus_sdk/transfer/client.py
@@ -1604,12 +1604,95 @@ class TransferClient(BaseClient):
                 an iterable of :class:`GlobusResponse
                 <globus_sdk.response.GlobusResponse>`
 
+        **Filters**
+
+        The following filters are supported (passed as keyword arguments in ``params``).
+        For any query that doesn’t specify a filter_status that is a subset of
+        ("ACTIVE", "INACTIVE"), at least one of filter_task_id or filter_endpoint is
+        required.
+
+        ====================== ================ ========================
+        Query Parameter        Filter Type      Description
+        ====================== ================ ========================
+        filter_status          equality list    |filter_status|
+        filter_task_id         equality list    |filter_task_id|
+        filter_owner_id        equality         |filter_owner_id|
+        filter_endpoint        equality         |filter_endpoint|
+        filter_is_paused       boolean equality |filter_is_paused|
+        filter_completion_time datetime range   |filter_completion_time|
+        filter_min_faults      int              |filter_min_faults|
+        filter_local_user      equality         |filter_local_user|
+        ====================== ================ ========================
+
+        .. |filter_status| replace::
+           Comma separated list of task statuses.  Return only tasks with any of the
+           specified statuses. Note that in-progress tasks will have status "ACTIVE" or
+           "INACTIVE", and completed tasks will have status "SUCCEEDED" or "FAILED".
+
+        .. |filter_task_id| replace::
+           Comma separated list of task_ids, limit 50. Return only tasks with any of
+           the specified ids. If any of the specified tasks do not involve an endpoint
+           the user has an appropriate role for, a ``PermissionDenied`` error will be
+           returned. This filter can't be combined with any other filter.  If another
+           filter is passed, a ``BadRequest`` will be returned.
+
+        .. |filter_owner_id| replace::
+           A Globus Auth identity id. Limit results to tasks submitted by the specified
+           identity, or linked to the specified identity, at submit time.  Returns
+           ``UserNotFound`` if the identity does not exist or has never used the Globus
+           Transfer service. If no tasks were submitted by this user to an endpoint the
+           current user has an appropriate role on, an empty result set will be
+           returned. Unless filtering for running tasks (i.e. ``filter_status`` is a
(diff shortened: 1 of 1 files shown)


globus_sdk/transfer/client.py
```diff
@@ -1618,7 +1701,8 @@ def endpoint_manager_task_list(self, num_results=10, **params):
         status:
 
         >>> tc = TransferClient(...)
-        >>> for task in tc.endpoint_manager_task_list(num_results=None):
+        >>> for task in tc.endpoint_manager_task_list(
+        >>>     num_results=None, filter_status="ACTIVE"):
```
rudyardrichter, line 1705: I'm not sure what the desired style is here. This seemed to most closely resemble other examples.

globus_sdk/transfer/client.py
```diff
@@ -1604,12 +1604,95 @@ def endpoint_manager_task_list(self, num_results=10, **params):
...
+        filter_is_paused       boolean equality |filter_is_paused|
+        filter_completion_time datetime range   |filter_completion_time|
+        filter_min_faults      int              |filter_min_faults|
+        filter_local_user      equality         |filter_local_user|
+        ====================== ================ ========================
+
+        .. |filter_status| replace::
+           Comma separated list of task statuses.  Return only tasks with any of the
+           specified statuses. Note that in-progress tasks will have status "ACTIVE" or
+           "INACTIVE", and completed tasks will have status "SUCCEEDED" or "FAILED".
```
rudyardrichter, line 1630: Under scrutiny it looks funny, but I see the Sphinx docs use this sort of weird style with the three-space indent, to match the `|x|` text.

sirosen: That's fine. The alignment looks nice, IMO, but if you chose to do 4 spaces instead, I wouldn't bat an eye.

The table above, where you used these, is really nice to read. I'm going to double-check the text one last time before merging though, since there's a lot of doc here.

globus_sdk/transfer/client.py
```diff
@@ -1604,12 +1604,95 @@ def endpoint_manager_task_list(self, num_results=10, **params):
...
+           tasks with ``completion_time`` in the specified range. If the end date is
+           blank, it will also include all active tasks, since they will complete some
+           time in the future.
+
+        .. |filter_min_faults| replace::
+           Minimum number of cumulative faults, inclusive.  Return only tasks with
+           ``faults >= N``, where N is the filter value.  Use ``filter_min_faults=1`` to
+           find all tasks with at least one fault.  Note that many errors are not fatal
+           and the task may still be successful even if ``faults >= 1``.  See the
+           link:../task{outfilesuffix}#task_fields[faults field] documentation for
```
sirosen, line 1680: This needs adjustment, since the linkage from the docs site won't work here.
I'd be satisfied if we just remove this sentence.

globus_sdk/transfer/client.py
```diff
@@ -1604,12 +1604,95 @@ def endpoint_manager_task_list(self, num_results=10, **params):
...
+           ``filter_status`` is also passed and contains a subset of "ACTIVE" and
+           "INACTIVE". Completed tasks always have ``is_paused`` equal to "false" and
+           filtering on their paused state is not useful and not supported.  Note that
+           pausing is an async operation, and after a pause rule is inserted it will
+           take time before the is_paused flag is set on all affected tasks. Tasks
+           paused by id will have the ``is_paused`` flag set immediately.
+
+        .. |filter_completion_time| replace::
+           Start and end date-times separated by a comma. Each datetime should be
+           specified as a string in ISO 8601 format: YYYY-MM-DDTHH:MM:SS, where the "T"
```
sirosen, line 1665:
```suggestion
           specified as a string in ISO 8601 format: ``YYYY-MM-DDTHH:MM:SS``, where the "T"
```

It's probably worth making this tweak so that the format is easier to read within the block of text.

I understand that this is mostly copied from Transfer doc, so we don't want to diverge too much in this copy. But I find the way that the supported subset of ISO 8601 is described to be a bit confusing... I don't want to bog down this improvement with copyediting, so let's leave it for now. If I think of something, I can try getting it adjusted upstream in the Transfer docs, and then we'd copy the improvement downstream to here.

sirosen:
Looks great to me after the recent edits. I'm going to squash merge, since I don't think we need the history in this case.

Thanks for taking the time for this. First commits... Of many! 🚀

PR #414, opened 2021-05-20
Set pytest to "filterwarnings=error"

sirosen:
Add pytest config to setup.cfg which sets 'filterwarnings=error'. This converts any warnings -- or at least, any which we don't explicitly handle -- into errors during tests.

Also, move from `-v --cov=globus_sdk` in tox.ini to `--cov=globus_sdk` via addopts only. This is more concise output and keeps configuration more together.

There's only one warning in tests right now: `Iterable` moved to `collections.abc`. So fix that too.

(NB: will conflict with #413, but I'll rebase to handle.)

### DIFF
--- a/globus_sdk/auth/client_types/base.py
+++ b/globus_sdk/auth/client_types/base.py
@@ -1,4 +1,4 @@
-import collections
+import collections.abc
 import json
 import logging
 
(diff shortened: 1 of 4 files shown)


sirosen:
I tried the GitHub UI flow for fixing a merge conflict. It's a pleasant experience to use, but that unnecessary merge-back commit... yuck. I'll rebase to clean that up.

PR #416, opened 2021-05-24
Update to pyjwt 2.x and fix jwt tests

sirosen:
`decode_id_token` tests are now failing because the token is expired. This was an oversight -- we weren't doing this testing before because of this issue. The simplest fix is to allow verification of `exp` (the expiration information) to be disabled and then use that in the testsuite.

---

Add passthrough parameters for "the jwt library" (pyjwt) to decode_id_token. This allows the testsuite to pass `verify_exp=False` as one of the pyjwt options.

After testing, this works on both pyjwt v1.x and v2.x , as the option name is unchanged. Since the testing is done, do the library update as well.

### DIFF
--- a/globus_sdk/auth/token_response.py
+++ b/globus_sdk/auth/token_response.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import time
+import typing
 
 import jwt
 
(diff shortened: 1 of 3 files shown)


sirosen:
The force-push is from me accidentally removing the `[crypto]` from the requirements, then restoring it.

PR #417, opened 2021-05-24
Add transport handling HTTP details, payload encoding, and automatic retries via a configurable RetryPolicy object

sirosen:
`globus_sdk.transport` is a new subpackage intended to encapsulate HTTP request sending as much as is possible.
It doesn't know about custom response types and exceptions, but deals instead in terms of raw response and exception objects from the HTTP lib (`requests`).

Instead of holding a `requests.Session`, client objects hold a `RequestsTransport` which is responsible for request sending, encoding data, retry handling, and correct use of the GlobusAuthorizer objects to add auth headers.
The client object is still responsible for defining a set of parameters for each method and resolving methodnames + parameters to well-formed HTTP requests, but it is no longer responsible for any retry logic for authz failure or many of the other details of HTTP.

In aggregate, the following changes are achieved:

1. The `json_body` and `text_body` parameters are replaced with `data` and `encoding`

The `data` parameter is meant to be a dict or string. `encoding` is one of `None, "text", "json", "form"`.
Previously, callers would send `json_body={...}` to send JSON to our APIs.
However, JSON is the norm, and form-encoded data is only used as strictly necessary for compliance with the OAuth2 spec.

Therefore, it's more comfortable to have a single parameter for sending a payload, `data={...}`.

The `encoding` optional string can be used to request `form` encoding. The implementation leaves open the possibility of additional encodings in the future, but for now only "raw text", JSON, and form-data are implemented.

2. The special retry of 401s is replaced with a more generic framework for evaluating whether or not requests should be retried

Rather than a hardcoded check for an HTTP 401 (to handle expired tokens), a RetryPolicy implements `should_retry(RetryContext)`, where the `RetryContext` is an object designed specifically to carry all of the relevant information for this decision.

The `RetryContext` is passed to various check functions until either
- one returns True (do the retry)
- one returns False (explicitly do not retry)
- all have returned `None`, meaning no decision was reached (default -> do not retry)

`RetryContext` may carry either a response object or an exception, and the retry checkers vary from a "max attempts" check to a fancy one which interacts with the authorizer (if present) to do a retry.

When a retry condition is reached, the RetryPolicy does exponential backoff with jitter before resending the request.

3. A variety of retry checks are defined by default

Notably, we add a retry checker which handles any network errors as "should retry = true", as well as one which does the same for 500s, 502s, 503s, and several other "transient error" HTTP statuses.

4. `retry_401s` is removed as a parameter

This parameter was somewhat speculative. No APIs have required it. It ought to be possible to handle such cases with a custom retry policy henceforth.

5. The `transport_class` and `retry_policy` attributes of a client class can be used to customize behavior

You could subclass the retry policy class or instantiate a retry policy and add your own checkers.

Other included changes:
- documentation of `BaseClient` methods has been cleaned up
- `BaseClient._request` has changed to `BaseClient.request`
- `http_timeout` is no longer a parameter to client initializers, it's now part of the `transport_params` dict if needed

----

This is still a draft until I can add the two major missing parts:
- [x] tests which cover the various default retry checks and the ability to register custom checks
- [x] figure out where to document the various new components (potentially requires a larger doc refactor)


sirosen:
I've just done some new testing, to the point where I'd be comfortable merging this as-is and I'm going to un-mark it as a draft.
The missing piece here is documentation, but that could be sufficiently extensive that it might be necessary to treat that as follow-up.

src/globus_sdk/transport/retry.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+def _parse_retry_after(response: requests.Response) -> typing.Optional[int]:
+    val = response.headers.get("Retry-After")
+    if not val:
+        return None
+    try:
+        return int(val)
+    except ValueError:
+        return None
```
rudyardrichter, line 63:
```suggestion
    try:
        return int(val)
    except TypeError, ValueError:
        return None
```

src/globus_sdk/transport/retry.py
```diff
@@ -0,0 +1,254 @@
...
+        self.register_default_checks()
+
+    def compute_delay(self, ctx: RetryContext) -> float:
+        """
+        Given a retry context, compute the amount of time to sleep.
+        This is always the minimum of the backoff (run on the context) and the
+        ``max_sleep``.
+        """
+        delay = self.backoff(ctx)
+        return min(delay, self.max_sleep)
```
rudyardrichter, line 137: I don't know what your preference is for this style, so I'll just point it out:
```suggestion
        return min(self.backoff(ctx), self.max_sleep)
```

src/globus_sdk/transport/retry.py
```diff
@@ -0,0 +1,254 @@
...
+        """
+        for check in self.checks:
+            result = check(context)
+            if result is None:
+                continue
+            elif result is False:
+                return False
+            else:
+                time.sleep(self.compute_delay(context))
+                return True
```
rudyardrichter, line 152: Wouldn't this make this method sensitive to the order the checks are run? I would picture the logic I expect as the following:
```suggestion
        return not any(check(context) is False for check in self.checks)
```
so perhaps the first check says to retry, but a check towards the end determines there should be no retry.

rudyardrichter: I would also tend towards keeping the side effect `time.sleep(...)` out of a `.should_retry()` method—I would assume the caller of `should_retry` is responsible for actually doing the waiting.

src/globus_sdk/transport/retry_policy.py
```diff
@@ -0,0 +1,49 @@
...
+class RetryPolicy:
+    max_retries: int = 5
+    max_sleep: int = 10
+
+    def _backoff(self, attempt):
+        # (0.5s * 2^attempt) results in ~15s of total sleep for 5 retries
+        # 2^0 + 2^1 + 2^2 + 2^3 + 2^4 = 31
+        # note that additional time can be spent sending the request or waiting for
+        # a reply
+        time.sleep(0.5 * (2 ** attempt))  # exponential backoff
```
rudyardrichter, line 16: Similarly here, having the actual sleep kind of buried I think makes this overall harder to follow.

src/globus_sdk/transport/retry_policy.py
```diff
@@ -0,0 +1,49 @@
...
+        return False
+
+    def _parse_retry_after(self, response: requests.Response) -> typing.Optional[int]:
+        val = response.headers.get("Retry-After")
+        if val:
+            try:
+                return int(val)
+            except ValueError:
+                pass
+        return None
```
rudyardrichter, line 31: I think this would have the same behavior?
```suggestion
        try:
            return int(response.headers.get("Retry-After"))
        except TypeError, ValueError:
            pass
```

src/globus_sdk/transport/retry.py
rudyardrichter: (Also `should_retry` to me sounds innocuous and non-side-effectful.)

src/globus_sdk/auth/client_types/base.py
```diff
@@ -47,6 +47,7 @@ class AuthClient(BaseClient):
 
     service_name = "auth"
     error_class = exc.AuthAPIError
+    service_name = "auth"
```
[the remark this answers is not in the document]
sirosen: Yep, definitely. Must have slipped in during the rebase. I'll clean it out.

src/globus_sdk/transport/retry.py
sirosen: I found that `mypy` yelled at me when I did it this way, because `int(None)` was a possibility (even though TypeError gets caught). Do you not run into the same issue?

I'd prefer to do with a simple try-except if it can pass the type checking.

src/globus_sdk/transport/retry_policy.py
sirosen: Ach! I don't know how `retry_policy.py` crept back into this during the rebase. I'll remove it entirely. It was renamed to `retry.py`, but somehow rebasing after the `src/` work accidentally reintroduced it. I'll have to do a pass to scrub it.
Please ignore this file. I didn't notice it "reappear".

src/globus_sdk/transport/retry.py
rudyardrichter: That's weird, I don't get the same error (aside from the example I wrote needing parens on the except). That is `mypy==0.812`. It's a trivial enough thing though if the more verbose version is needed.

tests/unit/test_retry_policy.py
```diff
@@ -0,0 +1,62 @@
...
+
+def test_retry_policy_respects_retry_after():
+    policy = RetryPolicy()
+
+    dummy_response = mock.Mock()
+    dummy_response.headers = {"Retry-After": "5"}
+    dummy_response.status_code = 429
+    ctx = RetryContext(1, retry_state={}, response=dummy_response)
+
+    with mock.patch("time.sleep") as mocksleep:
```
rudyardrichter, line 16: The `mocksleep` fixture could be in a higher-level conftest to also use here

src/globus_sdk/transport/retry.py
```diff
@@ -0,0 +1,254 @@
...
+            return None
+        # the response code was a 401 and there's already been at least one reauth
+        # attempt
+        # this could be `False` instead, but that would mean that nobody else can add
+        # 401-retries even when reauth didn't work
+        if ctx.retry_state.get("has_done_reauth"):
+            return None
+
+        if not ctx.authorizer.handle_missing_authorization():
+            return None
```
rudyardrichter, line 251: Comments aside, could these all be lumped into one large conditional?

src/globus_sdk/base.py
```diff
@@ -361,42 +262,23 @@ def _request(
...
-        if self.authorizer is not None:
-            log.debug(
-                "request will have authorization of type {}".format(
-                    type(self.authorizer)
-                )
-            )
-            self.authorizer.set_authorization_header(rheaders)
+        # prepare data...
+        # copy headers if present
+        rheaders = {**headers} if headers else {}
```
rudyardrichter, line 281: Interesting, I hadn't seen the unpacking syntax used for copying before. Was there a particular reason you adopted it here?

rudyardrichter (commented):
It turns out I don't have many more comments since the first pass.

One thing I would remark on the stuff in `retry.py` is that the `True`/`False`/`None` logic for retries might be clearer as an enum—"do"/"don't"/"allow-retry". (Anecdotally, using the null type to represent any meaning whatsoever seems fraught to me.) I wouldn't object to merging this version.

The "user-facing' interface of this seems great and I'm excited to have this in the SDK.

src/globus_sdk/base.py
sirosen: The `**` syntax is really handy when you're copying and modifying a dict to add a single attribute. e.g.
`d = {"x": 1, **d}`

Because it gets used like that, I've started to prefer it as a syntax for plain copying. That way, the two usages are consistent throughout a codebase.
Plus, I think it's more obvious that this is a shallow copy than `dict.copy` or `copy.copy`. The syntax agrees with splat-expansion in other contexts (e.g. keyword args) in doing a shallow copy.

This is the sort of stylistic thing that I don't think is a big deal -- certainly not enough to try to add a lint for it. I personally prefer this way of writing it, and I have my reason for that, but it doesn't run any deeper than that.

src/globus_sdk/transport/retry.py
sirosen: They could, but I worry that combining _all_ of them would be optimizing for lines-of-code rather than clarity. 

Because `handle_missing_authorization` can have side-effects, I think it's better to keep that one separate.

How about this?

```python
if (  # is this check applicable?
    ctx.response is None or
    ctx.response.status_code not in self.EXPIRED_AUTHORIZATION_STATUS_CODES or
    ctx.authorizer is None
):
    return None

# ... blah blah long comment ...
if ctx.retry_state.get("has_done_reauth"):
    return None

# potentially do a token refresh, exiting if the authorizer can't or won't handle this
if not ctx.authorizer.handle_missing_authorization():
    return None
```

tests/unit/test_retry_policy.py
sirosen: Good call, I'll make that change.

src/globus_sdk/transport/retry.py
sirosen: I'll check again. It's always possible that I did something odd.

src/globus_sdk/transport/requests.py
```diff
@@ -0,0 +1,156 @@
...
+        return {"Accept": "application/json", "User-Agent": self.user_agent}
+
+    def _encode(
+        self,
+        method: str,
+        url: str,
+        params: typing.Optional[typing.Dict] = None,
+        data: typing.Union[typing.Dict, typing.List, str, None] = None,
+        headers: typing.Optional[typing.Dict[str, str]] = None,
+        encoding: typing.Optional[str] = None,
```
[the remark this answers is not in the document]
sirosen: Let's punt on this.

I like strings for this because it makes it possible to add a custom encoder. Imagine
```python
def add_validating_encoder(name: str, validator: Callable):
    class ValidatingEncoder(JSONRequestEncoder):
        def encode(self, method, url, params, data, headers):
            validator(data)
            return super().encode(method, url, params, data, headers)
    RequestsTransport.encoders[name] = ValidatingEncoder()
```

Maybe I'm speculating too much here and nobody would do this. We can switch to an enum, but let's consider that after we get the rest of this in, yeah?

src/globus_sdk/transport/retry.py
sirosen: Sure. I have no opinion one way or another, so the fact that you're suggesting this is enough.

src/globus_sdk/transport/retry.py
sirosen: They are order-sensitive today, so I don't think contracting like this works unless we restructure more dramatically. Mainly, the issue is that checks can have side-effects. In particular, the authorization one, but also the Retry-After one needs a way to pass back the parsed header value (which right now is done by mutating the context object).

I'm inclined to stick with the current approach for the purposes of getting a first version merged, but I'd be extremely happy to revisit and try to improve it.  _Ideally_, checks would be forbidden from having side-effects. I tried a version at some point (not sure if it's in the commit history or not) which had a `RetryCheckResult` type for passing back true/false and a float for seconds of delay. But I think doing something like that will require changes to the authorizers too (I'm not sure exactly what -- that's where I drew the line with that experiment).

src/globus_sdk/transport/retry.py
sirosen: I'm also using `mypy==0.812`, and this `repro.py` file fails for me:
```python
import typing


def foo(x: typing.Optional[str]):
    try:
        return int(x)
    except (TypeError, ValueError):
        return None
```

sirosen:
> ... logic for retries might be clearer as an enum ...

💯 Let me do that. `DO_RETRY`, `DO_NOT_RETRY`, `NO_DECISION` is so much clearer that it hardly needs explanation.

That's the only rewrite that I still intend to do. If there are other changes -- or you like the potential rewrite of the authz check discussed above -- we can do those. But otherwise, with a `RetryCheckResult(Enum)`, I think that's the last change.

src/globus_sdk/transport/retry.py
sirosen: Actually, I went ahead and did this refactor since I already needed to touch this to add the result enum.

sirosen:
Doing a refactor to use `RetryCheckResult` naturally led to a refactor of the authz retry check as well. So I think everything is tidy. I'll check in on Monday -- if we're reasonably happy with this and willing to accept the things that I want to punt on for the moment, we'll merge.

PR #418, opened 2021-05-24
New pagination tools with distinct paginator objects + PaginatorTable

### LINKED ISSUE #320: Redesign Pagination
I find the way we attacked pagination to be somewhat limited and potentially fraught under v1. Some thoughts on this:

- Having an easy way, like the `.data` property, of forcing many paginated calls is bad
- Only being able to access a nested list of data from paginated calls, and not any other parts of the pages, is limiting
- Paginators need to be iterators, but they don't need to look like GlobusResponse objects at all
- Paginated calls don't always need to be made under the layer of pagination -- sometimes results are known to fit on one page and direct calls are simpler
- The boto3 design for this would have clients all get a `get_paginator` method returning pages. Although it works, it feels like a clunky interface and can confuse novices easily.
  - One key benefit of their design is that a paginator returns pages, not direct contents of any kind
  - Another benefit: if a call is paginated, knowing the unpaginated version tells you how to invoke the paginated version
- Paginated calls and unpaginated calls should "look similar"

Consider this interface:
```python
tc = TransferClient(...)
# unpaginated
res = tc.endpoint_search(...)
for doc in res['DATA']:
    ...
# paginated
for page in tc.endpoint_search.paginator(...):
    # page is a GlobusResponse for the individual page,
    # so we get the benefits of that
    for doc in page['DATA']:
        ...
```

This makes pagination easy, intentionally guides new and exploratory usage away from making many paginated calls, and avoids some of the awkwardness of
```python
client = boto3.client('ec2')
paginator = client.get_paginator('describe_instances')
for page in paginator.paginate(Filters=[...]):
    ...
```

sirosen:
resolves #320 

The primary goal here is to break apart pagination into a collection of similar (but distinct) iterables which can be used to handle pagination.

Each paginator is only guaranteed to be an iterable. No other interfaces are promised by the `Paginator` base class.

The working model is then as follows:
- all clients have an optional `paging_spec` dict
- client init loads `self.paging_spec: SpecDictType` into `self.paginated: PaginatorTable`
- for any given method name where pagination is supported, the same name is a valid attribute of `self.paginated`. e.g. `self.endpoint_search()` and `self.paginated.endpoint_search()` are both valid calls
- a paginator iterates over pages, not data items
- `num_results` is removed from all methods -- clients just iterate until reaching their desired limit

Optional changes/extensions:
- `Paginator.iter_items()` could be used if we give a "data key" to the paginators from the spec
- `Paginator.__init__` could be defined with some minimal required interface (currently assumed but not defined)
- `Paginator` could support `limit_less_than_available_results` if there's an easy way, but for now this is only known to be needed by the CLI, so it can be implemented there
- `Paginator.has_next_page()` isn't guaranteed to be easy to implement for all pagination types, but perhaps it could be defined as an `Optional[bool]` where `None` means there is no clear implementation -- would this be useful ever? e.g. instead of `limit_less_than_available_results`

----

I'm opening this as a draft until I can at least test out the CLI using it. But this is the current state of my work on improving the pagination situation.


src/globus_sdk/services/transfer/client.py
```diff
@@ -39,6 +36,25 @@ class TransferClient(client.BaseClient):
...
+    Using ``TransferClient.endpoint_search`` as an example::
+
+        from globus_sdk import TransferClient
+        tc = TransferClient(...)
+
+        # this is the unpaginated version
+        for x in tc.endpoint_search("tutorial"):
+            print("Endpoint ID: {}".format(x["id"]))
+
+        # this is the paginated version
```
aaschaer, line 53: Should this example use .items() instead of a nested for loop? Or is that not our recommended usage?

aaschaer (approved):
Looks good to me!
Left one comment on an example, but nothing worth blocking on

src/globus_sdk/services/transfer/client.py
sirosen: Yeah, let's make sure to come back to this.

I actually want to put together more work on the sphinx extension to automatically trawl methods with `_paginator_class` to generate more doc, so there's a whole pass on doc refinement we can make. I'll write it up in a story.

PR #421, opened 2021-05-29
Simplify responses to avoid complex typing and avoid "callback" style use of classes

sirosen:
This is a draft because it's based on #417. (And now we see why I'm excited to get that merged!)

The commits tell a 3-part story, but the last one is the version of this work that I really like.

[First](https://github.com/globus/globus-sdk-python/pull/421/commits/56c41ceb3849e731b83ef0800d27f0204347ce47), I tried to make `BaseClient` a `Generic`, and there were exceedingly clever games I was trying to play with bounded type vars to explain to `mypy` what I actually wanted to do. This was pretty much a disaster. Less of a disaster than some of the other solutions I tried, but still bad. `mypy` can overload `__new__` and `__init__`, and that allows for generics with default types as instance variables, but we want a default for a generic as a class var, and I just couldn't get it to work. On top of that, pulling the arguments to the base generic type at runtime requires the use of undocumented magic attributes which are currently in flux. I don't want us building something which could break in unfixable ways on 3.10 or 3.11 .

The next two steps are simplification.

[Second](https://github.com/globus/globus-sdk-python/pull/421/commits/2c4333fc5672e87e0d78f7828bf10f62dec1be36), we remove `default_response_class`. There's no notion of a default, just `GlobusHTTPResponse` as the class which wraps response data. Unless you specify `response_class=...`. This means that the class-level type annotations can be written correctly (in the first version, things are still kind of wonky), and generally makes things at least _doable_. This ditches `TransferResponse` because it's not particularly useful, but that's minor. More importantly, it puts us on our path to...

[Third and finally](https://github.com/globus/globus-sdk-python/pull/421/commits/efd786b69c993a51d18898eb685ad348d7d7695f), remove support for `response_class=...` from `BaseClient`. Client methods return `GlobusHTTPResponse` objects. That's what they do.
This causes a bit of a problem though -- we have several fancy response objects like the `ActivationRequirementsResponse` and the `OAuthTokenResponse` which offer important attributes in addition to the base `data` for JSON data. There are a few was to solve this, but the simplest one for now is to make these classes proxies which can wrap a `GlobusHTTPResponse`. This requires a small shim class, `GlobusHTTPResponseProxy`, and it becomes valid to write `IterableTransferResponse(client.get("/foo"))`. The types are harmonized -- `mypy` understands this trivially -- and it's very easy to read as a replacement for `client.get("/foo", response_class=IterableTransferResponse)`.

The commits have a little more detail, like `client` becoming a required parameter when constructing `GlobusHTTPResponse`, but that's all minor. The main drive of the work is to convert from the current `default_response_class` machinery, which offers relatively little value and blows up type checking, to something simpler.


sirosen:
Splitting this work to do a partial merge (first two commits), per discussion.

PR #424, opened 2021-06-03
Rewrite custom response classes as wrappers

sirosen:
This rewrite net-simplifies the combination of responses and various HTTP methods. The overall gist is that base client methods always return GlobusHTTPResponse objects, and the ability to pass `response_class` is removed. In order to still support advanced response types (ActivationRequirementsResponse, OAuthTokenResponse, etc), these types become proxy or wrapper objects which contain a GlobusHTTPResponse.

To make defining these simpler, a new class, GlobusHTTPResponseProxy is added, from which these inherit. The proxy just pulls attributes off of the inner response and re-exposes them. Both GlobusHTTPResponse and GlobusHTTPResponseProxy inherit from GlobusResponse, which lets them share implementations of `__getitem__`, `__contains__`, and
`get()` with respect to their `data` property.
GlobusResponse becomes abstract, with `data` as an abstract property.

As a result, where we would previously provide a desirable response class like so:

    self.get("/foo", response_class=FooResponse)

The same idea is expressed in a more "ordinary" and idiomatic way like so:

    FooResponse(self.get("/foo"))

The following included changes are related to this:

- PaginatedResponse no longer needs to take a `response_class` to pass through -- it just wraps data in IterableTransferResponse

- The once-optional `client` argument to GlobusHTTPResponse is now required, since all of this was being refactored anyway

- Type overloads for the BaseClient methods are all removed, because they now are guaranteed to always return GlobusHTTPResponse

- AuthClient.oauth2_token needs to allow the use of variable wrapper types to support dependent token responses while still returning a default of an OAuthTokenResponse. It therefore implements and has type overloads for `response_class`, similar to what was previously done in BaseClient

### DIFF
--- a/src/globus_sdk/response.py
+++ b/src/globus_sdk/response.py
@@ -1,39 +1,24 @@
+import abc
 import logging
+from typing import Mapping, Optional
 
-logger = logging.getLogger(__name__)
+log = logging.getLogger(__name__)
 
 
-class GlobusResponse:
+class GlobusResponse(abc.ABC):
     """
-    Generic response object, with a single ``data`` member.
+    A GlobusResponse is an abstract class which defines ``__getitem__``,
+    ``__contains__``, and ``get()`` with respect to some inner ``data``.
 
-    The most common response data is a JSON dictionary. To make
-    handling this type of response as seemless as possible, the
-    ``GlobusResponse`` object also supports direct dictionary item
-    access, as an alias for accessing an item of the underlying
-    ``data``. If ``data`` is not a dictionary, item access will raise
-    ``TypeError``.
-
-    >>> print("Response ID": r["id"]) # alias for r.data["id"]
-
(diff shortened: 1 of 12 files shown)


src/globus_sdk/response.py
```diff
@@ -44,35 +29,40 @@ def __getitem__(self, key):
...
+        return self.data.get(key, default)
 
-    def get(self, *args, **kwargs):
-        """
-        ``GlobusResponse.get`` is just an alias for ``GlobusResponse.data.get``
-        """
-        return self.data.get(*args, **kwargs)
+    @property
+    @abc.abstractmethod
+    def data(self) -> Optional[Mapping]:
```
sirosen, line 64: I just noticed that this type annotation can be wrong in the case that a response contains JSON data which is not a dict. Maybe it should just be `Any`.

Some APIs return arrays (I think that should always, always, always be avoided, but my opinion doesn't change the reality).

PR #429, opened 2021-06-10
Improve tox config for multipython coverage

sirosen:
'--cov-append' must be used to gather coverage info from multiple python versions. e.g. If we have code which conditionally executes on some versions and not others, it will only be properly recorded in this way.

Tox invocation via github actions will still produce distinct coverage reports by passing a `--cov-report=...` option.
However, local runs of `tox` should run tests on all interpreters and only print a final coverage report.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -34,7 +34,7 @@ jobs:
         if: ${{ matrix.python-version == '3.9' }}
         run: python -m tox -e lint
       - name: Run Tests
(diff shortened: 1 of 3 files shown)


rudyardrichter:
Missing 3.7 & 3.9 gives this output:
```
tests/unit/responses/test_token_response_pickleability.py .                                                                                                            [100%]Coverage.py warning: No data was collected. (no-data-collected)
WARNING: Failed to generate report: No data to report.
```

PR #430, opened 2021-06-10
Combine RetryPolicy and RequestsTransport

sirosen:
All of the properties of the retry policy are included in the transport, and it is solely responsible both for request sending and handling retries. As a result, more options are available as retry params via the `transport_params` argument to client construction.
It's still possible to set various settings with a custom `transport_class` as well.

i.e. Both imperative and declarative styles are supported well by the reduction in distinct objects.

---

This is a little bit experimental as a change, which is why I've marked it as a draft (i.e. I don't want us to merge this).
I'm looking at the retry+transport interface and wondering about how to improve it.

Combining the two objects makes the transport a little bit more complicated than I would like. But it does simplify a number of things, especially the user-facing interface. (`transport_class` and `transport_params` become sufficient to specify everything)

There are some other benefits too. We only have `max_retries` in one place with this, and I took the time to split the authz handling off of the other retry handlers too.

I'm still looking for ways of simplifying this further, so ideas are welcome.


sirosen:
I've taken another look at this and I think struck a balance here which makes this change a worthwhile improvement.
In the latest revision, elements of the `RetryPolicy` object and the `RetryContext` have been replaced with a system of `RetryCheckFlags` (an enum of bitwise-and-able flags) and `RetryCheckRunner` (an object responsible for running the checks and doing any necessary interpretation of the flags).

The authz checker then, instead of manipulating a dict attached to the context, is simply declared as
```python
@set_retry_check_flags(RetryCheckFlags.RUN_ONCE)
def default_check_expired_authorization(...): ...
```

The check body is simpler, with no extra logic for interpreting data in that `retry_state` dict.

The "runner" is an object instantiated once per request, and primarily just wraps a dict for tracking which checks have already run.

> Aside: The runner requires that the checks are hashable as keys in a dict, but as most callables are hashable I didn't deem it important to document this rigorously. If anyone is concerned, we could keep track of their positions in the list of checks and use the index as a key instead of the callable itself. I don't think that's necessary, however.

The runner could technically be merged into the transport, but it makes the logic for retries more confusing to read because some kind of state for the request needs to be tracked. We'd end up carefully declaring a state dict outside of the retry loop, very similar to what's in `main` today.

---

Net-net, we don't have _fewer_ objects involved (in fact, counting the flags we have more), but the objects we have are simpler and the interface we provide to users is superior.

sirosen:
I've just rebased to resolve conflicts.

I'd like to merge this as the last requisite change before releasing globus-sdk 3.0b1.
(I do not think it makes sense to do the first beta with a known backwards-incompatible change in the pipe.)

PR #433, opened 2021-06-30
move passthrough paramaters out of top level params

aaschaer:
A few questions / notes

- Is `Dict[str, str]` better than just `Dict` or `Dict[str, Any]` for the type here? I had to do a decent amount of `str()` casting and I'm not sure it affected anything other than making the type checker happy.
- A lot of functions that need to add their own query params previously assumed that the dictionary existed and now need `if query_params is None: query_params = {}` logic for that assumption to hold. Is there a cleaner way of doing that without running into issues with mutable defaults?
- I also changed the passthrough behavior for `TransferData` and `DeleteData` with the name of `additional_fields` since they're not query params. Not sure if we wanted this work to extend to there or not but it seemed reasonable?
- I added `limit` and `offset` as named parameters to several functions since their paginators expected them to accepted at the top level. Might have been cleaner to just have the paginateor use `query_params`?

### DIFF
--- a/src/globus_sdk/client.py
+++ b/src/globus_sdk/client.py
@@ -111,7 +111,7 @@ class BaseClient:
         self,
         path: str,
         *,
-        params: Optional[Dict] = None,
+        query_params: Optional[Dict[str, str]] = None,
         headers: Optional[Dict] = None,
     ) -> GlobusHTTPResponse:
         """
@@ -122,14 +122,14 @@ class BaseClient:
         :return: :class:`GlobusHTTPResponse \
         <globus_sdk.response.GlobusHTTPResponse>` object
         """
-        log.debug(f"GET to {path} with params {params}")
-        return self.request("GET", path, params=params, headers=headers)
+        log.debug(f"GET to {path} with query_params {query_params}")
+        return self.request("GET", path, query_params=query_params, headers=headers)
 
     def post(
         self,
         path: str,
         *,
-        params: Optional[Dict] = None,
+        query_params: Optional[Dict[str, str]] = None,
         data: Optional[Dict] = None,
         headers: Optional[Dict] = None,
         encoding: Optional[str] = None,
@@ -142,16 +142,21 @@ class BaseClient:
         :return: :class:`GlobusHTTPResponse \
         <globus_sdk.response.GlobusHTTPResponse>` object
         """
-        log.debug(f"POST to {path} with params {params}")
+        log.debug(f"POST to {path} with query_params {query_params}")
         return self.request(
-            "POST", path, params=params, data=data, headers=headers, encoding=encoding
+            "POST",
+            path,
+            query_params=query_params,
+            data=data,
+            headers=headers,
+            encoding=encoding,
         )
 
     def delete(
         self,
         path: str,
         *,
-        params: Optional[Dict] = None,
+        query_params: Optional[Dict[str, str]] = None,
         headers: Optional[Dict] = None,
     ) -> GlobusHTTPResponse:
         """
@@ -162,14 +167,14 @@ class BaseClient:
         :return: :class:`GlobusHTTPResponse \
         <globus_sdk.response.GlobusHTTPResponse>` object
         """
-        log.debug(f"DELETE to {path} with params {params}")
-        return self.request("DELETE", path, params=params, headers=headers)
+        log.debug(f"DELETE to {path} with query_params {query_params}")
+        return self.request("DELETE", path, query_params=query_params, headers=headers)
 
     def put(
         self,
         path: str,
         *,
-        params: Optional[Dict] = None,
+        query_params: Optional[Dict[str, str]] = None,
         data: Optional[Dict] = None,
         headers: Optional[Dict] = None,
         encoding: Optional[str] = None,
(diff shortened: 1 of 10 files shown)


src/globus_sdk/services/auth/client_types/base.py
```diff
@@ -129,37 +135,40 @@ def _convert_listarg(val):
 
         log.info("Looking up Globus Auth Identities")
 
+        if query_params is None:
+            query_params = {}
```
sirosen, line 139: To your question about ways of handling the default and guaranteeing a dict, I think this is the least-bad way to do it.
If you prefer, this is a line shorter but not in any way superior:
```python
query_params = query_params or {}
```

src/globus_sdk/services/auth/client_types/base.py
```diff
@@ -219,17 +230,19 @@ def oauth2_refresh_token(self, refresh_token, additional_params=None):
...
+        :param query_params: A dict of extra params to encode in the refresh call.
+        :type query_params: dict, optional
         """
         log.info("Executing token refresh; typically requires client credentials")
         form_data = {"refresh_token": refresh_token, "grant_type": "refresh_token"}
 
-        if additional_params:
-            form_data.update(additional_params)
+        if query_params:
+            form_data.update(query_params)
```
sirosen, line 240: Since this case is a request body, we shouldn't call it `query_params`.

What about `form_params` or `body_params`?

src/globus_sdk/services/auth/client_types/base.py
```diff
@@ -288,11 +301,11 @@ def oauth2_validate_token(self, token, additional_params=None):
             log.debug("Validating token with unauthenticated client")
             body.update({"client_id": self.client_id})
 
-        if additional_params:
-            body.update(additional_params)
+        if query_params:
+            body.update(query_params)
```
sirosen, line 305: Likewise, form or body in this case and the one below.

src/globus_sdk/client.py
```diff
@@ -111,7 +111,7 @@ def get(
         self,
         path: str,
         *,
-        params: Optional[Dict] = None,
+        query_params: Optional[Dict[str, str]] = None,
```
sirosen, line 114: I'll just comment on this first case, but this holds through.
I think `Dict[str, Any]` is better for these.

That handles integers and bools gracefully. True, it allows for incorrect things as well, but mypy is always giving us trade-offs.

Based on what's done in the stdlib and typeshed, I think this is a good case for Any.

src/globus_sdk/services/transfer/client.py
```diff
@@ -213,7 +216,9 @@ def endpoint_search(
         self,
         filter_fulltext: Optional[str] = None,
         filter_scope: Optional[str] = None,
-        **params,
+        limit: Optional[int] = None,
+        offset: Optional[int] = None,
```
sirosen, line 220: I can see why you found adding these noteworthy, but I actually think it's a good change.
It makes the SDK more verbose, but it puts more parameters directly into our documentation. I think you made the right call on how to handle this.

src/globus_sdk/services/transfer/data.py
```diff
@@ -80,8 +81,9 @@ class TransferData(dict):
         task to fail.
         [default: ``False``]
     :type fail_on_quota_errors: bool, optional
-
-    Any additional parameters are fed into the dict being created verbatim.
+    :param additional_fields: additional fields to be added to the transfer
+        document. Mostly intended for internal use
+    :type additional_fields: dict, optional
```
sirosen, line 86: I wonder if maybe we should just remove this?
Maybe leave it for now, but I'd be okay with removing it.

src/globus_sdk/services/transfer/data.py
```diff
@@ -197,13 +199,13 @@ def __init__(
...
-        for option, value in kwargs.items():
-            logger.info(
-                "TransferData.{} = {} (option passed in via kwargs)".format(
-                    option, value
+        if additional_fields is not None:
+            self.update(additional_fields)
+            for option, value in additional_fields.items():
+                logger.info(
+                    "TransferData.{} = {} (option passed in via "
+                    "additional_fields)".format(option, value)
```
sirosen, line 207: The format only gets applied to the second string here. I think it would be good to switch to an f-string anyway.

src/globus_sdk/services/transfer/data.py
```diff
@@ -212,7 +214,7 @@ def add_item(
         recursive=False,
         external_checksum=None,
         checksum_algorithm=None,
-        **params,
+        additional_fields: Optional[Dict[str, Any]] = None,
     ):
```
sirosen, line 218: Unlike the top-level extra parameters, this one is more useful to keep.

I'm still ruminating about a better name for it, but nothing comes to mind as obviously better.

sirosen (commented):
Overall I think this looks great! Comments and ideas inline.

I want a less verbose name for added parameters in transfer helpers, but I'm not sure what would work nicely. It isn't completely catastrophic of we conflict with a name that Transfer later uses, but we should try to avoid it. I think that rules out some obvious ones like `include` as too generic. I'll keep thinking about this for a little while we get this finalized.

aaschaer:
Added changes as we discussed, should be good to go now?

src/globus_sdk/services/transfer/data.py
sirosen: As we discussed, this name is actually pretty good. Marking this resolved.

sirosen:
I'm just doing a final once-over, and then I'll merge.

PR #434, opened 2021-06-30
Add support for get_shared_endpoint_list

aaschaer:
New paginator interface seems to work really well for this!
Had to make slight adjustments to IterableTransferResponse as well to allow an arbitrary key for iteration.

Not ready for merge yet since I based this on #433 so I wouldn't have to re-do the query_params on the new function

### DIFF
--- a/src/globus_sdk/client.py
+++ b/src/globus_sdk/client.py
@@ -111,7 +111,7 @@ class BaseClient:
         self,
         path: str,
         *,
-        params: Optional[Dict] = None,
+        query_params: Optional[Dict[str, str]] = None,
         headers: Optional[Dict] = None,
     ) -> GlobusHTTPResponse:
         """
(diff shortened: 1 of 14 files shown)


sirosen:
👍 6698ac6 looks perfect!
Let's sort out #433 and then come back and merge this.

aaschaer:
changed to Dict[str, Any] and rebased to match #433

sirosen:
I'm not sure quite why, but even though e9a819a is now merged, GitHub is showing the full diff. I'm going to do a final check, and then merge this and see what GitHub does. It _ought_ to only merge the head of this branch.

sirosen:
Looks like the merge produced the right history (the same as if I'd done local `git merge`s), strange display on github notwithstanding. So no worries about that. 🤷

PR #435, opened 2021-07-01
Groups v2 Client

joshbryan-globus:
This is a functional groups v2 client with most of the endpoints and options now supported.  I will likely add a few more tests to get better coverage, but wanted to get some eyes on it.

[ch8498](https://app.clubhouse.io/globus/story/8498/add-support-for-groups-v2-in-sdk)


src/globus_sdk/services/groups/client.py
```diff
@@ -1,8 +1,161 @@
...
+    `Perform actions on group members
+    <https://groups.api.globus.org/redoc#operation/group_membership_post_actions_v2_groups__group_id__post>`_.
+
+    .. automethodlist:: globus_sdk.GroupsClient
+    """
+
+    def __init__(self):
+        self._payload: dict = {}
+
+    def accept_invites(self, identity_ids: Sequence[str]):
```
sirosen, line 55: Because `str` is a `Sequence[str]`, I think these methods should check `isinstance(..., str)` before iterating.

It's not a declared/valid usage to pass a single string, but sooner or later someone would pass a single string, `mypy` would pass, but then the result would be a call with an "unexpected" bad payload.

src/globus_sdk/services/groups/client.py
```diff
@@ -1,8 +1,161 @@
...
+        """
+        Accept invites for identities.  The identities must belong to
+        the identity set of authenticated user.
+        """
+        self._payload.setdefault("accept", []).extend(
+            {"identity_id": identity_id} for identity_id in identity_ids
+        )
+        return self
+
+    def add_members(self, identity_ids: Sequence[str], role: Role = Role.member):
```
sirosen, line 65: As I look at this PR, and try to weigh the basic flexibility vs safety tension that surrounds use of strict types (enums, helper objects) over built-in, more generic types (str, dict), this `def` strikes me as possibly a place where we could strike a balance.

The helper object could require the enums, and make no allowances for strings or other objects, but then the client methods could accept `Union[<helper_object>, Dict]`.

That way, you can always call `to_dict()` and add explicit values to a payload, but the helper itself will have guardrails for the 99% use-case.

What do you think?

src/globus_sdk/services/groups/client.py
```diff
@@ -13,3 +166,157 @@ class GroupsClient(client.BaseClient):
...
+        """
+        return self.post(f"/groups/{group_id}", data=actions.to_dict())
+
+    def accept_invite(self, group_id: str, identity_id: str):
+        """
+        Accept invite for an identity.  The identity must belong to
+        the identity set of the authenticated user.
+        """
+        actions = BatchMembershipActions().accept_invites([identity_id])
+        return self.batch_membership_action(group_id, actions)
```
sirosen, line 256: I'm torn about the inclusion of these wrapper methods. They will certainly be convenient, but they seem to duplicate a lot -- especially the docstrings -- of relevant `BatchMembershipActions` methods.

I'm wondering if long-term maintenance would be better served, especially as things change and new parameters get added over time, by not having these and expecting usage to always be of the form
```python
client.batch_membership_action(
    groupid,
    BatchMembershipActions().accept_invites([id1, id2]),
)
```

Simply because we have limited resources for maintenance, I am somewhat concerned about drift and staleness. These projects are receiving attention right now, but in general they tend to be pushed down in priority.

sirosen (commented):
There are some relatively minor things that need fixing. And methods need docstrings with parameters listed and links to external API docs, but that can wait until everything else is finalized to save effort.

Primarily, I'm concerned about the GroupsClient being inflexible as the API changes over time.

This is the first client to be fleshed out without python2 support as a consideration, so it introduces some new things which we may want to backport to the other clients as strategies. But also, these introduce some potential pitfalls.

The first is the addition of enums for known enumerated strings. I think these are good for the vast majority of use-cases, but we've also been able to use new API features in projects (e.g. globus-cli) before an SDK release with support is done. I would like to preserve that flexibility, so the question becomes: how do we allow passing a string for callers who really know what they're doing?
If we're willing to add runtime string checks (`isinstance`), we could make the convention to pass `typing.cast(Role, "new-role")`. Or we could allow strings, but emit warnings.

The other, related, item is the introduction of annotations requiring that a custom type be passed for payload data (by contrast, [`TransferClient.submit_data()`](https://github.com/globus/globus-sdk-python/blob/ff84efab6e621aae0db40ae8c817b0db507bafc0/src/globus_sdk/services/transfer/client.py#L1088) is not annotated and accepts a dict). Again, this touches on flexibility and our ability to use new API behaviors without requiring an SDK release.

As I suggested in some inline comments, there might be a solution in which we're strict in one context but not another. We'll probably have to play with it a bit and see what produces good developer ergonomics.

sirosen:
#438 takes the approach here from the helper object and applies it to SearchQuery, TransferData, and DeleteData.
If we go forward with that, it might be nice to pull in `PayloadWrapper` as a base class for the helper object here, after a rebase.

src/globus_sdk/services/groups/client.py
```diff
@@ -1,8 +1,161 @@
...
+    """
+    An object used to represent a batch action on memberships of a group.
+    `Perform actions on group members
+    <https://groups.api.globus.org/redoc#operation/group_membership_post_actions_v2_groups__group_id__post>`_.
+
+    .. automethodlist:: globus_sdk.GroupsClient
+    """
+
+    def __init__(self):
+        self._payload: dict = {}
```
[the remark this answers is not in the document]
sirosen: This turned into #438, so I'm marking this comment resolved.

sirosen:
After a quick reread of my review, some of these comments may be looking to apply flexibility in the wrong places.
With #433 merged and included in 3.0b1, one thing that has become clear is the utility of an `additional_fields: Optional[Dict]` argument to payload builders. So long as an explicit passthrough is allowed, it's easy to handle new parameters. The comments may not be entirely on-target regarding _where_ things need to be flexible.

Also, a minor note: #438 makes it the responsibility of `BaseClient.request()` to handle conversion of `PayloadWrapper` objects to dicts. That might make being open to `Union[Dict, BatchMembershipActions]` more appealing.

sirosen:
Per our discussion this morning, I've
1. moved things around a little and done some minor renames so that we can `from globus_sdk import GroupRole`, etc.
2. Created a new payload type, `GroupPolicies`
3. Rebased onto `main` to pickup on new type annotation work + a commit to fix mypy linting
4. Noticed that I didn't use `--amend` in (2) and had to do an extra force-push on that (sorry)

I'm going to watch to ensure that the build passes, and then merge.

PR #438, opened 2021-07-08
Convert data helper classes to "PayloadWrapper"

sirosen:
Inspired by #435 and from discussion of getting rid of `dict` inheritance.

Rather than defining classes which inherit from `dict` directly, define PayloadWrapper, a utility class which wraps a `_payload` dict.
The PayloadWrapper inherits from MutableMapping and implements methods against its underlying dict, meaning that the range of usages for these objects is not diminished.

TransferData, DeleteData, and SearchQuery are all updated to inherit from this object and, other than invoking `super().__init__()`, they are almost entirely unchanged.
In the case of SearchData, __init__ needed to be defined to allow passing known parameters as keyword args. (Some previously valid usages will now require the use of `additional_fields`).

In terms of methods which pass data as a dict, the changes made are to allow any PayloadWrapper object to pass down to
BaseClient.request(...). At that stage, `PayloadWrapper`s will render with a `to_dict()` call. As a result, almost no concrete code, only type annotations needed changes. Mostly of the form `Optional[Dict]` -> `Union[None, Dict, PayloadWrapper]`

A new test exercises the methods of PayloadWrapper to make sure they all function as expected, and a minor update ensures that SearchQuery is well-tested.

The Transfer data module has been split in two, to make it easier to read/maintain the two helper classes.

### DIFF
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -57,29 +57,21 @@ def test_safe_stringify(value):
 
 
 def test_payload_wrapper_methods():
-    # comment on first use of each MutableMapping method to note which ones are ours
-    # (explicit) and which ones are inherited from MutableMapping
+    # just make sure that PayloadWrapper acts like a dict...
     data = utils.PayloadWrapper()
-
-    assert "foo" not in data  # __contains__() inherited
+    assert "foo" not in data
     with pytest.raises(KeyError):
-        data["foo"]  # __getitem__ explicitly defined
-
-    data["foo"] = 1  # __setitem__ explicitly defined
+        data["foo"]
+    data["foo"] = 1
     assert "foo" in data
-    assert data["foo"] == 1  # __delitem__ explicitly defined
+    assert data["foo"] == 1
     del data["foo"]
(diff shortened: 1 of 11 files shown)


tests/unit/test_utils.py
```diff
@@ -54,3 +54,24 @@ def test_safe_stringify(value):
     safe_value = utils.safe_stringify(value)
     assert safe_value == "1"
     assert type(safe_value) == str
+
+
+def test_payload_wrapper_methods():
```
aaschaer, line 59: I'd honestly be fine not having this test since we don't actually define any methods in PayloadWrapper now, but it doesn't hurt to have I guess

aaschaer (approved):
Looks good to me!

tests/unit/test_utils.py
sirosen: Yeah, I feel the same. It's kind of a 🤷 for me, but it does validate that the rewrite to `UserDict` didn't break anything? Not that I was exactly worried... I left it there, but if it were removed in the future that would be fine too.

PR #439, opened 2021-07-09
Add method and URL to stack traces

rudyardrichter:
- Change args passed to `Exception` to include method and URL
- Update tests for the different _get_args interface.

I would characterize this PR as slightly draft-ish in that suggestions for any other `Exception` arguments to include are welcome. As is this just adds the method and URL.

I wondered if checking for the response to have a request and URL attached would be necessary, but looking at the requests code it seems [a request object is always attached to the response](https://github.com/psf/requests/blob/33cf965f7271ab4978ed551754db37865c4085db/requests/adapters.py#L286-L287).

### DIFF
--- a/src/globus_sdk/exc.py
+++ b/src/globus_sdk/exc.py
@@ -34,6 +34,8 @@ class GlobusAPIError(GlobusError):
                    suitable for display to end users.
     """
 
+    RECOGNIZED_AUTHZ_SCHEMES = ["bearer", "basic", "globus-goauthtoken"]
+
     def __init__(self, r, *args, **kw):
         self._underlying_response = r
         self.http_status = r.status_code
@@ -61,8 +63,7 @@ class GlobusAPIError(GlobusError):
             # fallback to using the entire body as the message for all
             # other types
             self._load_from_text(r.text)
-        args = self._get_args()
-        GlobusError.__init__(self, *args)
+        super().__init__(*self._get_args())
 
     @property
     def raw_json(self):
@@ -95,12 +96,32 @@ class GlobusAPIError(GlobusError):
         """
         return self._underlying_response.text
 
+    def _get_request_authorization_scheme(self):
+        try:
+            authz_h = self._underlying_response.request.headers["Authorization"]
+            try:
+                authz_scheme = authz_h.split()[0]
+                if authz_scheme.lower() in self.RECOGNIZED_AUTHZ_SCHEMES:
+                    return authz_scheme
+            except IndexError:
+                pass
+        except KeyError:
+            pass
+        return None
+
     def _get_args(self):
         """
         Get arguments to pass to the Exception base class. These args are
         displayed in stack traces.
         """
-        return (self.http_status, self.code, self.message)
+        return [
+            self._underlying_response.request.method,
+            self._underlying_response.url,
+            self._get_request_authorization_scheme(),
+            self.http_status,
+            self.code,
+            self.message,
+        ]
 
     def _load_from_json(self, data):
         """
--- a/src/globus_sdk/services/search/errors.py
+++ b/src/globus_sdk/services/search/errors.py
@@ -14,9 +14,6 @@ class SearchAPIError(exc.GlobusAPIError):
         self.error_data = None
         super().__init__(r)
 
-    def _get_args(self):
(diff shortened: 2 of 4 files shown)


sirosen:
First off, I can confirm that this works and that it's helpful.

The only other thing which occurs to me is to look at `request.headers["Authorization"].split()[0]` (properly checked) as the "authorization scheme". I tried this out as a conditional insert at position 1, so we get this:
```
>>> ac = globus_sdk.AuthClient()
>>> ac.get("/foo")
  File "/home/sirosen/dev/sdk_and_cli/sdk/src/globus_sdk/client.py", line 123, in get
    return self.request("GET", path, query_params=query_params, headers=headers)
  File "/home/sirosen/dev/sdk_and_cli/sdk/src/globus_sdk/client.py", line 286, in request
    raise self.error_class(r)
globus_sdk.services.auth.errors.AuthAPIError: ('GET', 'https://auth.globus.org/foo', 404, 'NOT_FOUND', 'Resource not found: /foo')
>>> ac2 = globus_sdk.ConfidentialAppAuthClient("x", "y")
>>> ac2.get("/foo")
  File "/home/sirosen/dev/sdk_and_cli/sdk/src/globus_sdk/client.py", line 123, in get
    return self.request("GET", path, query_params=query_params, headers=headers)
  File "/home/sirosen/dev/sdk_and_cli/sdk/src/globus_sdk/client.py", line 286, in request
    raise self.error_class(r)
globus_sdk.services.auth.errors.AuthAPIError: ('GET', 'Basic', 'https://auth.globus.org/foo', 404, 'NOT_FOUND', 'Resource not found: /foo')
```

Does that seem good? Or maybe unnecessary?

<details><summary>git diff of a version with auth scheme and a check to avoid logging credentials</summary>

```diff
diff --git a/src/globus_sdk/exc.py b/src/globus_sdk/exc.py
index 9001bf0..cacdaaf 100644
--- a/src/globus_sdk/exc.py
+++ b/src/globus_sdk/exc.py
@@ -34,6 +34,8 @@ class GlobusAPIError(GlobusError):
                    suitable for display to end users.
     """

+    RECOGNIZED_AUTHZ_SCHEMES = ["bearer", "basic", "globus-goauthtoken"]
+
     def __init__(self, r, *args, **kw):
         self._underlying_response = r
         self.http_status = r.status_code
@@ -94,18 +96,35 @@ class GlobusAPIError(GlobusError):
         """
         return self._underlying_response.text

+    def _get_request_authorization_scheme(self):
+        try:
+            authz_h = self._underlying_response.request.headers["Authorization"]
+            try:
+                authz_scheme = authz_h.split()[0]
+                if authz_scheme.lower() in self.RECOGNIZED_AUTHZ_SCHEMES:
+                    return authz_scheme
+            except IndexError:
+                pass
+        except KeyError:
+            pass
+        return None
+
     def _get_args(self):
         """
         Get arguments to pass to the Exception base class. These args are
         displayed in stack traces.
         """
-        return [
+        args = [
             self._underlying_response.request.method,
             self._underlying_response.url,
             self.http_status,
             self.code,
             self.message,
         ]
+        authz = self._get_request_authorization_scheme()
+        if authz:
+            args.insert(1, authz)
+        return args

     def _load_from_json(self, data):
         """
```
</details>

rudyardrichter:
I can see that being really helpful to distinguish whether a call was authorized at all or not. I would maybe put it at 2, so `METHOD URL` are still together.

rudyardrichter:
Ok, also, you have to teach me how to include that collapsible diff section, because that seems really convenient.

sirosen:
>  I would still include a `None` value in the args to specify that there was no authorization.

Yeah, that seems like a good idea. In retrospect, it's better than having the arg list resize.

If you commit those changes, I'm onboard with doing a quick reread to check and merging.

---

Looks like you figured out the `<details>` trickery! 😄 
I find it super-useful. Feel free to try out this bookmarklet too:

<details><summary>a bookmarklet for inserting an empty details block</summary>

```javascript
javascript:(function(x){x.value=x.value.slice(0,x.selectionStart)+"<details><summary></summary>\n\n</details>"+x.value.slice(x.selectionEnd);})(document.activeElement);
```
</details>

rudyardrichter:
@sirosen I'm good with seeing this version merged now.

PR #440, opened 2021-07-09
Add scopes as client class attributes; doc scopes

sirosen:
- Add `.scopes` to client classes, pointing back at the relevant ScopeBuilder objects
- Document use of `globus_sdk.scopes` in a new doc page, adjacent to service client docs
- Add a custom sphinx directive for extracting "known_scopes" from a ScopeBuilder into a doc list and decorating the resulting doc
- Update `globus_sdk.scopes` to use the new `listknownscopes` directive to generate doc comments on all scope builder constants

Example doc screenshot:
image

### DIFF
--- a/src/globus_sdk/_sphinxext.py
+++ b/src/globus_sdk/_sphinxext.py
@@ -12,6 +12,11 @@ from docutils.statemachine import ViewList
 from sphinx.util.nodes import nested_parse_with_titles
 
 
+def _extract_known_scopes(scope_builder_name):
+    sb = locate(scope_builder_name)
+    return sb._known_scopes
+
+
 def _classname2methods(classname, include_methods):
     """resolve a class name to a list of (public) method names
     takes a classname and a list of method names to avoid filtering out"""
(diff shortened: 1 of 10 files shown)


src/globus_sdk/_sphinxext.py
```diff
@@ -73,30 +108,41 @@ def gen_rst(self):
...
+        sb_basename = sb_name.split(".")[-1]
+
+        add_scopes = []
+        example_scope = None
+        for arg in self.arguments[1:]:
+            if arg.startswith("add_scopes="):
+                add_scopes = [x for x in arg.split("=")[1].split(",")]
+            if arg.startswith("example_scope="):
+                example_scope = arg.split("=")[1]
+        known_scopes = _extract_known_scopes(sb_name) + add_scopes
```
rudyardrichter, line 128: I might put it in this order, since in the current case the `add_scopes` are the most "interesting".
```suggestion
        known_scopes = add_scopes + _extract_known_scopes(sb_name)
```

src/globus_sdk/_sphinxext.py
```diff
@@ -73,30 +108,41 @@ def gen_rst(self):
...
+        if example_scope is None:
+            example_scope = known_scopes[0]
+
+        yield ""
```
rudyardrichter, line 132: I don't necessarily find one better than the other, but one could `yield from [...]` here to be able to write a bunch of lines with just the content and no `yield` prefix.

src/globus_sdk/_sphinxext.py
sirosen: I waffled on this a bit. I think either order is fine, but the fact that you suggest it is enough reason for me to change it.

src/globus_sdk/_sphinxext.py
sirosen: I'm actually somewhat tempted, now that you point this out, to write a helper to `yield from textwrap.dedent(...).split("\n")` on a block of text. And then I could write such things in a single multiline format string, as
```python
yield from _tolines("""...""")
yield from [f"* ``{s}``" for s in known_scopes]
yield from ["", ""]
```

Why don't we save that for now / leave it alone, and I can put in a follow up PR to try that out, to see if it makes the extension code much cleaner?

sirosen: By way of demonstration, I tried this out.
<details><summary>tolines.py example</summary>

```
$ cat tolines.py && echo "---" && python tolines.py
import textwrap


def _tolines(*args):
    for x in args:
        if isinstance(x, str):
            yield from textwrap.dedent(x).split("\n")
        else:
            yield from x


print(
    list(
        _tolines(
            """
            A lovely string
            to report
            about a wonderful thing
            """,
            [f"x={x}" for x in ("a", "b", "c")],
            "",
            "",
        )
    )
)
---
['', 'A lovely string', 'to report', 'about a wonderful thing', '', 'x=a', 'x=b', 'x=c', '', '']
```
</details>

rudyardrichter (approved):

sirosen:
Tests are failing on this I believe because of changes in the latest release of `charset_normalizer`. In particular, [this check appears to be new](https://github.com/Ousret/charset_normalizer/blob/480fee6239bb333f5ff46bc8032a7460a1201b60/charset_normalizer/api.py#L94-L95).

I haven't figured out the full cause and what to do yet, but I don't think we should block this change on it. Something is going wrong with the test response building code, but it's broken on `main` too. I'm going to merge in spite of the failing build.

PR #441, opened 2021-07-14
Refactor error parsing to have clearer flow and handoff to "ErrorInfo" objects

sirosen:
This is a refactor meant to improve the flow of control through error instantiation to be clearer and easier to follow.

It also addresses the fact that error parsing is currently built to only be extensible via inheritance, and is not composable. For the first part, it's just a matter of general cleanup and reorganizing. But for composability, this PR introduces a set of "info" objects which can be attached to an error.

For example, it allows use of `err.info.consent_required`, the "consent required info" for an error.
Code can then be structured as
```python
if err.info.consent_required:
    consent_required_handler(err.info.consent_required.required_scopes)
```

Summary of the changes:

- `exc.py` is now a subpackage, to let it grow a bit bigger
- Rewrite error parsing from `GlobusAPIError.__init__` so that the initializer doesn't have any internal control-flow logic, just a series of assignments and calls
- Use `raw_json` when parsing data, rather than repeating content-type checking in multiple places
- Be more careful in general about checking for missing keys or unexpected datatypes in error bodies
- `GlobusAPIError._parse_response` is responsible for handling multi-error and single error responses, `GlobusAPIError._load_from_json` is still the hook for handling a JSON body if one is present
- The message fields for an error class is now a classvar, simplifying the `AuthAPIError` at no great cost to the base class
- Add `ErrorInfo` objects, with `__bool__` defined in terms of an inner attribute, `_has_data`
- Specific `ErrorInfo` objects for ConsentRequired and AuthorizationParameters. See [the relevant globus-cli logic](https://github.com/globus/globus-cli/blob/0b0565bd3124451916f744e6bc965d5d91aae0b1/src/globus_cli/parsing/excepthook.py#L204-L220) for the motivation for these
- `ErrorInfo` objects are wrapped in an `ErrorInfoContainer` and attached to API errors
- New doc on the exceptions page

The end result is that as additonal classes of errors are added to Globus services, we could easily add more `ErrorInfo` objects to the `ErrorInfoContainer` at minimal cost.


src/globus_sdk/exc/api.py
```diff
@@ -0,0 +1,167 @@
...
+
+        _parse_response() pulls the JSON body and does the following:
+        - if the data is not a dict, ensure _load_from_json is not called (so it only
+          gets called on dict data)
+        - if the data contains an 'errors' array, pull out the first error document and
+          pass that to _load_from_json()
+        - log a warning on non-dict JSON data
+        """
+        json_data = self.raw_json
+        if json_data is None:
```
rudyardrichter, line 120: It looks like this could be organized to "de-dent" the main block here, which I'm a proponent of:
```python
if json_data is None:
    ...
    return
if not isinstance(json_data, dict):
    log.warning("could not be parsed as JSON")
    return
# de-dented `isinstance(json_data, dict)` block goes here
...
```

src/globus_sdk/exc/api.py
```diff
@@ -0,0 +1,167 @@
...
+        try:
+            authz_h = self._underlying_response.request.headers["Authorization"]
+            try:
+                authz_scheme = authz_h.split()[0]
+                if authz_scheme.lower() in self.RECOGNIZED_AUTHZ_SCHEMES:
+                    return authz_scheme
+            except IndexError:
+                pass
+        except KeyError:
+            pass
```
rudyardrichter, line 90: This would be the same, right?
```suggestion
        try:
            authz_h = self._underlying_response.request.headers["Authorization"]
            authz_scheme = authz_h.split()[0]
            if authz_scheme.lower() in self.RECOGNIZED_AUTHZ_SCHEMES:
                return authz_scheme
        except (IndexError, KeyError):
            pass
```

src/globus_sdk/exc/api.py
```diff
@@ -0,0 +1,167 @@
...
+        """
+        Get the verbatim error message received from a Globus API, interpreted
+        as JSON data
+
+        If the body cannot be loaded as JSON, this is None
+        """
+        r = self._underlying_response
+        if "Content-Type" in r.headers and (
+            "application/json" in r.headers["Content-Type"]
+        ):
```
rudyardrichter, line 47: This could also "if NOT thing, return none; now main block here, de-dented." (I get that it's a bit of a subjective thing, especially in smaller chunks like here.)

sirosen: Yeah, I tend towards that style as well. I think the flatter we can manage to make things, the better. I'll do this one for sure.

src/globus_sdk/exc/api.py
sirosen: Yeah, I think the original was "stream of consciousness error handling". There's no need for separate handlers here.

src/globus_sdk/exc/api.py
sirosen: Yep, I'm in favor. 👍

src/globus_sdk/exc/api.py
sirosen: Writing it this way leads me to add
```python
    def _json_content_type(self) -> bool:
        r = self._underlying_response
        return "Content-Type" in r.headers and (
            "application/json" in r.headers["Content-Type"]
        )
```

so that it can be `if not self._json_content_type(): ...` and read more easily

rudyardrichter (approved):

PR #442, opened 2021-07-26
Add type annotations

rudyardrichter:
- Add `mypy.ini` with setting to require signatures on all function definitions
- Add type annotations everywhere to satisfy said mypy setting
- Several things flagged with `REVIEW`, a couple TODOs, mostly relating to things where I can't figure out what the annotations should be or why something is working—should be explained at those locations

- A point of discussion: the mypy setting requires a `-> None` on init methods if they do not have any type-annotated arguments, and otherwise doesn't. For uniformity I added it on everything whether it was strictly required or not, but we could remove some of those.


rudyardrichter:
Also note: build currently failing is expected for the moment because this needs to be addressed (the main things I couldn't figure out):
```
src/globus_sdk/services/search/data.py:59: error: Function is missing a type annotation for one or more arguments
src/globus_sdk/services/search/data.py:83: error: Function is missing a type annotation for one or more arguments
src/globus_sdk/services/search/data.py:93: error: Function is missing a type annotation for one or more arguments
```

sirosen:
I'll take a look at the content of this starting with items marked for review.

> the mypy setting requires a `-> None` on init methods if they do not have any type-annotated arguments, and otherwise doesn't. For uniformity I added it on everything whether it was strictly required or not

I think this is the right approach. It's easier to make writing `-> None` the rule than to try to remember when it is or is not required.

src/globus_sdk/services/search/data.py
```diff
@@ -55,15 +55,16 @@ def set_advanced(self, advanced: bool) -> "SearchQuery":
         self["advanced"] = advanced
         return self
 
+    # TODO: type signature
```
sirosen, line 58: Ah, yes, this requires knowledge of the service.
FWIW, I'd go with
```
name: str
field_name: str
size: int
date_interval: str
histogram_range: Tuple[Any, Any]
```

src/globus_sdk/services/search/data.py
```diff
@@ -55,15 +55,16 @@ def set_advanced(self, advanced: bool) -> "SearchQuery":
...
         self,
         name,
         field_name,
-        type="terms",
+        type: str = "terms",
         size=None,
         date_interval=None,
         histogram_range=None,
-        **kwargs
+        **kwargs: Any
```
sirosen, line 67: This should probably become an `additional_fields` dict...

Why don't we kind of ignore `SearchData` for the moment, and I can circle back with a PR to tidy it up and add more annotations after this PR merges? That way, we don't get bogged down in oddities like this. Plus, `histogram_range` being a tuple is odd and makes me want to take another look (but maybe it's fine...).

src/globus_sdk/types.py
```diff
@@ -0,0 +1,12 @@
...
+
+from typing_extensions import Protocol
+
+# From: https://docs.python.org/3/glossary.html#term-bytes-like-object
+BytesLike = Union[array, bytes, bytearray, memoryview]
+
+
+class ToStr(Protocol):
+    def __str__(self) -> str:
+        ...
```
sirosen, line 12: Since there is a default definition of `__str__ = __repr__`, is this protocol significantly different from `Any`?

Does this enforce that the object must define an explicit, non-default `__str__`?

src/globus_sdk/services/transfer/client.py
```diff
@@ -5,17 +5,19 @@
...
+from globus_sdk.types import ToStr
 
 from .data import DeleteData, TransferData
 from .errors import TransferAPIError
 from .response import ActivationRequirementsResponse, IterableTransferResponse
 
 log = logging.getLogger(__name__)
 
+# REVIEW: can this be ToStr?
 ID_PARAM_TYPE = Union[bytes, str, uuid.UUID]
```
sirosen, line 17: My response to this is that I want to be more certain that `ToStr` gives a similar level of safety. (See my comment there, about the default `__str__`.)

For example, would `ToStr` prevent this?
```python
epid = "foo",  # whoops, you made a tuple!
tc.get_endpoint(epid)
```

src/globus_sdk/services/transfer/client.py
```diff
@@ -739,10 +743,11 @@ def add_endpoint_role(
         path = self.qjoin_path("endpoint", endpoint_id_s, "role")
         return self.post(path, data=role_data)
 
+    # REVIEW: why do we call str() on server ID above but not on role_id here?
```
sirosen, line 746: I'm pretty sure we do not need to do so. I would guess that the source of this inconsistency is that `server_id` values are integers, but `role_id` values are UUIDs.

Perhaps the right solution is `server_id: Union[str, int]`, `role_id: ID_PARAM_TYPE`, and `safe_stringify()` or `str()` on everything.

src/globus_sdk/services/transfer/client.py
```diff
@@ -2177,14 +2182,16 @@ def endpoint_manager_cancel_status(
         <https://docs.globus.org/api/transfer/advanced_endpoint_management/#get_cancel_status_by_id>`_
         in the REST documentation for details.
         """
+        # REVIEW:
+        admin_cancel_id = utils.safe_stringify(admin_cancel_id)
```
sirosen, line 2186: I'm surprised that you were able to write it like this, rather than needing a new var name.

If you want review on the proper types for `admin_cancel_id` in the first place, I think we should ask for @aaschaer to weigh in. I'm pretty sure it's a UUID, but it would be good to confirm.

src/globus_sdk/services/transfer/data/delete_data.py
```diff
@@ -66,14 +71,14 @@ class DeleteData(utils.PayloadWrapper):
...
-        label=None,
-        submission_id=None,
-        recursive=False,
-        deadline=None,
+        transfer_client: "TransferClient",
+        endpoint: str,
+        label: Optional[str] = None,
+        submission_id: Optional[str] = None,
+        recursive: bool = False,
+        deadline: Optional[Union[str, datetime.datetime]] = None,
```
sirosen, line 79: Seeing this made me suspicious of the call to `str()` below. Somewhat unrelated to typing, but I think we should we swap that to a call to `strftime()` if we're taking in `datetime` objects, so that we strictly control the format.

I'm not sure if `str(datetime)` guarantees us a good value.

src/globus_sdk/services/transfer/data/transfer_data.py
```diff
@@ -184,6 +188,7 @@ def __init__(
         # garbage overnight
         if sync_level is not None:
             sync_dict = {"exists": 0, "size": 1, "mtime": 2, "checksum": 3}
+            # TODO: sync_level not allowed to be int?
```
sirosen, line 191: I think `sync_level: Union[None, str, int]` would be best. We use the `get()` call below to handle the `int` case, and we should allow the int since that's what the API and API docs use.

src/globus_sdk/transport/encoders.py
```diff
@@ -1,4 +1,7 @@
+from typing import Any, Dict, Optional
+
 import requests
+from requests.models import Request
```
sirosen, line 4: Was it necessary to switch from `requests.Request` to `requests.models.Request`?
We should prefer the top-level namespace of external packages since there's a stronger expectation of stability there (even across major versions).

src/globus_sdk/transport/retry.py
```diff
@@ -72,8 +72,8 @@ def set_retry_check_flags(flag: RetryCheckFlags):
     >>> def foo(ctx): ...
     """
 
-    def decorator(func):
-        func._retry_check_flags = flag
+    def decorator(func: Callable) -> Callable:
+        func._retry_check_flags = flag  # type: ignore
```
sirosen, line 76: This reminds me. If we're pulling in `typing_extensions`, we can use `Protocol`. That lets us do an alternate fix in these cases (function attribute addition) in which we cast the func to a protocol class. Not sure if it's worth it.

But also, I'm 👍 on using an ignore for this -- I think I've done so elsewhere for the same purpose.

src/globus_sdk/types.py
```diff
@@ -0,0 +1,12 @@
+from array import array
+from typing import Union
+
+from typing_extensions import Protocol
+
+# From: https://docs.python.org/3/glossary.html#term-bytes-like-object
+BytesLike = Union[array, bytes, bytearray, memoryview]
```
sirosen, line 7: I don't see this being used anywhere. Am I missing it?

`bytes` are often returned or required by python APIs -- DBs, wsgi, etc. -- so there's some definite value in supporting that type for string-like parameters. But these other types are more obscure. I'm not sure we want to worry about supporting them.

src/globus_sdk/utils.py
```diff
@@ -1,13 +1,17 @@
...
 
-def sha256_string(s):
+
+def sha256_string(s: str) -> str:
     return hashlib.sha256(s.encode("utf-8")).hexdigest()
 
 
-def safe_b64encode(s):
+# REVIEW
+def safe_b64encode(s) -> str:  # type: ignore
```
sirosen, line 14: I think the intent -- based on the function body -- is `s: Union[bytes, str]`.

src/globus_sdk/utils.py
```diff
@@ -32,7 +36,8 @@ def slash_join(a, b):
     return a + "/" + b
 
 
-def safe_stringify(value):
+# REVIEW: bytes has a str method but that isn't how this converts bytes here.
+def safe_stringify(value: ToStr) -> str:
```
sirosen, line 40: I think this comment touches on my point of concern with `ToStr` as a type.

`value: Any` is accurate for what this function does and the broadness of its usage. Plus, it's an internal method in a `utils` module, so it's not like we need an annotation which explains it to users.

mypy.ini
```diff
@@ -0,0 +1,3 @@
+[mypy]
+
+disallow_untyped_defs = True
```
sirosen, line 3: We could probably get more aggressive on these (see the config for globus-cli for example: https://github.com/globus/globus-cli/blob/0b0565bd3124451916f744e6bc965d5d91aae0b1/setup.cfg#L15-L22 ), but we can save any updates as a future refinement.

But I think it would be better to put this into `setup.cfg` to try to stay consistent between repos.

sirosen (commented):
I know I said I'd start with the TODO/REVIEW items, but I ended up reading everything anyway because I had to find those items.

For the most part, I'm very pleased. There are some fiddly bits with comments inline (e.g. pyjwt).
The only broad/overarching thing that I think we need to address is `ToStr`. I have some inline comments on it, but, in short, I think `object` and most (all?) primitive types satisfy `ToStr`, which makes it very similar to `Any`. e.g.
```python
>>> import typing as t
>>> class ToStr(t.Protocol):
...     def __str__(self) -> str: ...
...
>>> rt = t.runtime_checkable(ToStr)
>>> x = object()
>>> isinstance(x, rt)
True
>>> def foo(): return 1
...
>>> isinstance(foo, rt)
True
>>> y = (1, 1)
>>> isinstance(y, rt)
True
```

So I think we may need to rethink that approach. For some cases, `Any` may actually be appropriate. And in others, we may want to define a smaller `Union` type, like `ID_PARAM_TYPE`.

src/globus_sdk/services/search/data.py
sirosen: I just realized, rereading your original comments, that this is why we're failing the mypy untyped defs check.
I can queue up some suggested annotations like the above right now, if you like.

src/globus_sdk/paging/table.py
```diff
@@ -108,7 +111,7 @@ def __getattr__(self, attrname):
 
     # customize pickling methods to ensure that the object is pickle-safe
 
-    def __getstate__(self):
+    def __getstate__(self) -> Dict[str, "AuthClient"]:
```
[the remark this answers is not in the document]
rudyardrichter: Yep, this was probably a `monkeytype` over-specification I missed.

src/globus_sdk/services/auth/client_types/base.py
```diff
@@ -350,14 +359,22 @@ def oauth2_revoke_token(self, token, body_params: Optional[Dict[str, Any]] = Non
...
+        self, form_data: Union[dict, utils.PayloadWrapper]
+    ) -> OAuthTokenResponse:
         ...
 
-    def oauth2_token(self, form_data, response_class=OAuthTokenResponse):
+    def oauth2_token(
+        self,
+        form_data: Union[dict, utils.PayloadWrapper],
+        response_class: Union[Type[OAuthTokenResponse], Type[T]] = OAuthTokenResponse,
+    ) -> Union[OAuthTokenResponse, T]:
```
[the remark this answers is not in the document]
rudyardrichter: I had the same thought, but that produces
```
Incompatible default for argument "response_class" (default has type "Type[OAuthTokenResponse]", argument has type "Type[T]")
```
which seems to be a symptom of [this issue](https://github.com/python/mypy/issues/3737)

src/globus_sdk/services/auth/client_types/base.py
```diff
@@ -452,4 +500,5 @@ def get_jwk(self, openid_configuration=None, as_pem=False):
                 json.dumps(jwk_data["keys"][0])
             )
             log.debug("JWK PEM decoding finished successfully")
-            return jwk_as_pem
+            # type-ignore because should never be private key
```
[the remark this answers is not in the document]
rudyardrichter: `from_jwk` can return a public or private key (though not in our case). `cast` is a good suggestion.

src/globus_sdk/services/auth/client_types/native_client.py
```diff
@@ -99,7 +101,10 @@ def oauth2_start_flow(
         )
         return self.current_oauth2_flow_manager
 
-    def oauth2_refresh_token(self, refresh_token):
+    # REVIEW: this method has to also include slot for body_params kwarg
```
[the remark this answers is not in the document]
rudyardrichter: I hadn't realized it, but this is a good point from at least the "ethical" perspective. However, I'm not sure I follow what you mean about support in `oauth2_token`?

src/globus_sdk/services/auth/oauth2_flow_manager.py
```diff
@@ -29,11 +36,10 @@ def get_authorize_url(self) -> str:
 
         :rtype: ``string``
         """
-        raise NotImplementedError(
-            f"{type(self)} does not implement get_authorize_url()"
-        )
+        pass
```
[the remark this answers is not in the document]
rudyardrichter: I don't mind it, it should be obvious from the decorators what's going on here.

src/globus_sdk/services/auth/client_types/base.py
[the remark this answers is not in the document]
sirosen: Oy vey... I've run into that minefield before. Let's leave it with your solution, in that case.

rudyardrichter:
@sirosen thanks for a lot of useful comments _and_ a quick typeshed PR! I'm still going through some of the "low-level" stuff at the moment, so to move along the major question that is `ToStr`: I think the practical description of things with that type is "stuff we can feed to `safe_stringify` and get a sensible result". So `str` or `BytesLike` (this is what I was trying to use that for originally) is OK; arbitrary objects with `__str__` are not. Do you think we could restrict the functionality within `safe_stringify` itself to only work on some set of types, and then we could have something like:
```
ToStr = Union[str, bytes, uuid.UUID]
```

src/globus_sdk/services/auth/client_types/native_client.py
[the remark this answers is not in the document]
sirosen: `AuthClient.oauth2_token` doesn't support `body_params`, but if we added that, we could take `body_params` here and then call `self.oauth2_token(form_data, body_params=body_params)`.

src/globus_sdk/services/transfer/client.py
rudyardrichter: Pointing to this as a specific example, let me return to this question with a better phrasing: if we make `ToStr` a more specific interface, perhaps a slight superset of `ID_PARAM_TYPE` here, does it capture the original purpose of `ID_PARAM_TYPE` to use `ToStr` instead?

sirosen: I think the thing to compare against is something like `server_id`, which is an integer.
Without worrying about generalizing, these are probably the desirable types:
```
server_id: int | str | bytes
endpoint_id: UUID | str | bytes
```

So there's no one type we can write which will express this for us. Why not just write them down as separate types?
We could define these globally, e.g. `from globus_sdk.types import INT_T, UUID_T`.

rudyardrichter: So, "this type or a string of it"? And then probably `ToStr` is no longer needed. What about `IntLike`/`UUIDLike`?

src/globus_sdk/services/search/data.py
rudyardrichter: If you have the time to add the annotations for the search modules, that would be stellar and, as you noticed, clear up the current warnings. Otherwise we can postpone the strict checks on those modules.

src/globus_sdk/services/transfer/client.py
sirosen: Yeah, I think that's how we want to think about this: "an integer, or a string of it; a uuid, or a string of it; etc".

I'm cool with those names for the types. `IntLike` seems clear to me.

I'll go ahead and ask, since it seems related: do we think supporting `bytes` is even important? If we discard that, we could probably replace all calls to `safe_stringify` with just `str()`. (There's a note about string duplication in that function, but I think that was a python2 behavior. `str()` of a string just returns the string now.)

I know I mentioned that there are python interfaces which handle bytes, but dropping support for this from the SDK would just mean that users with unencoded bytes have to call `encode()` themselves. It's not a big deal for them and it's a simpler surface for us to maintain.

src/globus_sdk/services/search/data.py
sirosen: I'll open a PR against your fork with the addition of these types. Hopefully that will get things passing.

sirosen:
I just noticed that we have mypy options in `setup.cfg` already. I think adding the `mypy.ini` config suppresses the use of those options. We should merge the two together so that all of those settings are enabled (like `warn-unreachable`).

I'll see if I can just take care of this in a single commit, along with annotations for SearchClient.

src/globus_sdk/services/transfer/client.py
aaschaer: Yes, admin_cancel_id must be a UUID

src/globus_sdk/services/auth/client_types/base.py
```diff
@@ -1,15 +1,20 @@
 import collections.abc
 import json
 import logging
-from typing import Any, Dict, Optional, Type, TypeVar, overload
+from typing import Any, Dict, List, Optional, Type, TypeVar, Union, overload
 
 import jwt
+from cryptography.hazmat.primitives.asymmetric.rsa import RSAPublicKey
+from typing_extensions import Literal
```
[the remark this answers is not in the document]
sirosen: Per discussion, use quoted types as in `"Literal['foo']"`, and try to grab from stdlib.

i.e.
```python
if TYPE_CHECKING:
    try:
        from typing import Literal
    except ImportError:
        from typing_extensions import Literal
```

src/globus_sdk/services/auth/client_types/base.py
```diff
@@ -395,10 +417,13 @@ def oauth2_token(
         log.info("Fetching new token from Globus Auth")
         # use the fact that requests implicitly encodes the `data` parameter as
         # a form POST
+        data = dict(form_data)
+        if body_params:
+            data.update(body_params)
```
sirosen, line 418: FWIW, `utils.PayloadWrapper` supports `update()`, so I think you could skip the conversion. But also, this is not at all important. 🤷

src/globus_sdk/utils.py
```diff
@@ -46,7 +46,11 @@ def safe_stringify(value: ToStr) -> str:
     if isinstance(value, str):
         return value
     if isinstance(value, bytes):
-        return value.decode("utf-8")
+        try:
+            return value.decode("utf-8")
+        except UnicodeDecodeError:
+            # this raises its own ValueError if value doesn't decode into a UUID
+            return str(uuid.UUID(bytes=value))
```
sirosen, line 53: I'd rather not make this conversion. I don't know if UUID bytes are guaranteed to _not_ be valid UTF8.

Also, it just doesn't seem very necessary to me? If a caller has the bytes for a UUID, they can do this themselves.

src/globus_sdk/services/auth/client_types/base.py
rudyardrichter: Actually the thought was to copy to not side-effect update `form_data`. Probably doesn't matter; just a defensive habit.

src/globus_sdk/services/auth/client_types/base.py
```diff
@@ -69,8 +69,8 @@ def __init__(self, client_id: Optional[str] = None, **kwargs: Any) -> None:
 
     def get_identities(
         self,
-        usernames: Optional[Union[List[ToStr], ToStr]] = None,
-        ids: Optional[Union[List[ToStr], ToStr]] = None,
+        usernames: Optional[Union[Sequence[ToStr], ToStr]] = None,
+        ids: Optional[Union[Sequence[ToStr], ToStr]] = None,
```
sirosen, line 73: I think I'd prefer to see
```
usernames: Sequence[AnyStr] | AnyStr | None
ids: Sequence[UUIDLike] | UUIDLike | None
```
([typing.AnyStr](https://docs.python.org/3/library/typing.html#typing.AnyStr))

Otherwise, you can pass `ids=4` or `usernames=UUID(...)`, neither of which seem valuable to allow.

src/globus_sdk/utils.py
rudyardrichter: Yeah, I wanted to put this in as a demonstration of "get as close to we can to stringifying whatever the user puts in". I'm similarly happy with a more stringent version.

src/globus_sdk/services/auth/client_types/base.py
rudyardrichter: Yes, these and some other current `ToStr` uses (especially in TransferClient) still need updating to use the more specific versions. I'm in the middle of that pass now.

rudyardrichter: Sorry for the noise from the reviewing perspective, I was trying to link specific commits so you wouldn't have to troll through everything and sort out what still did or didn't need looking at.

src/globus_sdk/services/auth/client_types/base.py
sirosen: Ah! A good habit.

When I do this (and I don't always, which is maybe a _bad_ habit), I comment on it even if it's just `# copy`. As evidenced here, I easily find it confusing otherwise. 😅

src/globus_sdk/services/auth/client_types/base.py
sirosen: No worries at all. I'll wait to look until later. I'm just trying to get you feedback before I go on break!

sirosen:
I think there are a handful of remaining items which need resolution here.
I'll do a "final pass" and note or bump anything which still needs addressing.

src/globus_sdk/utils.py
sirosen: I accidentally resolved this at one point.
Needs the adjustment + removal of `REVIEW` comment.

src/globus_sdk/utils.py
```diff
@@ -41,7 +45,10 @@ def safe_stringify(value):
     if isinstance(value, str):
         return value
     if isinstance(value, bytes):
-        return value.decode("utf-8")
+        try:
+            return value.decode("utf-8")
+        except UnicodeDecodeError:
+            raise ValueError("couldn't decode bytes as utf-8: {value}")
```
sirosen, line 51: This would need to be an f-string to render. But I don't think it's even desirable, since rendering bytes as part of an error about failure to render bytes is... confusing.

Why not just stick with the original line here, `value.decode("utf-8")`?

If it's important to reraise something else here, it should chain from the original error (`raise from`).

sirosen (changes requested):
There are only a few more changes which I think are needed, and then we'll be done.

rudyardrichter:
I forgot to raise in our sync that the try import typing/except import typing_extensions has an issue:
```
src/globus_sdk/services/auth/client_types/base.py:25: error: Incompatible import of "Literal" (imported name has type "typing_extensions._SpecialForm", local name has type "typing._SpecialForm")
```
Per [mypy's advice here](https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-new-additions-to-the-typing-module), I think this is the desired 
```python
if TYPE_CHECKING:                                                                                                                                                             
    if sys.version_info >= (3, 8):
        from typing import Literal
    else:
        from typing_extensions import Literal
```

src/globus_sdk/services/auth/token_response.py
```diff
@@ -168,10 +169,10 @@ def by_scopes(self):
 
     def decode_id_token(
         self,
-        openid_configuration=None,
-        jwk=None,
+        openid_configuration: Optional[Dict[str, Any]] = None,
```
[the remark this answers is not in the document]
rudyardrichter: That makes a lot of sense, as you note, `oidc_config` is like that anyways and `get_jwk` accepts that type.

sirosen:
> Per [mypy's advice here](https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-new-additions-to-the-typing-module), I think [importing `Literal` from `typing` on py3.8+] is the desired approach

The `mypy` advice here seems good. Do we also want to specify `typing_extensions; python_version<"3.8"` in the dependencies, the way they suggest? It hardly matters, but it will make it easier to remember to remove it when support for py3.7 eventually drops.

rudyardrichter:
Sure, we might as well.

sirosen (approved):
It will always be hard to guarantee that something this big and broad is perfect, but I think we've been as diligent and careful as we can be. I'm going to merge it, and we can make any necessary further adjustments in the coming weeks or even, in most cases, after v3.0 .

PR #443, opened 2021-08-13
Make group_id and identity_id params UUIDLike types

sirosen:
Applies mostly to GroupsClient and GroupsManager.

Also included here: `mypy` no longer requires that we rename a var which is a union (including a string) which is then being assigned a string value, so there's some cleanup on TransferCleint. The "intuitive" type narrowing is now applied and the variable's deduced type can be narrowed without a rename. `endpoint_id_s = safe_stringify(endpoint_id)` is no longer necessary, and we now write `endpoint_id = safe_stringify(endpoint_id)`.


rudyardrichter (approved):
Looks fine to me. Nice that we can eliminate the `x_id_s = ...` everywhere.

PR #466, opened 2021-09-01
Add method to get local GCP owner from gridmap

sirosen:
Minimally parse the gridmap file, find a GCP DN, and then handle the result. It can be a globusid.org username or auth ID, so these are flagged on return. If users want a uniform(ish) interface, they can pass in an AuthClient, which is used to do an get-identities call to lookup the user info.

Skip coverage reporting on overloads for more accurate coverage info.

---

@michaellink, this is the feature we discussed the other day, as implemented by the SDK.
If you can spare a minute to look at how I've handled the GCP gridmap file, any feedback or info you can give would be helpful.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -13,3 +13,5 @@ exclude_lines =
     if __name__ == .__main__.:
     # don't check coverage on type checking conditionals
     if TYPE_CHECKING:
+    # skip overloads
+    @overload
--- a/docs/local_endpoints.rst
+++ b/docs/local_endpoints.rst
@@ -21,7 +21,10 @@ Globus Connect Personal
 Globus Connect Personal endpoints belonging to the current user may be accessed
 via instances of the following class:
 
-
 .. autoclass:: globus_sdk.LocalGlobusConnectPersonal
+   :members:
+   :member-order: bysource
+
+.. autoclass:: globus_sdk.GlobusConnectPersonalOwnerInfo
    :members:
    :member-order: bysource
--- a/src/globus_sdk/__init__.py
+++ b/src/globus_sdk/__init__.py
@@ -17,7 +17,7 @@ from .exc import (
     GlobusTimeoutError,
     NetworkError,
 )
-from .local_endpoint import LocalGlobusConnectPersonal
+from .local_endpoint import GlobusConnectPersonalOwnerInfo, LocalGlobusConnectPersonal
 from .response import GlobusHTTPResponse
 from .services.auth import (
     AuthAPIError,
@@ -90,6 +90,7 @@ __all__ = (
     "GroupsManager",
     "GCSClient",
     "GCSAPIError",
+    "GlobusConnectPersonalOwnerInfo",
     "LocalGlobusConnectPersonal",
 )
 
--- a/src/globus_sdk/local_endpoint/__init__.py
+++ b/src/globus_sdk/local_endpoint/__init__.py
@@ -1,3 +1,6 @@
-from globus_sdk.local_endpoint.personal import LocalGlobusConnectPersonal
+from .personal import GlobusConnectPersonalOwnerInfo, LocalGlobusConnectPersonal
 
-__all__ = ("LocalGlobusConnectPersonal",)
+__all__ = (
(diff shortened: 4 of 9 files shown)


michaellink:
LGTM.  Probably a good idea to document that the local info (both the owner and endpoint id) is editable, so it shouldn't be trusted for things like granting access without forcing the user to verify via login/endpoint ls.  Can't quite come up with an example scenario -- app creds with an admin role that adds acls based on those values or something.

aaschaer (commented):
Two things

1. I'm not sure how I feel about the str, bool tuple for determining if the value is an id or username. I might prefer returning a UUID for ids and asking the user to rely on the return value's type to know what they got, but I guess that's not any easier for the user than checking the bool we return.

2. I expect no-one will have this issue but me, but I use custom directories for GCP since I run multiple GCPs against multiple environments for testing so the code didn't work out of box for me. Once I copied one of my directories into the expected path everything worked fine

sirosen:
> I'm not sure how I feel about the str, bool tuple ...

I agree. I'm not clear on what the best interface for users is.
The only other good alternative I can come up with is a namedtuple or object with two (nullable) slots. e.g.:
```python
>>> GCPOwnerInfo = namedtuple("GCPOwnerInfo", ("username", "id"))
>>> GCPOwnerInfo(username="sirosen@globusid.org", id=None)
```

If you prefer this, let me know and I'm happy to swap it in for the current `(str, bool)` implementation.

> I expect no-one will have this issue but me, but I use custom directories for GCP since I run multiple GCPs against multiple environments for testing so the code didn't work out of box for me. Once I copied one of my directories into the expected path everything worked fine

I think users might be using the GCP config dir options as well, come to think of it.
Let's not do this right now, but instead make the GCP config dir an `__init__` option on the local endpoint object. If set, we'll use the specified dir in lieu of the expanduser/appdata one. That will solve it both for the new feature and for `client-id.txt`. I'll write this up as an improvement.

aaschaer:
Oh I like that named tuple! Makes usage more similar to when passing the auth client as well.

rudyardrichter:
Not knowing how this is expected to be used, I think `dataclass` would also work well for that, and allow defining a `__str__` method for if the caller doesn't care which it is:
```python
from dataclasses import dataclass


@dataclass
class GCPOwner:
    username: Optional[str]
    id: Optional[str]

    def __str__(self):
        if self.username:
            return self.username
        else:
            return self.id
```

sirosen:
`dataclass` is py3.7+ @rudyardrichter, always with these new features we want to use! 😜 

But maybe a custom dataclass-like object is better, since that lets us slap our own `__str__` on there, as you say.

We could even make it hold the `AuthClient` object and use that to magically lookup `id` or `username` if provided. Something like
```python
class GCPOwnerInfo:
    # imagine the __init__ with me, if you will...
    _auth_client: Optional[AuthClient]
    _id: Optional[str]
    _username: Optional[str]
    identity_info: Optional[Dict]

    @property
    def id(self): ...

    @property
    def username(self): ...

    def resolve(self, auth_client: Optional[AuthClient]): ...
```

rudyardrichter:
Argh! Sorry. One day I will break my habit...

sirosen:
I converted to a dedicated object and decided not to worry about passing in the AuthClient or doing "smart lookups" or anything like that. I felt that it had gotten complicated enough, and wanted to rein in the complexity a bit.

aaschaer (commented):
The only thing I might change is to make GlobusConnectPersonalOwnerInfo a dict-like object, so adding a client doesn't change usage:

`local.get_owner_info()["username"]`
`local.get_owner_info(client)["username"]`

But if we don't want to do that then I think this is good to go

sirosen:
It's tempting to add `__getitem__`, but it might be confusing.
`local.get_owner_info()["username"]` can fail (or "soft fail" by returning `None`, if we prefer). `local.get_owner_info(client)["username"]` will almost never fail -- the only way it can is when handling a legacy identity which is set to private.

Let's leave it out for now, and we can always add `__getitem__` later if we think it's worth it.

aaschaer (approved):

PR #467, opened 2021-09-02
Add filter_is_error to advanced task event list

sirosen:
A user reported this parameter missing in 3.0b4. It's a casualty of the `**params` => `query_params={...}` passthrough change, so this adds it as an explicit parameter.

The param is added as a bool. However, testing covers setting it as an int. This is the usage which was seen in the wild under SDK v2.x which prompted this addition, and checking that it is supported (although not documented and not to be encouraged) is simple enough.


aaschaer (approved):
Looks good to me

PR #469, opened 2021-09-10
Replace regex match in UnpackingGCSResponse

sirosen:
Instead of passing `1\.\d+\.\d+` to match a version range, this improves the interface to allow passing a simple version spec.

`collection>=1,<2` will check for `collection#x.y.z` where `x=1`, and so forth. The version specs work basically the same way that pypa package version specifiers work, and version specifiers for most packaging systems.

In order to parse these, a module is added which splits up the string and converts it into a "version range". Tests cover all of the major conditions.

---

Since we're going to release 3.0 final soon, it's our last chance to improve things like this which are backwards-incompatible changes. I'd like to get this in, or we'll be stuck with the regex matching usage for a long time.


sirosen:
@rudyardrichter, based on our conversation about this, and the follow-up I did with the GCS folks, I've pared this back to the non-version-spec variant.

It's now just `UnpackingGCSResponse(..., "collection")` and that "does the right thing" (as far as we've been able to determine).
It's "future compatible" with `collection>=1,<2`, etc. specifications.

Since we agreed that we're comfortable with this for v3.0, I'm going to merge.

PR #470, opened 2021-09-11
Add support for specifying config_dir to local GCP

sirosen:
@aaschaer, this is the follow up on what you suggested for the local endpoint object when we were looking at the owner-info change. Adds `config_dir`.
If you can get a moment to look at this before we cut the release, that's great, but we can also add this feature to v3.1 or later of the SDK since it's purely additive.

The local endpoint object now supports `config_dir: str`. Because this value is detected when `None`, it effectively replaces some of the `_local_data_dir` logic with something a bit more readable.

In order to align with GCP Linux's `-dir` option, the directory on non-Windows platforms is `~/.globusonline/`, not
`~/.globusonline/lta`. This means that there is still a local data dir property which joins with `"lta"` on non-Windows platforms.
(Don't blame the SDK for this. It's a very, very old GCP behavior and we're just aligning with it here.)

### DIFF
--- a/src/globus_sdk/local_endpoint/personal/endpoint.py
+++ b/src/globus_sdk/local_endpoint/personal/endpoint.py
@@ -25,33 +25,46 @@ class LocalGlobusConnectPersonal:
 
     These objects do *not* inherit from BaseClient and do not provide methods
     for interacting with any Globus Service APIs.
+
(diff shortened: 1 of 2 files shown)


sirosen:
I've just self reviewed this and I'm going to merge. After thinking about it a few more days, I'm concerned that releasing the feature without this support will be useless for too many users. So I'd like this in 3.0 .

PR #472, opened 2021-09-14
mypy bugfix: allow dynamic ScopeBuilder attrs

sirosen:
A custom `__getattr__` implementation tells `mypy` to expect `str` for unknown attributes of a ScopeBuilder object.

This is the approach documented in the mypy cheat sheet [1] as a way of telling mypy the type for "unknown attributes".

[1] https://mypy.readthedocs.io/en/latest/cheat_sheet_py3.html#when-you-re-puzzled-or-when-things-are-complicated

---

I ran `mypy` with this change before and after on the documented usage. Fails before; works after.

### DIFF
--- a/src/globus_sdk/scopes.py
+++ b/src/globus_sdk/scopes.py
@@ -29,8 +29,8 @@ class ScopeBuilder:
     #       from globus_sdk.scopes import TransferScopes
     #       x = TransferScopes.all
     #
-    # the assignment to `x` will fail type checking without this because `all` will be
-    # unknown to `mypy`
+    # without this method, the assignment to `x` would fail type checking
+    # because `all` is unknown to mypy
     #
     # note that the implementation just raises AttributeError; this is okay because
     # __getattr__ is only called as a last resort, when __getattribute__ has failed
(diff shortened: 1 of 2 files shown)


src/globus_sdk/scopes.py
```diff
@@ -22,6 +22,22 @@ def __init__(
...
 
+    # custom __getattr__ instructs `mypy` that unknown attributes of a ScopeBuilder are
+    # of type `str`, allowing for dynamic attribute names
+    # to test, try creating a module with
+    #
+    #       from globus_sdk.scopes import TransferScopes
+    #       x = TransferScopes.all
+    #
+    # the assignment to `x` will fail type checking without this because `all` will be
+    # unknown to `mypy`
```
rudyardrichter, line 33: A small nitpick/wording suggestion that I think is more clear:
```suggestion
    # without this method, the assignment to `x` would fail type checking
    # because `all` is unknown to mypy
```

sirosen: Thanks; I agree that the reword is better. Committed!

rudyardrichter (approved):
Good explanation

PR #473, opened 2021-09-14
Fix remaining type ignore comments in src/ tree

sirosen:
Just to follow-up on #472, this "fixes" the remaining usages of `type: ignore` comments.

In the first commit, this is an actual fix: setting the bound on a type var correctly specifies the range of classes which can be passed in to `AuthClient.oauth2_token` as `response_class`. With the bound set, the ignore comment becomes unnecessary.

The other changes are less obviously an improvement. Decorated functions raise warnings about undefined attributes -- and rightly so -- when we decorate to add metadata to them. However, `mypy` is pacified if we first cast to a stub class which defines the desired attributes.
This has three benefits, though they are all marginal:
1. `git grep 'type: ignore' src/` will be empty
2. The assignment statements themselves don't need "annoying"/"ugly" ignore comments
3. We could, if we ever want, cast in other contexts once we test for the presence of these attributes, to have nicer type-checked code at usage sites

(Note: to fully follow through on that last one, we'd probably need to make the stubs callable.)

### DIFF
--- a/src/globus_sdk/paging/table.py
+++ b/src/globus_sdk/paging/table.py
@@ -1,6 +1,14 @@
-from typing import Any, Callable, Dict, Optional, Type
+from typing import Any, Callable, Dict, Optional, Type, cast
 
 from .base import Paginator
+
+
+# stub for mypy
+class _PaginatedFunc:
+    _has_paginator: bool
+    _paginator_class: Type[Paginator]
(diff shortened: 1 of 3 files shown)


rudyardrichter (approved):
Looks cool. As a matter of fact, having the `_PaginatedFunc` interface defined seems kind of nice and good for documentation.

PR #482, opened 2021-09-29
Update the minimum requests and crypto versions

### LINKED ISSUE #481: Raise the lower bound on `cryptography` in the requirements data
While double-checking the CLI v3.0 release, I saw an error after upgrading a very old install:
```
ImportError: cannot import name 'RSAPublicKey' from 'cryptography.hazmat.backends.openssl.rsa' 
```

from `services/auth/response.py`.

We should find a reasonable older version which we want to support, and set that as the lower bound.

It's not a major issue, as you can just reinstall to grab a fresh `cryptography` version, but it means the package data is not quite right.

sirosen:
This sets a lower bound on `cryptography` (missing), and adds a `py36-mindeps` env to tox. Run in this mode, we verify the SDK testsuite against the minimum dependency versions we specify, to ensure compatibility.

resolves #481

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -35,6 +35,10 @@ jobs:
         run: python -m tox -e lint
       - name: Run Tests
         run: python -m tox -e py -- --cov-report="term-missing:skip-covered"
+      - name: Mindeps Test
+        # mindeps runs on py36, as "the oldest everything"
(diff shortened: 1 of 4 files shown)


sirosen:
I've also updated the import of `RSAPublicKey` from `cryptography` to pull from `primitives` instead of `backends`, since that's where it's declared (and therefore it may be more stable).

Additionally, the version requirements now have `!=3.4.0` for `cryptography`, since that's know to have issues.

PR #486, opened 2021-10-13
Adjust cryptography version bounds to be more accurate

sirosen:


sirosen:
I'm actually going to update this to adjust the lower-bound as well. It turns out that `pyjwt` requires `cryptography>=3` and, by my reading of the situation, might actually need `>=3.1` because it instantiates rsa pubkeys without specifying a backend. (Which was a new usage supported in 3.1)

I also want to take another look at the `mindeps` build, to make sure it's doing the right thing.

PR #488, opened 2021-10-28
Add iter_items to transfer helpers

sirosen:
TransferData and DeleteData now support `iter_items()` as a method which iterates over the items which have been added to the inner "DATA" arrays.

These are just "nice helpers" for something which can already be done today. However, it's non-obvious to users that once data has been put into a TransferData or DeleteData, it can be read back out. Making it part of the public API contract for these objects makes it clear that this is a supported usage.

The target use-case is submitting a Transfer or Delete task, and then enumerating the files/directories which were transferred or deleted.

---

I also corrected a minor error in a docstring.

NB: this was inspired by a support ticket.


aaschaer (approved):
Looks good to me

PR #489, opened 2021-11-05
Add resource server property to BaseClient

sirosen:
The end-goal here was to add `BaseClient.resource_server` as an alias for `BaseClient.scopes.resource_server`. It turned into something of a boondoggle, but it works.

In python3.9+, we can write the property as follows:
```python
class BaseClient:
    @classmethod
    @property
    def resource_server(cls) -> Optional[str]:
        return cls.scopes.resource_server if cls.scopes is not None else None
```
🎉 Done!

Everything beyond that in this PR is about handing 3.6-3.8.

Because `classmethod` cannot wrap another descriptor prior to 3.9, my first approach was to write a custom descriptor which acts like a class property, carefully type annotate it, test it, and then drop it in place of `classmethod(property(...))`. That, unfortunately, does not work cleanly. The major issue is sphinx support, wherein there are a number of uses of `inspect` and checks which look specifically for `classmethod` wrapping `property`. The commits after the first in this PR show the path I took towards a working solution with minimal (?) crazy stuff.

So the second commit here backtracks some of that work and converts it to a descriptor that works as a backport of the `classmethod` behavior from 3.9 into 3.6-3.8, called `chainable_classmethod`. We use it on those pythons, and on 3.9+ just set `chainable_classmethod = classmethod`.
This makes sphinx happy (because it sees `classmethod` properly), and a little trickery with `typing.TYPE_CHECKING` lets us always show `classmethod` to `mypy`.
But there's a lot of typing weirdness with this, and `mypy`'s support for class properties is a little weak in any case. Instance-based access to the property is correctly typed, but class-based access is not. So that needed sorting out.

In the final draft, `utils.classproperty` is a decorator, and its behavior is tied to whether or not the invoking executable was named `sphinx-build` (sphinx doesn't provide a flag which I could find documented for identifying the sphinx build context). In `sphinx-build`, `classproperty = classmethod(property(...))`, which is what sphinx wants. But for `mypy` and runtime logic, it's just a decorator which returns a descriptor with the simplest possible definition for a class property.

Tests cover the correctness of the property behavior, and we can check a `tox -e docs` run to see the resulting docs. But where this gets particularly interesting for long-term usage is the inclusion of `reveal_type()` tests which use the mypy API on generated python files. These are using some new fixtures I put together to let us assert that `mypy` will deduce the types of the class attribute and instance attribute named `resource_server` as `Union[builtins.str, None]`.


aaschaer (approved):
I'm not sure I completely understand how `_classproperty` is doing what its doing, but I understand what it is doing and why, and the tests confirm all is working as expected, so I think this is good to go

sirosen:
Let me know if you want to chat about descriptors. It's a part of the language which lets you define your own objects like `Property`, where get and set are redefined. For most cases, I think they're confusing, vs redefining getattr or setattr. This is a (very) rare case where they do a good job.

Issue #490, opened 2021-11-05
Paginated methods lose typing information for type-checking


sirosen:
```python
import globus_sdk

reveal_type(globus_sdk.TransferClient().paginated.task_event_list)
```
shows
`Revealed type is "def (*Any, **Any) -> Any"`

This is because the type found for attributes of `paginated` comes from `__getattr__` on the paginator table, which is annotated as `-> Callable` (because the precise type depends on the name being used).

mypy issues details why this is complicated, e.g. https://github.com/python/mypy/issues/5523
I don't think we can expect `__getattr__` to solve this correctly for us.

I tried some trivial hacks, but nothing is easily working for this case.
A paginator table is a partial proxy for a client, so there's always the possibility of casting it to the relevant client class. Although odd, it would get you the right signature on the resulting function.

sirosen:
Just to record it, I did try putting together a mypy plugin to see if it's possible to get paginator types with a plugin.
I'd like it -- even if it's experimental -- for use in the CLI. The following plugin code _doesn't_ work because I wasn't able to solve getting the attribute name from the AttributeContext:
```python
"""
Experimental: a mypy plugin for special types within the Globus SDK
"""

import typing as t

from mypy.plugin import AttributeContext, Plugin
from mypy.types import Type, get_proper_type


def _paginated_method_callback(ctx: AttributeContext) -> Type:
    # do not modify the type if the object type is not generic
    if len(ctx.type.args) != 1:
        return ctx.default_attr_type
    # if there is no name for this node, do not change the type
    if ctx.type.fullanme is None:
        return ctx.default_attr_type

    client_class_type = get_proper_type(ctx.type.args[0])
    attribute_name = ...  # UNKNOWN: how do we get the attribute name in this context?

    # TODO: this returns the underlying client method type; should transform to a
    # paginator as the result type
    return get_proper_type(getattr(client_class_type, attribute_name))


class GlobusSDKPlugin(Plugin):
    def get_attribute_hook(
        self, fullname: str
    ) -> t.Optional[t.Callable[[AttributeContext], Type]]:
        if fullname.startswith("globus_sdk.paging.table.PaginatorTable"):
            return _paginated_method_callback
        return None


def plugin(version: str) -> t.Type[Plugin]:
    return GlobusSDKPlugin
```

sirosen:
Closing as resolved by #494

PR #491, opened 2021-11-05
Fix preservation of type info via has_paginator

sirosen:
This is a minor bugfix, mostly changing `Callable[[Callable], Callable]` into `Callable[[C], C]` where C is a TypeVar.
However, it also provided a good opportunity to pull in the `run_reveal_type` testing utility I wrote for #489 in a separate PR.

`has_paginator` was meant to preserve type information, but it was not doing so properly.

Because the inner decorator was correctly annotated but the outer layer was not, all functions were being converted to generic Callable types when decorated with this.

Additionally, although some cases for preservation of type information in PaginatorTable are difficult (see GitHub #490), the simple case of runtime typing for paginated methods can be handled easily by using functools.wraps on the generated paginated methods.

It's possible in theory that this is inaccurate in some boundary cases, but it's definitely more accurate than the current type information of a generic callable accepting any arguments.

In order to test this change, a new mypy-testing utility is introduced, `run_reveal_type`, specifically for testing type annotations. It runs mypy's reveal_type() on a string input and reports back the result.

### DIFF
--- a/changelog.d/20211105_171713_sirosen_type_preserving_has_paginator.rst
+++ b/changelog.d/20211105_171713_sirosen_type_preserving_has_paginator.rst
@@ -1,0 +1,1 @@
+* Fix type annotations on client methods with paginated variants (:pr:`NUMBER`)
--- a/src/globus_sdk/paging/table.py
+++ b/src/globus_sdk/paging/table.py
@@ -1,3 +1,4 @@
+import functools
 from typing import Any, Callable, Dict, Optional, Type, TypeVar, cast
 
 from .base import Paginator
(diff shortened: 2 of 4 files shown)


rudyardrichter (approved):
The `run_reveal_type` fixture is a great trick. This test is a good example to have on hand; I would say bonus points for a `tests/unit/type_annotations/README` or something, or perhaps the rest of us could use reminders for a bit if we have features or PRs which should be making use of this.

sirosen:
Let me think more about docs for that fixture. I think maybe we should split some things up into a dir of docs, so if I wrote a testing guide it could go there. The contrib doc can easily get too long (a previous version of it was so long that it was almost unusable), so it doesn't necessarily fit in there.

sirosen:
I've merged for now, but like I said, I'll be thinking about more internal docs.

PR #494, opened 2021-11-15
Add a method for getting a Paginator which preserves argument types

sirosen:
In parallel with exploration of a mypy plugin for paginators, we discussed last week the possibility of a Callable interface which could preserve the type information about a paginated method. Because it doesn't rely on `__getattr__` proxying to do the lookup, it can be type-preserving without a plugin.

---

This adds a first draft of 'typed_paginator' as a client method which takes other client methods as inputs.

Unfortunately, because mypy v0.910 does not yet implement support for ParamSpec, this code is not yet usable. `mypy` will error on ParamSpec usage, indicating that it believes that the parameter specification for a Callable must be a list or ellipsis. The implementation appears correct (with the caveat that it uses `assert`s and is minimally tested and documented) using mypy master, v0.920, but this means it is not yet ready for general use.

In order to make things work out better, this makes the paginators generics over their "page type". This allows a paginator over a method returning `IterableTransferResponse` to indicate that it has `pages() -> Iterable[IterableTransferResponse]` correctly.
This change could be separated and merged prior to the rest of this work.

---

When `mypy` releases general support for `ParamSpec`, this can be cleaned up with the following:
- figure out where we need `typing_extensions` (and potentially always install it for simplicity)
- replace those `assert`s with usage errors
- add some `reveal_type` tests
- add docs which instruct users on usage `paginator = tc.typed_paginator(tc.endpoint_search)("foo")`, noting that this is only necessary if you want `mypy` to provide full type information

### DIFF
--- a/src/globus_sdk/paging/base.py
+++ b/src/globus_sdk/paging/base.py
@@ -1,13 +1,16 @@
 import abc
+import functools
 import inspect
 from typing import (
     Any,
     Callable,
     Dict,
+    Generic,
     Iterable,
     Iterator,
     List,
     Optional,
+    Type,
     TypeVar,
     cast,
 )
@@ -19,6 +22,15 @@ from globus_sdk.response import GlobusHTTPResponse
 PageT = TypeVar("PageT", bound=GlobusHTTPResponse)
 P = te.ParamSpec("P")
 R = TypeVar("R", bound=GlobusHTTPResponse)
+C = TypeVar("C", bound=Callable[..., GlobusHTTPResponse])
+
+
+# stub for mypy
+class _PaginatedFunc(Generic[PageT]):
+    _has_paginator: bool
+    _paginator_class: Type["Paginator[PageT]"]
+    _paginator_items_key: Optional[str]
+    _paginator_params: Dict[str, Any]
 
 
 class Paginator(Iterable[PageT], metaclass=abc.ABCMeta):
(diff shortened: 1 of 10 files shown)


sirosen:
I'm going to try to pull out the changes which make the `Paginator` class generic now, into a separate PR. Using a `TypeVar` with a bound of `GlobusHTTPResponse` means that we'll preserve the ability of the paginator pages to be accessed with the attributes of a base response.

Once that's merged, I can rebase this (and add the necessary newsfile).

sirosen:
In response to the feedback I've gotten talking to several people, I've moved the implementation to `Paginator.wrap`.
I eventually settled on `wrap` vs other names because it's evocative of `functools.wraps` and other decorator and function wrapper tools.

This led to an interesting realization: rather than `Paginator.wrap` referring to a `PaginatorTable`, `Paginator.wrap` can be the canonical implementation of the function -> paginated function transformation.
So instead of
`Paginator.wrap -> PaginatorTable -> "add binding" helper`
the source for that "add binding" helper now _is_ `Paginator.wrap` and we have
`PaginatorTable -> Paginator.wrap`
a much clearer and more obvious implementation.

This PR now includes a couple of tests to make sure the new interface works at runtime as expected, and I've made the appropriate documentation updates.

src/globus_sdk/paging/base.py
```diff
@@ -1,9 +1,36 @@
...
+    Iterable,
+    Iterator,
+    List,
+    Optional,
+    Type,
+    TypeVar,
+    cast,
+)
+
+import typing_extensions as te
```
sirosen, line 18: This is the only thing which needs attention still, as far as I know. The PR is basically ready, but we should only import `ParamSpec` from typing-extensions on sufficiently old python versions.

PR #495, opened 2021-11-15
Make Paginator a Generic type over PageT

sirosen:
This work was extracted from #494. I don't think it makes a lot of sense purely on its own because there's no clear motivation. #494 (or something like it) _is_ the motivator. We will want to be able to indicate that a paginated call to Endpoint Search returns a paginator _of `IterableTransferResponse`_ objects. To describe that kind of container-relationship in the types, a Generic is most appropriate.

---

This makes Paginator a generic over PageT, a type var with the type bound of GlobusHTTPResponse. This means that hte iterable elements of a paginator are still expected to be response objects, and they will still type-check as having the necessary methods and attributes of a response. However, this also makes it possible for a paginator to be constructed with PageT bound to a specific value, e.g. `IterableTransferResponse`.

This would mean that the pages of the paginator have a more specific type, and can be expected to respond to those methods.

Under the current implementation of pagination, a lot of type information is being lost. However, if a more type-specific pagination interface is added (PR #494), then the type-preserving paginator will be able to indicate the page type which it is producing.

Because `mypy --strict` disallows implicit `Any` on generics, several locations need to be updated with an explicit type variable for the paginators, usually Any.
One minor change this raised is that it allowed us to turn off a `type: ignore` in the paginator table. The result, however, is that we
must add `**kwargs: Any` to the base paginator class (because subclasses may implement additional keyword args) in order for that method to type-check.

### DIFF
--- a/src/globus_sdk/paging/table.py
+++ b/src/globus_sdk/paging/table.py
@@ -9,13 +9,13 @@ C = TypeVar("C", bound=Callable[..., Any])
 # stub for mypy
 class _PaginatedFunc:
     _has_paginator: bool
-    _paginator_class: Type[Paginator]
+    _paginator_class: Type[Paginator[Any]]
     _paginator_items_key: Optional[str]
     _paginator_params: Dict[str, Any]
 
 
 def has_paginator(
-    paginator_class: Type[Paginator],
+    paginator_class: Type[Paginator[Any]],
     items_key: Optional[str] = None,
     **paginator_params: Any,
 ) -> Callable[[C], C]:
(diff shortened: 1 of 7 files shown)


src/globus_sdk/paging/table.py
```diff
@@ -9,13 +9,13 @@
 # stub for mypy
 class _PaginatedFunc:
     _has_paginator: bool
-    _paginator_class: Type[Paginator]
+    _paginator_class: Type[Paginator[Any]]
```
rudyardrichter, line 12: Could these be bound to some `Type[Paginator[PaginatedT]]` where `PaginatedT` is a `PageT` iterator or something like that?

sirosen: I've put the test case we discovered for why `Paginator[Any]` is bad into the commit message for the fix.

rudyardrichter (approved):

PR #496, opened 2021-11-17
Publish `get_service_url`, add `get_webapp_url`

rudyardrichter:
This updates `get_service_url` to default to the environment settings to find the environment for a service, lets `get_service_url` return a sensible URL for the webapp and adds `get_webapp_url` for specifically that purpose.

Before:
```python
>>> from globus_sdk.config import get_service_url
>>> get_service_url("test", "app")
'https://app.api.test.globuscs.info/'
```

After:
```python
>>> from globus_sdk.config import get_service_url, get_webapp_url
>>> get_service_url("app", environment="test")
'https://app.test.globuscs.info/'
```
Or just `export GLOBUS_SDK_ENVIRONMENT=test` and then:
```python
>>> from globus_sdk.config import get_webapp_url
>>> get_webapp_url()
'https://app.test.globuscs.info/'
```

### DIFF
--- a/changelog.d/20211117_174041_rudyardrichter_config_get_app_url.rst
+++ b/changelog.d/20211117_174041_rudyardrichter_config_get_app_url.rst
@@ -1,0 +1,10 @@
+..
+.. A new scriv changelog fragment
+..
+.. Add one or more items to the list below describing the change in clear, concise terms.
+..
+.. Leave the ":pr:`...`" text alone. When you open a pull request, GitHub Actions will
+.. automatically replace it when the PR is merged.
+..
+
+* Tweak `get_service_url` to be able to return correct webapp URLs (:pr:`NUMBER`)
--- a/changelog.d/20211117_174041_rudyardrichter_config_get_app_url.rst
+++ b/changelog.d/20211117_174041_rudyardrichter_config_get_app_url.rst
@@ -7,4 +7,7 @@
 .. automatically replace it when the PR is merged.
 ..
 
-* Tweak `get_service_url` to be able to return correct webapp URLs (:pr:`NUMBER`)
+* Document ``globus_sdk.config.get_service_url`` and ``globus_sdk.config.get_webapp_url``
+  (:pr:`NUMBER`)
+    * Internally, these are updated to be able to default to the ``GLOBUS_SDK_ENVIRONMENT`` setting,
+      so specifying an environment is no longer required
--- a/src/globus_sdk/config/environments.py
+++ b/src/globus_sdk/config/environments.py
@@ -57,10 +57,10 @@ def get_service_url(service: str, environment: Optional[str] = None) -> str:
     Return the base URL for the given service in this environment. For example:
 
     >>> from globus_sdk.config import get_service_url
-    >>> get_service_url("auth", environment="sandbox")
-    'https://auth.sandbox.globuscs.info/'
-    >>> get_service_url("search", environment="test")
-    'https://search.api.test.globuscs.info/'
+    >>> get_service_url("auth", environment="preview")
+    'https://auth.preview.globus.org/'
+    >>> get_service_url("search", environment="production")
+    'https://search.api.globus.org/'
 
     If no ``environment`` is specified, this will use the ``GLOBUS_SDK_ENVIRONMENT``
     environment variable.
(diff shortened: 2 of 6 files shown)


sirosen:
Let's discuss this at our meeting in a few minutes.

The change itself is good -- I don't see any issue with merging it -- but it makes me wonder about a few things. Here's what I want to discuss:

- What is the context in which this is being used? Is the usage an indicator of a broader need?
- Do we need to document some methods of `globus_sdk.config` for public consumption? (right now they are intentionally _not_ documented, meaning they are not covered under our semver promise)
- "app" isn't a very clear/obvious name -- the hostname in this case does not match what we typically call the service, colloquially or in documentation. Should there be a wrapper, `get_webapp_url = get_service_url(<env>, "app")`?

changelog.d/20211117_174041_rudyardrichter_config_get_app_url.rst
```diff
@@ -0,0 +1,10 @@
+..
+.. A new scriv changelog fragment
+..
+.. Add one or more items to the list below describing the change in clear, concise terms.
+..
+.. Leave the ":pr:`...`" text alone. When you open a pull request, GitHub Actions will
+.. automatically replace it when the PR is merged.
+..
+
+* Tweak `get_service_url` to be able to return correct webapp URLs (:pr:`NUMBER`)
```
sirosen, line 10:
```suggestion
* Tweak ``get_service_url`` to be able to return correct webapp URLs (:pr:`NUMBER`)
```

The RST use of double-backticks is a frequent source of annoyance for me -- I make this mistake constantly, even after many years of working with sphinx.

sirosen:
We concluded:
- we will add public doc (maybe on the config page of the docs) for `get_service_url`
- we will add (and publicly doc) `get_webapp_url` as a wrapper/alias
- docs should also note that no other importable names from `globus_sdk.config` are considered public

rudyardrichter:
@sirosen looking at this again, should we also change `get_service_url`  from
```python
def get_service_url(environment: str, service: str) -> str:
```
to something like
```python
from globus_sdk.config.env_vars import get_environment_name

def get_service_url(service: str, environment: Optional[str] = None) -> str:
    environment = environment or get_environment_name()
```

It seems like that would make more sense, for callers to be able to just say e.g. `get_webapp_url()` and trust the environment settings to work.

sirosen:
Yes, that seems like a better interface. Please go ahead and make the change.

rudyardrichter:
@sirosen updated the PR, description, changelog. etc—should be good for a re-review

rudyardrichter:
cc-ing @kurtmckee @jakeglobus since this is to support [a fix on the flows service](https://app.shortcut.com/globus/story/11728/flows-on-test-activation-url-returned-when-a-run-requires-v4-endpoint-to-be-activated-points-to-production)

src/globus_sdk/config/environments.py
```diff
@@ -50,8 +52,21 @@ def get_by_name(cls, envname: str) -> Optional[Type["EnvConfig"]]:
...
-def get_service_url(environment: str, service: str) -> str:
+def get_service_url(service: str, environment: Optional[str] = None) -> str:
+    """
+    Return the base URL for the given service in this environment. For example:
+
+    >>> from globus_sdk.config import get_service_url
+    >>> get_service_url("auth", environment="sandbox")
+    'https://auth.sandbox.globuscs.info/'
+    >>> get_service_url("search", environment="test")
+    'https://search.api.test.globuscs.info/'
```
sirosen, line 63:
```suggestion
    >>> get_service_url("auth", environment="preview")
    'https://auth.preview.globus.org/'
    >>> get_service_url("search", environment="production")
    'https://search.api.globus.org/'
```

Because only `preview` and `production` are "public" (exactly what this means is subject to debate) and this is going into docs, I think we should avoid using other environments here.

src/globus_sdk/config/environments.py
```diff
@@ -68,6 +83,17 @@ def get_service_url(environment: str, service: str) -> str:
     return url
 
 
+def get_webapp_url(environment: Optional[str] = None) -> str:
+    """
+    Return the URL to access the Globus web app in the given environment. For example:
+
+    >>> get_webapp_url("sandbox")
+    'https://app.sandbox.globuscs.info/'
```
sirosen, line 91:
```suggestion
    >>> get_webapp_url("preview")
    'https://app.preview.globus.org/'
```

As above, using `preview` or `production` only.

sirosen (commented):
Before we merge, I'm suggesting that we switch the doc a little bit to refer to `production` and `preview` only. Other environments (e.g. `sandbox`) are internal-only, so less suitable for public documentation.

Otherwise this looks great!

src/globus_sdk/config/environments.py
rudyardrichter: Good point, thanks. I'm just going to apply these suggestions directly.

sirosen (approved):

PR #497, opened 2021-12-01
Expose `headers` from response objects

sirosen:
`GlobusHTTPResponse.headers` is just a pointer at the `requests.Response.headers` mapping.

To enable this, an internal-only property, `_raw_response` is added which traverses wrapped response objects until it reaches the original response object. We can then use that property to implement `headers` as a property which returns `self._raw_response.headers`. This pattern is also applicable to some existing data -- `http_status` and `content_type` -- so these are changed to be uniform.

This means that `http_status` and `content_type` are now read-only, which is a minor interface change. It is therefore noted explicitly in the changelog.

---

The added test uses http status code to demonstrate that the wrapping preserves properties as intended.

### DIFF
--- a/changelog.d/20211201_202405_sirosen_propogate_response_properties.rst
+++ b/changelog.d/20211201_202405_sirosen_propogate_response_properties.rst
@@ -1,0 +1,8 @@
+* The implementation of several properties of ``GlobusHTTPResponse`` has
+  changed (:pr:`NUMBER`)
+
+  * Responses have a new property, ``headers``, a case-insensitive
+    dict of headers from the response
+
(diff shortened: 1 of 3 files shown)


aaschaer (approved):
Code looks good to me, another test for the headers wouldn't hurt, but I don't think that needs to block this since the tests cover _raw_response's recursive nature which is the most complex piece of this.

sirosen:
> another test for the headers wouldn't hurt

I went ahead and added one, as it's nearly identical to the http-status test. I think it's a good idea, since that's the main purpose of this PR. Assuming the build passes, I'll go ahead and squash-merge.

PR #498, opened 2021-12-01
update type of scopes for ClientCredentialsAuthorizer

aaschaer:
Ran into this while linting client credentials work for the CLI.

The `ClientCredentialsAuthorizer` was being too strict in the allowed types for `scopes` since they are just being passed to `oauth2_client_credentials_tokens` which accepts `Optional[Union[str, Iterable[str]]]`

### DIFF
--- a/changelog.d/20211201_145304_aaschaer_scopes_type.rst
+++ b/changelog.d/20211201_145304_aaschaer_scopes_type.rst
@@ -1,0 +1,1 @@
+* ClientCredentialsAuthorizer now accepts Union[str, Iterable[str]] as the type for scopes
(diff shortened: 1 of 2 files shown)


changelog.d/20211201_145304_aaschaer_scopes_type.rst
```diff
@@ -0,0 +1 @@
+* ClientCredentialsAuthorizer now accepts Union[str, Iterable[str]] as the type for scopes
```
sirosen, line 1:
```suggestion
* ClientCredentialsAuthorizer now accepts Union[str, Iterable[str]] as the type for scopes (:pr:`NUMBER`)
```

The `:pr:` role should be included here. When it's merged, the github actions workflow will populate it. (Technically, now that the PR is open, we could just write `498` instead of `NUMBER`. But the norm should be, I think, to let github actions do the rewrite.)

aaschaer: Ah thanks!

sirosen:
This looks very much okay / reasonable to me.

I do want the `:pr:` link in the changelog. Either as suggested or just amended to
```
:pr:`498`
```
I also just noticed that it might be a good idea to double-backtick the union type in the changelog, as
```
``Union[str, Iterable[str]]``
```

With the changelog entry minimally edited, I'm happy to merge. We can get a release out shortly.

PR #499, opened 2021-12-10
Fix to be py3.10 compatible

sirosen:
Replace `distutils.util.strtobool` with a dedicated function so that we don't get deprecation warnings on py310.

Add py3.10 to the build matrix (tox and github actions).

Update RTD to use py3.10

Add 3.10 to classifiers.

@kurtmckee, as the original reporter of this issue, would you be willing to take a quick look at this?
I'm inclined to do a release with it as soon as it merges since it will cause downstream failures for anyone trying to use `filterwarnings=error` in a py3.10 build.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -3,7 +3,7 @@ sphinx:
   configuration: docs/conf.py
 formats: all
 python:
-  version: 3.8
+  version: "3.10"
   install:
     - method: pip
       path: .
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ["3.9"]
+        python-version: ["3.10"]
         include:
           - os: ubuntu-latest
             python-version: 3.6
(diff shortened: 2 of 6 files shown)


.readthedocs.yml
```diff
@@ -3,7 +3,7 @@ sphinx:
   configuration: docs/conf.py
 formats: all
 python:
-  version: 3.8
+  version: "3.10"
```
kurtmckee, line 6: SUPER clutch catch. I'd hate for that to accidentally normalize as a floating point 3.1 before being interpreted as a version!

kurtmckee (approved):
Code-wise this looks good to me. I also checked out your branch and grep'd for `"distutils"` and found nothing else in `src/` or `tests/` so it looks like this should resolve that warning.

Thanks for the quick turnaround!

.readthedocs.yml
sirosen: Yeah, it was a close call, actually! I messed up and amended to fix.

There's a [JSON Schema for readthedocs](https://raw.githubusercontent.com/readthedocs/readthedocs.org/6.3.0/readthedocs/rtd_tests/fixtures/spec/v2/schema.json), so I'll be looking into pulling this in and checking it via `check-jsonschema`.

PR #500, opened 2021-12-20
Allow IdentityMap to consume a cache on init

sirosen:
This allows multiple IdentityMap instances to share a cache by allowing users to pass a cache to the IdentityMap.

Naive usage still looks identical, with an automatic dict wrapped inside of the map object.

Add more tests to explore this behavior.

---

This isn't really relevant to any work we have in progress. It's an old branch I've had kicking around for a long time, to improve the behavior of IdentityMap. I figured I might as well clean it up and get it merged.

### DIFF
--- a/src/globus_sdk/services/auth/identity_map.py
+++ b/src/globus_sdk/services/auth/identity_map.py
@@ -1,5 +1,5 @@
 import uuid
-from typing import Any, Dict, Iterable, Optional, Set, Tuple
+from typing import Any, Dict, Iterable, MutableMapping, Optional, Set, Tuple
 
 from .client import AuthClient
 
@@ -120,6 +120,10 @@ class IdentityMap:
     :param id_batch_size: A non-default batch size to use when communicating with Globus
         Auth. Leaving this set to the default is strongly recommended.
     :type id_batch_size: int, optional
+    :param cache:  A dict or other mapping object which will be used to cache results.
+        The default is that results are cached once per IdentityMap object. If you want
+        multiple IdentityMaps to share data, explicitly pass the same ``cache`` to both.
+    :type cache: MutableMapping, optional
 
     .. automethodlist:: globus_sdk.IdentityMap
         include_methods=__getitem__,__delitem__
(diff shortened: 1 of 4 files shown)


tests/functional/auth/test_identity_map.py
```diff
@@ -167,3 +167,63 @@ def test_identity_map_del(client):
...
+            "46bd0f56-e24f-11e5-a510-131bef46955c",
+            "ae341a98-d274-11e5-b888-dbae3a8ba545",
+        ),
+    ],
+)
+def test_identity_map_shared_cache_mismatch(client, lookup1, lookup2):
+    register_api_route("auth", "/v2/api/identities", json=IDENTITIES_MULTIPLE_RESPONSE)
+    cache = {}
+    idmap1 = globus_sdk.IdentityMap(client, [lookup1], cache=cache)
+    idmap2 = globus_sdk.IdentityMap(client, [lookup2], cache=cache)
```
sirosen, line 217:
```suggestion
    idmap1 = globus_sdk.IdentityMap(client, [lookup1, lookup2], cache=cache)
    idmap2 = globus_sdk.IdentityMap(client, cache=cache)
```

src/globus_sdk/services/auth/identity_map.py
```diff
@@ -219,6 +225,15 @@ def __getitem__(self, key: str) -> Any:
...
             self._fetch_batch_including(key)
+        # if the key was in the cache, double-check that it's not in the unresolved
+        # usernames or IDs for the IdentityMap
+        # otherwise, when a cache is being shared between IdentityMap instances, it's
+        # possible to call out to Auth unnecessarily
+        else:
+            if key in self.unresolved_ids:
+                self.unresolved_ids.remove(key)
+            elif key in self.unresolved_usernames:
+                self.unresolved_usernames.remove(key)
```
sirosen, line 236: I want to circle back and check that this is the right place to do this bookkeeping. It seems like we should be checking for _any_ values being in the cache already when we are doing a lookup against Auth; not just on `__getitem__`.

PR #501, opened 2022-01-06
Update dev dependencies, including mypy and docs tools

sirosen:
Most things are current. I think I recently updated things.

Update to mypy==0.930 and handle new warnings (unused ignore, unnecessary casts).

This also switches from `pytest-cov` to `coverage`, which is just to simplify the number of tools on which we need to check for updates. Even though `coverage` is a second-order dependency today, it makes a lot of sense to get rid of `pytest-cov` and just cut out the middleman.


sirosen:
I'm having local testing annoyances because `mypy` is failing with the errors resolved here, so I've gone ahead and merged.

PR #502, opened 2022-01-10
Add Transfer notification options to helpers

sirosen:
Add notify_on_{succeeded,failed,inactive} to keyword opts, with defaults of `True` (equivalent to the API default).

Also fix some minor doc issues in nearby code:
- typo in param type doc for `additional_fields` on TransferData
- missing `additional_fields` from param docs for a TransferData method

Begin running `dead` on `transfer_data.py` and `delete_data.py`, which requires the addition of several `dead: disable` comments.

---

As an aside, I think I would like to make these objects dataclasses when py3.7 is our lowest supported python. We could, at least in theory, have an autogenerated `__init__` method. We'd have to explore it a bit.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,7 +30,14 @@ repos:
       #     superclasses or risk silently swallowing/ignoring arguments)
       # (3) we are prepared to commit to keeping `dead` happy and using
       #     `# dead: disable` comments as needed
-      args: ["--files", "src/globus_sdk/services/gcs/data.py"]
+      args:
+        - "--files"
+        - |
+          (?x)^(
+            src/globus_sdk/services/gcs/data.py|
+            src/globus_sdk/services/transfer/data/delete_data.py|
+            src/globus_sdk/services/transfer/data/transfer_data.py
+          )
 - repo: https://gitlab.com/pycqa/flake8
   rev: 3.9.2
   hooks:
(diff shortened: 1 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -30,7 +30,14 @@ repos:
       #     superclasses or risk silently swallowing/ignoring arguments)
       # (3) we are prepared to commit to keeping `dead` happy and using
       #     `# dead: disable` comments as needed
-      args: ["--files", "src/globus_sdk/services/gcs/data.py"]
+      args:
```
rudyardrichter, line 33: This is my first time noticing the `dead` hook here. What's the reason to add these but then mostly add ignores to the changes here? Was that sort of a separate issue that overlapped prominently with this PR?

sirosen: I'm not fully committed to using `dead` long term. The hook is something of an experiment.
(I also looked at `vulture`, which seems to me to have more false positives.)

`dead` very aggressively flags unused variables, including unused function parameters. I think it can help ensure that these long parameter lists for `__init__` are all properly used. For example, it would be easy to make the mistake where one of these is not assigned:
```python
def __init__(..., notify_on_succeeded=True, notify_on_failed=True, notify_on_inactive=True):
    ...
    self["notify_on_succeeded"] = notify_on_succeeded
    self["notify_on_inactive"] = notify_on_inactive
```

Unfortunately, it also flags unused methods. There's nothing elegant about those ignores, but it gets the hook to pass.

I've also been experimenting a bit with `pylint` to catch things like this, and perhaps other stuff like undocumented parameters. It's all up-in-the-air as far as I'm concerned, but I'd like some kind of check on this if possible.

PR #506, opened 2022-01-24
add update_group method to groups client

aaschaer:
Noticed this was missing while working on CLI support


sirosen (approved):
I see nothing that would cause us to hold this back. Let's get it into the next release.

PR #507, opened 2022-01-26
Add paginated `post_search` and `scroll` to SearchClient, plus role methods

sirosen:
The SearchClient gains three new methods here: `paginated.post_search`, `scroll`, and `paginated.scroll`

This also begins the long road towards having more complete test coverage of the SearchClient, its helpers, and methods.
These APIs are slightly different from some of the existing paginated ones, in that their pagination parameters are body params, not query params. However, because the paginator objects "just" pass their relevant arguments through to the client methods (`marker`, `limit`, etc), it is possible to support these only by ensuring that `marker`, `limit`, and `offset` are supported as keyword args.
Some care is taken not to mutate the data which was passed in to these methods, but otherwise the implementation does not require special efforts.

In order to make `scroll` more usable, an additional helper object is added, `SearchScrollQuery`, for this specialized query type.
The newly fleshed out documentation on this and `SearchQuery` should clarify the difference.

### DIFF
--- a/tests/functional/search/test_search.py
+++ b/tests/functional/search/test_search.py
@@ -3,6 +3,7 @@ import urllib.parse
 import uuid
 
 import pytest
+import responses
 
 import globus_sdk
 from tests.common import get_last_request, register_api_route_fixture_file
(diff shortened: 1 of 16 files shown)


sirosen:
I've added the methods which would be needed for https://github.com/globus/globus-cli/pull/597 to this PR, and updated the title. It was easier to add them here than to deal with the (inevitable) merge conflicts that would arise from adding them separately.

The methods are extremely simple and easy to test.

tests/functional/search/test_search.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+@pytest.mark.parametrize(
+    "query_doc",
+    [
+        {"q": "foo"},
+        globus_sdk.SearchScrollQuery("foo"),
+    ],
+)
+def test_search_paginated_scroll_qeury(search_client, query_doc):
```
sirosen, line 110:
```suggestion
def test_search_paginated_scroll_query(search_client, query_doc):
```

PR #510, opened 2022-01-27
Add `mapped_collection_id` and `filter` query args on `GCSClient.get_collection_list`

rudyardrichter:
- Add said qargs as keyword arguments for said method.
- Add simple test case that these are passed through as expected.

### DIFF
--- a/tests/functional/gcs/test_collections.py
+++ b/tests/functional/gcs/test_collections.py
@@ -40,6 +40,16 @@ def test_get_collection_list_include_param(client):
     assert "include" in req.params
     assert req.params["include"] == "foo,bar"
 
+    client.get_collection_list(mapped_collection_id="MAPPED_COLLECTION")
+    assert get_last_request().params.get("mapped_collection_id") == "MAPPED_COLLECTION"
+
+    filters = ["mapped_collections", "created_by_me"]
+    client.get_collection_list(filter=filters)
+    assert get_last_request().params.get("filter") == ",".join(filters)
+
+    client.get_collection_list(filter="created_by_me")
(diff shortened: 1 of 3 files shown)


tests/functional/gcs/test_collections.py
```diff
@@ -40,6 +40,13 @@ def test_get_collection_list_include_param(client):
     assert "include" in req.params
     assert req.params["include"] == "foo,bar"
 
+    client.get_collection_list(mapped_collection_id="MAPPED_COLLECTION")
+    assert get_last_request().params.get("mapped_collection_id") == "MAPPED_COLLECTION"
+
+    filters = ["mapped_collections", "created_by_me"]
+    client.get_collection_list(filter=filters)
+    assert get_last_request().params.get("filter") == ",".join(filters)
```
sirosen, line 48:
```suggestion
    assert get_last_request().params.get("filter") == ",".join(filters)
    
    client.get_collection_list(filter="created_by_me")
    assert get_last_request().params.get("filter") == "created_by_me"
```

Let's add one more for the string case, yeah?

sirosen (approved):
I'm (more than) happy to merge as-is, but it's so easy to add one more test case for 100% branch coverage that I figured I should suggest it.

tests/functional/gcs/test_collections.py
rudyardrichter: Oops, yep. Good suggestion, thanks!

sirosen: Sorry that my suggestion had trailing whitespace! I didn't even realize that GitHub had done that.

sirosen:
Thanks for handling the amend. I'm merging now.

PR #512, opened 2022-01-31
Add more explanation on BaseClient v3 upgrade

rudyardrichter:
Expand v3 upgrading docs to note that the `BaseClient` interface is different.

### DIFF
--- a/docs/upgrading.rst
+++ b/docs/upgrading.rst
@@ -97,11 +97,12 @@ Change:
     tc = globus_sdk.TransferClient(...)
     response = tc.get_endpoint(foo)  # again, just an example operation
 
-Import BaseClient from globus_sdk
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Updates to BaseClient Usage
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 You may be using the globus-sdk ``BaseClient`` object to implement a custom
-client or for type annotations.
+client or for type annotations. Firstly, ``BaseClient`` is available from the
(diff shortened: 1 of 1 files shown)


rudyardrichter:
The example isn't super compelling, but I think this is sufficient to get the point across?

Also, this should have a `no-news...` label which I forgot, but it doesn't look like I'm allowed to edit PR metadata...

sirosen:
> Also, this should have a `no-news...` label which I forgot, but it doesn't look like I'm allowed to edit PR metadata...

I'll double check that you're in the right teams.

Anyway, if you amend, it should build clean. But no need -- we know this is passing.

docs/upgrading.rst
```diff
@@ -101,7 +101,8 @@ Import BaseClient from globus_sdk
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
sirosen, line 101: We should update the title for this section. Maybe
> Updates to BaseClient Usage

?

sirosen (commented):
We need to adjust the section header now that it no longer matches.
I'm not too fussy about exactly what.

Or, alternatively, keep the title and add a second section heading. That might require a slight rephrase here or there, but seems equally good.

docs/upgrading.rst
rudyardrichter: Good suggestion, I'll use that.

sirosen:
Awesome! Thanks for putting this together!

Also, I made sure you have write access on this repo, so your permissions should be all set now.

PR #513, opened 2022-02-08
add gcs role methods

aaschaer:
Adds get_role_list, create_role, get_role and delete_role to the GCSClient

I mostly only added basic success tests, not sure if we want anything in addition to those.


tests/functional/gcs/fixture_data/role_list.json
```diff
@@ -0,0 +1,25 @@
...
+            "id": "{role_id_1}",
+            "collection": null,
+            "principal": "rn:globus:auth:identity:{user_id_1}",
+            "role": "owner"
+        },
+        {
+            "DATA_TYPE": "role#1.0.0",
+            "id": "{role_id_2}",
+            "collection": "{collection_id_1}",
+            "principal": "rn:globus:auth:group:{group_id_1}",
```
sirosen, line 16:
```suggestion
            "principal": "urn:globus:groups:id:{group_id_1}",
```

The format for a Group URN is `urn:globus:groups:id:`.
As above, our tests won't pass or fail on the basis of this, but it's good to be consistent with the real data format.

sirosen (changes requested):
The code in `src/` is good, and I have no objection to merging it.

And the test code itself is all good and checks out. However, there are a couple of typos in the test data which I'd like us to resolve now, before they can be lost in the repo. Otherwise, we might write new tests against these same data files in the future and be surprised by the results.

There's one optional change to the GCS payload classes which I suggested. If you don't think it's worth trying to do right now -- or you try it and run into trouble -- feel free to ignore it.

src/globus_sdk/services/gcs/data.py
```diff
@@ -441,3 +441,51 @@ def __init__(
...
+
+    def _set_value(
+        self, key: str, val: Any, callback: Optional[Callable[[Any], Any]] = None
+    ) -> None:
+        if val is not None:
+            self[key] = callback(val) if callback else val
+
+    def _set_optstrs(self, **kwargs: Any) -> None:
+        for k, v in kwargs.items():
+            self._set_value(k, v, callback=str)
```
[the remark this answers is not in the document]
aaschaer: I considered doing this, but it seemed not worth it atm. If we add any more documents with thus behavior though I think it would be a good idea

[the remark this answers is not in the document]
sirosen: Sounds good to me. I'll mark this resolved.

changelog.d/20220208_114626_aaschaer_gcs_roles.rst
```diff
@@ -0,0 +1,14 @@
...
+.. A new scriv changelog fragment
+..
+.. Add one or more items to the list below describing the change in clear, concise terms.
+..
+.. Leave the ":pr:`...`" text alone. When you open a pull request, GitHub Actions will
+.. automatically replace it when the PR is merged.
+..
+
+* Add role methods to ``GCSClient`` (:pr:`NUMBER`)
+    ** ``GCSClient.get_role_list`` lists endpoint or collection roles
```
rudyardrichter, line 11: The `**` syntax is asciidoc, right? I thought RST used indentation.

src/globus_sdk/services/gcs/client.py
```diff
@@ -212,3 +212,93 @@ def delete_collection(
...
+        :type include: str, optional
+        :param query_params: Additional passthrough query parameters
+        :type query_params: dict, optional
+        """
+        if query_params is None:
+            query_params = {}
+        if include is not None:
+            query_params = {**query_params, "include": include}
+        if collection_id is not None:
+            query_params = {**query_params, "collection_id": collection_id}
```
rudyardrichter, line 246:
```suggestion
        if include is not None:
            query_params["include"] = include
        if collection_id is not None:
            query_params["collection_id"] = collection_id
```

changelog.d/20220208_114626_aaschaer_gcs_roles.rst
sirosen: _Holds head in hands_. Which non-markdown markdown language were we using again? 😂 

I think you're right. Indent + a single asterisk should be correct.

sirosen (approved):
I'd like to squash-merge this one when we're done, to simplify the history.

I noticed one thing that I missed in my prior pass, so I left a comment. It is, however, extremely minor and optional to address.

src/globus_sdk/services/gcs/client.py
```diff
@@ -212,3 +212,93 @@ def delete_collection(
...
+        """
+        if query_params is None:
+            query_params = {}
+        if include is not None:
+            query_params["include"] = include
+        if collection_id is not None:
+            query_params["collection_id"] = collection_id
+
+        path = "/roles"
+        return IterableGCSResponse(self.get(path, query_params=query_params))
```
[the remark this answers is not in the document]
aaschaer: I think I prefer having the path as a separate variable to keep the REST call as simple as possible. For short paths like this one its not really an issue, but for longer paths I feel it makes it the REST call feel messy.
Not a strong preference one way or the other though.

PR #517, opened 2022-02-11
First draft of an internal _testing package

sirosen:
This is my first crack at putting this together and using it for some of the SDK's own tests.
The main idea is that a "RegisteredResponse" object describes a response. It primarily wraps the `responses.add` ~and `responses.replace`~ method, recording values which will later be passed to this method.
The additional `metadata` field is included, as it's been very useful in the CLI testsuite.

In order to organize, names, and provide access to these fixtures, define a `ResponseSet` class which thinly wraps a dict. A `ResponseSet` also has it's own `metadata` dict, so we have flexibility here.

There are two layers of "registry" involved. One is very simple and explicit:
- a dict which maps names to `ResponseSet`s
- a set of modules which can be loaded via importlib to provide `ResponseSet`s which are always named `RESPONSES`

This lets us use module and subpackage names to namespace and organize the fixture data.
The result is that we can create a response set in a module like `_testing/data/scenarios/all_500s.py`, and then load it later _by name_ with `load_response_set("scenarios.all_500s")`! ✨ 

I've also included some speculative tooling for registering response sets and loading them back up.
This will hopefully let us easily transition existing testsuites with fixtures like the globus-cli YAML data by formatting those data as appropriate and passing them to `register_fixture_set`. Those data could be ported into the `_testing/data/` space in the future.

To demonstrate the new tooling, set it up and begin using it for some of the simpler Auth tests.
~I also pulled in some of the CLI's fixtures, which I played with to see what kinds of semantics were nice. Check out `data/auth/unauthorized.py` and `data/transfer/unauthorized.py` to see these -- they aren't used, but demonstrate the kind of things we're able to support by using python as the definition language (rather than YAML, JSON, etc).~

---

EDITS:
- Drop `responses.replace` support. `replace` isn't as important as `add` and can always be done manually
- added resolution based on an object's `__qualname__` so that you can pass a client method and have that resolve to a ResponseSet, e.g. `load_response(AuthClient.get_identities)`
- cut some of the unused fixture data to keep review of this focused on the toolchain


rudyardrichter (commented):
I'm going to spend the rest of today installing what's here and trying to dogfood it against flows (is it still dogfooding if flows is the dog here??), and I'll see if I can come back with any suggestions based on that.

sirosen:
> is it still dogfooding if flows is the dog here??

I think if you have > 5 commits in both repos, it counts. 😉 

I'm looking forward to what you figure out from trying this!

src/globus_sdk/_testing/README.rst
```diff
@@ -0,0 +1,152 @@
...
+    from globus_sdk._testing import load_response_set, register_response_set
+    import uuid
+
+    # register a scenario under which Globus Auth get_identities and Globus
+    # Transfer operation_ls both return payloads of `{"foo": "bar"}`
+    # use an autogenerated endpoint ID and put it into the response metadata
+    # register_response_set takes dict data and converts it to fixtures
+    endpoint_id = str(uuid.uuid1())
+    register_response_set(
+        "foobar",
```
rudyardrichter, line 126: I think it could be clearer in the doc here that while `"foobar"` is your own made-up tag for this, the `get_identities` etc. have to actually map to methods.

On that note, I sort of think it would be clearer organized like:
```python
{
  "auth": {
    "get_identities": ...,
  }
  "transfer": {
    etc.
  }
}
```

rudyardrichter (commented):
OK, I have not so much to report yet (it turns out flows needs a *lot* of updates to the test suite), other than this works pretty great out-of-the-box. Very exciting.

src/globus_sdk/_testing/registry.py
```diff
@@ -0,0 +1,147 @@
...
+            headers = {"Content-Type": "application/json"}
+        self.headers = headers
+
+        self.metadata = metadata or {}
+        self.kwargs = kwargs
+
+    def add(self) -> "RegisteredResponse":
+        kwargs: Dict[str, Any] = {
+            "headers": self.headers,
+            "match_querystring": False,
```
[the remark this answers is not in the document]
sirosen: There's a bit of a snag here. `0.18.0` fixed this buggy deprecation warning, but it also dropped support for py3.6 .

So for the SDK and CLI, we can't update until we drop 3.6 (which I think we should do soon; I'll queue that up for discussion).

We could try to support `0.17.0` _and_ the later (fixed) versions, and add a test env which runs with the latest `responses`, to make sure it works on both. Otherwise, we need everything which uses this to pin to `responses==0.17.0` until we can drop py3.6 support.
I don't think there are other significant fixes and features in `responses` since `0.17.0`, so pinning appeals a bit for its simplicity.

src/globus_sdk/_testing/README.rst
sirosen: Let me see what I can do about the doc to clarify in the next couple of days.

---

I also prefer the "service name" rather than the bare `__qualname__`.
How can we get from `AuthClient.get_identities` to `"auth", "get_identities"` best? Unbound methods are just functions (since py3), so `__qualname__` might still be best.

What about mapping from `__qualname__` to attributes of the `globus_sdk` top-level module, like so?
```python
name = func.__qualname__
module_name = f"globus_sdk._testing.data.{name}"
if "." in name:
    prefix, suffix = name.split(".")
    if hasattr(globus_sdk, prefix):
        service_name = getattr(globus_sdk, prefix).service_name
        module_name = f"globus_sdk._testing.data.{service_name}.{suffix}"
```

aaschaer (approved):
I think I'm happy with this. I imagine some more things might come up with further usage, but that's why its experimental and internal for now

rudyardrichter (approved):

src/globus_sdk/_testing/README.rst
sirosen: I took this out into a dedicated function here: https://github.com/globus/globus-sdk-python/pull/517/commits/a1f083a440e2618fbd2dcf23845c4a2896674b19

As soon as CI passes, I plan to merge.

PR #520, opened 2022-02-18
Fix pagination behaviors for "next_marker" APIs

sirosen:
Transfer has several APIs which use a nullable `next_marker` field to control pagination. In order to support it, make the following changes:

- MarkerPaginator can be given a `marker_key` (defaults to `"marker"`)
- MarkerPaginator has a method used to determine if there is another page of results, defaulting to the previously hardcoded has_next_page check
- NullableMarkerPaginator(MarkerPaginator) is a new class which alters the has_next_page check to look for a null marker
- task_skipped_errors, task_successful_transfers, endpoint_manager_task_successful_transfers, and endpoint_manager_task_skipped_errors now all use a NullableMarkerPaginator and set `marker_key="next_marker"`

Add a test for task_skipped_errors specifically, using generated data, to confirm the behavior of the NullableMarkerPaginator with an altered marker_key.

### DIFF
--- a/src/globus_sdk/paging/marker.py
+++ b/src/globus_sdk/paging/marker.py
@@ -16,6 +16,7 @@ class MarkerPaginator(Paginator[PageT]):
         method: Callable[..., Any],
         *,
         items_key: Optional[str] = None,
+        marker_key: str = "marker",
         client_args: List[Any],
         client_kwargs: Dict[str, Any]
     ):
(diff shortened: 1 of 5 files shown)


src/globus_sdk/paging/marker.py
```diff
@@ -34,5 +39,18 @@ def pages(self) -> Iterator[PageT]:
...
             yield current_page
-            self.marker = current_page.get("marker")
-            has_next_page = current_page.get("has_next_page", False)
+            self.marker = current_page.get(self.marker_key)
+            has_next_page = self._check_has_next_page(current_page)
+
+
+class NullableMarkerPaginator(MarkerPaginator[PageT]):
+    """
+    A paginator which uses `has_next_page` and `marker` from payloads, sets the `marker`
```
aaschaer, line 48: Help text needs a fix here since this class explicitly does not use a `has_next_page`

aaschaer (commented):
Looks good minus a minor docstring issue

src/globus_sdk/paging/marker.py
sirosen: Oof! Good catch. It's a clear case of copy-paste-replace gone wrong. I'll edit this to cleanup and then we should be good to go.

PR #522, opened 2022-02-22
Custom TransferResponseTransport that does not retry ExternalErrors

aaschaer:
For https://app.shortcut.com/globus/story/12770/fix-how-retries-are-done-in-the-sdk-on-transfer-502s-to-not-retry-externalerrors

I went with a custom class that inherits from RequestsTransport, seems to apply pretty cleanly.

Hacked this into my local CLI to do a bit of live API testing and all seemed well.

### DIFF
--- a/changelog.d/20220222_110056_aaschaer_transfer_transport.rst
+++ b/changelog.d/20220222_110056_aaschaer_transfer_transport.rst
@@ -1,0 +1,10 @@
+..
+.. A new scriv changelog fragment
+..
+.. Add one or more items to the list below describing the change in clear, concise terms.
+..
+.. Leave the ":pr:`...`" text alone. When you open a pull request, GitHub Actions will
+.. automatically replace it when the PR is merged.
+..
+
+* Add TransferRequestsTransport class that does not retry ExternalErrors. This fixes cases in which the TransferClient incorrectly retried requests (:pr:`NUMBER`)
--- a/src/globus_sdk/services/transfer/client.py
+++ b/src/globus_sdk/services/transfer/client.py
@@ -1,7 +1,7 @@
 import logging
 import time
 import uuid
-from typing import Any, Dict, Iterable, List, Optional, Tuple, Union
+from typing import Any, Dict, Iterable, List, Optional, Tuple, Type, Union
 
 from globus_sdk import client, exc, paging, response, utils
 from globus_sdk.scopes import TransferScopes
@@ -10,6 +10,7 @@ from globus_sdk.types import DateLike, IntLike, UUIDLike
 from .data import DeleteData, TransferData
 from .errors import TransferAPIError
 from .response import ActivationRequirementsResponse, IterableTransferResponse
+from .transport import TransferRequestsTransport
 
 log = logging.getLogger(__name__)
 
@@ -97,6 +98,7 @@ class TransferClient(client.BaseClient):
     """
     service_name = "transfer"
     base_path = "/v0.10/"
+    transport_class: Type[TransferRequestsTransport] = TransferRequestsTransport
     error_class = TransferAPIError
     scopes = TransferScopes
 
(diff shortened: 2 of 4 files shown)


src/globus_sdk/services/transfer/client.py
```diff
@@ -97,6 +98,7 @@ class TransferClient(client.BaseClient):
     """
     service_name = "transfer"
     base_path = "/v0.10/"
+    transport_class: Type[TransferRequestsTransport] = TransferRequestsTransport
```
sirosen, line 101: Is it necessary to annotate this with `Type[...]`? I would have thought that this wouldn't be needed, but maybe `mypy` can't understand without it.

changelog.d/20220222_110056_aaschaer_transfer_transport.rst
```diff
@@ -0,0 +1,10 @@
+..
+.. A new scriv changelog fragment
+..
+.. Add one or more items to the list below describing the change in clear, concise terms.
+..
+.. Leave the ":pr:`...`" text alone. When you open a pull request, GitHub Actions will
+.. automatically replace it when the PR is merged.
+..
+
+* Add TransferRequestsTransport class that does not retry ExternalErrors (:pr:`NUMBER`)
```
sirosen, line 10: I think it would help to have another sentence here to explain the external impact a little. Maybe adjust like so?
```suggestion
* Add TransferRequestsTransport class that does not retry ExternalErrors. This fixes cases in which the TransferClient incorrectly retried requests (:pr:`NUMBER`)
```

src/globus_sdk/services/transfer/transport.py
```diff
@@ -0,0 +1,25 @@
...
+        """
+        check for transient error status codes which could be resolved by
+        retrying the request. Does not retry ExternalErrors as those are
+        unlikely to actually be transient.
+        """
+        if ctx.response is not None and (
+            ctx.response.status_code in self.TRANSIENT_ERROR_STATUS_CODES
+        ):
+            try:
+                if "ExternalError" in ctx.response.json()["code"]:
```
sirosen, line 19: I feel like there's a lot happening in one line here. JSON parsing and indexing and a comparison.

Can we break it up a little, maybe like
```
try:
   code = ctx.response.json()["code"]
except (ValueError, KeyError):
    code = ""
if code.startswith("ExternalError"): ...  # also, is changing to `startswith` a good idea or not?
```

sirosen (commented):
This looks great! It's even simpler than I had hoped, which has me super-happy.
The tests look good too. Nothing too elaborate, just checks that the modified transport does what we want.

I had one thought about the changelog to make it clearer, and one question about a type annotation. (Mostly wondering if we can do more to pacify `mypy` via annotations on `BaseClient` in the future.)

Other than that, I'm looking at the check function and wondering if we can make it simpler, but if you prefer it as it is now, I'm happy to leave it as-is.

src/globus_sdk/services/transfer/client.py
aaschaer: I just did what BaseClient did, was having trouble getting mypy to run locally for some reason

src/globus_sdk/services/transfer/transport.py
aaschaer: Sure, I'll break it up
I believe the Transfer error code may be RestExternalError in some cases, so probably best to stick with `in`

src/globus_sdk/services/transfer/client.py
sirosen: Okay, if I remember I might play with this later.
(Aside: did `tox -e mypy` work for you? `mypy` with and without virtualenvs active gets tricky, so I'd rely on the tox target to setup a "good" install where libraries load correctly.)

src/globus_sdk/services/transfer/transport.py
sirosen: That is somewhat unfortunate, but understandable. If we need to refine this in the future we can always `code.split(".")`, and so forth.

sirosen (approved):

PR #525, opened 2022-02-26
Add support for skipping the registry in _testing

sirosen:
`load_response_set` and `load_response` can now take their relevant object types directly. This allows use of the RegisteredResponse interface even under parametrized tests and other scenarios where it's not useful or desirable to register a ResponseSet for multiple tests to use.

Convert the transport retry tests to use this behavior in order to demonstrate. There's also an example in the `_testing` readme.

Also, while doing this, reorganize `_testing` a little. The registry module now contains all of the code for adding to and reading from the registry, and the models have been moved into a dedicated module, separate from the registry. The main goal of this cleanup is to make the registry more readable, as new and more nuanced behaviors are being added here.

---

@rudyardrichter, I thought this change might solve the problem you described running multiple registration calls under a parametrized test by providing a more "direct" outlet for your scenario. I initially imagined that your case would be solved with direct `responses` usage, but it occurs to me that

1. Once you start using `globus_sdk._testing`, you might prefer to use _only_ that, rather than a mix of some tests using `_testing` and some using `responses`
2. We are offering some value by handling "service name" resolution, and that might be desirable even when registering explicit responses

I still think that the original usage we discussed of multiple registrations *should* work -- the registry is just a global dict and all you'd be doing is overwriting values in it -- but we can dig into that specific case more later.

### DIFF
--- a/src/globus_sdk/_testing/README.rst
+++ b/src/globus_sdk/_testing/README.rst
@@ -48,9 +48,9 @@ activated by name:
     from globus_sdk._testing import load_response
 
     # load_response will add the response to `responses` and return it
-    load_response("AuthClient.get_identities")
+    load_response("auth.get_identities")
     # "case" is used to have a single name map to multiple responses
-    data = load_response("AuthClient.get_identities", case="multiple")
+    data = load_response("auth.get_identities", case="multiple")
 
 Responses can also be activated by passing an SDK client method, bound or
 unbound, as in:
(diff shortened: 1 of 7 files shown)


rudyardrichter:
This definitely is useful. I've been messing around with it yesterday & today a bit, and the interface seems good.

However, I should note I'm still getting that error about `match_querystring`, despite trying various different `responses` pins. It would be nice to get that resolved so whenever this gets published it'll be easy to slot the published version into flows.

sirosen:
> However, I should note I'm still getting that error about `match_querystring`, despite trying various different `responses` pins.

Independent of this comment, I wanted to take another look at that. Let me do it in a follow-up if this is good to merge.

I _think_ that in `_testing` I've set `match_querystring=False`, and that the right way to avoid the warning on `responses==0.17.0` is `match_querystring=None`. If it were incorrect here though, I'd expect us to get warnings when running the SDK testsuite, and for those to be flagged. I'll check this, and there might be another PR to fix it.

rudyardrichter (approved):

PR #528, opened 2022-02-28
Add support for 'include' as a param to get_group

sirosen:
The primary change here is to add support for `get_group(..., include="memberships")` and related usages. We'll need this for `globus group members list` which will pass this parameter.

However, there are two "adjacent" changes being included here for expediency. One is to move some of the test fixture data into `_testing` -- only for tests which are being modified now. The other is to ensure that the docstrings on all GroupsClient methods include param docs.


sirosen:
To show a bit of the effects of the documentation change...

before | after
---|---
image | image
image | image

src/globus_sdk/_testing/data/groups/_common.py
```diff
@@ -0,0 +1,27 @@
+from typing import Any, Dict
+
+GROUP_ID = "d3974728-6458-11e4-b72d-123139141556"
+
+BASE_GROUP_DOC: Dict[str, Any] = {
+    "name": "Claptrap's Rough Riders",
+    "description": "No stairs allowed.",
```
rudyardrichter, line 7: Your reference has not gone unappreciated :)

sirosen: That warms my frosty heart. This is a real group in production, which I created when the reference was very fresh. 😸

PR #536, opened 2022-03-14
Add "known_url_scopes" to ScopeBuilder

sirosen:
Additionally, allow a `str` instead of only `list[str]`.

Initially I only intended to update to add `known_url_scopes`, but I accidentally passed a string in the tests (which failed). Rather than fixing the test, I chose to fix the behavior of ScopeBuilder to allow for a string -- passing a string instead of a string iterable is a rather common error with surprising behavior. (`foo` -> `['f', 'o', 'o']`)


rudyardrichter (approved):
Super! This actually unblocks a mostly-fleshed-out Timer client, since right now we have just the URL scope.

PR #537, opened 2022-03-14
Add TransferEndpointType enum example to docs

sirosen:


docs/examples/endpoint_type_enum.rst
```diff
@@ -0,0 +1,87 @@
...
+            else:
+                gcsv5 = False
+
+            if gcsv5:
+                if has_host:
+                    return EndpointType.GUEST_COLLECTION
+                else:
+                    return EndpointType.MAPPED_COLLECTION
+
+            elif shared:
```
aaschaer, line 84: I think this should be `has_host`

aaschaer (commented):
Found a few issues in the example code.

There seems to be mixed usage between EndpointType and TransferEndpointType. Not sure which one is "correct" but only one should be used

docs/examples/endpoint_type_enum.rst
sirosen: Yes, thank you for the save. I was editing the CLI version of this (where we called it `shared`) and trying to adjust the names for a broader audience.

sirosen:
@aaschaer, could you take a second look at this?

I've just corrected the issues you pointed out, and I think this is now clean.

aaschaer (approved):

PR #548, opened 2022-04-05
Add TimerClient, TimerJob payload, TimerAPIError, docs, and tests

rudyardrichter:
* Add a client for the Timer service
    * Add `TimerClient` class, along with `TimerJob` for constructing data
      to pass to the Timer service for job creation, and ``TimerAPIError``
    * Modify `globus_sdk.config` utilities to provide URLs for Actions and
      Timer services

The docs additions might need some RST/sphinx proofreading—I'll check out the built version for the PR and see if it works.


rudyardrichter:
Let me know if there's any docs additions I missed. I think I covered it, but I'm not sure if there are other places like `index.rst` where I should also link out to Timer now.

Also, I have a larger functional example script which runs most of the methods on the `TimerClient`, where would be a good place to put that?

rudyardrichter:
(Force-pushing a bit to try to fix up the docs.)

src/globus_sdk/services/timer/data.py
```diff
@@ -0,0 +1,113 @@
...
+            self["stop_after"] = stop_after.isoformat()
+        if stop_after_n is not None:
+            self["stop_after_n"] = stop_after_n
+        if scope is not None:
+            self["scope"] = scope
+
+    @classmethod
+    def from_transfer_data(
+        cls,
+        transfer_data: TransferData,
```
sirosen, line 80: I think we also want to support dict data, since transfer submission supports that. It will complicate things regarding `environment` support (see the note below), but I'm looking for the SDK to generally "feel consistent" about accepting `dict | <fancy-helper>` in all locations where it's reasonable to do so. It's also an important escape-valve if the helper class, in this case `TransferData`, doesn't support some valid but unexpected format for the data.

```suggestion
        transfer_data: Union[TransferData, Dict[str, Any]],
```

src/globus_sdk/services/timer/data.py
```diff
@@ -0,0 +1,113 @@
...
+        start: Union[datetime.datetime, str],
+        interval: Union[datetime.timedelta, int, None],
+        **kwargs: Any,
```
sirosen, line 83: I think we should not use `**kwargs` and instead expand out the full list of args (exactly as in `__init__`).

It's somewhat annoying as a maintainer, but it has good effects on the type annotations because the signature for this method will then contain things like `name: str | None`. Without doing this, a caller can pass `name=0` here and it will pass `mypy`, but then behave in a bad way at runtime.

src/globus_sdk/services/timer/data.py
```diff
@@ -0,0 +1,113 @@
...
+            self["name"] = name
+        if stop_after is not None:
+            self["stop_after"] = stop_after.isoformat()
+        if stop_after_n is not None:
+            self["stop_after_n"] = stop_after_n
+        if scope is not None:
+            self["scope"] = scope
+
+    @classmethod
+    def from_transfer_data(
```
sirosen, line 78: I'm leaving my comments below because they'll show my thought process on this a bit, but I'd also like to offer an alternative way forward. I have questions/concerns about handling environments nicely, and I want various changes which will be tedious/laborious.

If this doesn't seem like a quick solve and we want to unblock things, we can remove this helper and add the following method:

```python
@classmethod
def get_transfer_action_url(
    cls, *, environment: Optional[str] = None
) -> str:
    return slash_join(
        get_service_url("actions", environment=environment), "transfer/transfer/run"
    )
```

plus an example added to the `docs/examples/` dir which shows how to use this to construct a timer task.

That will set us up to implement something that does this assembly step but we can think about how to handle env names being set at runtime without blocking the rest of the work.

This isn't the only way forward. If there's a way of refactoring this which satisfies my desire to support setting `environment` at runtime, I'm okay with keeping this more sophisticated helper.

src/globus_sdk/services/timer/errors.py
```diff
@@ -0,0 +1,60 @@
...
+class TimerAPIError(GlobusAPIError):
+    """
+    Error class to represent error responses from Timer.
+
+    Has no particular additions to the base ``GlobusAPIError``, but implements a
+    different method for parsing error responses from Timer due to the differences
+    between pydantic-generated errors and other errors.
+    """
+
+    MESSAGE_FIELDS: List[str] = []  # we are overriding `_load_from_json` instead
```
sirosen, line 15: I don't think you need to set this classvar at all. It's not necessary to remove it, but if you want to drop a couple of lines I think it's fine to omit.

src/globus_sdk/services/timer/errors.py
```diff
@@ -0,0 +1,60 @@
...
+        the request. It might say `["body", "start"]` for example, or it might start
+        with "query" to point to a value in the query string.
+        """
+        if "error" in data:
+            self.code = data["error"].get("code", "ERROR")
+            self.message = data["error"].get("detail", "")
+        elif "detail" in data:
+            self.code = "Validation Error"
+            self.message = "; ".join(
+                e["msg"] + ": " + ".".join(k for k in e["loc"]) for e in data["detail"]
```
sirosen, line 56: Not important, but this triggered an idea which I don't want to forget. If you wanted to build a jsonpath from `loc`, you could do
```python
path = "$"
for elem in e["loc"]:
    if isinstance(elem, int):
        path += "[" + str(elem) + "]"
    else:
        path += "." + elem
```

We could add this to `utils` as a helper for all data which has path-like elements like this.

src/globus_sdk/_testing/data/timer/get_job.py
```diff
@@ -0,0 +1,126 @@
+# flake8: noqa
```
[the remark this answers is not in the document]
sirosen: I think it's fine to not try to share data between the fixtures, at least for now. We could maybe work something out later.

I assume the noqa is for line length?
Could we maybe dedent a bit and avoid that by assigning to some vars which we then use in the fixture? Like

```python
_ACTION_DATA = {
    "data": {
        ...
    }
}
```
?

sirosen (changes requested):
With only a couple of exceptions this looks merge-ready! 👍

My major area of interest is the `TimerJob` object, and the classmethod for converting a `TransferData`. I have detailed comments there.

One meta comment/question: should there also be an `ActionsClient` object? If there's going to be a future one, that may have bearing on `TimerJob`. e.g. Perhaps `get_transfer_timer_callback_url` is a method which belongs on `ActionsClient`, not on `TimerJob`. (I haven't thought that through thoroughly.)

src/globus_sdk/_testing/data/timer/get_job.py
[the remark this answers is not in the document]
rudyardrichter: Yes, exactly. I'll give that a shot.

src/globus_sdk/_testing/models.py
```diff
@@ -13,6 +13,7 @@ class RegisteredResponse:
         "search": "https://search.api.globus.org/",
         "gcs": "https://abc.xyz.data.globus.org/api",
         "groups": "https://groups.api.globus.org/v2/",
+        "timer": "https://timer.automate.globus.org/",
```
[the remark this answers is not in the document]
rudyardrichter: I was thinking about that. Maybe a stub `ActionsClient` would be reasonable. However, for the callback URL issue, it would still be a little odd, I think? Since there are several different actions, which have differing URLs. I think we'd need something like an Action Provider enum which stores a bunch of the URLs for various action providers. What do you think?

src/globus_sdk/config/environments.py
```diff
@@ -103,11 +103,15 @@ class ProductionEnvConfig(EnvConfig):
     envname = "production"
     domain = "globus.org"
     nexus_url = "https://nexus.api.globusonline.org/"
+    timer_url = "https://timer.automate.globus.org/"
+    actions_url = "https://actions.automate.globus.org/"
```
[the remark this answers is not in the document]
rudyardrichter: No, that seems good. Since at minimum timer + actions + flows have this in common I think it makes sense. I'll make the change.

src/globus_sdk/services/timer/client.py
```diff
@@ -0,0 +1,113 @@
...
+        """
+        ``GET /jobs/<job_id>``
+
+        **Examples**
+
+        >>> timer_client = globus_sdk.TimerClient(...)
+        >>> job = timer_client.get_job(job_id)
+        >>> assert job["job_id"] == job_id
+        """
+        log.info(f"TimerClient.get_job({job_id})")
```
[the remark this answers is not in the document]
rudyardrichter: Yeah, that would be nice. So, these will be left in for this PR, and we'll expect to remove them at some point?

src/globus_sdk/services/timer/data.py
```diff
@@ -0,0 +1,113 @@
...
+        :type start: datetime.datetime or str
+        :param interval: The interval at which the Timer job should recur.
+        :type interval: datetime.timedelta
+        :param \**kwargs: Keyword args to pass to define the Timer job; these are the
+            same as the keyword args for ``__init__`` (``name``, `stop_after``, etc.).
+            See the init docs.
+        :type \**kwargs: any
+        """
+        transfer_action_url = slash_join(
+            get_service_url("actions"), "transfer/transfer/run"
```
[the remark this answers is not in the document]
rudyardrichter: What if this just includes an `environment` kwarg?

src/globus_sdk/services/timer/data.py
rudyardrichter: Hah, I started with kwargs as a hack, and found pylint complained about the docstring (apparently it recently changed that `\**kwargs` is the correct way to write it for sphinx?), so I was distracted by my little kwargs win. Yeah, I agree expanding the args is preferable.

src/globus_sdk/services/timer/errors.py
rudyardrichter: Hmm, sure. I wrote this in partly to indicate that the attr on the base class would not be correct. This version seems slightly more correct to me in that respect (even if it isn't used)? I would set it to `None` but that would be a type error.

sirosen: Yeah, I thought that might be the reasoning here -- it's not used but this is the only way to write "there are no `MESSAGE_FIELDS`". Makes sense. Let's mark this resolved; I'm good with keeping it.

I just wanted to make sure you knew it's not technically necessary.

src/globus_sdk/services/timer/data.py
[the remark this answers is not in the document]
sirosen: I thought about that too, as a possible solution.

I'm just not sure what the right interface is here. A `TransferData` object actually has a relevant environment because it has a `submission_id`. So, just to write down what I'm thinking, if you are setting the environment at runtime, usage might be like so...
```python
env: str = get_env_from_config_file()

# not looking things up, usages might be a little incorrect
ac = ConfidentialAppAuthClient(
    client_id=creds[env].client_id,
    client_secret=creds[env].client_secret,
    environment=env
)

tc = TransferClient(
    authorizer=ClientCredentialsAuthorizer(ac, TransferClient.scopes.all),
    environment=env
)

tdata = TransferData(tc, ...)
tdata.add_item(...)

timerc = TimerClient(
    authorizer=ClientCredentialsAuthorizer(ac, TimerClient.scopes.timer),
    environment=env
)
hourly_job = TimerJob.from_transfer_data(tdata, utcnow(), 3600, environment=env)

timerc.create_job(hourly_job)
```

Looking at this, I see three things.
1. You need to specify `environment=env` _many times_, and if you make a mistake things will not work in somewhat opaque ways
2. Each time, it's for a client init except for `TimerJob`, so that still stands out to me
3. The interface in this case is already not _great_, so maybe it doesn't matter much to make it slightly worse in the name of simplicity

In that light, this seems not that critical to me to try to get this "perfect" because it will be a little bit awkward no matter what we do. `environment: str | None = None` as a new parameter seems like the easiest solution.

rudyardrichter:
Ok, I think I've addressed all the suggestions, and this should be good for re-review. I tested the non-prod-env case again and it's still good.

Also, I added an example page in the docs with the script I mentioned above.

src/globus_sdk/services/timer/data.py
```diff
@@ -0,0 +1,133 @@
...
+
+    :param callback_url: URL for the action which the Timer job will use.
+    :type callback_url: str
+    :param callback_body: JSON data which Timer will send to the Action Provider on
+        each invocation
+    :param start: The datetime at which to start the Timer job.
+    :type start: datetime.datetime or str
+    :param interval: The interval at which the Timer job should recur.
+    :type interval: datetime.timedelta
+    :type callback_body: dict
```
sirosen: Let's tweak to put this line up with the `:param:`.

src/globus_sdk/services/timer/data.py
```diff
@@ -0,0 +1,133 @@
...
+    Class for specifying parameters used to create a job in the Timer service. Used as
+    the ``data`` argument in
+    :meth:`create_job <globus_sdk.TimerClient.create_job>`.
+
+    Timer operates through the `Globus Automate API
+    <https://docs.globus.org/globus-automation-services/>`_. Crucially, the
+    ``callback_url`` parameter should always be the URL used to run an action provider.
+    Currently the only supported action provider for this is for Transfer. Thus, users
+    should generally only use the :meth:`~from_transfer_data` method here. Any other
+    usage is meant for internal purposes; proceed with caution!
```
sirosen: Maybe format with
```
.. warning:
    some text
```

so that this is very attention-get-y. Otherwise, users are probably going to try `__init__` and get confused

src/globus_sdk/services/timer/data.py
```diff
@@ -0,0 +1,132 @@
...
+        :param interval: The interval at which the Timer job should recur.
+        :type interval: datetime.timedelta
+        :param name: A (not necessarily unique) name to identify this job in Timer
+        :type name: str, optional
+        :param stop_after: A date after which the Timer job will stop running
+        :type stop_after: datetime.datetime, optional
+        :param stop_after_n: A number of executions after which the Timer job will stop
+        :type stop_after_n: int
+        :param scope: The scope Timer will need in order to authorize against the Action
+            Provider specified in the callback URL.
```
sirosen, line 111:
```suggestion
        :param scope: Timer defaults to the Transfer 'all' scope. Use this parameter to
            change the scope used by Timer when calling the Transfer Action Provider
```

docs/examples/timer_operations.rst
```diff
@@ -0,0 +1,54 @@
...
+    timer_client = TimerClient(authorizer=timer_authorizer)
+
+    start = datetime.utcnow()
+    # We're going to run this job only a single time (``stop_after_n=1``), so the
+    # interval must be specified as ``None``, meaning it does not repeat. If we were
+    # going to set up a recurring job, we would use, for example,
+    # ``interval = timedelta(days=7)``.
+    interval = None
+    name = "example-timer-job"
+    job = TimerJob.from_transfer_data(data, start, interval, stop_after_n=1, name=name)
```
sirosen, line 41: After discussion, we agree that this might "explain better" if in the example,

```
interval = timedelta(minutes=30)
stop_after_n = 2
```

so that we don't need to explain the `n=1` case here.

But wait, there's more! We should also explain the `n=1` behavior in the TimerJob docs.

docs/examples/timer_operations.rst
```diff
@@ -0,0 +1,54 @@
...
+    response = timer_client.create_job(job)
+    assert response.http_status == 201
+    job_id = response["job_id"]
```
sirosen, line 44: Why don't we print the job_id here, so that delete can happen later?

src/globus_sdk/services/timer/data.py
```diff
@@ -0,0 +1,133 @@
...
+
+    :param callback_url: URL for the action which the Timer job will use.
+    :type callback_url: str
+    :param callback_body: JSON data which Timer will send to the Action Provider on
+        each invocation
+    :type callback_body: dict
+    :param start: The datetime at which to start the Timer job.
+    :type start: datetime.datetime or str
+    :param interval: The interval at which the Timer job should recur.
+    :type interval: datetime.timedelta
```
sirosen, line 33: I just noticed this, but should this be documented as `timedelta or int` and "The interval in seconds ..."?
Otherwise, the only way I'd know what `interval=300` means is by reading source here and seeing that the delta is converted to seconds.

Also applies to the same param in the Transfer helper.

rudyardrichter:
Made the updates discussed earlier today— let me know in particular if the new version of the example script makes sense.

sirosen:
The revised example script looks great to me -- I realize the login flow makes it a bit longer, but I think it's worth it.

I've gone back over each of the things we discussed and everything seems to be resolved.
I'm going to go ahead and merge, and we can always make further adjustments if it proves necessary.

sirosen (approved):

PR #549, opened 2022-04-08
Fix annotations to allow data:str

sirosen:
I found this annotation bug while working on https://github.com/globus/globus-cli/pull/626

### DIFF
--- a/changelog.d/20220408_044943_sirosen_allow_str_data.rst
+++ b/changelog.d/20220408_044943_sirosen_allow_str_data.rst
@@ -1,0 +1,2 @@
+* Fix annotations to allow request data to be a string. This is
+  supported at runtime but was missing from annotations. (:pr:`NUMBER`)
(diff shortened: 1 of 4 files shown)


aaschaer (approved):
Looks good to me. Not sure why anyone would rather pass a string than a dict, but if we support it the types should reflect that.

sirosen:
> Not sure why anyone would rather pass a string than a dict

The reason this came up is that `globus api ...` is going to want to be able to pass an arbitrary payload.

I think for _normal_ cases, nobody wants to pass a string. But it's important for the corner cases.

Two quick examples to serve as food for thought:
- You might already have form-encoded data like `foo=bar%20baz` and want to send it with `Content-Type: application/json` to see that the application rejects it with a correct error
- An API which takes non-JSON, non-form data. e.g. Some bulk APIs are implemented using newline-delimited JSON objects (I don't think we have any of these today) because it lets each line parse independently

PR #551, opened 2022-04-15
Expose more details of HTTP requests

sirosen:
Motivation:
In the `globus api` command, I wanted to disable redirects and found we couldn't. I also wanted verbose mode to reconstruct the response string (as best we can) and was missing several pieces of data. [This commit](https://github.com/globus/globus-cli/pull/626/commits/ea767b3b1fd1d4178077a06ffbbc17376459f200) on the [`globus api` PR](https://github.com/globus/globus-cli/pull/626) shows what is possible with the new options and values.

`stream` is not used anywhere today, but I've added it in case we need it in the future. While looking at docs on `allow_redirects`, I examined the other options of `requests.Session.send` and on prepared request objects. `stream` was the only one that I judged might have some utility.

I did consider exposing the underlying `requests.Response` objects rather than adding the new attributes, but have decided not to do so. Right now, the use of `requests` is encapsulated by the SDK -- the thinking behind that has always been that we could plug-and-play an alternate transport like urllib3 or even (if we can sort out how) aiohttp. It's probably a far-off possibility that we'll do this in the near term, but for now I'm not going to disrupt the status-quo with respect to this. `requests` is "somewhat hidden" and usually shouldn't need to be touched directly.

---

New options:

- allow_redirects: bool on `request` methods
- stream: bool on `request` methods

These are underlying `requests` options which are re-exposed via the SDK methods for raw request sending. Because they are "advanced", they are not also re-exposed on the `get()`, `post()`, etc helper methods.

New properties:

- http_reason on responses and errors
- headers on errors

This lets the caller get more detail about the HTTP interaction in the success and error cases.

Tests exercise the new functionality, with the caveat that `allow_redirects=False` doesn't really do anything when running under `responses`. If we find a better test, there's a comment which will help us know that we can remove the old test.

For the tests, there are some status->reason mappings that "do the normal thing", e.g. `200 -> OK`, but these aren't baked into the SDK anywhere or even into `_testing`.

---

One final, minor note: you may note that `http_reason` is of type `str`, not `str | None`, even though `requests.Response.reason` is given a value of `None` on init. The `reason` will actually always be populated, and it took some digging to find this out. `http.client.HTTPResponse` is used, and the reason string is pulled from the status line of the response.

### DIFF
--- a/changelog.d/20220415_153608_sirosen_expose_more_http_details.rst
+++ b/changelog.d/20220415_153608_sirosen_expose_more_http_details.rst
@@ -1,0 +1,8 @@
+* Several changes expose more details of HTTP requests (:pr:`NUMBER`)
+  * ``GlobusAPIError`` has a new property ``headers`` which provides the
+    case-insensitive mapping of header values from the response
+  * ``GlobusAPIError`` and ``GlobusHTTPResponse`` now include ``http_reason``,
+    a string property containing the "reason" from the response
+  * ``BaseClient.request`` and ``RequestsTransport.request`` now have options
+    for setting boolean options ``allow_redirects`` and ``stream``, controlling
+    how requests are processed
--- a/src/globus_sdk/client.py
+++ b/src/globus_sdk/client.py
@@ -242,6 +242,8 @@ class BaseClient:
         data: DataParamType = None,
         headers: Optional[Dict[str, str]] = None,
         encoding: Optional[str] = None,
+        allow_redirects: bool = True,
+        stream: bool = False,
     ) -> GlobusHTTPResponse:
         """
         Send an HTTP request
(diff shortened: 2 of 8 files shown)


aaschaer (approved):
Code looks good to me!

I had to read up on `stream` since "Do not immediately download the response content" didn't click right away, but that is a reasonable description of what the option does, so I'm not sure if/how that text could be more clear. I might be slightly happier if we somehow labeled it as an advanced option somehow? Just so anyone else who goes "wait what does that mean exactly?" can know it's safe to ignore

sirosen:
> I might be slightly happier if we somehow labeled it as an advanced option somehow?

I decided against "for advanced use-cases" in the docstrings, but I could be convinced. I just didn't want the text to get too long and potentially harder to read.

For now, I put these into `request()` but not `get()`, etc. Which is how I marked them "advanced". We could maybe refine that more in the future -- I'm open to ideas on this.

PR #553, opened 2022-04-26
Add MutableScope type for manipulating scopes

sirosen:
In order to define optional scopes and scoes with dependencies, define a new type, MutableScope, from the scopes module. Anywhere that scope strings are accepted, MutableScopes are now accepted too.

MutableScope.collection2str is a static method attached to the class for converting strings, iterables of strings, and mixed iterables of strings and MutableScopes into the desired (space-separated) string.

Additionally, `ScopeBuilder.make_mutable` is added as a method for ScopeBuilders to create MutableScope objects from their existing scope attributes.

Narrative docs are added and MutableScope is added to the `scopes` module's reference documentation.

---

A brief note on the scope of this feature-set:
I considered allowing `add_dependency(MutableScope(...))` and several other conveniences and variants when I did the first implementation. However, as I worked through various cases, the complexity it introduced seemed greater than the benefit of supporting it. This is therefore fairly "bare bones" and can be expanded in the future. One thing I kept from those efforts is the check for a leading `*` on a scope string. I think it provides some value, given that `foo[**bar]` is not valid, but `foo[bar]` and `foo[*bar]` both are.

### DIFF
--- a/changelog.d/20220426_121025_sirosen_mutable_scopes.rst
+++ b/changelog.d/20220426_121025_sirosen_mutable_scopes.rst
@@ -1,0 +1,9 @@
+* New tools for working with optional and dependent scope strings (:pr:`NUMBER`)
+
+  * A new class is provided for constructing optional and dependent scope
+    strings, ``MutableScope``. Import as in
+    ``from globus_sdk.scopes import MutableScope``
+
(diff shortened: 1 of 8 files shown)


rudyardrichter (approved):
I'm good with this interface :+1:

PR #555, opened 2022-04-29
Minor tweaks to test config

sirosen:
- add a `find` to `make clean` so that it scrubs `__pycache__` dirs as well
- set pytest testpaths=... for faster test collection (avoids scanning any other dirs for test modules)
- set a needed ignore for cryptography warnings for py3.6 builds

Regarding the last item, the constraints for handling the ignore are tied up in the way that the `warnings` module is initialized. We cannot give a very precise ignore, so use string-based matching for now. If this turns out to be unstable, we can revisit and potentially disable all warnings for the 3.6 builds via tox factor config.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,4 +22,5 @@ release:
 
 .PHONY: clean
 clean:
-	rm -rf dist build *.egg-info .tox
+	rm -rf dist build *.egg-info .tox .venv
+	find . -type d -name '__pycache__' -exec rm -r {} +
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,10 +20,17 @@ warn_no_return = true
 
 
 [tool:pytest]
+testpaths = tests
 norecursedirs = tests/non-pytest
 filterwarnings =
     # warnings are errors, like -Werror
     error
+    # ignore cryptography warnings about end of 3.6 support
+    # we are aware, but we only see this warning because we still support py3.6
+    # DO NOT refer to `cryptography.utils.CryptographyDeprecationWarning` for
+    # this because that would cause failures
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -22,4 +22,5 @@ release:
 
 .PHONY: clean
 clean:
-	rm -rf dist build *.egg-info .tox
+	rm -rf dist build *.egg-info .tox .venv
+	find . -type d -name '__pycache__' -exec rm -r {} +
```
rudyardrichter, line 26: I think I've seen this as `\+` at the end of the find exec, is there a difference?

aaschaer (approved):

Makefile
rudyardrichter: Ahh, looks like the semicolon version _does_ need escaping, while the plus one does not (but people write it that way anyways).

sirosen: A very subtle one. If `+` is a special character in your shell, `\+` would ensure the literal `+` makes it to `find`.

I don't think this is an issue in make, since the default shell is `/bin/sh`.
We could tweak it if you like, but I've already merged because the cryptography thing is in the way of getting everything else merge-ready.

rudyardrichter: Yep, totally fine. Just a curiosity.

sirosen: > Just a curiosity.

Like the rest of the syntax of `find -exec`! 😹 

You'll also sometimes see single quotes, for similar reasons. It gets... very confusing. But it's the "easiest" way I know of to clean up files and dirs in a tree.

The change made it in here because I was tinkering a lot with the warnings stuff and thought maybe a pycache dir was causing issues. (It was not.)

PR #559, opened 2022-05-13
Implement lazy imports for the top-level __init__.py for faster imports

sirosen:
I'm opening this for us to discuss. I believe that this could be very much worthwhile, and I realized that the py36-compat doesn't make it more complex. But there are downsides, not only in terms of code complexity for us (which I think is all manageable).

The first commit contains my first (working) crack at lazy imports using module-level `__getattr__` as introduced in python3.7 . The second commit contains a rework to autogenerate the code and support python3.6 .

I suspect we will _not_ pursue this. However, I have another (alternative/competing) change which I think I _would_ like to pursue. More on this later.

## Why it matters

`globus-cli` is used to drive CLI completions. That means that every time you hit tab with completions enabled, you need to wait for the interpreter to start up, for the CLI to "put itself together", and then for it to generate the completion response. This is slow, so our completions don't "feel snappy".

A common assumption that people have is that python interpreter start is slow. But it turns out that they're probably compounding _actual_ interpreter start with the import of modules they use. i.e. They're measuring the time between invocation and the first line of their `main()` func executing.
How slow is python interpreter start? 26ms. How slow is importing `globus_sdk`? 225ms. 26ms is about twice the lower detectable limit of human perception -- i.e. "the blink of an eye". 225ms is slower than human reaction times -- i.e. "a perceptible delay". If the CLI is taking another 100ms (it's probably slower), we're crossing from 126ms (noticeable but fast) to 351ms ("very laggy").

You can see a breakdown of the slow modules for imports with `PYTHONPROFILEIMPORTTIME=1 python -c 'import globus_sdk'
`requests` and the underlying `urllib3` tree of modules will stand out as the most significant delay.

This may not seem significant for other applications, but also note that SDK names may be imported for use in type annotations. A simple module may read:
```
from __future__ import annotations
import globus_sdk

def get_group_ids(client: globus_sdk.GroupsClient) -> list[str]:
    ...
```

With deferred import semantics (a future import today, becoming the default), `import globus_sdk` could be a no-op at import-time. With the current, imperative, import semantics, whenever this module executes, there's a delay if it's the first import of `globus_sdk`.

Basically, this change turns into a speed boost for any application which wants to use `globus_sdk` but not all of it. For example, `TransferClient` takes a long time to initialize due to its size. If an application does not use `TransferClient` ever, this changeset will result in a small perf improvement.

## py3.6 compatibility and type compatibility

`mypy` cannot grasp what a module-level `__getattr__` does. So if we just do things with `__getattr__`, `mypy` won't be able to handle our types correctly. This is the same general problem that `__getattr__` has in type checking.

However, we can enumerate exactly what `mypy` _should_ see, as it matches the data for `__getattr__`. Put that set of imports under `if TYPE_CHECKING` and `mypy` "understands" what we're telling it. Nicely, this is also the code which needs to execute eagerly under py3.6 where there is no module-level `__getattr__`. So the check is for `TYPE_CHECKING` or py3.6.

## Are we fast yet?

Some real numbers from my machine:
- `import globus_sdk` on python3.7 without this change: 220ms
- `import globus_sdk` on python3.7 with this change: 17ms

On python3.6, we see times of ~220ms on either version.

## Downsides: Correctness and Perf Spikes

If there are import cycles in python, they can be safe or unsafe depending on where you enter the cycle.
With lazy import semantics, we no longer (as the library authors) control where cycles are entered. It means that `from globus_sdk import AuthClient; from globus_sdk import GlobusAuthorizer` could work while `from globus_sdk import GlobusAuthorizer; from globus_sdk import AuthClient` fails on an import cycle. This is hard to test and catch.

Performance in this case is a double-edged sword. We aren't really making the SDK _faster_ per se. We're just avoiding the slow paths until they're really needed. This could manifest as performance spikes, which is especially relevant for our web APIs which use the SDK. Some requests (e.g. the first request to a new web worker) will be slower because they're the first in the relevant process to hit an import path. We know the total import time is over 200ms. We might need to provide some method like `globus_sdk._force_eager_imports()` for these cases, to protect against undesired spikiness.


sirosen:
Closing -- superseded by #571.

PR #560, opened 2022-05-13
Implement partial lazy importing in globus-sdk

sirosen:
This defers imports of 3rd party packages and some stdlib modules until they are first used. Where necessary, the packages are imported at type checking time.

This means that, for example, applications which do not use `globus_sdk.GlobusConnectPersonalOwnerInfo` do not necessarily end up importing `shlex`. As a result, the startup cost associated with importing that module is not paid. In particular, `requests` is a slow import, and if an application has codepaths which do not send requests at all (e.g. globus-cli with completions) we can save over 100ms by deferring the import.

However, the cost of imports must be paid at some point when these modules and packages are needed. As a result, the changelog calls out `requests` explicitly as an area of interest for "very latency-sensitive" applications.

3rd party packages deferred:
- requests
- jwt
- cryptography

stdlib modules deferred:
- sqlite3
- shlex
- inspect

After some study of the dependencies within the stdlib, no other modules or packages of interest were easily identifiable.

The result is a reduction in the time to `import globus_sdk` of over 100ms.

---

For some related background, see also: #559 

The numbers here are only somewhat less extreme. While that showed an improvement from 220ms to 17ms, this change shows a reduction to 76ms on py3.7

### DIFF
--- a/changelog.d/20220513_214357_sirosen_half_lazy_imports.rst
+++ b/changelog.d/20220513_214357_sirosen_half_lazy_imports.rst
@@ -1,0 +1,6 @@
+* ``globus_sdk`` now defers many 3rd party imports until they are needed. This
+  may lead to faster import times, but there may be a new delay when libraries
+  are first used. In particular, ``requests`` can be slow to import (~100ms).
+  Applications which are highly sensitive to operation latencies may prefer to
+  ``import requests`` explicitly at some point during application startup to
+  avoid unexpected performance spikes. (:pr:`NUMBER`)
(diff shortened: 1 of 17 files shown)


aaschaer (approved):
The actual code all looks fine to me.

Part of me wants to say if this matters the SDK shouldn't be written in python, but I do vaguely remember users complaining about the CLI feeling sluggish and this directly improves that without us having to rewrite the SDK.

Importing inside functions is in my head as "bad" but I don't seem to remember why I think that, and none of the reasons I found on a quick google seem to really apply here, other than possibly causing issues with threading?

sirosen:
Closing -- superseded by #571.

PR #562, opened 2022-05-19
Add support for more StorageGatewayPolicies docs

sirosen:
All of the remaining (current) types from gcs-cli have been added here. Most are a matter of copy-paste-replace with minor adjustments to docstrings.

One item in here merits more attention. `OneDriveStoragePolicies` supported "EXPLICIT_NULL" in the gcs-cli. This type of value has been used in globus-cli as well to indicate a value which should be set to `null`/`None` instead of being omitted, but it has never had any representation in the SDK. Now, `utils.missing` is a sentinel value (with type `utils.MissingType` which can be used for this purpose. The field in question defaults to `utils.missing` and is checked for `None` explicitly. The PayloadWrapper helper methods recognize `missing` as an alternative to `None` and omit it.

In the future, the SDK may want to support `missing` as the default for all helper object fields, so that setting `x=None` explicitly always has an effect. This could be used throughout the SDK, even beyond helper objects. However, that change is breaking and would require a major release.


sirosen:
Per discussion, I'm removing the `missing` behavior because it makes the various fields for different document types behave inconsistently. This is likely to be confusing to users down the line.

In our internal tracker, I've written up the possibility of an enhancement for SDK v4.0 which would add the `str | None | missing` behavior to all payload types for all fields. We agree that the behavior it provides is much nicer, so we may want to pay the cost of handling that change. But it would also change many usages, so the cost of such a change is pretty high.

PR #567, opened 2022-05-25
Several test improvements to get module coverage

sirosen:
This is a collection of "one line fixes". Single lines from modules which were not tested, several of which were just `if query_params: ...` or similar.

As a result, this batch of improvements makes the coverage report shorter and more focused.

### DIFF
--- a/tests/unit/helpers/test_auth_flow_managers.py
+++ b/tests/unit/helpers/test_auth_flow_managers.py
@@ -1,10 +1,14 @@
 import base64
 import hashlib
 import os
+from unittest import mock
 
 import pytest
 
 from globus_sdk import GlobusSDKUsageError
+from globus_sdk.services.auth.flow_managers.authorization_code import (
+    GlobusAuthorizationCodeFlowManager,
+)
 from globus_sdk.services.auth.flow_managers.native_app import make_native_app_challenge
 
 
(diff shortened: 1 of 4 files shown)


tests/unit/helpers/test_auth_flow_managers.py
```diff
@@ -52,3 +56,23 @@ def mock_b64encode(b: bytes):
...
+    mock_client.base_url = "https://auth.globus.org/"
+    flow_manager = GlobusAuthorizationCodeFlowManager(
+        mock_client, redirect_uri="https://foo.example.org/authenticate"
+    )
+
+    silly_string = "ANANAS_IS_PINEAPPLE_BUT_BANANE_IS_BANANA"
+
+    authorize_url = flow_manager.get_authorize_url()
+    assert authorize_url.startswith("https://auth.globus.org")
+    assert silly_string not in authorize_url
```
rudyardrichter, line 73: I'm confused what the purpose of this assertion is?

tests/unit/helpers/test_transfer.py
```diff
@@ -103,6 +104,18 @@ def test_transfer_add_item(transfer_data):
...
+    addfields = {"foo": "bar", "bar": "baz"}
+    tdata.add_item(source_path, dest_path, additional_fields=addfields)
+    assert len(tdata["DATA"]) == 4
+    fields_data = tdata["DATA"][3]
+    assert fields_data["DATA_TYPE"] == "transfer_item"
+    assert fields_data["source_path"] == source_path
+    assert fields_data["destination_path"] == dest_path
+    assert not fields_data["recursive"]
+    for k, v in addfields.items():
+        assert fields_data[k] == v
```
rudyardrichter, line 117: Can't resist suggesting a onelinerization; feel free to disregard:
```suggestion
    assert all(fields_data[k] == v for k, v in addfields.items())
```
(similar one below)

tests/unit/helpers/test_auth_flow_managers.py
sirosen: It's not terribly important; just a little habit or quirk of how I write many of my tests.
If I want to assert that calling some callable `foo` with a parameter `bar` causes a desired behavior, I like to assert the converse -- that calling `foo` without `bar` does not exhibit that behavior.

There are a lot of cases where this is sensible and makes the test better. And then there are cases like this one, in which it's hard to imagine the assertion ever failing.

rudyardrichter: I see! Makes sense. Just making sure I'm not missing anything.

rudyardrichter (approved):

tests/unit/helpers/test_transfer.py
sirosen: I liked this too much to ignore. 😄 I hardly ever have use-cases for `any` and `all`.

PR #568, opened 2022-05-26
Add __str__ for ScopeBuilder

sirosen:
These objects are fiddly to inspect. Make it nicer to do so with a friendly `__str__` method.

I also slipped in a check for the new `https` property-scope.


sirosen:
The mypy failure is not related to this PR. That's a new check introduced by mypy v0.960 . I'll resolve that separately, and I think we need to be pinning mypy to avoid these false-negatives.

rudyardrichter (approved):

PR #573, opened 2022-05-27
Add skip_activation_check and add_filter_rule

sirosen:
Add skip_activation_check as a bool arg to both TransferData and DeleteData. `add_filter_rule` is a new method on TransferData for adding exclude filter rules.

Also, start using Literal to type-check `sync_level` string values.

---

Cleaning up old branches, I found this in my local repo. Rebased and improved for our current codebase, I believe this is a set of missing options for these helpers.

### DIFF
--- a/tests/non-pytest/mypy-ignore-tests/transfer_data.py
+++ b/tests/non-pytest/mypy-ignore-tests/transfer_data.py
@@ -1,0 +1,24 @@
+import uuid
+
+from globus_sdk import TransferClient, TransferData
+
+# simple usage, ok
+tc = TransferClient()
+TransferData(tc, "srcep", "destep")
+
+# can set sync level
+TransferData(tc, "srcep", "destep", sync_level=1)
+TransferData(tc, "srcep", "destep", sync_level="exists")
+# unknown int values are allowed
+TransferData(tc, "srcep", "destep", sync_level=100)
+# unknown str values are rejected (Literal)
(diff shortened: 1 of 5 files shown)


tests/non-pytest/mypy-ignore-tests/transfer_data.py
```diff
@@ -0,0 +1,24 @@
...
+tc = TransferClient()
+TransferData(tc, "srcep", "destep")
+
+# can set sync level
+TransferData(tc, "srcep", "destep", sync_level=1)
+TransferData(tc, "srcep", "destep", sync_level="exists")
+# unknown int values are allowed
+TransferData(tc, "srcep", "destep", sync_level=100)
+# unknown str values are rejected (Literal)
+TransferData(tc, "srcep", "destep", sync_level="sizes")  # type: ignore[arg-type]
```
aaschaer, line 15: shouldn't this line be raising a ValueError?

sirosen:
> Are the mypy-ignore-tests actually run, or do we just have mypy look at them to make sure types are behaving correctly?

It is the latter. Those files are explicitly never run -- they're _only_ type-checked.

I'd expect a bunch of errors if you tried to run them (e.g. missing authorization).

tests/non-pytest/mypy-ignore-tests/transfer_data.py
sirosen: It would, you're correct. See below 👇 . I'm going to mark this resolved, but we can discuss more.

aaschaer (approved):

PR #582, opened 2022-06-15
add user credential methods to GCSClient

aaschaer:
One notable addition alongside the CRUD operations is the addition of a `connector_id_to_name` helper to the GCSClient, which I added since the UserCredential response document includes a connector_id that is opaque without some kind of lookup table like this. If there is a better place for this to live I'm happy to move it, but I think we need something like this at least to support the CLI.

### DIFF
--- a/src/globus_sdk/services/gcs/client.py
+++ b/src/globus_sdk/services/gcs/client.py
@@ -5,7 +5,12 @@ from globus_sdk import client, paging, response, scopes, utils
 from globus_sdk._types import UUIDLike
 from globus_sdk.authorizers import GlobusAuthorizer
 
-from .data import CollectionDocument, GCSRoleDocument, StorageGatewayDocument
+from .data import (
+    CollectionDocument,
+    GCSRoleDocument,
+    StorageGatewayDocument,
+    UserCredentialDocument,
+)
 from .errors import GCSAPIError
 from .response import IterableGCSResponse, UnpackingGCSResponse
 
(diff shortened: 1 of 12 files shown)


src/globus_sdk/services/gcs/data/user_credential.py
```diff
@@ -0,0 +1,41 @@
...
+
+class UserCredentialDocument(utils.PayloadWrapper):
+    """
+    Convenience class for constructing a UserCredential document
+    to use as the `data` parameter to `create_user_credential` and
+    `update_user_credential`
+
+    :param DATA_TYPE: Versioned document type.
+    :type DATA_TYPE: str
+
```
sirosen, line 15: We need the full parameter list in here. Even stubby descriptions are fine, but we want full doc on all parameters.

Other datatypes in this dir should have good example text for `additional_fields`.

sirosen (changes requested):
There are two minor changes which I want to see, but this is basically in good shape (I'll continue reading more carefully).

I'm onboard with `GCSClient.connector_id_to_name` since I have no better home for that functionality. Was there an inverse of that operation, name-to-id? Should we consider supporting that so that the CLI can support usage like `--connector-type box` in the future?

src/globus_sdk/services/gcs/client.py
```diff
@@ -97,6 +102,27 @@ def get_gcs_collection_scopes(
...
+            "976cf0cf-78c3-4aab-82d2-7c16adbcc281": "Google Drive",
+            "145812c8-decc-41f1-83cf-bb2a85a2a70b": "POSIX",
+            "7643e831-5f6c-4b47-a07f-8ee90f401d23": "S3",
+            "7e3f3f5e-350c-4717-891a-2f451c24b0d4": "SpectraLogic BlackPearl",
+        }
+
+        if str(connector_id) in connector_dict:
+            return connector_dict[str(connector_id)]
+        else:
+            return "Unknown Connector"
```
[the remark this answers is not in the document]
sirosen: To clarify, I suggested this before thinking that `Optional[str]` might be better for this method. So the above would be `get(...)` with no default if we go that route.

src/globus_sdk/services/gcs/data/user_credential.py
aaschaer: Ack, can't believe I forgot param docs, good catch

tests/functional/gcs/test_user_credential.py
```diff
@@ -0,0 +1,77 @@
+import json
+
+from globus_sdk._testing import get_last_request, load_response
+from globus_sdk.services.gcs import UserCredentialDocument
```
sirosen, line 4: I just noticed this import. To me, this means that we're missing the name exported in `src/globus_sdk/__init__.py`.

It should be sufficient to append it to [the list in the init generator](https://github.com/globus/globus-sdk-python/blob/574bf8e22913d9dc8eb465bc647b8749a1841fd1/src/globus_sdk/_generate_init.py#L106) and then run that script.

sirosen:
I just did another once-over and nothing stands out to me as needing more tweaking. 👍 

I'm not going to rush to do a release, but I think I'd like to do one next week.
That means that if you can finish up the example usage doc in the next week days, we could push it out with the release (so that it would be on the `/stable/` RTD site). I'm not sure if that's significant extra motivation or not -- it's fine if the doc is only in `/latest/` for a while.

PR #583, opened 2022-06-17
remove nonexistant monitor_ongoing transfer scope

aaschaer:
I haven't checked that one cannot actually request this scope, but it is at least completely unused by the Transfer service.


aaschaer:
Just confirmed the scope can't even be requested, so this should definitely be removed.

sirosen (approved):
As discussed in slack, I think this is the best thing to do. I haven't decided if the next version should be a minor or patch for this... But that's fine. I'll noodle on that a bit longer.

PR #584, opened 2022-06-21
Improve testing subpackage with more responses support and inherited metadata

sirosen:
- Use `responses==0.21.0` (current latest) on py3.7+ and update testing readme to note support
- RegisteredResponse objects inherit `metadata` from the containing ResponseSet where appropriate (unspecified)

I consider more complex behaviors for inheritance, e.g. merging. However, if you can load the response object, you can access `parent`, so `x = load_response(...); x.parent.metadata` is valid as a way of getting at all of the data.

Net-net, the inheritance change lets us cut enough lines that the diff is still negative! 🥳

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,11 @@ TEST_REQUIREMENTS = [
     "pytest<7",
     "coverage<7",
     "pytest-xdist<3",
-    "responses==0.17.0",
+    # 'responses' requirement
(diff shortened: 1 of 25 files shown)


rudyardrichter (approved):
This is a nice change, and definitely more elegant. I'm on board with having a little more complexity in the scaffolding to save the effort in writing the hopefully higher-traffic files, namely the actual test data.

PR #585, opened 2022-06-22
Add `_testing.data.transfer.get_endpoint` module

rudyardrichter:
Paste in an example transfer endpoint document from the API docs to use in `_testing`.


sirosen (approved):
Yep, yep, yep!

As always, if it turns out that we need to tweak the data in the future, we'll do so.

I intend to look at a follow-up change to do a test call on `TransferClient.get_endpoint` to test the response.

PR #590, opened 2022-06-30
add guest collection creation example

aaschaer:


sirosen:
It reads cleanly, it's understandable (except for the difficulty of understanding the series of calls needed), and I see no reason to delay. If there are nits to pick, I'm not seeing them.

I'm merging and we'll see it in the stable docs next release. 🚀

PR #594, opened 2022-07-13
Fix creating timer job from TransferData

rudyardrichter:
- Fix `TimerJob.from_transfer_data` to correctly translate JSON fields from TransferData to the action provider format
- Improve timer create job test to catch the above issue

### DIFF
--- a/src/globus_sdk/services/timer/data.py
+++ b/src/globus_sdk/services/timer/data.py
@@ -123,8 +123,29 @@ class TimerJob(PayloadWrapper):
         transfer_action_url = slash_join(
             get_service_url("actions", environment=environment), "transfer/transfer/run"
         )
-        # dict will either convert a `TransferData` object or leave us with a dict here
-        callback_body = {"body": dict(transfer_data)}
+        body = dict(transfer_data)
+        # Check, if `transfer_data` is a dict, whether it's just dictified
+        # TransferData (which we still need to convert) or actually the correct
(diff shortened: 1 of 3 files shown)


src/globus_sdk/services/timer/data.py
```diff
@@ -123,8 +123,26 @@ def from_transfer_data(
         transfer_action_url = slash_join(
             get_service_url("actions", environment=environment), "transfer/transfer/run"
         )
-        # dict will either convert a `TransferData` object or leave us with a dict here
-        callback_body = {"body": dict(transfer_data)}
+        body = dict(transfer_data)
+        # Use `"DATA" in body` as a proxy to check, if `transfer_data` is a
+        # dict, whether it's just dictified TransferData (which we still need
+        # to convert) or actually the correct data to send to Transfer AP
+        if isinstance(transfer_data, TransferData) or "DATA" in body:
```
rudyardrichter, line 130: `body.get("DATA_TYPE") == "transfer"` would be a slightly less hacky way of doing this.

sirosen:
We've been making changes to the AP to harmonize it with the Transfer Task format, and #602 represents the work to make it possible for a TransferData to be passed pretty much verbatim to that. Additionally, it preps us for the removal of `TransferClient` as a parameter to `TransferData`, which will happen in v4.

As a result, this is a path we're not going to go down -- neither in terms of fixing the current method nor in terms of removing or rewriting it.

PR #596, opened 2022-07-20
Allow use of non-default requests_mock in _testing

sirosen:
responses has an interface which allows you to customize the mocking of `requests` with an object. The `_testing` package was all built using the default mocks, which meant that it couldn't accept a `RequestsMock` as an input and gracefully attach to it.
This adds `requests_mock=...` as an argument to all response loading methods.

Also, extend the readme with
- doc on RequestsMock usage
- a short "index" of _testing contents

---

I ran into this as an issue on a project which had a customized RequestsMock for some tests. I was surprised to realize that I couldn't use `_testing` at all.

### DIFF
--- a/src/globus_sdk/_testing/README.rst
+++ b/src/globus_sdk/_testing/README.rst
@@ -34,6 +34,43 @@ guaranteeing that requests are sent to the production hostnames:
         responses.stop()
         responses.reset()
 
+Methods and Classes
+-------------------
+
+Users of ``globus_sdk._testing`` have the following methods and classes
+available:
+
+``get_last_request``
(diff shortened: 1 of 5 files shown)


aaschaer (approved):
Looks good to me!

aaschaer:
Approved after reviewing code, but tests failing should have given me more pause. Not sure what's going on there

sirosen:
> tests failing should have given me more pause. Not sure what's going on there

I vaguely know what's up. The tests I added for this behavior only work on newer versions of `responses`, so they fail on py3.6. I'll figure it out and -- unless it's a very intricate fix -- merge on the strength of your prior review. 🙂

PR #598, opened 2022-07-22
Fix index id key in doc example

rpwagner:
Minor correction to a documentation example. The key for the index ID is `id` in the response for `get_index`.

Also, those short examples are really useful.


sirosen (approved):
Thanks for the catch!

PR #602, opened 2022-08-15
Update TransferData and DeleteData to allow no `submission_id`, then fit this with TimerJob.from_transfer_data

sirosen:
### Motivation

We want to rework `TransferData` and `DeleteData` to not take a `TransferClient` as a parameter. For the near term, this means "allow `None`" and longer term it means "in SDKv4, remove this parameter".

Additionally, `submission_id` is a mismatch between the Transfer Action Provider[^1] and Transfer Task Submission APIs. The Action Provider is visible via Timer, so it's effectively a mismatch between what `TransferData` provides and what Timer requires.

Additionally, and somewhat relatedly, `TransferData` and `DeleteData` will pass explicit `null` values today (which the Transfer API accepts/ignores), but we'd like to remove this. That makes the documents better match user expectations.

[1]: I'm talking about the upcoming changes to the Transfer AP, not the current behavior.

### Implementation

Several fields in TransferData and DeleteData are good fits for the (relatively new) `_set_optstrs` and `_set_optbools` helpers. `sync_level` is actually a good fit for `_set_value` with a custom callback. Converting to these "automatically" handles removal of null values. This also uniformizes behavior between fields like `label` which had special handling to avoid passing `null` and various other fields.

The switch to allow for a `transfer_client` of `None` is relatively simple, and wraps the `get_submission_id` call with an `if None` check.

With this done, `TimerJob` can start rejecting documents which set `submission_id` or `skip_activation_check`, which are not valid for Transfer Action Provider inputs. I chose `ValueError` for this behavior because it seems the most straightforward approach.

As we have discussed on other occasions, this is "technically backwards incompatible", but I'm advocating for it in the "spirit rather than letter of semver".

### DIFF
--- a/changelog.d/20220815_171620_sirosen_update_transfer_data.rst
+++ b/changelog.d/20220815_171620_sirosen_update_transfer_data.rst
@@ -1,0 +1,12 @@
+* Adjust behaviors of ``TransferData`` and ``TimerJob`` to make
+  ``TimerJob.from_transfer_data`` work (:pr:`NUMBER`)
+** ``TransferData`` avoids passing ``null`` for several values when they are
+   omitted, ranging from optional parameters to ``add_item`` to
+   ``skip_activation_check``
+** ``TransferData`` and ``DeleteData`` now support usage in which the
+    ``transfer_client`` parameters is ``None``. In these cases, if
+    ``submission_id`` is omitted, it will be omitted from the document,
+    allowing the creation of a partial task submsision document with no
+    ``submission_id``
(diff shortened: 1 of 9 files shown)


sirosen:
@aaschaer, in accordance with our discussion, I've made two further changes to include in this changeset:

- allow "kwarg-only" usage, as in `TransferData(source_endpoint=foo, destination_endpoint=bar)` by making several args optional
- auto-set `submission_id` in the submit methods if it's missing, making the "client-less" init a good default usage everywhere for the data helpers

aaschaer (approved):
tests all look good to me, think this is ready to go

PR #603, opened 2022-08-19
Add __dir__ to the top-level SDK module

sirosen:
Explicit __dir__ is part of the PEP 562 spec which added module-level __getattr__. We were missing this function, which had limited negative impact other than the fact that `globus_sdk.Tr<TAB>` was not handled nicely by an interactive interpreter.

---

I encountered this while working on a distinct PR and finding that tab-complete was missing. A quick investigation revealed the missing function. `dir()` may also be used by pycharm and other IDEs to inspect the package, so it's worth including our implementation.

### DIFF
--- a/changelog.d/20220819_190305_sirosen_add_module_dirfunc.rst
+++ b/changelog.d/20220819_190305_sirosen_add_module_dirfunc.rst
@@ -1,0 +1,3 @@
+* Implement ``__dir__`` for the lazy importer in ``globus_sdk``. This
+  enables tab completion in the interpreter and other features with
+  rely upon ``dir(globus_sdk)`` (:pr:`NUMBER`)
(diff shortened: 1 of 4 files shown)


sirosen:
I realized just after opening this that it might be writable as `return list(__all__)`. I'm going to double-check the semantics of this.

sirosen:
The PEP itself suggests `sorted(__all__)` as part of a solution, so that's definitely considered safe.
After more reading, it seems that many of the values like `__package__` are considered optional and they'll be dropped by taking that approach. I've therefore cherry-picked `__all__` (itself), `__file__`, and `__path__` as potentially useful attributes, but omitted everything else. The other attributes are still there, but they won't appear in `dir(globus_sdk)`.

The new solution is in place here now.

PR #604, opened 2022-08-19
Add a minimal FlowsClient supporting list_flows

sirosen:
FlowsClient is defined with the attributes and configuration needed to talk to the Globus Flows API. A single paginated method named "list_flows" is supplied, with only some of its (many) parameters provided as keyword arguments.

A data stub in _testing and a single test for the "simple listing case" are included the exercise the code and ensure the resolution of the service URL and other plumbing works as expected.

`_generate_init` is updated with the new module and its attributes.

One notable quirk of this client: the method is paginated, but the pagination parameter is named "pagination_token" rather than the more typical "marker". For this reason, the parameter name is now configurable when instantiating a MarkerPaginator. The paginated usage is not tested as of yet.

### DIFF
--- a/src/globus_sdk/services/flows/client.py
+++ b/src/globus_sdk/services/flows/client.py
@@ -1,9 +1,10 @@
 import logging
 from typing import Any, Callable, Dict, Optional, TypeVar
 
-from globus_sdk import client, paging, response, scopes, utils
+from globus_sdk import client, paging, scopes, utils
 
 from .errors import FlowsAPIError
+from .response import IterableFlowsResponse
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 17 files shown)


src/globus_sdk/services/flows/client.py
```diff
@@ -0,0 +1,85 @@
...
+        the ``flow_starter``, ``flow_administrators`` or ``flow_owner`` roles will be
+        returned.
+        """
+
+        if query_params is None:
+            query_params = {}
+        if filter_role is not None:
+            query_params["filter_role"] = filter_role
+        if filter_fulltext is not None:
+            query_params["filter_role"] = filter_role
```
sirosen, line 81:
```suggestion
            query_params["filter_fulltext"] = filter_fulltext
```

sirosen:
Because Flows is planning to update to make `marker` a supported pagination parameter, I'm testing a change right now to back out the pagination tweaks and to remove the pagination decorator from `list_flows`.

ada-globus (approved):
This is great, and with the last commit to temporarily hold back the pagination support, this is GTG per our 8/23 group review.

PR #607, opened 2022-08-30
Add mapped collection "policies" document helpers

sirosen:
These are additional data-builders for GCS, ported from gcs-cli.


aaschaer (approved):
Looks good to me, are there other storage types that will need these at some point, or all these all of them?

sirosen:
Although there might be other types (today or in the future), these were the only ones I saw defined in gcs-cli, so I'm starting with them. We can always add more.

PR #612, opened 2022-09-20
Remap flows resource server

derek-globus:
* Remapped the flows service resource server from it's uuid to the configured dns
   * This change is ultimately to support flows ingress into the broader globus-cli. The flows resource server registered in auth is the dns `flows.globus.org` rather than it's UUID client id `eec9b274-0c81-4334-bdc2-54e90e689b9a`. See the code snippet below for a demonstration of this

**Code Snippet demonstrating scopes-associated resource server name** *(relevent tokens/secrets removed for obvious reasons)*
```
import globus_sdk
...
client = globus_sdk.ConfidentialAppAuthClient(CLIENT_ID, CLIENT_SECRET)
requested_scopes = 'https://auth.globus.org/scopes/eec9b274-0c81-4334-bdc2-54e90e689b9a/view_flows'
client.oauth2_start_flow('https://localhost', requested_scopes=requested_scopes)
client.oauth2_get_authorize_url()
...
token_response = client.oauth2_exchange_code_for_tokens(AUTH_CODE)
token_response.by_resource_server.keys()
> dict_keys(['flows.globus.org'])
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ dist
 globus_sdk.egg-info
 eggs/
 .eggs/
+.idea
 
 # virtualenv
 .venv/
--- a/changelog.d/20220920_132048_derek_remap_flows_resource_server.rst
+++ b/changelog.d/20220920_132048_derek_remap_flows_resource_server.rst
@@ -1,0 +1,1 @@
+* Remapped the flows service resource server from it's uuid to the configured dns (flows.globus.org). (:pr:`NUMBER`)
--- a/src/globus_sdk/scopes.py
+++ b/src/globus_sdk/scopes.py
@@ -299,8 +299,37 @@ AuthScopes = _AuthScopesBuilder(
 """
 
 
-FlowsScopes = ScopeBuilder(
+class _FlowsScopeBuilder(ScopeBuilder):
+    """
+    The Flows Service breaks the scopes/resource server convention:
+      its resource server is a domain name but its scopes are built around the client id
+    Given that there isn't a simple way to support this more generally
+      (and we shouldn't encourage supporting this more generally), this class serves to
+      build out the scopes accurately specifically for Flows
+    """
+
+    def __init__(
+        self,
+        domain_name: str,
+        client_id: str,
+        known_scopes: Union[List[str], str, None] = None,
+        known_url_scopes: Union[List[str], str, None] = None,
+    ) -> None:
+        self._client_id = client_id
+        super().__init__(
+            domain_name, known_scopes=known_scopes, known_url_scopes=known_url_scopes
+        )
+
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -7,6 +7,7 @@ dist
 globus_sdk.egg-info
 eggs/
 .eggs/
+.idea
```
aaschaer, line 10: part of me wants to say this should be its own commit or something, but its pretty innocuous so this is probably fine.

aaschaer (commented):
One concern would be if this could be a breaking change since the `FlowsClient.resource_server` value is changing, and that value could be used for token storage or something.

CLI isn't using that value yet, and I don't think we have enough users using this in the SDK that it's particularly likely to break anyone, so its probably fine?

derek-globus:
> One concern would be if this could be a breaking change since the FlowsClient.resource_server value is changing, and that value could be used for token storage or something.
CLI isn't using that value yet, and I don't think we have enough users using this in the SDK that it's particularly likely to break anyone, so its probably fine?

Fair callout. This PR is actually coming from a [CLI PR](https://github.com/globus/globus-cli/pull/681) to start using the `FlowsScope.resource_server` field. In that PR I've hardcoded the flows resource server (breaking convention from the other clients (eg: [cli timer client instantiation](https://github.com/globus/globus-cli/blob/main/src/globus_cli/login_manager/manager.py#L278))).

With that in mind, this change could be classified as breaking (and have it's release delayed until appropriate) without blocking the underlying CLI task. It'll just add some code smell to CLI until this does get merged.

My inclination is to classify this as a bugfix as auth doesn't associate the flows client id as the FlowsScope resource server so this fixes that incorrect association but it's certainly possible customers are relying on this as the source of reference for some other usage of flows client id.

aaschaer:
> With that in mind, this change could be classified as breaking (and have it's release delayed until appropriate) without blocking the underlying CLI task. It'll just add some code smell to CLI until this does get merged.

Ah just saw the CLI PR! I think I would prefer avoiding the CLI code smell since the chance of breakage seems pretty low. @sirosen do you agree?

aaschaer (approved):
`_FlowsScopeBuilder` looks good, pending results of any auth tinkering I think this is good to go

derek-globus:
>I have to apologize for not raising this earlier, but I have one concern which we should check, but which I do not think we should treat as blocking:
>
>Does this apply equally well in all environments?
I worry that the resource_server string might not be consistent outside of production. If there are environments where it doesn't work because resource_server is set to the client ID, we should normalize them to match production.

Just checked & while I found some timer-related discrepencies, I can attest that the following two invariants hold true for all flows environments:
1. fqdn is registered resource server
2.client id-composed url is scope string

PR #621, opened 2022-10-12
Add pagination back to FlowsClient.list_flows

sirosen:
The bulk of this change is a response data generator which produces test data. Although it would be nice to put this set of responses into the `_testing` data, it's not 100% clear *where* it should go. For now, this changeset punts on that question and puts the test data into the test module.

Now that Flows is supporting `marker`, the most barebones application of the pagination decorator for marker pagination is sufficient.

### DIFF
--- a/changelog.d/20221012_190435_sirosen_paginated_flows_list.rst
+++ b/changelog.d/20221012_190435_sirosen_paginated_flows_list.rst
@@ -1,0 +1,1 @@
+* ``FlowsClient.list_flows`` now supports pagination (:pr:`NUMBER`)
--- a/tests/functional/flows/test_list_flows.py
+++ b/tests/functional/flows/test_list_flows.py
@@ -18,7 +18,10 @@ OWNER_ID = "e061df5a-b7b9-4578-a73b-6d4a4edfd66e"
 
 def generate_hello_world_example_flow(n: int) -> Dict[str, Any]:
     flow_id = str(uuid.UUID(int=n))
-    base_time = datetime.datetime.fromisoformat("2021-10-18T19:19:35.967289+00:00")
+    # base_time = datetime.datetime.fromisoformat("2021-10-18T19:19:35.967289+00:00")
+    base_time = datetime.datetime(
+        2021, 10, 18, 19, 19, 35, 967289, tzinfo=datetime.timezone.utc
+    )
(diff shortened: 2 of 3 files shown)


sirosen:
There were a few changes here after the first push because I made some mistakes in terms of support for older python versions.

changelog.d/20221012_190435_sirosen_paginated_flows_list.rst
```diff
@@ -0,0 +1 @@
+* ``FlowsClient.list_flows`` now supports pagination (:pr:`NUMBER`)
```
kurtmckee, line 1: Just to confirm, a hook runs on GH during merge to swap out `` :pr:`NUMBER` ``, right?

tests/functional/flows/test_list_flows.py
```diff
@@ -1,8 +1,135 @@
...
+
+OWNER_ID = "e061df5a-b7b9-4578-a73b-6d4a4edfd66e"
+
+
+def generate_hello_world_example_flow(n: int) -> Dict[str, Any]:
+    flow_id = str(uuid.UUID(int=n))
+    # base_time = datetime.datetime.fromisoformat("2021-10-18T19:19:35.967289+00:00")
+    base_time = datetime.datetime(
+        2021, 10, 18, 19, 19, 35, 967289, tzinfo=datetime.timezone.utc
+    )
```
kurtmckee, line 24: Is this comment here to suggest a future code swap with a later version of Python? Would this be a candidate for the `if sys.version_info >= (3, x)` that allows pyupgrade to auto-remove this compatibility call at some point in the future?

sirosen: Ooh, yes, I had forgotten about that trick; I'll make that change.

The comment was left there to suggest a swap in the future. I originally pulled the string from an API response (and then wanted to do datetime math on it), and parsed it in this way.

changelog.d/20221012_190435_sirosen_paginated_flows_list.rst
sirosen: Yes, it runs after any merge and replaces `NUMBER` with the source PR. The workflow for that is here:
https://github.com/globus/globus-sdk-python/blob/main/.github/workflows/update_pr_references.yaml

kurtmckee (approved):

Issue #14, opened 2016-04-03
How much error handling should the SDK do before requests are sent to the Transfer API?


sirosen:
@jimPruyne mentioned in #13 that he thinks (or at least thought at one time) that the SDK should catch `label=''` before a task is submitted to the Transfer API.
This is part of a more general class of feature, which is handling for invalid requests to the Transfer API prior to API interactions.

I think this *only* applies to Transfer, since it's our oldest and most mature API, so others should just rely on non-200 responses for now.
Even if you don't agree, please keep this particular issue focused only on our handling of the Transfer API.
Otherwise, there are many rabbit holes and we might not be able to drive this to resolution.

I actually created an error class for this exact type of error handling (perhaps poorly named as `InvalidDocumentBodyError`), but it's currently unused because I stripped out the code that was triggering it.

In the particular case of `label=''` for a Task, I don't think the SDK should do validation.
It's a fairly subtle and non-generalizable error specific to task submission, and I don't see that a single request to the API matters much.
However, it becomes more of a concern if we think that a lot of work might be done between assembling the Task document and task submission.
Causing an error when `label=''`, or perhaps when paths do or do not have requisite trailing slashes, would, in that case, move the failure to be earlier.
I'm not really convinced that's a likely scenario and important use case on which to focus.

There are some potential areas for this type of handling that make a bit more sense client-side. That could be because the SDK has a better idea of what went wrong and how to fix it in an SDKish way, or because it's not the same kind of error (or even an error at all) when it arrives at the API.
Some examples, in no particular order,

- POST body has wrong `DATA_TYPE` for the API endpoint, possibly recommend using a relevant `make_X_data` helper
- Transfer Task is missing submission ID, also recommend using `make_X_data` helper
- Non-UUID string supplied in a field that must be a UUID (e.x. endpoint ID). This often results in a 404 from the API because the requested resource doesn't exist, but the SDK could treat as "invalid" and throw a different error.
- Similar, attempt to use canonical endpoint name for a call that only supports endpoint IDs

Although there will always be exceptions to our rules, we shouldn't be handling these decisions on a pure case-by-case basis.
There's a little bit of bleed-over into this issue from the fact that the Transfer API gives 404s for certain requests which are definitively "wrong", like `GET /endpoint/nonuuidstring`, whereas maybe those should result in 400 Bad Request.

[...]

PR #624, opened 2022-10-18
Convert to 'import typing as t' style

sirosen:
- apply `flake8-typing-as-t` in pre-commit
- make everything pass

---

I considered using this as an opportunity to learn about new automated refactoring tools, but clever search-and-replace worked too well for it to be worth it to me. 😄


kurtmckee (approved):
If unit tests pass, I'm going to assume that everything was imported correctly! (Got to about file #14 and thought "Nope. It's fine.") :smile:

PR #628, opened 2022-10-27
Support python 3.11

sirosen:
In the changelog, this is called out as "official support" since the code all worked on 3.11 prior to this change.

Places touched:
- CI testing job on linux (all pythons)
- CI testing on windows (latest python only)
- CI testing on macos (latest python only)
- readthedocs config
- pypi classifiers
- tox testenvs

CI builds should be examined to see what might fail in particular. Local testing under pyenv+tox works, but it's always possible that something will be different in the CI environment.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -31,14 +31,14 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11"]
         # we do not want a large number of windows and macos builds, so
         # enumerate them explicitly
         include:
           - os: windows-latest
-            python-version: "3.10"
+            python-version: "3.11"
           - os: macos-latest
-            python-version: "3.10"
+            python-version: "3.11"
     name: "test py${{ matrix.python-version }} on ${{ matrix.os }} "
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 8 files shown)


kurtmckee (approved):
Looks good to me!

I see the Windows CI is encountering that most wonderful of all issues, "the `shutil.rmtree()` that fails for no discernible reason". In the past I "fixed" this by sleeping for .1 seconds in an `except` clause, at which point trying again would succeed.

That kind of fix is only acceptable in a production environment, though, so you might just try re-running the test or confirming that everything touching that sqlite file has released / closed. :wink:

sirosen:
I think I'm going to explicitly close the sqlite connections created by the tests. I'm somewhat surprised this hasn't been an issue before now -- we're opening those connections and then trusting GC to clean them up. I might even want to add a `close()` method to the storage adapter if we don't have one already.

sirosen:
Whew! That was more of a rewrite to the tests than I expected, but I've just added `SQLiteAdapter.close()` and made sure tests call it. 🤞 that this fixes this build _and_ makes the tests more stable in the long run.

kurtmckee (approved):
On Linux (and probably Mac), opening a file for read/write doesn't prevent the file from being deleted. Try this:

```
# Terminal 1
f = open("abc", "w+")

# Terminal 2
rm abc

# Back to terminal 1
f.write("not backed by the original file anymore? no problem!")
f.close()

# And in terminal 2...
ls abc  # finds nothing!
```

On Windows, though, an open file cannot be deleted so you're depending on garbage collection. Note that PyPy and CPython [garbage collect differently](https://github.com/python-poetry/poetry-core/pull/122), so it's best to explicitly close resources always! :door: :partying_face:

sirosen:
Yeah, IMO this was a resource leak and we should probably double-check SQLiteAdapter usage sites.

It makes me want to add a context manager interface which closes on `__exit__` as well, as a future change.

PR #632, opened 2022-10-31
Make the `has-changelog` action reusable

kurtmckee:
Exactly what it says on the tin.

### DIFF
--- a/.github/workflows/has_changelog.yaml
+++ b/.github/workflows/has_changelog.yaml
@@ -1,5 +1,6 @@
 name: changelog
 on:
+  workflow_call:
   pull_request:
     types:
       - labeled


.github/workflows/has_changelog.yaml
```diff
@@ -1,5 +1,6 @@
 name: changelog
 on:
+  workflow_call:
```
derek-globus, line 3: Is this intended to point `workflow_call` at null?

```
{
  "on": {
     "workflow_call": Null,
     "pull_request": {
       ...
}
```

kurtmckee: Yes, that appears to match with [GitHub's documentation](https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow), which offers this fragment:

```yaml
on:
  workflow_call
```

kurtmckee: Notably, this syntax worked after my initial effort was rejected by a pre-commit hook for checking JSON schemas...

derek-globus: Weird, okay

derek-globus (approved):

.github/workflows/has_changelog.yaml
kurtmckee: After merging this and testing in another repo, this change is confirmed. I'm marking this resolved.

sirosen: I believe that error would have been from `check-jsonschema` (of which I am the author 😄 ).

If you _want_ (no pressure), share the error you got from valid syntax with me [on the issue tracker](https://github.com/python-jsonschema/check-jsonschema/issues) for that project. I'll treat it as an "upstream schema issue" and see if I can find a way to mark it valid.

kurtmckee: I don't think it _was_ valid syntax! It seems to have worked a-okay. :+1:

PR #630, opened 2022-10-29
Streamline testing of Flow CRUD operations

sirosen:
1. Move multiple tests into a single file, consolidating simple crud tests to avoid modules with single test functions
2. Introduce a new object to the `globus_sdk._testing` models: ResponseList. This object allows a series of responses to be registered to a ResponseSet and named under a given case name.
3. Apply ResponseList to separate the data requirements for paginated flow list testing from the test itself

---

`ResponseList` is imperfect, but overall I think it's better than not introducing a new type.
I tried a version in which `ResponseList` was not included, and `list[RegisteredResponse]` lists were used instead. The results were much harder to read because more cases needed explicit handling.

"XList" it's not a great name alongside "XSet" when the difference between the two does not align with list vs set semantics. So I want to rename things in the future, but that doesn't necessarily mean `ResponseList` gets renamed. Maybe `ResponseSet` should be `ResponseMap` or `NamedResponseSet`.

One design note is that I decided against allowing recursive ResponseSets (😵), as it keeps things simpler and reduces the number of possible ways things can be nested.

PR #634, opened 2022-10-31
Begin test restructuring into module-per-method

sirosen:
This is a two-step refactor to our tests to accord with our conversation today about the thread which started in #630 .

1. Move all functional client tests into a `services/` subdir

This makes the structure of the tests mirror the structure of the source more closely, so it hopefully makes the test contents more intuitive to navigate.

2. Split up the GroupsClient tests into per-method tests

There's more refactoring I'd like to do here to avoid using the older fixture loader, but for now this just does the update we talked about into a test module per method.

I'll also note that there's a `GroupsClient` and `GroupsManager`, and we should probably split things further to separate tests of the two, but I haven't done this yet.

---

You may need to `make clean` before tests pass. I found that moving the test modules around resulted in old -- but not invalidated -- pycache directories.


kurtmckee (approved):
Whew! Thank goodness this was mostly a bunch of renames! :wink:

sirosen:
Indeed, it's basically all renames!

I'm going ahead with the merge pretty fast because I'd like to add another change on top of this.

PR #635, opened 2022-10-31
Convert GCS get_collection_list tests to new style

sirosen:
This change moves the tests for this method into a dedicated module, and moves the test data from the older `fixture_data/` dir to the `globus_sdk._testing.data` subpackage.

The oddity here is that get_collection_list is being used to test error parsing for a GCS 403. It's not immediately clear where this should go, so it's bundled with these other tests for now. (Perhaps this test is good but should be augmented by additional unit tests.)

---

This is how I want to churn through the whole testsuite. We have test data in an older format (and therefore not available via `_testing`), and we have a lot of test cases in what I'll also call our "older" (as of today, hah!) format. It seems like we can go through both of them to find missing pieces.


src/globus_sdk/_testing/data/globus_connect_server/get_collection_list.py
```diff
@@ -0,0 +1,70 @@
+import uuid
+
+from globus_sdk._testing.models import RegisteredResponse, ResponseSet
+
+identity_id = str(uuid.uuid1())
+collection_ids = [str(uuid.uuid1()), str(uuid.uuid1())]
+gateway_ids = [str(uuid.uuid1()), str(uuid.uuid1())]
```
kurtmckee, line 7: `uuid.uuid1()` has serious collision problems. I recommend always using `uuid.uuid4()` to help ensure that values are unique.

```pycon
>>> import uuid
>>> 
>>> print("\n".join(str(uuid.uuid1()) for _ in range(5)))
c35c7812-59ee-11ed-8f60-67bc0b767b36
c35c7813-59ee-11ed-8f60-67bc0b767b36
c35c7814-59ee-11ed-8f60-67bc0b767b36
c35c7815-59ee-11ed-8f60-67bc0b767b36
c35c7816-59ee-11ed-8f60-67bc0b767b36
>>> 
>>> print("\n".join(str(uuid.uuid4()) for _ in range(5)))
00d6ceb0-0391-44dd-90aa-7fd9a1578299
f21a6479-5946-4b59-b82a-c1bda912fcf4
941b4f1b-f8e5-4646-a7d1-f2a07fb235b9
d07ee957-284d-4d83-aeb6-e4cd5bea847f
bde3dec8-f609-4c25-be17-217a042d5e0c
```

I'd be willing to bet that `collection_ids` and `gateway_ids` don't contain unique values when this runs.

```suggestion
identity_id = str(uuid.uuid4())
collection_ids = [str(uuid.uuid4()), str(uuid.uuid4())]
gateway_ids = [str(uuid.uuid4()), str(uuid.uuid4())]
```

tests/functional/services/gcs/test_get_collection_list.py
```diff
@@ -0,0 +1,68 @@
...
+    # sanity check some fields
+    assert res["DATA_TYPE"] == "result#1.0.0"
+    for index, item in enumerate(res):
+        assert item["DATA_TYPE"] == "collection#1.0.0"
+        assert item["id"] == meta["collection_ids"][index]
+        assert item["storage_gateway_id"] == meta["gateway_ids"][index]
+        assert item["display_name"] == meta["display_names"][index]
+
+
+def test_get_collection_list_include_param(client):
```
kurtmckee, line 22: It looks like this could be broken into four tests, two of which are parametrized:

1. Test without an `include` parameter
2. Test the `include` parameter
3. Test the `mapped_collection_id`
4. Test the `filter` parameter

Bringing the three `include` parameters and expected values to the top would help highlight what's actually getting checked:

```python
@pytest.mark.parametrize(
    "include, expected",
    (
        ("foo", "foo"),
        ("foo,bar", "foo,bar"),
        (("foo", "bar"), "foo,bar"),
    )
)
def test_get_collection_list_include_param(client, include, expected):
    load_response(client.get_collection_list)
    client.get_collection_list(include=include)
    req = get_last_request()
    assert "include" in req.params
    assert req.params["include"] == expected

**Review by kurtmckee — CHANGES_REQUESTED** [review]

File: src/globus_sdk/_testing/data/globus_connect_server/get_collection_list.py (hunk 1 above)
Comment (sirosen) [reply src/globus_sdk/_testing/data/globus_connect_server/get_collection_list.py:7]: Ah, right, good catch! UUID1 is local timestamp based, and the resolution isn't good enough to guarantee uniqueness at this level.

I think of UUID4 as "slower", but it's so marginal that you need a very large number of calls to notice.

Just because I think it's interesting...
<details><summary>UUID v1 vs v4 vs v5 timing</summary>

This is what I ran, expecting v5 to be significantly faster. I was surprised to see that it's the slowest by far.
```
$ python -m timeit 'import uuid; namespace = uuid.uuid4(); x = [uuid.uuid5(namespace, "foo") for _ in range(100)]'
500 loops, best of 5: 496 usec per loop
$ python -m timeit 'import uuid; x = [uuid.uuid4() for _ in range(100)]'
500 loops, best of 5: 377 usec per loop
$ python -m timeit 'import uuid; x = [uuid.uuid1() for _ in range(100)]'
1000 loops, best of 5: 302 usec per loop
```

So our speedups are...

|      | v1      | v4      | v5    |
|-----|---------|---------|---------|
| v1 | 100% | 20%   | 39%  |
| v4 |  -        | 100% | 24%  |
| v5 |  -        | -         | 100% |


Of course, the scale here is microseconds! So it doesn't really matter. 😁 

</details>

kurtmckee: I wasn't aware of `uuid5()`, but it looks like v5 uses SHA-1 under the hood, which helps explain its "slowness". :grinning:

kurtmckee (approved):

tests/functional/services/gcs/test_get_collection_list.py
sirosen: I copied your sample and modified it very minimally for that case, and fleshed out the other two tests.

The only significant difference between what you suggested and what I did was that I didn't include a separate test for `"include" not in req.params`. I instead made that a case in which `include_param, expected = None, None`. I know it puts conditional logic into the test, but it feels to me like the "same thing" being checked.

(Also, I saw you approved, but I just pushed this, so you may want to voice an opinion at this point. 🙂 )

sirosen:
I pushed the change, and then realized I had accidentally dropped a case (`assert "include" not in req.params`). I then amended to add that but didn't check my work. That broke the tests, and I amended again to fix the addition.

kurtmckee (approved):

tests/functional/services/gcs/test_get_collection_list.py
kurtmckee: This looks good to me!

PR #636, opened 2022-10-31
Add stubs for missing Auth client tests

sirosen:
For Globus Auth client classes, identify untested methods and list each one in a distinct module with an xfail test which raises a NotImplementedError.

Include a "stubber" script which was used to generate these modules.

---

This is a first step towards enumerating all of the existing untested methods.

The stubber is a script containing what I used (I put it in a shell function, but it's the same thing).
To get which lines to inspect, I did a full `tox` run, pulled the lines from the coverage report which list client modules, and then started going chunk-by-chunk until I found missing tests.
The list is long enough that I didn't want to do everything all at once, but the AuthClient classes provide a small-to-moderate sized dataset to start with.

One (annoyingish) detail is that we need `__init__.py` files to ensure that modules with the same name are treated as distinct tests by pytest discovery, but once those were in place this reports out as desired:
```
596 passed, 1 skipped, 10 xfailed
```

### DIFF
--- a/tests/stub-testmod.sh
+++ b/tests/stub-testmod.sh
@@ -1,0 +1,9 @@
+name="$1"
+cat > "test_$name.py" <<EOH
+import pytest
+
+
+@pytest.mark.xfail
+def test_$name():
+    raise NotImplementedError
+EOH
--- a/tests/functional/services/auth/base/test_oauth2_refresh_token.py
+++ b/tests/functional/services/auth/base/test_oauth2_refresh_token.py
@@ -1,0 +1,6 @@
+import pytest
+
+
+@pytest.mark.xfail
+def test_oauth2_refresh_token():
+    raise NotImplementedError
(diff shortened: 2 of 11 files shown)


tests/stub-testmod.sh
```diff
@@ -0,0 +1,9 @@
+name="$1"
+cat > "test_$name.py" <<EOH
```
kurtmckee, line 2: :heart: `EOH` as the identifier here

kurtmckee (approved):
Thank you very much for explaining the value of including an `__init__.py` file in the test directories! I knew that pytest could collect tests absent a dunder-init, but wasn't sure if there were benefits or drawbacks to its inclusion. Great info!

derek-globus (commented):
Just to make sure I'm understanding correctly here; the script is basically just a missing-test template right?

Like there's no automated discovery piece?

sirosen:
> Just to make sure I'm understanding correctly here; the script is basically just a missing-test template right?
> 
> Like there's not intended to be any automated discovery piece?

Correct. I had spoken in the past about automating more of this, but after starting the work I decided it wasn't worth it.

PR #637, opened 2022-11-01
Add missing test stubs

sirosen:
This adds all of the stubs for missing tests in the SDK for client methods.

### DIFF
--- a/tests/functional/services/transfer/endpoint_manager/test_endpoint_manager_acl_list.py
+++ b/tests/functional/services/transfer/endpoint_manager/test_endpoint_manager_acl_list.py
@@ -1,0 +1,6 @@
+import pytest
+
+
+@pytest.mark.xfail
+def test_endpoint_manager_acl_list():
+    raise NotImplementedError
--- a/tests/functional/services/groups/manager/test_accept_invite.py
+++ b/tests/functional/services/groups/manager/test_accept_invite.py
@@ -1,0 +1,6 @@
+import pytest
+
+
+@pytest.mark.xfail
+def test_accept_invite():
+    raise NotImplementedError
--- a/tests/functional/services/groups/manager/test_decline_invite.py
+++ b/tests/functional/services/groups/manager/test_decline_invite.py
@@ -1,0 +1,6 @@
+import pytest
+
+
+@pytest.mark.xfail
+def test_decline_invite():
+    raise NotImplementedError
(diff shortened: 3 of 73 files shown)


tests/functional/services/transfer/endpoint_manager/test_endpoint_manager_acl_list.py
```diff
@@ -0,0 +1,6 @@
+import pytest
+
+
+@pytest.mark.xfail
+def test_endpoint_manager_acl_list():
```
kurtmckee, line 5: Just to confirm:

* `endpoint_manager_` should be included in the filename and test name (here, and in other files in this directory)? Or should this be `test_acl_list()` or similar?
* The lack of an active verb in this particular name (and a few others) is anticipated, i.e. there's no "get" verb that should be here?

aaschaer (approved):
I like it! I'm not even sure the transfer ones need to be further split up, its a lot, but not unmanageable since having an explicit pattern makes things easy to find even when there are 70+ methods in a client.

On the off chance that some of these did actually have tests hiding somewhere, I think that could be discovered pretty quickly when going through to implement these?

kurtmckee (approved):

derek-globus (approved):

tests/functional/services/transfer/endpoint_manager/test_endpoint_manager_acl_list.py
sirosen: These are intentional, but the test function name is something we didn't specify.

The pattern we agreed upon (sorry that you missed the relevant meeting!) was that we'd have a module named by the _method name_ on the relevant client.
So the method is defined as `TransferClient.endpoint_manager_acl_list`, which determines the module name without any `get_` prefix as `test_endpoint_manager_acl_list`.

Absent any decision about what to name the functions inside, I decided to keep them matching the method names. I don't think it matters _too much_ since these are mostly here to flag the missing-ness of any tests in the modules.

sirosen:
> On the off chance that some of these did actually have tests hiding somewhere, I think that could be discovered pretty quickly when going through to implement these?

We don't have to worry about that, actually. I only produced these stubs for methods which were missing from the coverage reports entirely.

We _do_ now have a kind of "inverse discovery" problem to consider, finding the existing tests which don't match our desired pattern. But we can solve that problem later, I think.

sirosen:
> Edit: just noticed the `endpoint_manager` directory, I'm fine leaving that as is given how many methods fall under there, but also think a flat structure might be okay?

Just to say this before I merge -- I'm 100% open to flattening the TransferClient tests into a single directory. But let's save that for later. 😄

PR #641, opened 2022-11-03
Allow scope names to be different from their value

derek-globus:
* Scope Names can be set explicitly in a `ScopeBuilder`
  *  Idea behind this is that scope values (whether they should or shouldn't be) can have a different human-understandable name than their `scope_value`.
  *  With this change, a ScopeBuilder instantiation can attach a more human understandable `scope_name` to a less human understandable `scope_value` to be formatted into either a urn or url
* Introduced `ScopeBuilder.scope_names` property
  *   Moved _sphinxext.py to use this instead of ScopeBuilder internal properties
* Fixed SpecificFlowClient scope string
  * Proof that this is accurate:
    * The scope string has it's operation string replaced from `-` to `_` [here](https://github.com/globusonline/flows/blob/main/globus/flows/scope_utils.py#L80)
    * An example get flow truncated call
    ```
    % globus-automate flow get f73633d2-9445-4558-8228-977f79d71971
    {
      ...trunc...
      "globus_auth_scope": "https://auth.globus.org/scopes/f73633d2-9445-4558-8228-977f79d71971/flow_f73633d2_9445_4558_8228_977f79d71971_user",
      ...trunc...
    }
    ```

### Testing
* Added unit test for re-named scope strings
* Tested out the new SpecificFlowClient usage manually:
```
SpecificFlowClient("7edaaa8a-72cd-4491-b7bd-b1f6160f44db").scopes.user
'https://auth.globus.org/scopes/7edaaa8a-72cd-4491-b7bd-b1f6160f44db/flow_7edaaa8a_72cd_4491_b7bd_b1f6160f44db_user'
```
^ this scope string is verified accurate above

### DIFF
--- a/changelog.d/20221103_144520_derek_fix_specific_flow_scope_string.rst
+++ b/changelog.d/20221103_144520_derek_fix_specific_flow_scope_string.rst
@@ -1,0 +1,3 @@
+
+* Scope Names (used to set ``ScopeBuilder`` attributes) can be set explicitly
+* Fixed SpecificFlowClient scope string
--- a/changelog.d/20221103_144520_derek_fix_specific_flow_scope_string.rst
+++ b/changelog.d/20221103_144520_derek_fix_specific_flow_scope_string.rst
@@ -1,3 +1,4 @@
 
-* Scope Names (used to set ``ScopeBuilder`` attributes) can be set explicitly
-* Fixed SpecificFlowClient scope string
+* Scope Names can be set explicitly in a ``ScopeBuilder`` (:pr:`NUMBER`)
+* Introduced ``ScopeBuilder.scope_names`` property (:pr:`NUMBER`)
+* Fixed SpecificFlowClient scope string (:pr:`NUMBER`)
--- a/src/globus_sdk/scopes.py
+++ b/src/globus_sdk/scopes.py
@@ -147,7 +147,7 @@ class ScopeBuilder:
         None => {}
         "my-str" => {"my-str": "my-str"}
         ["my-list"] => {"my-list": "my-list"}
-        ("my-tuple-key": "my-tuple-val") => {"my-tuple-key": "my-tuple-val"}
+        ("my-tuple-key", "my-tuple-val") => {"my-tuple-key": "my-tuple-val"}
         """
         if items is None:
             return {}
@@ -316,7 +316,6 @@ AuthScopes = _AuthScopesBuilder(
 """Globus Auth scopes.
 
 .. listknownscopes:: globus_sdk.scopes.AuthScopes
-    add_scopes=openid,email,profile
     example_scope=view_identity_set
 """
 
--- a/src/globus_sdk/scopes.py
+++ b/src/globus_sdk/scopes.py
@@ -1,7 +1,5 @@
 import typing as t
 
-from globus_sdk import utils
-
 # this type alias is meant for internal use, which is why it's named with an underscore
 _ScopeCollectionType = t.Union[
     str,
@@ -116,29 +114,28 @@ class ScopeBuilder:
         self,
         resource_server: str,
         *,
-        known_scopes: t.Union[t.List[str], str, None] = None,
-        known_url_scopes: t.Union[t.List[str], str, None] = None,
+        known_scopes: t.Union[
+            t.List[t.Union[str, t.Tuple[str, str]]], str, None
+        ] = None,
+        known_url_scopes: t.Union[
+            t.List[t.Union[str, t.Tuple[str, str]]], str, None
+        ] = None,
     ) -> None:
         self.resource_server = resource_server
-        self._known_scopes = (
-            list(utils.safe_strseq_iter(known_scopes))
-            if known_scopes is not None
-            else []
-        )
-        self._known_url_scopes = (
-            list(utils.safe_strseq_iter(known_url_scopes))
-            if known_url_scopes is not None
-            else []
-        )
+
+        known_scopes_dict = self._consolidate_collections(known_scopes)
+        known_url_scopes_dict = self._consolidate_collections(known_url_scopes)
+
+        self._known_scopes = list(known_scopes_dict.values())
+        self._known_url_scopes = list(known_url_scopes_dict.values())
+
         self._known_scope_names: t.List[str] = []
-        if self._known_scopes:
-            for scope_name in self._known_scopes:
-                self._known_scope_names.append(scope_name)
-                setattr(self, scope_name, self.urn_scope_string(scope_name))
-        if self._known_url_scopes:
-            for scope_name in self._known_url_scopes:
-                self._known_scope_names.append(scope_name)
-                setattr(self, scope_name, self.url_scope_string(scope_name))
+        for scope_name, scope_val in known_scopes_dict.items():
+            self._known_scope_names.append(scope_name)
+            setattr(self, scope_name, self.urn_scope_string(scope_val))
+        for scope_name, scope_val in known_url_scopes_dict.items():
+            self._known_scope_names.append(scope_name)
+            setattr(self, scope_name, self.url_scope_string(scope_val))
 
     # custom __getattr__ instructs `mypy` that unknown attributes of a ScopeBuilder are
     # of type `str`, allowing for dynamic attribute names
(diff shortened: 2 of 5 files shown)


src/globus_sdk/services/flows/client.py
```diff
@@ -312,9 +312,10 @@ def __init__(
             transport_params=transport_params,
         )
         self._flow_id = flow_id
+        user_scope_value = f"flow_{str(flow_id).replace('-', '_')}_user"
         self.scopes = ScopeBuilder(
             resource_server=str(self._flow_id),
-            known_url_scopes=[f"flow_{flow_id}_user"],
+            known_url_scopes=[("user", user_scope_value)],
```
NickolausDS, line 318: I see the purpose for specifying the resource server and user_scope_value, but I don't think I understand why the `"user"` value is now being passed for `known_url_scopes`. Does that have a special meaning?

NickolausDS: Explained to me in chat, how "user" here is a reference for how to refer to the scope later, for instance `self.scopes.user`. Thanks for explaining @derek-globus!

NickolausDS (approved):

src/globus_sdk/services/flows/client.py
derek-globus: Yep, specifically because `self.scopes.user` is easier than `self.scopes.flow_8705ad07_e5a0_4645_b2c7_80167bf84656_user`

sirosen (changes requested):
The core of the change is good, but there are some minor tweaks I want here. Specifically, after #643, I think we can fixup the sphinx extension to avoid using some of the internal names which you've preserved.
I'm testing a fix to the sphinx extension now:
```diff
diff --git a/src/globus_sdk/_sphinxext.py b/src/globus_sdk/_sphinxext.py
index c92d894..243bb51 100644
--- a/src/globus_sdk/_sphinxext.py
+++ b/src/globus_sdk/_sphinxext.py
@@ -13,7 +13,7 @@ from sphinx.util.nodes import nested_parse_with_titles

 def _extract_known_scopes(scope_builder_name):
     sb = locate(scope_builder_name)
-    return sb._known_scopes
+    return sb._known_scope_names
```

It looks to me like this works.

(NB: I'm aware that there are no unit tests which enforce that the sphinx extension code handles objects correctly; maybe we should add some in the future.)

changelog.d/20221103_144520_derek_fix_specific_flow_scope_string.rst
```diff
@@ -0,0 +1,3 @@
+
+* Scope Names (used to set ``ScopeBuilder`` attributes) can be set explicitly
+* Fixed SpecificFlowClient scope string
```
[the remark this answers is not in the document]
kurtmckee: Bonus: Your local branch is not instantly out-of-sync when you open a PR. :grinning:

kurtmckee (approved):

changelog.d/20221103_144520_derek_fix_specific_flow_scope_string.rst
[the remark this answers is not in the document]
derek-globus: Oh whoops. Didn't mean to remove that.
New rev

src/globus_sdk/scopes.py
```diff
@@ -116,29 +114,28 @@ def __init__(
...
-            if known_url_scopes is not None
-            else []
-        )
+
+        known_scopes_dict = self._consolidate_collections(known_scopes)
+        known_url_scopes_dict = self._consolidate_collections(known_url_scopes)
+
+        self._known_scopes = list(known_scopes_dict.values())
+        self._known_url_scopes = list(known_url_scopes_dict.values())
+
```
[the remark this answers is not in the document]
derek-globus: Didn't catch the spinx extension usage case. Thanks for calling it out

The existing sphinx extension only worked for urn-based scopes (those were registered at `_known_scopes` while url-based ones were at `_known_url_scopes`)

In new rev I updated this paradigm to
* consolidate all instance specific scopes into a shared `_registered_scope_names` list
* expose accessible scope names (including classattr ones) via a `scope_names` property (which sphinx uses now)

src/globus_sdk/scopes.py
```diff
@@ -116,29 +121,52 @@ def __init__(
...
+            setattr(self, scope_name, transform_func(scope_val))
+
+    def _scopes_input_to_dict(self, items: ScopeBuilderScopes) -> t.Dict[str, str]:
+        """
+        ScopeBuilders accepts many collection-style types of scopes. This function
+          normalizes all of those types into a standard {scope_name: scope_val} dict
+
+        Translation Map:
+        None => {}
+        "my-str" => {"my-str": "my-str"}
```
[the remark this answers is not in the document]
derek-globus: I suppose I could make the comments accurate if you insist....  \s

Good catch. New rev.

src/globus_sdk/scopes.py
```diff
@@ -116,29 +121,52 @@ def __init__(
...
-                self._known_scope_names.append(scope_name)
-                setattr(self, scope_name, self.urn_scope_string(scope_name))
-        if self._known_url_scopes:
-            for scope_name in self._known_url_scopes:
-                self._known_scope_names.append(scope_name)
-                setattr(self, scope_name, self.url_scope_string(scope_name))
+
+        self._registered_scope_names: t.List[str] = []
+        self._register_scopes(known_scopes, self.urn_scope_string)
+        self._register_scopes(known_url_scopes, self.url_scope_string)
```
sirosen, line 131: I really like how small and simple `__init__` became in the course of these changes! 👍

src/globus_sdk/scopes.py
```diff
@@ -116,29 +121,52 @@ def __init__(
...
+    def _scopes_input_to_dict(self, items: ScopeBuilderScopes) -> t.Dict[str, str]:
+        """
+        ScopeBuilders accepts many collection-style types of scopes. This function
+          normalizes all of those types into a standard {scope_name: scope_val} dict
+
+        Translation Map:
+        None => {}
+        "my-str" => {"my-str": "my-str"}
+        ["my-list"] => {"my-list": "my-list"}
+        ("my-tuple-key": "my-tuple-val") => {"my-tuple-key": "my-tuple-val"}
```
sirosen, line 150:
```suggestion
        ("my-tuple-key", "my-tuple-val") => {"my-tuple-key": "my-tuple-val"}
```

I think this is a (minor) typo.

src/globus_sdk/scopes.py
```diff
@@ -218,15 +246,9 @@ def make_mutable(self, scope: str) -> MutableScope:
         """
         return MutableScope(getattr(self, scope))
 
-    def _iter_scopes(self) -> t.Iterator[t.Tuple[str, str]]:
-        for name in self._classattr_scope_names:
-            yield (name, getattr(self, name))
-        for name in self._known_scope_names:
-            yield (name, getattr(self, name))
```
sirosen, line 225: 👍 for this removal, in favor of the `scope_names` property.

sirosen (changes requested):
Everything looks great, but we have one final fix to apply, to the sphinx extension.

See the following screenshot of docs for AuthScopes:
[doc build, scopes page](https://globus-sdk-python--641.org.readthedocs.build/en/641/scopes.html)
image

Where did the duplicates come from? (I knew they were there, which is why I went to the doc build. 😉 )

A failing of the `ScopeBuilder` has been fixed. `ScopeBuilder.scope_names` fetches the classattr scopes in addition to the instance-assigned scopes. This was missing in the past mechanism.
The scopes were being explicitly listed with `add_scopes=...`. (I think the classattr list was added after the extension was written.)

The fast fix is to just do this:
```diff
 .. listknownscopes:: globus_sdk.scopes.AuthScopes
-    add_scopes=openid,email,profile
     example_scope=view_identity_set
```

The _optional_ thing to add on here is to remove the `add_scopes=...` feature from the extension, but that can be done in a follow-up. I leave it to your discretion as the PR author whether you'd like us to do that in a separate PR or if you want to include it here.

src/globus_sdk/scopes.py
derek-globus: I love small methods! Don't make me think about what this thing is doing!

src/globus_sdk/scopes.py
derek-globus: This line of the comment will be the end of me

derek-globus:
Good catch on AuthScopes Stephen, I generated the docs locally after adding `scope_names` but must've missed that one in my inspection.

I'll update that in a new rev.

sirosen (approved):

PR #642, opened 2022-11-04
Add scope parsing to Scope object

sirosen:
I'm not sure that it's all perfect, but I wrote a parser and I can't seem to find a valid scope which breaks it, so that seems like a good sign.

I _did_ strongly consider using `pyparsing` for this. I've had good experiences using `pyparsing` and it seemed like it might fit this use-case. There were a couple of reasons I decided not to use it here.
1. Overkill. If it turns out that we can write a correct lexer trivially, and a correct parser easily in pure python without defining it as a recursive descent parser.
2. Dependency handling. We'd either add a dependency (possibility of conflicts), or add it as an optional dependency (more complicated codepaths).
3. Globally scoped ParserElement changes. The whitespace sensitivity of `pyparsing` can be adjusted on a global basis by certain method calls, and I know of at least one other bit of global state (`enable_packrat`). I didn't like the idea that our parsing behavior could change depending on what a user application might be doing with pyparsing independently.
4. Whitespace-sensitivity is messy. `pyparsing` defines all whitespace as non-semantic by default. This is fine when it fits your language, but Globus Auth requires that an optional marker be adjacent to the scope string (`*foo`, but not `* foo`). It seemed like this might lead to us using `Combine()` in ways that don't naturally fit our desired grammar.

---

The core of this change is that `MutableScope` has been renamed to `Scope` and now has a pair of parsing functions as classmethods: `Scope.parse` and `Scope.deserialize`. The first parses any valid scope string to request, meaning it may be a space delimited list of scopes, and the second parses any valid scope string with a single top-level scope.

As an inverse of `deserialize()`, the stringifier has been renamed to `serialize()`, and `__str__()` now just calls `serialize()`.

The parsing itself requires several new internal components. A tokenizer handles converting strings into lists of tokens, e.g. converting `foo[bar *baz]` into
`["foo", "[", "bar", "*", "baz", "]"]`.
The tokens are small objects pairing strings with a type, which makes it simpler for the parser to handle various cases. The parser is minimally stateful, has no recursion or backtracking, operates with limited lookahead, and is generally meant to be the smallest and simplest thing which can do the job that it does.

The parse result, a `Scope`, can now represent the true tree-like structure of scopes. `Scope.dependencies` is a list of child nodes in the tree, a list of  `Scope`s. This has implicitations for `Scope.add_dependency` in that a `Scope` is now a valid input. As a result of this introduction, the use of
`add_dependency("foo", optional=True)` is considered deprecated and now emits a DeprecationWarning. Building dependencies in the current version should be done either with (newly parseable) strings or with `Scope` objects. i.e. The aforementioned usage can be replaced with either `add_dependency("*foo")` or
`add_dependency(Scope("foo", optional=True))`.

Some related changes included here:
- `globus_sdk.scopes` is now a subpackage (it was getting a bit long)
- the internal type alias for ScopeCollectionType is now kept in `_types` for simpler universal access within the SDK
- many more unit tests are added for the scope parsing and serialization functionality
- documentation is updated to refer to these as "Scope objects" rather than "MutableScopes"
- various doc fixes, including missing whitespace, missing scope data in the scopes docs, and a bug in the listknownscopes custom directive


src/globus_sdk/scopes/scope_definition.py
```diff
@@ -236,6 +236,61 @@ def __repr__(self) -> str:
...
+        # top-level scope must match
+        if self._scope_string != other._scope_string:
+            return False
+        # if both are optional, okay
+        # if neither is optional, okay
+        # but if only one is optional...
+        if self.optional != other.optional:
+            # ... then make sure it is 'other'
+            if self.optional:
+                return False
```
kurtmckee, line 279: I had to stare at this for a moment, even with the comments. This may have been that the final explanatory comment mentions `other` but the actual condition was for `self`. Would this -- still with explanatory comments! -- improve the readability here?

```python
# Three out of the four possible combinations are acceptable:
# Both are optional; neither is optional; or only 'other' is optional.
# If 'other' is required but 'self' is optional,
# then 'self' cannot contain 'other'.
if not other.optional and self.optional:
    return False
```

src/globus_sdk/scopes/scope_definition.py
```diff
@@ -236,6 +236,61 @@ def __repr__(self) -> str:
...
+
+        # dependencies must all be contained -- search for a contrary example
+        for other_dep in other.dependencies:
+            found_match = False
+            for dep in self.dependencies:
+                if other_dep in dep:
+                    found_match = True
+                    break
+            if not found_match:
+                return False
```
kurtmckee, line 289: How do you feel about `else` clauses on `for` loops?

```python
for other_dep in other.dependencies:
    for dep in self.dependencies:
        if other_dep in dep:
            break
    else:
        # The inner loop ended without finding a match.
        return False
```

src/globus_sdk/scopes/scope_definition.py
sirosen: Ooh, I like how this flattens the conditions. It is written the way I _reasoned_ through what is correct. But now, with the conclusion in hand, I like it better with the explanation in the comment but the condition being this way around. It makes it clear that there's only one False case, and calls out what that is.
I almost want to put a truth table into the comment! 😁

changelog.d/20221104_145601_sirosen_add_scope_parsing.rst
```diff
@@ -15,3 +15,9 @@
     ``Scope(...).add_dependency("*foo")`` can be used to add an optional
     dependency as a string, or equivalently
     ``Scope(...).add_dependency(Scope("foo", optional=True))``
+
+  * ``Scope``\s support containment checking (``Scope("foo") in Scope("bar")``),
```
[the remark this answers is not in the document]
sirosen: I'm a bit torn. I'm actually not even sure I want us to implement this idea with `__contains__`.

Maybe this should be a set of comparators so that it's `scope1 < scope2`? That's more laborious to define, but maybe easier to explain.

Or we could stay away from the dunder methods and call this something like "contains()"?

src/globus_sdk/scopes/scope_definition.py
sirosen: I never think to use for-else -- and when reminded about it, I tend to want to be cautious that we keep things clear. However, this looks like an appropriate usage site. I'll switch it.

changelog.d/20221104_145601_sirosen_add_scope_parsing.rst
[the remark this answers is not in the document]
kurtmckee: I think the only difference between `.contains()` vs `.__contains__()` is that we get to write `smaller_thing in maybe_bigger_thing`. If "smaller" and "bigger" aren't meaningful ways to think about this, it might not be intuitive to write `scope1 < scope2`.

Spit-balling just to get the gears turning:

* `.is_superset()`
* `.will_get()`
* `.overlays()`

src/globus_sdk/scopes/scope_definition.py
kurtmckee: Haha, and I'm pretty sure I only think of for-else _after_ I've already written a bail-if-not-found loop. "Found" flags in `for` loops trigger the same thoughts as seeing a `while` loop that manages a counter ("use a `for` loop") or seeing a `for` loop that manages a counter ("use `enumerate()` in a `for` loop").

All that say that I, too, never first think to use a for-else...and it's not always the clearest way to solve something.

src/globus_sdk/scopes/scope_definition.py
kurtmckee: I'll rubber-stamp any change that adds a truth table, but only if the matrix intersection points are Unicode characters instead of "T/F" or "Y/N" or similar. 😆

changelog.d/20221104_145601_sirosen_add_scope_parsing.rst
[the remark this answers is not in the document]
sirosen: Per our discussion, we agreed on "permission coverage checking" as the textual name for this, but `__contains__` as the method to use such that `x in y` is described as "coverage checking".

I'll make sure that we get a docstring on `__contains__` which uses the same terminology, and that we include it in docs via the sphinx directive (I think `:special-members:`? but I'll double-check that).

src/globus_sdk/scopes/builder.py
```diff
@@ -0,0 +1,161 @@
...
+        client, in the Globus Auth URN format.
+
+        Note that this module already provides many such scope strings for use with
+        Globus services.
+
+        **Examples**
+
+        >>> sb = ScopeBuilder("transfer.api.globus.org")
+        >>> sb.urn_scope_string("transfer.api.globus.org", "all")
+        "urn:globus:auth:scope:transfer.api.globus.org:all"
```
kurtmckee, line 106: Are these docstrings extracted by autodoc?

src/globus_sdk/scopes/scope_definition.py
```diff
@@ -0,0 +1,323 @@
...
+        if c in "[]* ":
+            if current_token:
+                tokens.append(
+                    ParseToken("".join(current_token), ParseTokenType.scope_string)
+                )
+                current_token = []
+
+            if c == "*":
+                if peek == " ":
+                    raise ScopeParseError("'*' must not be followed by a space")
```
kurtmckee, line 61: While researching whether `*[scope]` was acceptable syntax I noticed that [scopes may be plus-separated](https://docs.globus.org/api/auth/reference/#grant_request), too. Should that be addressed here?

tests/unit/test_scopes.py
```diff
@@ -148,9 +154,187 @@ def test_scopebuilder_make_mutable_produces_same_strings():
...
+        # double brackets
+        "foo[[bar]]",
+        # unbalanced brackets
+        "foo[",
+        "foo[bar",
+        "foo[bar]]",
+        # space before brackets
+        "foo [bar]",
+        # missing space before next scope string after ']'
+        "foo[bar]baz",
```
kurtmckee, line 203: I don't see the following possibilities (which might not all be invalid):

* `"foo * "` (space after '*')
* `"foo[ bar]"` (space after '[')
* `" foo"` (leading space)
* `"foo]"` (unmatched closing bracket 1)
* `"foo bar]"` (unmatched closing bracket 2)
* `"foo[bar] baz]"` (unmatched closing bracket 3)

The reason there are multiple unmatched closing brackets examples is because each might exercise different types of "reset" failures.

Note that if plus-separated string parsing is added after reviewing a comment above, these tests will need to also be updated.

src/globus_sdk/scopes/builder.py
sirosen: Yes, they are. Do we need to `.. code-block: pycon` this? I thought that sphinx will read a leading `>>>` as the start of a console block, but maybe I'm not remembering the behavior correctly.

src/globus_sdk/scopes/scope_definition.py
sirosen: I believe that that's a misleading bit of documentation. Unless I'm mistaken, it's just referring to `+` as a style of url quoting.

tests/unit/test_scopes.py
sirosen: I just checked, and there are two of these which are valid: `" foo"` (equivalent to `"foo"`) and `"foo[ bar]"`. The others are all definitely invalid, and I'm more than happy to add them as test cases; thanks for enumerating them!
I'll also add those two valid cases to our tests, to document/enforce their validity.

At this point, I'm setup with a copy of Auth's scope parser to play with, so I can check these against the service to ensure we agree on things to accept or reject. So if you have other questionable cases that you think up, we can evaluate them and add as positive or negative tests.

src/globus_sdk/scopes/scope_definition.py
kurtmckee: I wondered if that might be the case. I'll mark this resolved. Thanks!

src/globus_sdk/scopes/builder.py
kurtmckee: Ah, yes, you're right. Cary on!

sirosen:
This work is something I'm considering ready to merge in its current state.

We discussed the differences between the parse tree and the true processing and representation of consents in Globus Auth. The end result of that discussion was that we agreed that parsing is still a very valuable utility to provide, but that one of the pieces here is something which we aren't ready to expose -- `__contains__`. For the immediate, short term, we are keeping this functionality, but under the internal name `_contains`. And as an extra measure, the docstring for that internal method declares it "non-authoritative", since it could always be misleading to look at a saved scope string rather than at consents in Auth.

`scope1 in scope2` has been removed from public docs, including the changelog.

Also, per a discussion around the `add_dependency` interface, we are not changing its chaining behavior, but that usage no longer appears in the docs. Removing the chaining behavior is slated for SDK v4.0, to ensure it won't be forgotten when we are next liberated to change interfaces.

tests/unit/test_scopes.py
kurtmckee: I just realized there's one more: `*foo`. I don't think this variant is accounted for yet.

sirosen: I just pushed [6a97871](https://github.com/globus/globus-sdk-python/pull/642/commits/6a97871b24709d4d381a9f838a67ad3858e6c4f3), which adds a few test cases for "can parse" as opposed to these for "can't parse".
It includes `*foo` and gives us a place for any future interesting cases which are valid.

kurtmckee (approved):

src/globus_sdk/scopes/builder.py
```diff
@@ -0,0 +1,161 @@
+import typing as t
+
+from .scope_definition import Scope
+
+ScopeBuilderScopes = t.Union[
+    None, str, t.Tuple[str, str], t.List[t.Union[str, t.Tuple[str, str]]]
+]
+
+
+class ScopeBuilder:
```
derek-globus, line 10: With the introduction of the class `Scope`, I think the name `ScopeBuilder` becomes a bit confusing.
`ScopeBuilder` builds a `string`, not a `scope`.

Probably beyond the fold for this change, but we should probably rethink what this class is responsible for & name it accordingly.
Ideas from the top of my head:
 * `ScopeRegistry`
 * `ScopeStringRegistry`
 * `ScopeStringFormatter`
 * `ScopeDict`

src/globus_sdk/scopes/scope_definition.py
```diff
@@ -0,0 +1,332 @@
...
+                ret.append(current_scope)
+    if parents:
+        raise ScopeParseError("unclosed brackets, missing ']'")
+
+    return ret
+
+
+class Scope:
+    """
+    A mutable scope is a representation of a scope which allows modifications to be
```
derek-globus, line 138: nit: ~~mutable scope~~

derek-globus (changes requested):
Overall looks good

The only additional piece I'd ask for before merging would be to include data in the unit tests that performs deserializing/serializing on real (or real-formatted) url & urn scope strings.

I doubt it would cause issues by reading the parser myself, but I wouldn't want us to unintentionally break in a way that we could successfully parse the scope string `"foo"` but  not `"urn:globus:auth:scope:auth.globus.org:foo"` or `"https://auth.globus.org/scopes/auth.globus.org/foo"`

src/globus_sdk/scopes/builder.py
sirosen: I want to address this in SDKv4 by renaming the object and making its members `Scope` objects. That would remove any need for `make_mutable` as a method, and provide a more consistent view that what the SDK provides are high-level helper objects which _render_ as strings.

The fact that the `ScopeBuilder` objects are attached to and re-exposed through client classes makes this fiddly to even attempt prior to v4.

derek-globus: Yeah, I agree with that plan.

Just a side-effect I thought of while reading & wanted to make sure that opinion was voiced.

src/globus_sdk/scopes/scope_definition.py
sirosen: I switched to "scope object".

derek-globus (approved):

PR #643, opened 2022-11-04
Fix doc warnings which error locally

sirosen:
And also set `fail_on_warning: true` so that the CI run of sphinx (done by readthedocs) will match local runs of sphinx (done by `tox -e docs`)

The warnings came from missing whitespace before lists.

---

This impacts #641 and #642. The latter is where I discovered this. In the former, I'd like to suggest a change which requires that we fix this in order to be able to run `tox -e docs` to test the change locally.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -2,6 +2,7 @@ version: 2
 
 sphinx:
   configuration: docs/conf.py
+  fail_on_warning: true
 
 build:
   os: ubuntu-20.04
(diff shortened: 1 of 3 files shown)


kurtmckee (approved):

sirosen:
This failed in RTD even though it worked for me locally on py3.11.
I tracked down the cause thanks to RTD providing this bit of debug output:
```
asdf global python 3.11.0b3
```

I didn't realize they were still using one of the betas. As such, I've rolled our RTD config back to 3.10 and am waiting for a hopefully clean build.

PR #645, opened 2022-11-04
Add an initial seed codeowners

sirosen:
This lists the personnel currently considered responsible for the SDK code. It intentionally does not list all of the impacted stakeholders, and makes no attempt to break down any service responsibilities other than those for flows and timer.

---

Please note that the listed reviewers != the list of personnel in codeowners and that this is intentional.
The order in which names are listed ~reflects who are my bestest friends~ is alphabetical for simpler comparisons and just to set some sort of rule.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -9,11 +9,9 @@
 * @aaschaer @ada-globus @derek-globus @kurtmckee @sirosen
 
 # Flows service
-/src/globus_sdk/services/flows/ @ada-globus @derek-globus @kurtmckee @sirosen
-/src/globus_sdk/_testing/data/flows/ @ada-globus @derek-globus @kurtmckee @sirosen
-/tests/functional/services/flows/ @ada-globus @derek-globus @kurtmckee @sirosen
+**/flows/ @ada-globus @derek-globus @kurtmckee @sirosen
+docs/services/flows.rst @ada-globus @derek-globus @kurtmckee @sirosen
 
 # Timer service
-/src/globus_sdk/services/timer/ @ada-globus @derek-globus @kurtmckee @sirosen
-/src/globus_sdk/_testing/data/timer/ @ada-globus @derek-globus @kurtmckee @sirosen
-/tests/functional/services/timer/ @ada-globus @derek-globus @kurtmckee @sirosen
+**/timer/ @ada-globus @derek-globus @kurtmckee @sirosen
+docs/services/timer.rst @ada-globus @derek-globus @kurtmckee @sirosen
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,19 @@
...
+# for ease of comparison, usernames are kept alphabetized
+#
+
+# default rule
+* @aaschaer @ada-globus @derek-globus @kurtmckee @sirosen
+
+# Flows service
+/src/globus_sdk/services/flows/ @ada-globus @derek-globus @kurtmckee @sirosen
+/src/globus_sdk/_testing/data/flows/ @ada-globus @derek-globus @kurtmckee @sirosen
+/tests/functional/services/flows/ @ada-globus @derek-globus @kurtmckee @sirosen
```
kurtmckee, line 14: Do you think this pattern would be usable here?

```suggestion
**/flows/ @ada-globus @derek-globus @kurtmckee @sirosen
```

kurtmckee (commented):
Interestingly, my own friendships' rankings correlate strongly with _reverse_ alphabetical order.

This is likely because I don't take time to scroll when I want to call or text somebody about trivial top-of-mind stuff, so the people at the top of my address book have heard quite enough from me.

aaschaer:
I'm wondering if it would make sense to use teams instead of individual members?

The sdk-and-cli-team is just Stephen and myself, so that wouldn't be sufficient as is, but I wouldn't mind having that team updated to reflect the people we want to be the default code owners here.

ada-globus:
> I'm wondering if it would make sense to use teams instead of individual members?
> 
> The sdk-and-cli team is just Stephen and myself, so that wouldn't be sufficient as is, but I wouldn't mind having that team updated to reflect the people we want to be the default code owners here.

Yeah, the unfortunate thing is that the way GitHub interprets teams in this context means that the first person from a team who engages on a PR becomes something of the single official representative of the team vis-a-vis that PR while the rest of the team is removed (IIRC), so it works if fine you want _anyone_ on a team to see a PR, but it's not great if you actually want _everyone_ on a team to have ongoing visibility.

ada-globus:
Do we also want to include e.g., `docs/services/flows.rst`?

.github/CODEOWNERS
sirosen: I wasn't sure. I guess the risk that a directory gets named `flows` for some other purpose is low, and importantly the impact of incorrect matching is also low.
Why don't we go ahead with changing to the `**` match and we can always re-evaluate if we ever find a need. I'll push the update in a few.

sirosen:
> Do we also want to include e.g., `docs/services/flows.rst`?

Yes, good catch! I'll push that change momentarily.

> I'm wondering if it would make sense to use teams instead of individual members?

Ada already answered the details on why it's this way, but I'll take a moment to say: I wish we could use teams and get the semantics people want. But it doesn't seem to be possible. (And listing a team + members is even more confusing.)

kurtmckee (approved):

PR #647, opened 2022-11-09
First version of LocalGlobusConnectServer object

sirosen:
This adds a start on a local GCS object which looks at the canonical `info.json` file and pulls attributes out of it in a type-safe way.

Primarily in order to support testing, the path to the `info.json` file can be set as an argument to the LocalGlobusConnectServer initializer.

### DIFF
--- a/src/globus_sdk/local_endpoint/server/endpoint.py
+++ b/src/globus_sdk/local_endpoint/server/endpoint.py
@@ -37,14 +37,14 @@ class LocalGlobusConnectServer:
         applications using the SDK may not be able to interact with Globus Connect
         Server on the container host.
 
-        :type: dict
+        :type: dict or None
         """
         if self._loaded_info is None:
             if self.info_path.is_file():
                 with open(self.info_path, encoding="utf-8") as fp:
                     try:
                         parsed_data = json.load(fp)
-                    except json.JSONDecodeError:
+                    except (UnicodeDecodeError, json.JSONDecodeError):
                         pass
                     else:
                         if isinstance(parsed_data, dict):
@@ -61,7 +61,7 @@ class LocalGlobusConnectServer:
         The endpoint ID of the local Globus Connect Server endpoint.
         None if the data cannot be loaded or is malformed.
 
-        :type: str
+        :type: str or None
         """
         if self.info_dict is not None:
             epid = self.info_dict.get("endpoint_id")
@@ -75,7 +75,7 @@ class LocalGlobusConnectServer:
         The domain name of the local Globus Connect Server endpoint.
         None if the data cannot be loaded or is malformed.
 
-        :type: str
+        :type: str or None
         """
         if self.info_dict is not None:
             domain = self.info_dict.get("domain_name")
--- a/src/globus_sdk/local_endpoint/server/endpoint.py
+++ b/src/globus_sdk/local_endpoint/server/endpoint.py
@@ -1,0 +1,84 @@
+import json
+import pathlib
+import typing as t
+
+
+class LocalGlobusConnectServer:
+    r"""
+    A LocalGlobusConnectServer object represents the available SDK methods
+    for inspecting and controlling a running Globus Connect Server
+    installation.
+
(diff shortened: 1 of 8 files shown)


src/globus_sdk/local_endpoint/server/endpoint.py
```diff
@@ -0,0 +1,84 @@
...
+        Server on the container host.
+
+        :type: dict
+        """
+        if self._loaded_info is None:
+            if self.info_path.is_file():
+                with open(self.info_path, encoding="utf-8") as fp:
+                    try:
+                        parsed_data = json.load(fp)
+                    except json.JSONDecodeError:
```
kurtmckee, line 47: `UnicodeDecodeError` needs to be caught as well.

```suggestion
                    except (UnicodeDecodeError, json.JSONDecodeError):
```

src/globus_sdk/local_endpoint/server/endpoint.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+        The info.json data for the local Globus Connect Server, as a dict.
+
+        If the info.json file is not present or cannot be parsed, the data is None.
+        This indicates that there is no local Globus Connect Server node, or if there is
+        one, it cannot be used or examined by the SDK. For example, containerized
+        applications using the SDK may not be able to interact with Globus Connect
+        Server on the container host.
+
+        :type: dict
```
kurtmckee, line 40: Should the annotations throughout (like this one) reflect that None is a failure-case possibility?

tests/functional/local_endpoint/test_server.py
```diff
@@ -0,0 +1,78 @@
...
+
+def test_info_dict_from_nonexistent_file_is_none(tmp_path):
+    info_path = tmp_path / "info.json"
+    gcs = LocalGlobusConnectServer(info_path=info_path)
+    assert gcs.info_dict is None
+
+
+def test_info_dict_from_non_json_file_is_none(tmp_path):
+    info_path = tmp_path / "info.json"
+    info_path.write_text("{")
```
kurtmckee, line 12: A failure case for invalid Unicode is needed here as well. Perhaps write a binary file with:

```python
bytes.fromhex("1BAD DEC0DE")
```

Eh? EH?

kurtmckee (changes requested):

src/globus_sdk/local_endpoint/server/endpoint.py
sirosen: I always assumed that the decoder would raise a `JSONDecodeError` on non-unicode data. I'm willing to bet that I have this handling missing from many bits of code I've written over the years.

I'll put together a test case and get this added.

src/globus_sdk/local_endpoint/server/endpoint.py
sirosen: I was on the fence about that, so I'm glad you brought it up. If the type is documented as `X or None`, is that sufficient? Or should each one also have a sentence in the docstring about "if the value cannot be read, this is `None`"?

tests/functional/local_endpoint/test_server.py
sirosen: I cannot turn down `1BAD DEC0DE`. I don't have that kind of willpower. Expect to see it later today, when I can circle back on this.

src/globus_sdk/local_endpoint/server/endpoint.py
kurtmckee: I think that it's worthwhile to both document that something _could be_ `X or None`, and to explain _why_ something might be None. The reason might not be obvious in some circumstances, particularly when a descriptive exception shouldn't be raised for each failure mode.

sirosen: That's very reasonable; I'll make sure it's documented clearly.
I will note, by way of explanation, that I'm not a big fan of properties which raise exceptions, which is why I chose the `X | None` paradigm for these.

kurtmckee (approved):

PR #650, opened 2022-11-14
add entity type to get endpoint _testing data

aaschaer:
Doing CLI work for https://app.shortcut.com/globus/story/19989/use-entity-type-when-determining-endpointish-type and some tests will depend on `entity_type` being set here.

Can just modify the fixtures CLI side if needed, but this seems like a particularly good field to add to _testing data anyways

### DIFF
--- a/src/globus_sdk/_testing/data/transfer/get_endpoint.py
+++ b/src/globus_sdk/_testing/data/transfer/get_endpoint.py
@@ -9,6 +9,7 @@ ENDPOINT_DOC = {
     "organization": "My Org",
     "username": "auser",
     "description": "Example gridftp endpoint.",
+    "entity_type": "GCSv4_host",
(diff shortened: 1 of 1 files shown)


sirosen (approved):
Maybe we should add a "data lint" test to check that our `_testing` data provides a specific set of required fields?

kurtmckee (approved):

aaschaer:
> Maybe we should add a "data lint" test to check that our _testing data provides a specific set of required fields?

I think keeping that linter up to date would be about the same amount of effort as keeping the data itself up to date, unless we point the linter at live API responses or something?

sirosen:
I think that's a reasonable concern. I'll noodle on this more. I think the saved response data will slowly drift out of date if we just let it, as it has in this case.

PR #654, opened 2022-11-21
Yank scope parsing, revert to MutableScope

sirosen:
After discussion of the ways in which a parse tree cannot accurately represent the meaning of consents and scopes in Auth, we have agreed to pull scope parsing from the upcoming SDK release. The code is remaining in-tree, but no longer provided via documented modules and import paths. The changelog for it has been pulled, replaced by a more modest set of changes to MutableScope, and the docs changes have been further modified to remove any references to scope parsing.

Tests continue to exercise the scope tree parsing logic and all of the utility of the Scope class as it was previously introduced.


src/globus_sdk/scopes/scope_definition.py
```diff
@@ -330,3 +332,99 @@ def scopes2str(obj: ScopeCollectionType) -> str:
...
+        return "MutableScope(" + ", ".join(parts) + ")"
+
+    def __str__(self) -> str:
+        return self.serialize()
+
+    @staticmethod
+    def scopes2str(obj: ScopeCollectionType) -> str:
+        """
+        Given a scope string, a collection of scope strings, a Scope object, a
+        collection of Scope objects, or a mixed collection of strings and
```
kurtmckee, line 427: ?

```suggestion
        Given a scope string, a collection of scope strings, a MutableScope object, a
        collection of MutableScope objects, or a mixed collection of strings and
```

kurtmckee (approved):
I approve of this! I left barbarically-short comments questioning if a few instances of the word "Scope" should change to "MutableScope", but this looks good to me.

sirosen:
> I left barbarically-short comments questioning if a few instances of the word "Scope" should change to "MutableScope",

I love that phrasing of this. 😹 

The cases that you picked out were meant to be reverted to `MutableScope`, so 👍  to both of them. I'm pushing through that commit right now.

PR #655, opened 2022-11-25
Update to flake8==6.0.0 , and update additional_dependencies

sirosen:
Working through various repos I can update nicely with [`upadup`](https://pypi.org/project/upadup/) now that I have it working for basic purposes; I was also poking at other elements of our linting config.

1. Update to the latest flake8 version (no impact for us)
2. Add `flake8-comprehensions`
3. Apply `upadup`

Regarding (2), I saw this as one of the plugins emulated by `ruff` and therefore took a look at it.
I don't think of myself as being very good about writing comprehensions consistently, so I thought it would possibly even call out something inefficient in a core codepath somewhere. I was a little disappointed that it saw so few things in our codebase, but I guess that's a good thing. 😉 
Even so, `flake8-comprehensions` will help protect against mistakes, and it _did_ find a couple of oddities (in places that don't matter much, but still...). So I kept the addition.

I'm more excited about `upadup`!

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,16 +18,16 @@ repos:
   rev: v1.12.1
   hooks:
     - id: blacken-docs
-      additional_dependencies: ['black==21.12b0']
+      additional_dependencies: ['black==22.10.0']
 - repo: https://github.com/PyCQA/flake8
   rev: 6.0.0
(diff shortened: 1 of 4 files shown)


ada-globus (approved):
This is great—excited to use upadup in other projects!

PR #658, opened 2022-12-05
add session_required_policy to AuthorizationParamaterInfo

aaschaer:


kurtmckee (dismissed):

kurtmckee (approved):

kurtmckee (approved):

sirosen (approved):
Yep, I'm good with this being our approach here.

If we find that some authorization_parameters fields are used exclusively in many (or all?) cases, we may want to pull that up into a (duplicate) `ErrorInfo` object, like `err.info.session_required_policies`. I'll need to learn more about the design here to be able to comment on what makes sense.

For now, this exposes the data which we know about, so I'm happy with it. 👍

PR #659, opened 2022-12-05
Fix py3.6 build in CI

sirosen:
Use Ubuntu 20.04 (instead of ubuntu-latest) to avoid build issues on 22.04 + py3.6.


aaschaer (approved):
"build / test py3.6 on ubuntu-20.04 (pull_request) Successful in 37s"

Looks good to me!

PR #670, opened 2023-01-20
Update tox.ini and handle fallout

sirosen:
Refactor the tox config to achieve the following:
- a wheel is built in a shared build_wheel environment for faster overall builds
- `usedevelop=true` is no longer used
- cov-clean is renamed 'coverage_clean'
- cov-cobmine, cov-report are combined into 'coverage_report'
- default tox envs now include `lint` and `mypy`
- update `mypy` in tox.ini to v0.991
- cleanup the 'depends' lists' and add test-lazy-imports to the default envlist

Effects from these changes are handled thusly:
- `responses` must also be updated to the latest (0.22.0) to fix an issue on newer `mypy` versions
- .coveragerc needs adjustment to work when `usedevelop` is not in use, normalizing '.tox/*' installs of globus_sdk to be treated identically to the src/ tree

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -8,7 +8,8 @@ envlist =
     py36-mindeps
     coverage_report
 skip_missing_interpreters = true
-minversion = 4.0.0
+# TODO: set a minversion after we drop support for python3.6
+# minversion = 4.0.0
 
 [testenv]
 # build a wheel, not a tarball, and use a common env to do it (so that the wheel is shared)
--- a/tox.ini
+++ b/tox.ini
@@ -1,40 +1,40 @@
 [tox]
 envlist =
-    cov-clean
-    cov-combine
-    cov-report
+    lint
+    mypy
+    test-lazy-imports
+    coverage_clean
     py{311,310,39,38,37,36}
     py36-mindeps
+    coverage_report
 skip_missing_interpreters = true
-minversion = 3.0.0
+minversion = 4.0.0
 
 [testenv]
-usedevelop = true
+# build a wheel, not a tarball, and use a common env to do it (so that the wheel is shared)
+package = wheel
+wheel_build_env = build_wheel
+
 extras = dev
 deps =
     mindeps: requests==2.19.1
     mindeps: pyjwt==2.0.0
     mindeps: cryptography==3.3.1
     mindeps: typing_extensions==4.0
-commands = coverage run -m pytest {posargs}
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -1,40 +1,41 @@
...
+wheel_build_env = build_wheel
+
 extras = dev
 deps =
     mindeps: requests==2.19.1
     mindeps: pyjwt==2.0.0
     mindeps: cryptography==3.3.1
     mindeps: typing_extensions==4.0
-commands = coverage run -m pytest {posargs}
+commands = coverage run --source globus_sdk -m pytest {posargs}
```
kurtmckee, line 25: Why is `--source globus_sdk` needed? Is the config in `.coveragerc` not producing expected results?

tox.ini
```diff
@@ -1,40 +1,41 @@
...
 depends =
-    py36,py37,py38,py39,py36-mindeps: cov-clean
-    cov-combine: py36,py37,py38,py39,py36-mindeps
-    cov-report: cov-combine
+    py{36,37,38,39,310,311,36-mindeps}: coverage_clean
+    coverage_report: py{36,37,38,39,310,311,36-mindeps}
```
kurtmckee, line 28: I think that the `-mindeps` marker can be left off of this, since it's a marker...but if not, then that may finally explain some challenges I've had in another codebase (and had thought were issues related to PyPy being slow)!

Does this work?

```suggestion
    py{36,37,38,39,310,311}: coverage_clean
    coverage_report: py{36,37,38,39,310,311}
```

tox.ini
sirosen: 🤦 It's not needed. I think I was toying with it earlier and didn't remove it.

tox.ini
sirosen: I don't really know on this one, since it seems like a detail of how `depends` evaluates. Let me toy with it a bit.

kurtmckee: If it _is_ needed then this will finally, _finally_ explain a problem I've had in another repo. I'm actually going to test that now. If it resolves a problem I've had with PyPy tests producing coverage files after coverage combine has already run, then it will indicate that tox requires explicit lists, and all markers must be provided!

kurtmckee: It is absolutely necessary. This knowledge fixes a problem I've had in another repo, too.

The moral of the story is: "All tox environment dependencies must be specified in full, including their markers."

kurtmckee (approved):

PR #672, opened 2023-01-24
Revise Example Timer Script

ada-globus:
dsc_3114_jt-1

# Description

Makes some changes to the example Timer script based on discoveries while supporting a user that was running into trouble getting this running:

- Instantiate the native client
- Handle mapped collections
- Remove Transfer client from TransferData invocation
- Use Transfer action provider in scope tree and revise dependencies
- Specifically send the constructed Transfer AP scope

### DIFF
--- a/docs/examples/timer_operations.rst
+++ b/docs/examples/timer_operations.rst
@@ -1,9 +1,8 @@
 Timer Operations Script
 -----------------------
 
-This example demonstrates the usage of methods on the Timer client. Note that we first
-need a Transfer client set up in order to create the ``TransferData`` we will use to
-define our recurring Timer job.
+This example demonstrates the usage of methods on the Timer client to
+schedule a recurring Transfer task.
 
 .. code-block:: python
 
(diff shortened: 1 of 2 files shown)


docs/examples/timer_operations.rst
```diff
@@ -15,33 +14,57 @@ define our recurring Timer job.
...
     timer_token = token_response.by_resource_server[TIMER_CLIENT_ID]["access_token"]
-    transfer_token = token_response.by_resource_server["transfer.api.globus.org"][
-        "access_token"
-    ]
-
-    # Set up a transfer client and create a `TransferData` object
-    transfer_authorizer = AccessTokenAuthorizer(transfer_token)
-    transfer_client = TransferClient(authorizer=transfer_authorizer)
-    data = TransferData(transfer_client, GOEP1, GOEP2)
-    data.add_item("/share/godata/file1.txt", "/~/file1.txt")
```
sirosen, line 44: I'm seeing this removal, but not the redefinition of `data`. Do you have an unpushed change for this, perhaps?

I'm guessing that we just want
```python
data = TransferData(source_endpoint=GOEP1, destination_endpoint=GOEP2)
data.add_item(...)
```

docs/examples/timer_operations.rst
```diff
@@ -15,33 +14,57 @@ define our recurring Timer job.
...
+
+    # If you declared and mapped collections above, add them to the transfer scope
+    if MAPPED_COLLECTION_IDS:
+        for id in MAPPED_COLLECTION_IDS:
+            gcs_data_access_scope = GCSCollectionScopeBuilder(id).make_mutable(
+                "data_access",
+                optional=True,
+            )
+            transfer_scope.add_dependency(gcs_data_access_scope)
+        transfer_action_provider_scope.add_dependency(transfer_scope)
```
sirosen, line 51: Should we maybe dedent this block and remove the `if MAPPED_COLLECTION_IDS` check?
As it currently reads, I got a little mixed up and thought there was a possibility that this `transfer AP -> transfer` dependency could be missing.

But I think that the Transfer AP depends on Transfer statically? So it would do no harm to add the dependency, but it would maybe reduce the number of distinct cases the user has to think about.

docs/examples/timer_operations.rst
```diff
@@ -51,7 +74,14 @@ define our recurring Timer job.
...
     interval = timedelta(minutes=30)
     name = "sdk-timer-example"
-    job = TimerJob.from_transfer_data(data, start, interval, stop_after_n=2, name=name)
+    job = TimerJob.from_transfer_data(
+        data,
+        start,
+        interval,
+        stop_after_n=2,
+        name=name,
+        scope=str(transfer_action_provider_scope),
```
sirosen, line 83: To clarify, this is passing the string-ized full dependency list. 
Is it also valid to pass Timer only this?
`scope="https://auth.globus.org/scopes/actions.globus.org/transfer/transfer"`

It seems to me that if Timer is going to take its token, do a dependent token grant, and then lookup values in the response, we really ought to be passing it the resource server name ("actions.globus.org", I think?) rather than a scope. Or is there some other purpose here, which requires the full scope be passed?

sirosen (changes requested):
I think there's a small missing piece, and then there are two discussion points for us to address.

The only one which must be addressed, as far as I'm concerned, is the `data = ...` one.

docs/examples/timer_operations.rst
ada-globus: Yep—this was missed in a a copy-paste error when I was moving bits from the script I wrote for the support ticket! 🤦‍♀️

docs/examples/timer_operations.rst
ada-globus: I tested this out to confirm, and you're right, there shouldn't be an issue with just allowing it to be added regardless of whether there are mapped collections. Reads better! 👍

docs/examples/timer_operations.rst
ada-globus: Yeah, good call—I had that wrong!

sirosen (approved):
Thanks for fixing this!

I'm going to merge right away. You'll see it publish to the `latest` docs as soon as RTD does the build, but it won't reach `stable` until we do a release.

PR #675, opened 2023-01-30
Add support for new GCS collection doc versions

sirosen:
Support versions '1.6.0' and '1.7.0' via `guest_auth_policy_id` and long `user_message` handling.

Internally, the `guest_auth_policy_id` is simply a mapped-collection-only field matching the current structure, but the `user_message` treatment requires a new structure.

DocumentWithInducedDatatype now defines `DATATYPE_VERSION_CALLBACKS` as a secondary versioning mechanism alongside the
`DATATYPE_VERSION_IMPLICATIONS` mapping. The callbacks take the entire object and return an optional version number tuple. This is an extremely generic way of allowing a callback to apply arbitrary criteria for the version numbering. In service of that change, the protocol now requires `__getitem__` (trivially satisfied by all instances) so that the contents can be checked.
The callbacks are defined as a classvar tuple.

### DIFF
--- a/src/globus_sdk/services/gcs/data/_common.py
+++ b/src/globus_sdk/services/gcs/data/_common.py
@@ -7,14 +7,23 @@ else:
     from typing import Protocol
 
 
+VersionTuple = t.Tuple[int, int, int]
+
+DatatypeCallback = t.Callable[["DocumentWithInducedDatatype"], t.Optional[VersionTuple]]
+
+
 class DocumentWithInducedDatatype(Protocol):
     DATATYPE_BASE: str
-    DATATYPE_VERSION_IMPLICATIONS: t.Dict[str, t.Tuple[int, int, int]]
+    DATATYPE_VERSION_IMPLICATIONS: t.Dict[str, VersionTuple]
+    DATATYPE_VERSION_CALLBACKS: t.Tuple[DatatypeCallback, ...]
 
     def __contains__(self, key: str) -> bool:  # pragma: no cover
         ...
 
     def __setitem__(self, key: str, value: t.Any) -> None:  # pragma: no cover
+        ...
+
+    def __getitem__(self, key: str) -> t.Any:  # pragma: no cover
         ...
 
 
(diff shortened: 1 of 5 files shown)


src/globus_sdk/services/gcs/data/_common.py
```diff
@@ -7,16 +7,25 @@
...
 
     def __contains__(self, key: str) -> bool:  # pragma: no cover
         ...
 
     def __setitem__(self, key: str, value: t.Any) -> None:  # pragma: no cover
         ...
 
+    def __getitem__(self, key: str) -> t.Any:  # pragma: no cover
+        ...
+
```
sirosen, line 28: A brief note for reviewers:

I'm well aware that this protocol looks like it could inherit from Mapping. I've avoided making that change for two basic reasons:
- I don't fully know what the implications are of protocol inheritance (normal class inheritance rules?)
- Protocols are a way of getting non-inheritance based polymorphism. Using inheritance to define them strikes me instinctively as wrong, although maybe it's fine.

On both points, I'm mostly unsure and avoiding the issue for now.

src/globus_sdk/services/gcs/data/collection.py
```diff
@@ -257,9 +281,10 @@ def __init__(
         public: t.Optional[bool] = None,
         # > common args end <
         # > specific args start <
-        storage_gateway_id: t.Optional[UUIDLike] = None,
         # strs
         domain_name: t.Optional[str] = None,
+        guest_auth_policy_id: t.Optional[UUIDLike] = None,
+        storage_gateway_id: t.Optional[UUIDLike] = None,
```
kurtmckee, line 287: It looks like these type-based comment markers (`# strs`, `# str lists`) are becoming inaccurate.

kurtmckee (approved):

src/globus_sdk/services/gcs/data/collection.py
sirosen: I'm not sure if they're useful or not, but I meant "is a string in the payload" (rather than the types accepted by the SDK helper). So the `storage_gateway_id` moved in part to try to make it accurate again.

Like I said, I'm not really sure if this is useful. I'm just trying to give us some handle for looking through these (long!) lists of arguments.

PR #676, opened 2023-01-31
Widen AuthClient client_id types to include UUID

sirosen:
Users are now permitted to pass UUID objects to AuthClient and its subclasses. The UUID is converted to a string on init, but this is more accepting in terms of inputs and matches some usage we have seen in the wild which *probably* works but is not guaranteed.

New unit tests simply assert that the UUID values are accepted and converted to str values on init.


kurtmckee (approved):

sirosen:
The pylint failure is _brand new_ and happens on main, so I'm going to merge in spite of it and then figure out the fix for that separately.

PR #678, opened 2023-02-06
Parse 'session_required_policies' to a list

sirosen:
Rather than keeping the string which an API may pass back, convert it to the list of IDs by comma-splitting the value. If the value is not a string, the comma split won't be possible, so this is handled by treating anything other than a string as invalid (and setting the attribute to None).

This also makes this property a little bit safer than many of our other computed properties, which are cast to the "right" types by the SDK but aren't validated.

The change is technically backwards incompatible, but surrounds the use of a feature which is not in production use yet, and we have therefore agreed upon pragmatic treatment of this case as not requiring a major version bump.

The test changes include updates to check the parsed value(s), but also a general update to the ErrorInfo tests to split a monolithic test into more individual cases.

### DIFF
--- a/src/globus_sdk/exc/err_info.py
+++ b/src/globus_sdk/exc/err_info.py
@@ -1,4 +1,7 @@
+import logging
 import typing as t
+
+log = logging.getLogger(__name__)
 
 
 class ErrorInfo:
@@ -66,13 +69,16 @@ class AuthorizationParameterInfo(ErrorInfo):
         )
 
         # get str|None and parse as appropriate
+        self.session_required_policies: t.Optional[t.List[str]] = None
         session_required_policies = data.get("session_required_policies")
         if isinstance(session_required_policies, str):
-            self.session_required_policies: t.Optional[
-                t.List[str]
-            ] = session_required_policies.split(",")
-        else:
-            self.session_required_policies = None
+            self.session_required_policies = session_required_policies.split(",")
+        elif session_required_policies is not None:
+            log.warning(
+                "during ErrorInfo parsing, got unexpected type for "
+                "'session_required_policies'. "
+                "Expected 'str', but got '{type(session_required_policies)}'"
+            )
 
 
 class ConsentRequiredInfo(ErrorInfo):
--- a/src/globus_sdk/exc/err_info.py
+++ b/src/globus_sdk/exc/err_info.py
@@ -64,9 +64,15 @@ class AuthorizationParameterInfo(ErrorInfo):
         self.session_required_single_domain = t.cast(
             t.Optional[t.List[str]], data.get("session_required_single_domain")
         )
-        self.session_required_policies = t.cast(
-            t.Optional[str], data.get("session_required_policies")
-        )
+
+        # get str|None and parse as appropriate
+        session_required_policies = data.get("session_required_policies")
+        if isinstance(session_required_policies, str):
+            self.session_required_policies: t.Optional[
+                t.List[str]
+            ] = session_required_policies.split(",")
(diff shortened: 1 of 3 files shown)


src/globus_sdk/exc/err_info.py
```diff
@@ -64,9 +64,15 @@ def __init__(self, error_data: t.Dict[str, t.Any]):
...
             t.Optional[t.List[str]], data.get("session_required_single_domain")
         )
-        self.session_required_policies = t.cast(
-            t.Optional[str], data.get("session_required_policies")
-        )
+
+        # get str|None and parse as appropriate
+        session_required_policies = data.get("session_required_policies")
+        if isinstance(session_required_policies, str):
+            self.session_required_policies: t.Optional[
```
aaschaer, line 71: Not a big deal, but it feels odd to only have the type information in the if branch and not the else branch for some reason. Does a pattern like this seem better or is it just me?

```
self.foo: t.Optional[t.fooType] = None
if condition_for_foo_existing:
    self.foo = bar
```

aaschaer (approved):
Looks good to me, part of me wants to somehow warn the user that there was `session_required_policies`, but it wasn't a string so something is wrong. But I can't think of a clean way to do that, and hopefully that issue would manifest in other ways like the session message not making sense.

I left a small style comment/question, but feel free to leave as is

src/globus_sdk/exc/err_info.py
sirosen: It's stopped looking odd to me to do the typing assignment like this, but that doesn't mean it isn't odd. It just means that I've done it too many times. 😉 

I think I'll make a modification here to be similar to your suggestion, but also add a warning log for the present-but-not-string case. (I don't think we can do more than that.)

src/globus_sdk/exc/err_info.py
```diff
@@ -64,9 +67,18 @@ def __init__(self, error_data: t.Dict[str, t.Any]):
...
-        )
+
+        # get str|None and parse as appropriate
+        self.session_required_policies: t.Optional[t.List[str]] = None
+        session_required_policies = data.get("session_required_policies")
+        if isinstance(session_required_policies, str):
+            self.session_required_policies = session_required_policies.split(",")
+        elif session_required_policies is not None:
+            log.warning(
+                "during ErrorInfo parsing, got unexpected type for "
```
kurtmckee, line 78: What are your thoughts on this phrasing?

```suggestion
                "During ErrorInfo instantiation, got unexpected type for "
```

src/globus_sdk/exc/err_info.py
```diff
@@ -64,9 +67,18 @@ def __init__(self, error_data: t.Dict[str, t.Any]):
...
+                "'session_required_policies'. "
+                "Expected 'str', but got '{type(session_required_policies)}'"
```
kurtmckee, line 80:
```suggestion
                f"Expected 'str', but got '{type(session_required_policies)}'"
```

kurtmckee (changes requested):

src/globus_sdk/exc/err_info.py
sirosen: D'oh! Thanks, excellent catch.

src/globus_sdk/exc/err_info.py
sirosen: I'm not sure I'm convinced that there's a difference between these, but I don't have any particular preference for the prior version. I'm taking the suggestion!

kurtmckee (approved):

PR #684, opened 2023-02-16
Add unit tests for TimerAPIError class

sirosen:
This also refactors some of the existing API error class unit tests in order to make the addition simpler and part of a more uniform and extensible style for testing various service-specific customizations.

### DIFF
--- a/tests/unit/errors/conftest.py
+++ b/tests/unit/errors/conftest.py
@@ -1,0 +1,84 @@
+import json
+from collections import namedtuple
+
+import pytest
+import requests
+
+_TestResponse = namedtuple("_TestResponse", ("data", "r", "method", "url"))
+
+
+def _mk_response(
+    data, status, method=None, url=None, headers=None, data_transform=None
+):
+    resp = requests.Response()
+
+    if data_transform:
+        resp._content = data_transform(data).encode("utf-8")
+    else:
+        resp._content = data.encode("utf-8")
+    resp.encoding = "utf-8"
+
+    if headers:
+        resp.headers.update(headers)
+
+    resp.status_code = str(status)
+    resp.reason = {
+        200: "OK",
+        400: "Bad Request",
+        401: "Unauthorized",
+        403: "Forbidden",
+        404: "Not Found",
+        500: "Server Error",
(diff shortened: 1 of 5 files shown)


tests/unit/errors/conftest.py
```diff
@@ -0,0 +1,84 @@
...
+
+    resp.status_code = str(status)
+    resp.reason = {
+        200: "OK",
+        400: "Bad Request",
+        401: "Unauthorized",
+        403: "Forbidden",
+        404: "Not Found",
+        500: "Server Error",
+    }.get(status, "Unknown")
```
kurtmckee, line 32: This dict [exists in the `http.client` module as `responses`](https://docs.python.org/3/library/http.client.html#http.client.responses):

```python
resp.reason = http.client.responses.get(status, "Unknown")
```

Note, though, that `500` translates to `"Internal Server Error"`.

tests/unit/errors/conftest.py
```diff
@@ -0,0 +1,84 @@
...
+    method = method or "GET"
+    url = url or "default-example-url.bogus"
+    resp.url = url
+    resp.request = requests.Request(method=method, url=url, headers=headers)
```
kurtmckee, line 36: Should the response headers be added to the headers of the request, too? This might be convenient behavior but I want to confirm whether `headers` should be split into customizable `response_headers` and `request_headers`.

tests/unit/errors/conftest.py
sirosen: 🥳

Truly a joy to replace a quick hack with something from the stdlib. I'll add that change.

tests/unit/errors/conftest.py
sirosen: I'll take a look. This was, as you may have guessed, largely a lift-and-shift.
I think maybe `headers=...` and `request_headers`, since this whole fixture is supposed to be a response builder.

kurtmckee: If something depends on this existing functionality, no need to pursue this at all, but I wanted to ask in case this is a good time opportunity to separate the two.

kurtmckee (approved):

PR #688, opened 2023-02-22
Add the RTD preview workflow

sirosen:
This GitHub Action appends a doc preview link to all PR descriptions.

See also: https://github.com/readthedocs/actions/tree/v1/preview

### DIFF
--- a/.github/workflows/readthedocs-pr-links.yaml
+++ b/.github/workflows/readthedocs-pr-links.yaml
@@ -1,0 +1,16 @@
+name: Read the Docs Pull Request Preview
+on:
+  pull_request_target:
+    types:
+      - opened
+
+permissions:
+  pull-requests: write
+
+jobs:
+  documentation-links:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: readthedocs/actions/preview@v1
+        with:
+          project-slug: "globus-sdk-python"


.github/workflows/readthedocs-pr-links.yaml
```diff
@@ -0,0 +1,16 @@
+name: Read the Docs Pull Request Preview
+on:
+  pull_request_target:
```
kurtmckee, line 3: Had to [read the docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) on why this was chosen vs simply `pull_request`. Summary for others' benefit:

> This event runs in the context of the base of the pull request, rather than in the context of the merge commit, as the pull_request event does. This prevents execution of unsafe code from the head of the pull request that could alter your repository or steal any secrets you use in your workflow. This event allows your workflow to do things like label or comment on pull requests from forks.

Nice work!

.github/workflows/readthedocs-pr-links.yaml
```diff
@@ -0,0 +1,16 @@
...
+    types:
+      - opened
```
kurtmckee, line 5: Should `paths` also be specified here? For example, if no files changed in `docs/` or `src/`, would it make sense to add add a preview link? I'm specifically thinking:

```yaml
paths:
  - "docs/"
  - "src/"
  - ".readthedocs.yml"
```

Would it also make sense to run this on `synchronize` as well?

sirosen: To the first question, restricting paths... maybe? But I'd rather not.
The class of PRs which don't impact is very narrow, and it would be easy for this to drift away from our codebase over time or miss things. e.g. We should add `setup.py` and probably `setup.cfg`. If we switch to something like hatch or flit, we'd need `pyproject.toml`, etc etc.
I think applying uniformly is easier, and at _worst_ adds a noisy link to a PR which didn't need it.

I don't think it's necessary to run on `synchronize` because the PR doc build done by RTD gets updated on sync. So the link from the initial PR opening remains accurate.

kurtmckee (approved):

ada-globus (approved):

PR #689, opened 2023-02-23
Enable use of MutableScopes in client methods

sirosen:
Two quick notes to set the stage:
- I ran into this being forbidden by `mypy` in globus-cli and was a bit surprised, which is what drives this change
- If you haven't seen the mypy "ignore tests" before, [the readme for these is here](https://github.com/globus/globus-sdk-python/tree/main/tests/non-pytest/mypy-ignore-tests)

---

The use of a MutableScope object, iterable thereof, or mixed iterable of MutableScope and str objects, was already supported by the underlying OAuth2 Flow Manager objects. However, the types used there were not "propagated upwards" to the auth client methods which wrap those manager objects.

Firstly, this change updates the annotations for the passthrough methods and adds a typing test which ensures that the correct types are allowed and defined by the ScopeCollectionType helper.

Secondarily, this adds the same capability to the
``oauth2_client_credentials_tokens`` method to make our scope interfaces more uniform. Unlike the other cases, this isn't just a typing change -- a new runtime behavior is being introduced here. As a result, a new test is wanted to validate that behavior. It turns out that this new test is the first test for this method, so fixture data must be included.

----
:books: Documentation preview :books:: https://globus-sdk-python--689.org.readthedocs.build/en/689/

### DIFF
--- a/src/globus_sdk/services/auth/client/confidential_client.py
+++ b/src/globus_sdk/services/auth/client/confidential_client.py
@@ -3,8 +3,8 @@ from __future__ import annotations
 import logging
 import typing as t
 
-from globus_sdk import exc, utils
-from globus_sdk._types import UUIDLike
+from globus_sdk import exc, scopes, utils
+from globus_sdk._types import ScopeCollectionType, UUIDLike
 from globus_sdk.authorizers import BasicAuthorizer
 from globus_sdk.response import GlobusHTTPResponse
 
@@ -51,7 +51,7 @@ class ConfidentialAppAuthClient(AuthClient):
 
     def oauth2_client_credentials_tokens(
         self,
-        requested_scopes: None | (str | t.Iterable[str]) = None,
+        requested_scopes: ScopeCollectionType | None = None,
     ) -> OAuthTokenResponse:
         r"""
         Perform an OAuth2 Client Credentials Grant to get access tokens which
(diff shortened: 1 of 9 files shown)


kurtmckee (approved):

src/globus_sdk/services/auth/client/confidential_client.py
```diff
@@ -75,13 +76,11 @@ def oauth2_client_credentials_tokens(
...
         """
         log.info("Fetching token(s) using client credentials")
-        requested_scopes = requested_scopes or DEFAULT_REQUESTED_SCOPES
-        # convert scopes iterable to string immediately on load
-        if not isinstance(requested_scopes, str):
-            requested_scopes = " ".join(requested_scopes)
-
+        requested_scopes_string: str = scopes.MutableScope.scopes2str(
+            requested_scopes or DEFAULT_REQUESTED_SCOPES
+        )
```
ada-globus, line 81: If I'm reading correctly, since the comparison is falsey rather than with `None` then a user requesting an empty set of scopes would get the default set of scopes. I highly doubt there's any reasonable use case for a user to want to do that, but I would find it a little surprising that it requests four scopes when I explicitly requested no scopes—rather than, say, throwing an error.

This is not a change from the existing behavior and in any case I don't think it warrants a change, but wanted to mention in case there was some nuance that it might be helpful for me to know!

ada-globus (approved):
This looks good! I had one minor comment/question, but it is tangential to the change here and shouldn't block merging.

src/globus_sdk/services/auth/client/confidential_client.py
kurtmckee: Great catch!

sirosen: I think there are a couple of locations which have this behavior on scopes.
Let me merge this PR to get the main improvement in and take a look at putting this behavior (`scopes2str` + error handling for `""` and maybe other bad values) into a helper somewhere.

PR #690, opened 2023-02-24
Catch empty and unspecified `requested_scopes`

sirosen:
Primarily, this changes a runtime behavior today to catch and error on an otherwise valid(ish) but undesirable condition. An empty scope string or collection of scopes would previously be replaced with the SDK's default scopes due to a loose "falsey" check. This is potentially confusing or even wrong. A user trying intentionally to pass `""` will find that their value is discarded in favor of the defaults. The new behavior is that any falsey non-None value, (e.g. `[]`) will be rejected with a usage error.

This is implemented in a new helper for handling the None check and default behavior. In that helper, we also now have a deprecation warning behavior inspired by the `sqlalchemy` 2.0 warnings and our generally positive experience integrating against and enabling/disabling those warnings.

The deprecation warning helpers define a warning class which inherits from `Warning` rather than `DeprecationWarning` so that it won't be suppressed by the default python warning filters. The warning is meant to be emitted via a helper which checks an environment variable to see if the "GLOBUS_SDK_V4_WARNINGS" has been set (enabling our v4 warnings). As a result, users won't see new warnings with current usage or while following our examples and tutorial, but they will be able to trivially enable the warnings by setting the env var.

New segments in the docs refer to the new env var and help users opt-in. Tests cover the warning behavior and explore the helper method a bit.

----
:books: Documentation preview :books:: https://globus-sdk-python--690.org.readthedocs.build/en/690/


src/globus_sdk/exc/warnings.py
```diff
@@ -0,0 +1,45 @@
...
+# therefore, inheriting from Warning is an "incorrect" but pragmatic choice
+#
+# in concert with the helpers below to make the warnings opt-in, this results in good
+# user-facing behavior
+class GlobusSDKLegacyBehaviorWarning(Warning):
+    def __init__(
+        self,
+        message: str,
+        *args: t.Any,
+        will_be_removed: str | None = "4.0.0",
```
sirosen, line 17: The behaviors around this are somewhat hard to access via warning emission, so I also considered making a multi-class hierarchy here, like
```python
class Foo(Warning):
    will_be_removed: str | None = None

class Bar(Foo):
    will_be_removed = "4.0.0"
```

but it seemed like overthinking things. I think we should feel okay about rewriting this class a bit if we need to in the future.

src/globus_sdk/exc/warnings.py
```diff
@@ -0,0 +1,45 @@
...
+        return
+
+    warnings.warn(message, GlobusSDKLegacyBehaviorWarning, stacklevel=stacklevel)
+
+
+def _deprecation_warnings_enabled() -> bool:
+    var = os.getenv("GLOBUS_SDK_V4_WARNINGS", "false").lower()
+    if var in ("true", "yes", "on", "1"):
+        return True
+    return False
```
sirosen, line 45: I looked at how sqlalchemy handles the warnings env var for inspiration, and it does this same kind of check and evaluation at import time. I'm okay with that if we prefer, and it would read like
```python
_DEPRECATION_WARNINGS_ENABLED = os.getenv("GLOBUS_SDK_V4_WARNINGS", "false").lower() in ("true", "yes", "on", "1")
```

But then that code can't really be tested and we just have to monkeypatch the module attribute in the tests.

src/globus_sdk/services/auth/_common.py
```diff
@@ -0,0 +1,31 @@
...
+
+def stringify_requested_scopes(requested_scopes: ScopeCollectionType | None) -> str:
+    if requested_scopes is None:
+        warn_deprecated(
+            "`requested_scopes` was not specified or was given as `None`. "
+            "A default set of scopes will be used, but this behavior is deprecated. "
+            "Specify an explicit set of scopes instead.",
+            stacklevel=3,
+        )
+        requested_scopes = _DEFAULT_REQUESTED_SCOPES
```
sirosen, line 24: We haven't talked about this, so there may be mixed feelings about deprecating this behavior, but I think it's for the best in the long run.

The current situation is that the interfaces for users look akin to
```python
tokens = do_login()
... = do_stuff_to_tokens(tokens)
tc = TransferClient(...)  # works
gc = GroupsClient(...)  # fails
```

and it's not obvious why this is the case.

Of course, we could keep growing and changing the default scopes, but that also has serious problems.

I think we ought to remove this in v4 and this puts us on a path in which we can start updating docs and educating users on what we want them to do (which is, I think, to use `globus_sdk.scopes`).

tests/unit/helpers/test_auth_scope_stringify.py
```diff
@@ -0,0 +1,62 @@
...
+        monkeypatch.delenv("GLOBUS_SDK_V4_WARNINGS", raising=False)
+    elif deprecation_mode == "disabled":
+        monkeypatch.setenv("GLOBUS_SDK_V4_WARNINGS", "false")
+    elif deprecation_mode == "enabled":
+        monkeypatch.setenv("GLOBUS_SDK_V4_WARNINGS", "true")
+    else:
+        raise NotImplementedError
+
+    # NOTE: the unset variant below will change branches when the default warning
+    # behavior changes
```
sirosen, line 49: Meaning, if we make warnings on by default (which I anticipate us doing someday), then we will want to swap `"unset"` to go with `"enabled"`, not `"disabled"`.

src/globus_sdk/exc/warnings.py
```diff
@@ -0,0 +1,47 @@
...
+# therefore, inheriting from Warning is an "incorrect" but pragmatic choice
+#
+# in concert with the helpers below to make the warnings opt-in, this results in good
+# user-facing behavior
+class GlobusSDKLegacyBehaviorWarning(Warning):
+    def __init__(
+        self,
+        message: str,
+        *args: t.Any,
+        will_be_removed: str | None = "4.0.0",
```
kurtmckee, line 19: My gut is that there should be no default for the version here; it should always be explicitly passed in at the location the warning is being issued.

It may also be appropriate to disallow `None` values.

src/globus_sdk/exc/warnings.py
```diff
@@ -0,0 +1,47 @@
...
+                f"{self.message} (support for this will be removed in "
+                f"globus-sdk version {self.will_be_removed})"
+            )
+
+
+def warn_deprecated(message: str, stacklevel: int = 2) -> None:
+    if not _deprecation_warnings_enabled():
+        return
+
+    warnings.warn(message, GlobusSDKLegacyBehaviorWarning, stacklevel=stacklevel)
```
kurtmckee, line 40: As part of the exercise to ensure the version is always passed, the call syntax will need to change so that the argument to the `message` parameter is an instance of the class, so the version can be explicitly passed:

```suggestion
def warn_deprecated(message: str, will_be_removed: str, stacklevel: int = 2) -> None:
    if not _deprecation_warnings_enabled():
        return

    warnings.warn(GlobusSDKLegacyBehaviorWarning(message, will_be_removed=will_be_removed), stacklevel=stacklevel)
```

src/globus_sdk/exc/warnings.py
kurtmckee: I agree with this design. It's testable, and cheap.

src/globus_sdk/exc/warnings.py
```diff
@@ -0,0 +1,47 @@
+from __future__ import annotations
+
+import os
+import typing as t
+import warnings
+
+
+# this is (by meaning) a type of DeprecationWarning, but because python suppresses
+# DeprecationWarning by default, most users would never see it if we inherited from that
```
kurtmckee, line 9: > by default, most users would never see it

To be clear, that is the point. _Developers_ are the ones who should be seeing these warnings and addressing them as they will. By inheriting from `Warning` we are guaranteeing that those developers' _users_ have to see these warnings. I disagree with this outcome.

I think this should be a `DeprecationWarning` so we're not forcing warning messages upon our developers' users.

kurtmckee (changes requested):

src/globus_sdk/exc/warnings.py
sirosen: What if I changed the way that this is declared to be
```
GlobusSDKLegacyBehaviorV4Warning
```
and hardcoded the `4.0.0` as part of the message format?

I don't see any reason to be deprecating things today which we'll remove in any other version. And when we release v4, we can replace this with the V5 variant!

sirosen: Oh, it could also be named more like `GlobusSDKRemovedInV4Warning` or other names like that.

src/globus_sdk/exc/warnings.py
sirosen: This is similar to the argument which has underpinned the decision to ignore `DeprecationWarning` by default in the interpreter.

While I respect this position, I have a significant concern about this is how it impacts users who are writing relatively small scripts and have no testing mechanisms. Such a user will typically _never_ run the interpreter with `-W` filters specified.

`Warning` as the base + the toggle env var ensures that these users will see the warnings when we flip the default for the env var flag.
The big alternative I see is to use `DeprecationWarning` as the base and remove the env var (why bother suppressing the warning explicitly if it's a `DeprecationWarning`?).

i.e. I think that for our user-base, the developer/user distinction is less clear than it is with some other libraries. We can and should discuss this further. I could be convinced, in part because there's a question of how often and why/when scripting users would be updating to the newest library version. And the end-goal is to have the smoothest upgrade experience possible.

src/globus_sdk/exc/warnings.py
derek-globus:
```suggestion
        removal_version: str | None = "4.0.0",
```

Not sure if my suggestion is perfect but I don't think `will_be_removed` conveys what this thing is well enough.

src/globus_sdk/exc/warnings.py
derek-globus: I like the no default suggestion.

Hardcoding a specific version of `4.0.0` (either via classname or default string) doesn't save much in terms of keystrokes (the only benefit I can really see) and in return introduces a potential error case.

src/globus_sdk/exc/warnings.py
```diff
@@ -0,0 +1,47 @@
...
+# therefore, inheriting from Warning is an "incorrect" but pragmatic choice
+#
+# in concert with the helpers below to make the warnings opt-in, this results in good
+# user-facing behavior
+class GlobusSDKLegacyBehaviorWarning(Warning):
```
derek-globus, line 14: Did you intentionally not name it `GlobusSdkDeprecationWarning` since we don't extend `DeprecationWarning`?

`Deprecation` more accurately states "we're taking this thing away" than `LegacyBehavior` imo

src/globus_sdk/services/auth/_common.py
derek-globus: Agree with this change.

I do think it'll be useful to keep in mind (read out-of-scope for this PR) as we go forward how customers should be referencing common groups of scopes. Are there enough common groups of scopes that aren't static scopes to make that into a useful SDK feature?

Mostly just concerned about SDK customers needing to learn too much down the scope trail before they can start using a simple api invocation.

src/globus_sdk/services/auth/_common.py
```diff
@@ -0,0 +1,31 @@
...
+
+    requested_scopes_string: str = MutableScope.scopes2str(requested_scopes)
+    if requested_scopes_string == "":
+        raise GlobusSDKUsageError(
+            "requested_scopes cannot be the empty string or empty collection"
+        )
```
derek-globus, line 30: Are we 100% there isn't a use case for this?

Feels like the type of thing where a customer will come back asking us to [re-enable spacebar heating](https://xkcd.com/1172/) and I want to be sure there aren't any actual valid usages for this edge case.

derek-globus (commented):
I have a broader concern around the deprecation warning path taken here (defaulting to False) which is that I don't think almost anybody will know to enable this env variable.

Maybe if they are a very large power user they'll read through all the docs & see one instructing them to set it but I don't think it covers most the people who we'd want to actually see deprecation warnings.

src/globus_sdk/services/auth/_common.py
sirosen: A user or two could, in theory, ask for us to re-enable support for `requested_scopes=""`, but my feeling is that this has been a bug for as long as we've treated that as valid, and we're fixing it. The incompatible change is that we used to treat `""` or `[]` as meaning the default as well as `None`, but that wasn't really intentional as far as I can tell.

I don't mean that there's no risk that a user is going to be bitten by this, but I'm just not inclined to treat the behavior of `requested_scopes=[]` as covered by our semver "contract" with the user.

To clarify why this is a bug IMO: a user could reasonably expect that `requested_scopes=""` means "request no scopes". Such a usage is never valid, and we silently converted it to valid usage.

They weren't really getting `""` as their scope string. They were getting the defaults. The better way for them to get the defaults is either to pass no scopes or to pass `None` explicitly.

---

This possibility -- that the _computed_ scope string is the empty string -- is actually a newly introduced situation in this change.

The only way I know of to get `""` out of the existing code in `main` is to define an iterable which is empty but which bools as "True". e.g.
```python
class MyList(list):
    def __bool__(self): return True

x = MyList()  # try in an interpreter, `x or "foo"` is []
...

requested_scopes=MyList()
```

At that point, I'm like "yeah, you got me! good job! gold star! go away" 😹

src/globus_sdk/exc/warnings.py
sirosen: It is intentional, and you guessed right about the naming, but both you and Kurt have raised concerns about how it plays out.
I plan for us to discuss this at the SDK meeting today, so we don't have to get too deep into it right now.

The one thing which I think is useful to know ahead of time is that the interpreter has a default filter for warnings which suppresses `DeprecationWarning` (and subclasses thereof). So you will only see a `DeprecationWarning` if you explicitly adjust the warning filters or use a program like `pytest` which sets warning filters for you.

src/globus_sdk/services/auth/_common.py
derek-globus: Ah yeah, I didn't think correctly about how this was handled previously when writing this comment.

Happy to resolve.

sirosen:
In response to our discussion, I've rewritten the warning behavior to simply use
```python
class RemovedInV4Warning(DeprecationWarning): pass
```

and reduced the `warn_deprecated` helper to emitting a warning of this class. (I still like the helper because it abstracts away any decisions about how/why we name and structure this.)

One of the unexpected side-effects (benefits) of this change is that a few SDK tests which violated this usage were flagged by pytest and needed handling.

---

I'm going to go back and mark as resolved the threads which I think this closes out.

src/globus_sdk/exc/warnings.py
sirosen: One of the key points which got me onboard with using `DeprecationWarning` as the base, and getting the standard Python behavior, is that doing so provides the best experience for anyone who is doing their work and using pytest to test their code.

src/globus_sdk/exc/warnings.py
sirosen: If we were keeping the notion of a warning which could take the version number as an input, I'm convinced that making it a required parameter with no default would be better.

But I've actually come to think that not having any notion of trying to report out the version number in this way is better, at least for now. I'm happy to revisit or discuss further, but for now I'm marking this as resolved.

src/globus_sdk/exc/warnings.py
sirosen: 👍 This is what we'd do if we were continuing down this path, and I like it.
However, my latest draft simply makes `warn_deprecated(message)` raise a `RemovedInV4Warning`, which I like better for its brevity and clarity.

As such, I'm marking this as resolved without taking the suggestion, even though I like what's being suggested here.

src/globus_sdk/exc/warnings.py
sirosen: I think "removal_version" may be a better name, but after our discussion and doing a refactor, I'm even happier with where this ended up (with a simple empty warning class whose name encodes this notion).

kurtmckee (approved):

PR #692, opened 2023-02-27
Switch to GHA automated package publishing

sirosen:
We run a workflow to publish to test-pypi on tag, and a workflow to publish to pypi on GitHub release.

The tox environment for publishing has been removed (it can always be restored if it turns out to be necessary in the future).

`make release` has been replaced with `make tag-release`, which is slightly more clear about where this step fits into the workflow, and no longer tries to run the publishing step.

The `RELEASING.md` doc has been updated to accord with these details.

----
:books: Documentation preview :books:: https://globus-sdk-python--692.org.readthedocs.build/en/692/

### DIFF
--- a/.github/workflows/publish_to_test_pypi.yaml
+++ b/.github/workflows/publish_to_test_pypi.yaml
@@ -1,0 +1,25 @@
+name: Publish Test PyPI Release
+
+on:
+  push:
+    tags: ["*"]
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+          cache: pip
+
+      - run: python -m pip install build
+      - run: python -m build .
+
+      - name: Publish to TestPyPI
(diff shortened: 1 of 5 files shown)


.github/workflows/publish_to_test_pypi.yaml
```diff
@@ -0,0 +1,25 @@
+name: Publish Test PyPI Release
+
+on:
+  push:
+    tags: ["*"]
```
ada-globus, line 5: Are we sure we don't want to scope this a little more narrowly?

ada-globus (approved):
This looks good! Had one question, but otherwise I think this is GTG, so marking approved.

.github/workflows/publish_to_test_pypi.yaml
sirosen: I considered it. We could use `*.*.*` or something else like that.
But we don't have any other tags that we create in this repo as part of our normal operating procedures, and since this workflow only publishes to test-pypi, I shrugged it off for now.

I'm happy to revisit this pattern in the future, but I didn't want to block here on figuring out what kinds of regexes are supported.

PR #694, opened 2023-02-28
Add sphinx directive to list _testing data

sirosen:
This adds a directive to our internal sphinx extension to list the data visible in `_testing`.

It uses the same helper as `AutoMethodList`, but
`EnumerateTestingFixtures` checks those fixtures against the `_testing` registry and emits ReST for those which have defined test cases.

The listing is done on a new method_list page, which follows the existing norms and forms of the ("alpha") testing docs.

----
:books: Documentation preview :books:: https://globus-sdk-python--694.org.readthedocs.build/en/694/

### DIFF
--- a/src/globus_sdk/_testing/models.py
+++ b/src/globus_sdk/_testing/models.py
@@ -161,6 +161,9 @@ class ResponseSet:
     ) -> t.Iterator[RegisteredResponse | ResponseList]:
         return iter(self._data.values())
 
+    def cases(self) -> t.Iterator[str]:
+        return iter(self._data.keys())
+
     def activate(
         self,
         case: str,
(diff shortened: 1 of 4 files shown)


sirosen:
Results are visible here:
https://globus-sdk-python--694.org.readthedocs.build/en/694/testing/method_list.html

src/globus_sdk/_testing/models.py
```diff
@@ -161,6 +161,9 @@ def __iter__(
     ) -> t.Iterator[RegisteredResponse | ResponseList]:
         return iter(self._data.values())
 
+    def cases(self) -> t.Iterator[str]:
+        return iter(self._data.keys())
```
kurtmckee, line 165: The call to `.keys()` is not required. Iterating over the dict itself returns the same iterator (a `dict_keyiterator` object).

```suggestion
        return iter(self._data)
```

kurtmckee (changes requested):
This is awesome! I've added just one request.

src/globus_sdk/_testing/models.py
sirosen: And so it shall be! 😄

PR #700, opened 2023-03-12
Begin using tabs to refactor method docs

sirosen:
This is a combined set of several minor changes which support a new way of writing docs for client methods. The overall goal is to make use of sphinx-inline-tabs to present several pieces of information in a way which is compact by default, expandable, and allows us to put some more structure to what is currently a long scroll of text.

To wit:
- include the sphinx-inline-tabs extension for sphinx
- cleanup our custom sphinx directives to use docutils' built-in option parsing
- add new custom directives for several small functions:
  - expandtestfixture: expands a JSON or text body from a test fixture as a code-block, for use as example data
  - extdoclink: link to an external doc location (only supports docs.globus.org right now)
  - paginatedusage: emit the standard paginated usage info text
- select two methods which are updated to use the new documentation capabilities: TransferClient.get_submission_id and SearchClient.search. These demonstrate the behavior and what is now possible

It should be noted that `get_submission_id` has no usage example while `search` does, that `search` is paginated while `get_submission_id` is not, and that both refer to docs site docs and have relevant test fixture data for examples. Therefore, they demonstrate the flexibility of this setup to handle different kinds of methods, but also share some commonalities.

The original drive of this work was `expandtestfixture`, but once that was achieved, it demands some solution for better presentation, like tabs. Once tabs are in place, they conflict with the default behavior for the pagination decorator (which appends to the docstring), so some accommodation is needed. A similar need precipitates the refactor of the external doc decorator into a sphinx directive.

----
:books: Documentation preview :books:: https://globus-sdk-python--700.org.readthedocs.build/en/700/

### DIFF
--- a/src/globus_sdk/_sphinxext.py
+++ b/src/globus_sdk/_sphinxext.py
@@ -232,7 +232,7 @@ class ExpandTestingFixture(AddContentDirective):
             for line in output_lines:
                 yield f"    {line}"
         elif response.body is not None:
-            yield ".. code-block::"
+            yield ".. code-block:: text"
             yield ""
             output_lines = response.body.split("\n")
             for line in output_lines:
@@ -254,7 +254,9 @@ class PaginatedUsage(AddContentDirective):
         yield "To use the paginated variant, give the same arguments as normal, "
         yield "but prefix the method name with ``paginated``, as in"
         yield ""
-        yield f">>> client.paginated.{self.arguments[0]}(...)"
+        yield ".. code-block:: pycon"
+        yield ""
+        yield f"    >>> client.paginated.{self.arguments[0]}(...)"
         yield ""
         yield "For more information, see"
         yield ":ref:`how to make paginated calls <making_paginated_calls>`."
--- a/src/globus_sdk/_sphinxext.py
+++ b/src/globus_sdk/_sphinxext.py
@@ -5,10 +5,11 @@ A Globus SDK Sphinx Extension for Autodoc of Class Methods
 from __future__ import annotations
 
 import inspect
+import json
 from pydoc import locate
 
 from docutils import nodes
-from docutils.parsers.rst import Directive
+from docutils.parsers.rst import Directive, directives
 from docutils.statemachine import ViewList
 from sphinx.util.nodes import nested_parse_with_titles
 
@@ -18,19 +19,22 @@ def _extract_known_scopes(scope_builder_name):
     return sb.scope_names
 
 
+def _locate_class(classname):
+    cls = locate(classname)
+    if not inspect.isclass(cls):
+        raise RuntimeError(
+            f"uh-oh, {classname} is not a class name? type(classname)={type(cls)}"
+        )
+    return cls
+
+
 def _classname2methods(classname, include_methods):
     """resolve a class name to a list of (public) method names
     takes a classname and a list of method names to avoid filtering out"""
-    klass = locate(classname)
-    if not inspect.isclass(klass):
-        raise RuntimeError(
-            "uh-oh, {} is not a class name? type(classname)={}".format(
-                classname, type(klass)
-            )
-        )
+    cls = _locate_class(classname)
 
     # get methods of the object as [(name, <unbound method>), ...]
-    methods = inspect.getmembers(klass, predicate=inspect.isfunction)
+    methods = inspect.getmembers(cls, predicate=inspect.isfunction)
 
     def _methodname_is_good(m):
         if m in include_methods:
@@ -39,7 +43,7 @@ def _classname2methods(classname, include_methods):
         if m.startswith("_"):
             return False
         # filter out any inherited methods which are not overloaded
-        if m not in klass.__dict__:
+        if m not in cls.__dict__:
             return False
         return True
 
@@ -85,15 +89,15 @@ class AddContentDirective(Directive):
 class AutoMethodList(AddContentDirective):
     has_content = False
     required_arguments = 1
-    optional_arguments = 1
+    optional_arguments = 0
+    option_spec = {"include_methods": directives.unchanged}
 
     def gen_rst(self):
         classname = self.arguments[0]
 
         include_methods = []
-        for arg in self.arguments[1:]:
-            if arg.startswith("include_methods="):
-                include_methods = arg.split("=")[1].split(",")
+        if "include_methods" in self.options:
+            include_methods = self.options["include_methods"].strip().split(",")
 
         yield ""
         yield "**Methods**"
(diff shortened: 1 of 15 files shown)


src/globus_sdk/_sphinxext.py
```diff
@@ -170,7 +174,96 @@ def gen_rst(self):
...
+    has_content = False
+    required_arguments = 1
+    optional_arguments = 0
+
+    def gen_rst(self):
+        yield "This method supports paginated access. "
+        yield "To use the paginated variant, give the same arguments as normal, "
+        yield "but prefix the method name with ``paginated``, as in"
+        yield ""
+        yield f">>> client.paginated.{self.arguments[0]}(...)"
```
kurtmckee, line 257: Should this be wrapped in a code block?

```suggestion
        yield ".. code-block:: pycon"
        yield ""
        yield f"    >>> client.paginated.{self.arguments[0]}(...)"
```

src/globus_sdk/_testing/data/transfer/_common.py
```diff
@@ -1,5 +1,10 @@
 import uuid
 
-SUBMISSION_ID = str(uuid.UUID(int=2020))
-ENDPOINT_ID = str(uuid.UUID(int=1234))
-TASK_ID = str(uuid.UUID(int=1001))
+
+def _as_uuid(s: str) -> str:
+    return str(uuid.UUID(int=int(s, 36)))
```
kurtmckee, line 5: :heart:

src/globus_sdk/paging/base.py
```diff
@@ -161,7 +161,10 @@ def decorate(func: C) -> C:
         as_paginated._paginator_items_key = items_key
         as_paginated._paginator_params = paginator_params
 
-        func.__doc__ = f"""{func.__doc__}
+        # do not append doc if the paginatedusage sphinx directive is in use
+        if ".. paginatedusage::" not in func.__doc__:
+            # but if not, append the desired text
+            func.__doc__ = f"""{func.__doc__}
```
kurtmckee, line 167: Although untouched by this PR, this string may need to be updated to use a `.. code-block:: pycon` as noted above.

Also, it may be useful to consider using [`.. rubric:: Paginated Usage`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-rubric) instead of simply marking the text as bold. The benefit is that it is rendered as a paragraph heading (so it _should_ be linkable) but won't generate a TOC entry.

src/globus_sdk/_sphinxext.py
```diff
@@ -170,7 +174,96 @@ def gen_rst(self):
...
+        if response.json is not None:
+            yield ".. code-block:: json"
+            yield ""
+            output_lines = json.dumps(
+                response.json, indent=2, separators=(",", ": ")
+            ).split("\n")
+            for line in output_lines:
+                yield f"    {line}"
+        elif response.body is not None:
+            yield ".. code-block::"
```
kurtmckee, line 235: If unspecified, code blocks can fallback to defaults. I recommend disabling syntax highlighting by specifying either "text" or "none".

```suggestion
            yield ".. code-block:: text"
```

kurtmckee (changes requested):

src/globus_sdk/paging/base.py
sirosen: Both points are good and well taken. I'll switch to pycon for sure.
Regarding the rubric rule, I want to take a look at how it looks to make sure I'm okay with it, but assuming it renders similarly to the existing bold text, I'll change that too.

src/globus_sdk/_testing/data/transfer/_common.py
sirosen: I learned this base 36 usage trick from someone. 😉

kurtmckee (approved):

src/globus_sdk/paging/base.py
sirosen: I'm refraining from doing the rubric change, at least for now.
A quick experiment shows that furo styles that as uppercase text, which makes it look a little inharmonious with the surrounding (bold text) "sections" in some docs. But I've amended to switch the code block here to `pycon` styling.

src/globus_sdk/services/search/client.py
```diff
@@ -174,33 +174,44 @@ def post_search(
...
-        >>>      "order": "asc"}
-        >>>   ]
-        >>> }
-        >>> search_result = sc.post_search(index_id, query_data)
+            .. tab-item:: Example Usage
+
+                .. code-block:: pycon
+
+                    >>> sc = globus_sdk.SearchClient(...)
+                    ... query_data = {
```
kurtmckee, line 184:
```suggestion
                    >>> query_data = {
```

src/globus_sdk/services/search/client.py
```diff
@@ -174,33 +174,44 @@ def post_search(
...
+                    ...         {
+                    ...             "name": "Publication Date",
+                    ...             "field_name": "path.to.date",
+                    ...             "type": "date_histogram",
+                    ...             "date_interval": "year",
+                    ...         }
+                    ...     ],
+                    ...     "sort": [{"field_name": "path.to.date", "order": "asc"}],
+                    ... }
+                    ... search_result = sc.post_search(index_id, query_data)
```
kurtmckee, line 203:
```suggestion
                    >>> search_result = sc.post_search(index_id, query_data)
```

src/globus_sdk/services/search/client.py
```diff
@@ -237,10 +245,25 @@ def scroll(
...
+        .. tab-set::
 
-        >>> sc = globus_sdk.SearchClient(...)
-        >>> scroll_result = sc.scroll(index_id, {"q": "*"})
+            .. tab-item:: Example Usage
+
+                .. code-block:: pycon
+
+                    >>> sc = globus_sdk.SearchClient(...)
+                    ... scroll_result = sc.scroll(index_id, {"q": "*"})
```
kurtmckee, line 255:
```suggestion
                    >>> scroll_result = sc.scroll(index_id, {"q": "*"})
```

kurtmckee (changes requested):

src/globus_sdk/services/search/client.py
```diff
@@ -174,33 +174,44 @@ def post_search(
...
-        >>>   ],
-        >>>   "sort": [
-        >>>     {"field_name": "path.to.date",
-        >>>      "order": "asc"}
-        >>>   ]
-        >>> }
-        >>> search_result = sc.post_search(index_id, query_data)
+            .. tab-item:: Example Usage
+
+                .. code-block:: pycon
```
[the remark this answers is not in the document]
sirosen: I agree, but I think this PR is starting to "wander off course".

Why don't we get the pycon blocks fixed up, as you've suggested, and we can possibly switch them to python blocks next. (Might be good for copy-paste too.)

src/globus_sdk/services/search/client.py
sirosen: Thanks for this batch of catches on the pycon blocks!
blacken-docs was unhappy with everything being `>>>`, and I started fixing them up with some find-replace for `>>>` to `...`. But then I realized that I had gone too far and went _back_ to fix several cases back to `>>>`, and clearly I missed a couple.

kurtmckee (approved):

PR #701, opened 2023-03-12
Refactor dev dependency handling

sirosen:
#700 is failing CI because it introduces a documentation dependency which breaks the py3.7 build.

I saw 4 solutions:
1. Remove the dependency
2. Apply a `python_version>=3.8` marker to the dependency
3. Add a separate 'docs' extra alongside 'dev' to separate out the dependency
4. this PR

I'll note that I considered (1) and (3) unacceptable. (1) gives up on a valuable doc improvement, and (3) takes us further down a bad road (extras to manage our dev needs).

~This PR adds a commit on top of the one in #700 that which solves the problem by separating the documentation dependencies, typing dependencies, and test dependencies into separate categories. Commit message follows.~
This has been rebased to separate it from #700, but was previously based on top of that PR.

---

[Switch dev deps from an extra to pip-compile/multi](https://github.com/globus/globus-sdk-python/pull/701/commits/76f4e38991bd1137deaf7d9a762824b8963d668f)

All dev dependencies are herein moved out of the `globus-sdk[dev]` extra, and into files in a `requirements/` dir.
These files are managed using pip-compile-multi and pip-compile, which can be run near-trivially with a new tox env: `tox r -e freezedeps`.

The pip-compile-multi workflow has been put forth in the pallets projects where it is well liked for providing build reproducibility, and is used by several other community projects as well. However, for our "mindeps" environments, pip-compile-multi is unsuitable because it explicitly protects against mismatched dependency versions.

Why do this rearrangement? And why now?

To the first question, this removes our development needs from published package metadata (extras). This is good for long-term maintenance: it simplifies what we are publishing to only what is useful, necessary, and supported, and the change means that the various collections of dev dependencies do not need to be installable into any singular virtualenv (e.g. 'docs' needs can conflict with 'test').

To the second question, an issue has arisen with the introduction of sphinx-inlne-tabs for the doc build. It currently requires python3.8+, but we are still testing on python3.7 . This needs some resolution, and this one is best because it provides us other benefits.

Some additional details shake loose here:
- unnecessary cast due to 'responses' update (mypy)
- The requests version can't be kept at 2.19.1 by this pinning technique for the mindeps build because that would also pin back the responses version and break tests. For now this is handled by using a slightly newer version, which might be for the best in any case

----
:books: Documentation preview :books:: https://globus-sdk-python--701.org.readthedocs.build/en/701/


kurtmckee:
As a part of the effort this sprint for small PR's, I'm not going to review this until the changes can be isolated.

One way to handle this is by pushing branches to this org's repo, rather than a separate repo, and opening dependent PR's like this against their parent git branches.

sirosen:
I'm going to rebase this to remove #700 from it, and that one can then be rebased after this merges if we go forward with it.

I thought I was following the spirit of small PRs here, since the two changes are isolated and this one exists as a singular commit. But in the cold light of day, having them mixed in GitHub may be a big part of what we wanted to solve.
_If_ the rebase doesn't resolve our desire for a smaller unit-of-review, let me know how you think we can break it down further. I'm not seeing an easy way to break this particular refactor into smaller chunks.

sirosen:
The rebase is done. It still looks like a large change because it introduces the pinned requirements files, but this should be easier to read now.

kurtmckee:
Rebasing got me where I wanted to be! :partying_face: :pray:

requirements/README.rst
```diff
@@ -0,0 +1,8 @@
+Requirements Data
+-----------------
+
+These ``*.txt`` data are autogenerated from the ``*.in`` and ``*.isolated``
+source files.
+
+Modify the sources manually, then regenerated the "pinned" or "locked"
```
kurtmckee, line 7:
```suggestion
Modify the sources manually, then regenerate the "pinned" or "locked"
```

requirements/test-mindeps.isolated
```diff
@@ -0,0 +1,23 @@
...
+# this version of requests is newer than the minimum in package data
+# however, if we pin back to 2.19.1, we break 'responses' behaviors used by the
+# testsuite
+requests==2.22.0
+
+pyjwt==2.0.0
+cryptography==3.3.1
+typing_extensions==4.0
+
+## replicate the test dependencies here, rather than including the test.in file
```
kurtmckee, line 11: Is this comment correct? It looks like the `test.in` file is getting included at the bottom of the file.

requirements/test-mindeps.isolated
```diff
@@ -0,0 +1,23 @@
...
+##   Package idna was resolved to different versions in different environments: 2.7 and 3.4
+##   Traceback (most recent call last):
+##     ...
+##   RuntimeError: Please add constraints for the package version listed above
+##
+#pytest
+#coverage
+#pytest-xdist
+#responses
+-r test.in
```
kurtmckee, line 23: Could this be at the top of the file? Even if pip-compile-multi / pip-tools interprets this correctly, I'd prefer for this file to look like its explicit versions are overriding the contents of `test.in`.

requirements/test-mindeps.txt
```diff
@@ -0,0 +1,56 @@
+#
+# This file is autogenerated by pip-compile with Python 3.11
+# by the following command:
+#
+#    pip-compile --output-file=/home/sirosen/dev/sdk-and-cli/sdk/requirements/test-mindeps.txt --resolver=backtracking /home/sirosen/dev/sdk-and-cli/sdk/requirements/test-mindeps.isolated
```
kurtmckee, line 5: Is there to set this up so the command uses relative paths?

Otherwise, many lines in this file (not just this one) will change each time another dev generates this file, because there are many absolute paths listed throughout the file.

requirements/typing-mindeps.txt
```diff
@@ -0,0 +1,43 @@
+#
+# This file is autogenerated by pip-compile with Python 3.11
+# by the following command:
+#
+#    pip-compile --output-file=/home/sirosen/dev/sdk-and-cli/sdk/requirements/typing-mindeps.txt --resolver=backtracking /home/sirosen/dev/sdk-and-cli/sdk/requirements/typing-mindeps.isolated
```
kurtmckee, line 5: This file is also affected by absolute paths.

tox.ini
```diff
@@ -102,6 +94,19 @@ commands_pre = rm -rf dist/
...
+# we're doing with the mindeps environments
+description = freeze development dependencies using pip-compile and pip-compile-multi
+skip_install = true
+deps =
+    pip-tools
+    pip-compile-multi
+commands =
+    pip-compile-multi
+    pip-compile -q --resolver=backtracking {toxinidir}/requirements/test-mindeps.isolated -o {toxinidir}/requirements/test-mindeps.txt
+    pip-compile -q --resolver=backtracking {toxinidir}/requirements/typing-mindeps.isolated -o {toxinidir}/requirements/typing-mindeps.txt
```
kurtmckee, line 108: This is where the absolute paths are getting introduced. Hopefully it's possible to strip the leading `{toxinidir}/` from this command and rely on the current working directory! :crossed_fingers:

kurtmckee (changes requested):

requirements/test-mindeps.isolated
sirosen: Oops... This is a development artifact. I was trying it both ways with pip-compile and pip-compile-multi.
pip-compile-multi was able to handle things if we listed the packages twice, which was interesting behavior to explore but not something I was very comfortable using.

I'll remove the whole commented out section.

requirements/test-mindeps.txt
sirosen: Oh boy, I hadn't noticed that... I should find a way of doing this which works with relative paths.
I'm sure it's possible -- I'll dig through the `pip-compile` options to see what we've got on hand.

tox.ini
sirosen: Nice catch!
I hadn't even considered that tox would expand that and that would be the cause of the problem.

requirements/test-mindeps.txt
sirosen: With your recommendation to remove the `{toxinidir}` variable, this resolved. Thanks for calling that out.

tox.ini
```diff
@@ -102,6 +94,19 @@ commands_pre = rm -rf dist/
 changedir = tests/non-pytest/poetry-lock-test
 commands = poetry lock
 
+[testenv:freezedeps]
```
kurtmckee, line 97: Guess what! You can modify the header output by modifying this environment variable:

### `CUSTOM_COMPILE_COMMAND`

https://github.com/jazzband/pip-tools/blob/6735e6271f489a73f490a1b28dd47f2616417e73/piptools/writer.py#L121-L134

This will allow you to put the tox command directly into these files. I think you can then remove pip-compile-multi and just use pip-tools exclusively.

sirosen: I found that this isn't supported by pip-compile-multi, the pip-compile-multi replacement option is for setting the whole header, and overall I'm not loving my experience trying to get this set for both tools.
I'm going to reduce this down to `pip-compile`-only and set the `CUSTOM_COMPILE_COMMAND`.

kurtmckee (changes requested):

requirements/docs.txt
```diff
@@ -0,0 +1,77 @@
...
+beautifulsoup4==4.11.2
+    # via furo
+certifi==2022.12.7
+    # via requests
+charset-normalizer==3.1.0
+    # via requests
+docutils==0.19
+    # via sphinx
+furo==2022.12.7
+    # via -r /home/sirosen/dev/sdk-and-cli/sdk/requirements/docs.in
```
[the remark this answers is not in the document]
sirosen: Grrr! I think this is because my wrapper uses pathlib paths (which str-ize as their abspaths).
I'll make them relative to cwd and fix this.

scripts/freezedeps.py
```diff
@@ -0,0 +1,47 @@
...
+        print(f"{source} -> {dest}")
+        subprocess.run(
+            [
+                sys.executable,
+                "-m",
+                "piptools",
+                "compile",
+                "-q",
+                "--resolver=backtracking",
+                str(source),
```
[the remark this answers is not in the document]
sirosen: Shame on me for replying to your comment above without reading this one! Yep, we're on the same wavelength. Working it up now.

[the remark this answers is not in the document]
sirosen: Since I set `cwd=REQS_DIR`, it needed to be `relative_to(REQS_DIR)`, but I just did this and think the result is what we'd like it to be.

kurtmckee (approved):

PR #705, opened 2023-03-14
Re-root where pip-compile runs from

kurtmckee:
This aligns with how Dependabot will submit changes.

----
:books: Documentation preview :books:: https://globus-sdk-python--705.org.readthedocs.build/en/705/


sirosen (approved):
Two quick notes:
- we'll see if we end up turning dependabot off, but this is definitely non-harmful
- I like seeing how this changed with a dir change + someone else running it, since it confirms that we eliminated a lot of the person-specific details we were concerned with seeing

PR #708, opened 2023-03-16
Improve extdoclink and convert GroupsClient docs

sirosen:
Improve the service selection feature of ExternalDocLink and list known service doc URLs within it.
"mode" is better named as "service". Add `:base_url:` as an optional override, and define `:service:` settings for the default base_url.

This allows for usage of

    .. extdoclink:: Some text
        :service: groups
        :ref: foo

to produce a link to Groups service documentation.

Apply these links, in dedicated API Info sections, to all GroupsClient method docs. This eliminates one usage of the `doc_api_method` docstring modifier.

----
:books: Documentation preview :books:: https://globus-sdk-python--708.org.readthedocs.build/en/708/

### DIFF
--- a/src/globus_sdk/_sphinxext.py
+++ b/src/globus_sdk/_sphinxext.py
@@ -57,6 +57,9 @@ def _is_paginated(func):
 class AddContentDirective(Directive):
     # Directive source:
     #   https://sourceforge.net/p/docutils/code/HEAD/tree/trunk/docutils/docutils/parsers/rst/__init__.py
+    #
+    # for information on how to write directives, see also:
+    #   https://docutils.sourceforge.io/docs/howto/rst-directives.html#the-directive-class
     def gen_rst(self):
         return []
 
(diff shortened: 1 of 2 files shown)


src/globus_sdk/_sphinxext.py
```diff
@@ -180,30 +183,34 @@ class ExternalDocLink(AddContentDirective):
...
+        "ref": directives.unchanged_required,
+    }
 
     def gen_rst(self):
         message = self.arguments[0].strip()
 
-        mode = "default"
-        if "mode" in self.options:
-            mode = self.options["mode"]
+        service: str | None = self.options.get("service")
```
derek-globus, line 195: Do we have sufficient confidence that `mode` is not used across the repo?

aaschaer: From a quick search I'm not seeing it used anywhere

aaschaer (commented):
Looking at the render, single tab tabs feel kinda strange. Don't know if its worth blocking on until we have more than one tab for these or not though

src/globus_sdk/_sphinxext.py
sirosen: It was only recently introduced -- a couple of days before this PR was opened -- so unless someone introduced usage and I didn't see it, it is not yet.

kurtmckee (approved):

sirosen:
> Looking at the render, single tab tabs feel kinda strange. Don't know if its worth blocking on until we have more than one tab for these or not though

I agree that it feels slightly awkward, but I thought it was preferable to put it in place as part of making the doc style uniform both internally and externally. (And it prepares for more tabs in the future.)

PR #718, opened 2023-04-17
Add codespell to pre-commit hooks

sirosen:
This catches and fixes several typos.

Also, fix 'mutabl->mutable' in docs, which was not caught by codespell.
(For reference, I've tried opening [a PR on codepsell](https://github.com/codespell-project/codespell/pull/2822), to see how hard or easy it is to get our relevant typos included there as we encounter things.)

---

Some background/meta: codepsell lists known words and typos, to keep false negatives to a minimum. For some problem-words like `deque` vs `dequeue`, it features a few different built-in dictionaries which we could apply.

I'd like to try it out here and on the CLI, and maybe replace the custom typo checker I wrote for one of our internal projects with it.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -38,6 +38,10 @@ repos:
   hooks:
     - id: pyupgrade
       args: ["--py37-plus"]
+- repo: https://github.com/codespell-project/codespell
+  rev: v2.2.4
+  hooks:
+    - id: codespell
 
 # custom local hooks
 - repo: local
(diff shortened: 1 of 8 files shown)


kurtmckee (approved):
1. Love that you used `codepsell` throughout this PR!
2. Found one additional while looking this over (might be a PR opportunity to the codespell project?): `SDK-provied` in the changelog!

sirosen:
> 1. Love that you used `codepsell` throughout this PR!

:facepalm: 

> 2. The `typos` project seems to catch more typos than `codespell` currently does. I recommend taking a look at that instead!

Yeah, I'm happy to swap this out in the future (as in, within the hour) if we try it out and like it better.

PR #719, opened 2023-04-17
Run `typos` against the codebase, too

kurtmckee:
Exactly what it says on the tin. Great tool, not what I want to set up as a pre-commit hook in this context.

### DIFF
--- a/tests/functional/tokenstorage/test_simplejson_file.py
+++ b/tests/functional/tokenstorage/test_simplejson_file.py
@@ -14,7 +14,7 @@ def filename(tempdir):
     return os.path.join(tempdir, "mydata.json")
 
 
-def test_file_dne(filename):
+def test_file_done(filename):
     adapter = SimpleJSONFileAdapter(filename)
     assert not adapter.file_exists()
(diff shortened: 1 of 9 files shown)


tests/functional/tokenstorage/test_simplejson_file.py
```diff
@@ -14,7 +14,7 @@ def filename(tempdir):
     return os.path.join(tempdir, "mydata.json")
 
 
-def test_file_dne(filename):
+def test_file_done(filename):
```
sirosen, line 17: `dne` was for "does not exist", so we should revert this one

sirosen (changes requested):
One minor item to revert, and I'll label this as no-changelog. Looking good!

tests/functional/tokenstorage/test_simplejson_file.py
kurtmckee, line 17:
```suggestion
def test_file_does_not_exist(filename):
```

sirosen (approved):

PR #722, opened 2023-04-26
Implement flows_client.get_run_logs

derek-globus:
[sc-19955](https://app.shortcut.com/globus/story/19955)

## Important Note
The story requests a method called `get_run_log`.
I instead implemented a method called `get_run_logs` and would advocate that we use this pluralized version.

Using the singular version has a strange implication that each individual event is not a discrete "log" but rather the entirety of the list is a "log". I think that will be confusing terminology to expose and would like to change it as we roll this out in a better supported sdk.

## What?
* Added FlowsClient.get_run_logs(...)
* Link to generated docs: [flows.get_run_logs(...)](https://globus-sdk-python--722.org.readthedocs.build/en/722/services/flows.html#globus_sdk.FlowsClient.get_run_logs)

## Testing
* Unit tests (verifies the testing data)
* Verified actual run log retrievals against the flows service

----
:books: Documentation preview :books:: https://globus-sdk-python--722.org.readthedocs.build/en/722/


src/globus_sdk/services/flows/client.py
```diff
@@ -448,6 +448,59 @@ def delete_flow(
...
+                .. paginatedusage:: get_run_logs
+
+
+            .. tab-item:: Example Response Data
+
+                .. expandtestfixture:: flows.get_run_logs
+
+            .. tab-item:: API Info
+
+                .. extdoclink:: Get Run Log
```
aaschaer, line 488: I feel like the link here should either be "Get Run Logs" to match the method name, or "Get execution details on a Run" to match the docs

derek-globus: Yeah you right; good catch.

Will update to `Get Run Logs` to match the method name (those docs need to be updated)

ada-globus (approved):

aaschaer (approved):

PR #725, opened 2023-05-02
Refactor error parsing to better support JSON:API

sirosen:
Core error parsing has been rewritten to be a better match for the multiple-errors paradigm of JSON:API and similar APIs.

Parsing is now broken down internally into the following phases:
- _parse_response: the main parsing method, which aborts or calls the following methods, listed in order
- _parse_errors_array: extract one or more sub-errors from the document
- _parse_code: extract a singular "code" field from the document
- _parse_messages: extract message strings from the document

GlobusAPIError and its subclasses now feature an `errors` attribute, which is a list of sub-errors, and a `messages` attribute, which is a list of message strings. `message` is a property which is defined as `"; ".join(messages)`, and has a deprecated setter.

`message` will be `None` when there is no parsed message, as opposed to the previous behavior which falls back to the text of the response.

`code` parsing can consequently be improved to have clearer semantics with respect to multiple sub-errors with different codes -- in this case, it will no longer assume that the first code from the sub-errors is accurate. This means that we won't show misleading results if an errors array comes back like
    {
      "errors": [
        {"code": "foo"},
        {"code": "bar"}
      ]
    }

Previously, this would parse the code as `foo`, but now it will leave it with the default of `Error`.

---

I originally wanted to "only" add support for the Flows data format, but when I started changing things, I took a really hard look at what we had for parsing and... :shrug: 

I'm missing a test-case for the other existing Flows data format, which I'll pull together now, but this is already ready for 90%+ of the meaningful review.

Note that this changes some things which users would see in error `repr()`s for malformed error data. Testsuite changes show, in several places, a change from error text to the HTTP Reason -- this is a consequence of the change to make `message` null when it can't parse.


ada-globus (changes requested):
(n.b, we discussed changing the `errors` array handling, and having subclasses deal with the `errror` scenarios.)

sirosen:
I've updated the uses of `raw_json` to `_dict_data`. However, after making the change to try putting the main error into the `errors` array, I almost immediately found the results awkward and confusing. Ultimately, even though that change sounded interesting while we discussed it, applying it does not have the clarity that we would want, so I decided against it.

changelog.d/20230501_212027_sirosen_fix_error_shapes_with_messages.rst
```diff
@@ -0,0 +1,34 @@
...
+  * A new attribute is added to API error objects, ``errors``. This is a list
+    of subdocuments parsed from the error data, especially relevant for
+    JSON:API errors and similar formats.
+
+  * A new attribute is now present on API error objects, ``messages``. This is
+    a list of messages parsed from the error data, for errors with multiple
+    messages. When there is only one message, ``messages`` will only contain
+    one item.
+
+  * The ``message`` field is now an alias for a joined string of of
```
kurtmckee, line 14:
```suggestion
  * The ``message`` field is now an alias for a joined string of
```

sirosen: 🤦

changelog.d/20230501_212027_sirosen_fix_error_shapes_with_messages.rst
```diff
@@ -0,0 +1,34 @@
...
+    ``messages``. Assigning a string to ``message`` is supported for error
+    subclasses, but is deprecated.
+
+  * ``message`` will now be ``None`` when no error message can be parsed.
```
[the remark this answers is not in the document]
sirosen: No objection from me. I think both ways read okay, but this update is certainly not _less_ clear. And if you think it's better, chances are other readers will agree.

sirosen:
To leave a record of why this is marked as a draft:

We're discussing making the parsing behaviors more standardized within our team, so that everyone is in full agreement on what the expected parse of a given error is.

Effectively, if we introduce this changeset without that discussion, we introduce a de-facto standard which is not JSON:API but something closely related (e.g. handling for singular "error" as a supported extension, but not a default).

sirosen:
I've made some final changes here to attempt to accord best with the draft error specification which we have in progress.
The main change is to do much more explicit detection of different error formats and fully split the three distinct parsing paths between JSON:API, "Type Zero", and Undefined formats.

There's a separate method for each parsing path, so subclasses (see TimerAPIError and FlowsAPIError) can override only the undefined path and have a stronger notion of which part is being customized.

Independently, there's a final parsing hook which can be used to compute attributes after successful error parsing finishes with any of those three paths. This suits the GCS and Search errors well.

I would very much like for us to consider this as a target for our next release, as it fixes a number of problematic behaviors.

There are some subtle cases here which are introduced but are, I think, no better or worse than what came before. For example, the JSON:API parse path will now set `code = None` on the error. Since this isn't reachable today without explicitly passing the Accept header to the SDK, I think this minor change is acceptable.

Code-wise, the structure changes significantly, but the observable changes with real error payloads are pretty minor.
There are some test changes -- e.g. changing a `request_id` from `123` to `"123"` -- which show some of the behavioral strictness at play now, but which aren't really issues with current API response formats.

sirosen:
I'm making another change here momentarily, to populate the `errors` array with the top-level error in the case of a `type_zero` parse.
My rationale comes from some discussions about how GlobusAPIError uniformizes the different error formats, and specifically this thought which I shared in a 1-on-1 conversation:

> Any adaptations should be done as necessary to make Type Zero errors appear like JSON:API errors, not vice versa. Since the `errors` array is a new attribute for globus-sdk, I think we should populate it with the top-level error document. The user expectation that `errors` is a non-empty array of subdocuments can thereby be maintained.

kurtmckee:
Ping me if this is ready for re-review.

src/globus_sdk/exc/api.py
```diff
@@ -157,67 +212,258 @@ def _get_args(self) -> list[t.Any]:
...
+        return True
+
+    def _extract_messages_from_error_array(
+        self, errors: list[ErrorSubdocument]
+    ) -> list[str]:
+        """
+        Extract 'messages' from an array of errors (JSON:API or otherwise)
+
+        This is done by checking each error document for the desired message fields and
+        popping out the first such message for each document.
```
derek-globus, line 384: This comment seems a bit out of date.

Root error document is no longer responsible for determining what constitutes a subdocument "message" right?

src/globus_sdk/services/gcs/errors.py
```diff
@@ -26,10 +26,10 @@ def _get_args(self) -> list[t.Any]:
             args.append(self.detail)
         return args
 
-    def _load_from_json(self, data: dict[str, t.Any]) -> None:
-        super()._load_from_json(data)
+    def _final_parse_callback(self) -> bool:
```
derek-globus, line 29: I like the idea of this extension for polymorphism a lot.

---

Loosely held naming opinion, if you disagree feel free to simply ignore

Seems very hook-y so I'd consider naming it with "post" over "final" (sets us up to add a "pre" if needbe).

```suggestion
    def _post_parse_hook(self) -> bool:
```

src/globus_sdk/services/flows/errors.py
```diff
@@ -1,7 +1,39 @@
...
+
+from globus_sdk.exc import ErrorSubdocument, GlobusAPIError
 
 
 class FlowsAPIError(GlobusAPIError):
     """
     Error class to represent error responses from Flows.
     """
+
+    def _parse_undefined_error_format(self) -> bool:
```
derek-globus, line 11: Question, not suggestion:

Would it make sense to explicitly add a "custom parse" overridable method like

```
class GlobusAPIError:
   ...
   def _custom_parse(self) -> bool:
       return False
```

So order of parsing would be like:

* JsonAPI
* TypeZero
* Custom (if exists)
* undefined

derek-globus (approved):
Overall looks good & happy with the direction this takes us.

Big fan of the inheritance structures in place to allow for custom overriding.

A few comments here & there about stylistic choices, comments, etc. but generally 👍

docs/core/exceptions.rst
```diff
@@ -97,6 +96,34 @@ Error Classes
...
+
+ErrorSubdocuments
+-----------------
+
+Errors returned from APIs may define a series of subdocuments, each containing
+an error object. This is used if there were multiple errors encountered and the
+API wants to send them back all at once.
+
+All instances of ``GlobusAPIError`` define an attribute, ``errors``, which is
+an array of ``ErrorSubdocument``\s. It may be empty if there were no subdocuments.
```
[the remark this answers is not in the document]
sirosen: It is still possible in a small selection of cases.

1. `errors` is initialized to `[]`. If the body isn't JSON or is the wrong top-level shape (e.g., top-level array), `errors` will stay `[]`
2. A Type Zero or undefined format can produce an errors array of `[]` if it explicitly sets`"errors": []`.

Regarding (2), we talked about this and I was waffling about it. But in another conversation, we looked again at the draft error spec and the current wording there is as follows:

```
Clients MAY ignore any `"errors"` key in Type Zero Globus Errors and always set `errors` to `[]`.

If the error payload has a top-level key `"errors"` containing an array, and every element of that array is an object, then errors on the parsed response MAY be populated with the `"errors"` in the response.

Absent an `"errors"` key, clients MAY populate `errors` with the root response object.

Clients MUST NOT populate the `errors` of a Parsed Globus Error from a Type Zero Globus Error with data which is not here specified as a valid parse.
```

Now, obviously we're talking about a draft spec that we're working on contemporaneously with this first implementation. So we could change the wording.

But as specified here, the spec was written to say that `errors=[<root_doc>]` is valid only if `"errors"` was not present as a key. If it is present, our choices are:
- respect any well-formed `errors` data (to the above definition of well-formed-ness)
- set `errors=[]`

I'm not sure it's convincing and perfect but it makes some sense in that an API which says `errors=[]` has explicitly told us that there are no sub-errors.

src/globus_sdk/exc/api.py
```diff
@@ -11,33 +12,67 @@
...
+        if self.messages:
+            return "; ".join(self.messages)
+        return None
+
+    @message.setter
+    def message(self, value: str) -> None:
+        warn_deprecated(
+            "Setting a message on GlobusAPIError objects is deprecated. "
+            "This overwrites any parsed messages. Append to 'messages' instead."
+        )
```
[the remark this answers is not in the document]
sirosen: No, users should never see this from our first-party globus-sdk behaviors. We are no longer setting `message` directly, so the SDK never triggers this codepath (except in the test for this).

The way it would likely happen is with third-party tooling built on top of the SDK which sets `message`. Or, if the user is maintaining that tooling, it's their own second-party code using globus-sdk.

Consider this one:

```python
class MyError(globus_sdk.GlobusAPIError):
    def __init__(self, r):
        super().__init__(r)
        self.message = "For the lulz"
```

That error class will emit the warning. If the maintainer is behaving as we hope and has some testing environment where they see those warnings (e.g. pytest), they'll see the warning when they test with the new SDK version.

src/globus_sdk/exc/api.py
sirosen: Let me review this docstring. I bet it needs a bit of revamping.

It gets a little difficult to describe perfectly -- you'll see cases where we pass `self.MESSAGE_FIELDS` to `ErrorSubdocument` objects. So the error class can influence behavior.

docs/core/exceptions.rst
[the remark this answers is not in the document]
derek-globus: I'm just trying to make sure our documentation reflects reality of what is in this change.

The line "It may be empty if there were no subdocuments" makes me expect that

```
{
  "code": "BadRequest",
  "message": "it was a bad request",
  "request_id": "b1ce01db-0b2b-486b-9602-5ddaf0351ad9"
}
```

Would be deserialized into an object with an empty "errors" (there were no subdocuments)

---

As I understand it, this is not the case though.

Am I misunderstanding the implementation? Reading that line unfairly?

src/globus_sdk/services/flows/errors.py
sirosen: Having written that, I'm still thinking about this.

Let me see how messy and hard it would be to rig this up as `() -> bool | None`.
Part of the need for `None` as a supported output is to indicate that parsing didn't detect whatever custom thing was wanted, and that the default undefined parsing needs to take over.

i.e. Calls to `super()._parse_undefined_error_format()` could become `return None`.

I'll give it a quick try, but I might ditch it if I don't like it or it turns out to be too hard.

derek-globus: If custom parse fails wouldn't we want to fall back to the default undefined parsing handler?

sirosen: I ditched it for now. I'm happy to circle back later though.

The end result of my first crack at this was that `return None` became a slightly obscure alias for `return super()._parse_undefined_error_format()`.

---

This is the sort of comparison I ended up with:
```python
if not isinstance(self._dict_data.get("error"), dict):
    return None
...
```
vs
```python
if not isinstance(self._dict_data.get("error"), dict):
    return super()._parse_undefined_error_format()
...
```

The path I was taking was that parsing needs to be able to signal success, failure, or "I don't know, call my parent's way of handling this", and that idea is better captured by overriding the undefined handler.

src/globus_sdk/exc/api.py
sirosen: Looking at it alongside the code, I described both what's happening in that method ("ask each subdoc for its message"), and the higher level fact that this means we're looking at fields which were chosen by the parsing path and the error class.
Let me know if you like the change and we can resolve this.

derek-globus (approved):

aaschaer (approved):
"Type Zero" seems dramatic for an error classification, but its not really exposed, and its kinda fun to say, so no real complaint there

I like the approach of classification and then parsing and couldn't find anything obviously wrong with one readhtrough

src/globus_sdk/exc/api.py
```diff
@@ -11,33 +12,67 @@
...
     RECOGNIZED_AUTHZ_SCHEMES = ["bearer", "basic", "globus-goauthtoken"]
 
     def __init__(self, r: requests.Response, *args: t.Any, **kwargs: t.Any):
+        self._cached_raw_json: t.Any = _CACHE_SENTINEL
+
         self.http_status = r.status_code
         # defaults, may be rewritten during parsing
-        self.code = "Error"
-        self.message = r.text
+        self.code: str | None = "Error"
```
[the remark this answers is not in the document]
sirosen: That was my intent. I think that we shouldn't be injecting values which we didn't get in the response when possible, but removing the default of "Error" seemed too breaking.

sirosen:
> "Type Zero" seems dramatic for an error classification, but its not really exposed, and its kinda fun to say, so no real complaint there

To this note, the name Type Zero isn't fully decided on yet. That's something I introduced because
1. As you say, it's fun
2. 0 is the first natural number
3. I'm not sure if "Legacy" is appropriate as a term when the future is still being charted out

But we should be careful not to make "Type Zero" part of the docs or public interface, so that the name can change in the future. It's an implementation detail -- not a secret one, but not a part of the public SDK surface area.

PR #726, opened 2023-05-02
Allow 'required_scope' singular variant

sirosen:
ConsentRequiredInfo now accepts 'required_scope' containing an individual string as an alternative to 'required_scopes'. The validation of these values has also been made slightly stricter.

Additionally, `ConsentRequiredInfo.required_scopes` is now the empty list rather than `None` in the falsy case. This is not a very significant change, since

1. the documented usage pattern is that the truthiness of `ConsentRequiredInfo` should always be checked before it is used
2. the resulting value is still itself falsy

This subtlety is not covered in the changelog, on the grounds that it would only bring undue attention to a minor implementation detail.

----
:books: Documentation preview :books:: https://globus-sdk-python--726.org.readthedocs.build/en/726/

### DIFF
--- a/tests/unit/errors/test_common_functionality.py
+++ b/tests/unit/errors/test_common_functionality.py
@@ -117,23 +117,44 @@ def test_info_is_falsey_on_non_dict_json(make_json_response):
     assert str(err.info) == "AuthorizationParameterInfo(:)|ConsentRequiredInfo(:)"
 
 
-def test_consent_required_info(make_json_response):
-    res = make_json_response(
-        {"code": "ConsentRequired", "required_scopes": ["foo", "bar"]}, 401
-    )
+@pytest.mark.parametrize(
+    "body, is_detected, required_scopes",
+    (
+        (
+            {"code": "ConsentRequired", "required_scopes": ["foo", "bar"]},
+            True,
+            ["foo", "bar"],
+        ),
+        (
+            {"code": "ConsentRequired", "required_scope": "foo bar"},
+            True,
+            ["foo bar"],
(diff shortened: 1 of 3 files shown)


tests/unit/errors/test_common_functionality.py
```diff
@@ -117,23 +117,44 @@ def test_info_is_falsey_on_non_dict_json(make_json_response):
...
+            ["foo bar"],
+        ),
+        ({"code": "ConsentRequired"}, False, None),
+        ({"code": "ConsentRequired", "required_scopes": []}, False, None),
+        ({"code": "ConsentRequired", "required_scopes": ["foo", 123]}, False, None),
+        ({"code": "ConsentRequired", "required_scope": 1}, False, None),
+    ),
+)
+def test_consent_required_info(make_json_response, body, is_detected, required_scopes):
+    res = make_json_response(body, 401)
```
ada-globus, line 140: I'm not sure that this matters as a practical concern here, but out of curiosity, why is this 401 rather than 403? It seems like the docs suggest this should canonically be a 403.

ada-globus (approved):
I had one question to help me improve my understanding, but afaict it's non-blocking. Looks great!

tests/unit/errors/test_common_functionality.py
sirosen: Yeah, it should be a 403 if we want to match what services do, but it doesn't matter much since the ConsentRequiredInfo detection isn't keyed to any particular status code.

I'll switch it since we're here though. :+1:

PR #727, opened 2023-05-08
Allow list_flows to take orderby as a tuple

sirosen:
The new type for orderby allows str or `tuple[str, Literal["ASC", "DESC"]]`.

The result is that the new ordering style can be validated by mypy, but the existing string style is kept in place.

The validation is intentionally very limited, and primarily provides a nicety for any type-based editor integrations and type checked codebases. The goal is not to fully validate possible `orderby` values.

---

This idea may or may not be beneficial. We should discuss before potentially merging it. We've done something like this for the filters on `TransferClient` because users very consistently had difficulty assembling correct filters.
But will users have trouble with `orderby`, and is this the right way to assist them? This PR makes this option, which was proposed in a conversation, more concrete so that we can evaluate it.

----
:books: Documentation preview :books:: https://globus-sdk-python--727.org.readthedocs.build/en/727/

### DIFF
--- a/src/globus_sdk/services/flows/client.py
+++ b/src/globus_sdk/services/flows/client.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import sys
 import typing as t
 
 from globus_sdk import GlobusHTTPResponse, client, paging, scopes
@@ -11,7 +12,14 @@ from globus_sdk.scopes import ScopeBuilder
 from .errors import FlowsAPIError
 from .response import IterableFlowsResponse
 
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
+
 log = logging.getLogger(__name__)
+
+OrderByTuple = t.Tuple[str, Literal["DESC", "ASC"]]
 
 
 class FlowsClient(client.BaseClient):
@@ -201,7 +209,7 @@ class FlowsClient(client.BaseClient):
         *,
         filter_role: str | None = None,
         filter_fulltext: str | None = None,
-        orderby: str | None = None,
+        orderby: str | OrderByTuple | None = None,
         marker: str | None = None,
         query_params: dict[str, t.Any] | None = None,
     ) -> IterableFlowsResponse:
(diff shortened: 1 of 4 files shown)


tests/non-pytest/mypy-ignore-tests/flows_orderby.py
```diff
@@ -0,0 +1,29 @@
+# test the `orderby` param for `FlowsClient.list_flows()`
```
derek-globus, line 1: What are non-pytest tests?

Is this a nose testsuite or something?

tests/functional/services/flows/test_flow_crud.py
```diff
@@ -48,6 +48,12 @@ def test_list_flows_simple(flows_client, filter_fulltext, filter_role, orderby):
     if orderby:
         add_kwargs["orderby"] = orderby
 
+    expect_orderby_param = None
```
derek-globus, line 51: nit:

```suggestion
    expected_orderby_param = None
```

`expect_orderby_param` -> sounds like a boolean flag answering "should I be expecting an orderby param?"
`expected_orderby_param` -> sounds like the value expected to be provided as an orderby param

src/globus_sdk/services/flows/client.py
```diff
@@ -214,7 +222,7 @@ def list_flows(
         :param filter_fulltext: A string to use in a full-text search to filter results
         :type filter_fulltext: str, optional
         :param orderby: A criterion for ordering flows in the listing
-        :type orderby: str, optional
+        :type orderby: str or tuple, optional
```
derek-globus, line 225: Can you add a piece to the docstring here indicating what the required string format is (since we still support it in addition to the tuple)?

tests/non-pytest/mypy-ignore-tests/flows_orderby.py
sirosen: We have a few tests which are checks that do not run as part of the normal testsuite. This is either for speed reasons or because they aren't written in pytest.
These are in a dedicated dir so that we can tell pytest not to recurse here.

`tox r -e test-lazy-imports` is a good example: those are _super_ slow to be running in a case like `tox run`, in which everything runs on each python version, but we still want/need them in CI.

In this case, the relevant run is `tox r -e mypy-test`, which runs `mypy` on the target files. But we don't want pytest to import/load these because :boom:

sirosen:
I'm making this a draft because we may want/need to allow for `Sequence[str]` to specify multiple orderings.

We should look at the simple form of that first, and then we can think about how to support this usage as well.

sirosen:
I'm closing this. I took a look at maybe trying to integrate this with the new support for iterable[str] and I don't see a clear path.

It would be nice to support something like `orderby=["title", ("description", "DESC"), ...]` but I didn't think up a good, unambiguous way to do it. And I thought about alternatives like a dataclass object but I think it would need a nicer access pattern than importing yet another name from `globus_sdk`. That namespace is crowded enough already.

PR #730, opened 2023-05-11
Support `orderby: Iterable[str]` for list_flows

sirosen:
Add support for an iterable of strings to pass this parameter multiple times.
The tests for `list_flows` have also been moved to a dedicated module, to keep testing modules more focused on a per-method basis in the suite of functional tests.

`orderby` checks for `str` values (passed verbatim), and converts any other value with `list`. This ensures that if a non-sequence iterable is passed, it will be coerced into a type which works well.

A new test is added which exercises several different supported types.

The `list_flows` documentation now includes example usage (which has been tested out-of-band).

----
:books: Documentation preview :books:: https://globus-sdk-python--730.org.readthedocs.build/en/730/


ada-globus (approved):
LGTM!

PR #732, opened 2023-05-23
Add doc note on resource_server lookups

sirosen:
This note is added in response to mention that it was not clear how to do these lookups.

----
:books: Documentation preview :books:: https://globus-sdk-python--732.org.readthedocs.build/en/732/

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -248,6 +248,20 @@ response, and for each one, the response contains the following info:
   access_token when it expires. ``None`` unless explicitly requested.
   Details on refresh_token are in the next section
 
+.. note::
+
+    The keys into ``by_resource_server`` are the registered ``resource_server``
+    value for the service.
+    In general, the ``resource_server`` value will by the ID of the client used
+    in a service.
+
+    However, for Globus services like Globus Auth and Globus Transfer, the
+    ``resource_server`` has been customized to be the hostname for the service.
+
+    You can lookup the ``resource_server`` for most globus-sdk supported
+    clients by looking at the ``resource_server`` class attribute, e.g.
+    ``globus_sdk.TransferClient.resource_server``.
+
 .. _tutorial_step5:
 
 Step 5: Do a login flow with Refresh Tokens
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -252,15 +252,15 @@ response, and for each one, the response contains the following info:
 
     The keys into ``by_resource_server`` are the registered ``resource_server``
     value for the service.
-    In general, the ``resource_server`` value will by the ID of the client used
-    in a service.
 
-    However, for Globus services like Globus Auth and Globus Transfer, the
-    ``resource_server`` has been customized to be the hostname for the service.
+    For Globus hosted services like Globus Auth and Globus Transfer, the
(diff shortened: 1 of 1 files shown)


docs/tutorial.rst
```diff
@@ -248,6 +248,20 @@ response, and for each one, the response contains the following info:
...
   Details on refresh_token are in the next section
 
+.. note::
+
+    The keys into ``by_resource_server`` are the registered ``resource_server``
+    value for the service.
+    In general, the ``resource_server`` value will by the ID of the client used
+    in a service.
+
+    However, for Globus services like Globus Auth and Globus Transfer, the
```
aaschaer, line 258: Feels a little odd to have Auth and Transfer be treated as exceptions to the general case?

docs/tutorial.rst
```diff
@@ -248,6 +248,20 @@ response, and for each one, the response contains the following info:
   access_token when it expires. ``None`` unless explicitly requested.
   Details on refresh_token are in the next section
 
+.. note::
+
+    The keys into ``by_resource_server`` are the registered ``resource_server``
+    value for the service.
+    In general, the ``resource_server`` value will by the ID of the client used
```
ada-globus, line 255:
```suggestion
    In general, the ``resource_server`` value will be the ID of the client used
```

docs/tutorial.rst
aaschaer: Maybe just move things around a bit? Something like:

For Globus hosted services like like Globus Auth and Globus Transfer, the ``resource_server`` value will be the hostname for the service.

For other services, including Globus Connect Server v5, the the ``resource_server`` value will be the ID of the service client.

docs/tutorial.rst
```diff
@@ -248,6 +248,20 @@ response, and for each one, the response contains the following info:
...
+    in a service.
+
+    However, for Globus services like Globus Auth and Globus Transfer, the
+    ``resource_server`` has been customized to be the hostname for the service.
+
+    You can lookup the ``resource_server`` for most globus-sdk supported
+    clients by looking at the ``resource_server`` class attribute, e.g.
```
ada-globus, line 262:
```suggestion
    clients by looking at the ``resource_server`` class attribute, e.g.,
```

ada-globus (approved):
Looks good! Added a couple very minor suggestions.

docs/tutorial.rst
sirosen: I'm just coming back to this and I agree with the suggestion. Let's put the services in as the "normal case" and everything else (the ID case) is "special". I'll also look at some of the other suggestions and probably fold it all together.

docs/tutorial.rst
sirosen: Oooh, I am definitely not consistent about that! Fixed.

docs/tutorial.rst
sirosen: The order and arrangement of the text I went with is a little different from Aaron's suggestion, since I think we need to go straight from "hostname of the service" to "and you can get that via `TransferClient.resource_server`".

Please take a look at the results and let me know what you think.

aaschaer (approved):

ada-globus (approved):
👍

PR #733, opened 2023-05-23
Update requirements

kurtmckee:
This addresses a security issue identified by dependabot, but also aligns our requirement file formatting with dependabot's, and adds an `--upgrade` flag so requirements are upgraded when `tox -e freezedeps` is run.

----
:books: Documentation preview :books:: https://globus-sdk-python--733.org.readthedocs.build/en/733/

### DIFF
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -8,15 +8,15 @@ alabaster==0.7.13
     # via sphinx
 babel==2.12.1
     # via sphinx
-beautifulsoup4==4.11.2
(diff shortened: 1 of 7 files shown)


sirosen:
I like the change to add `--upgrade`, but something seems to be failing in the tests with `urllib3>=2`. Let's track this down -- I'm putting it on my docket for today and hopefully I'll get to it.

sirosen (approved):
I expect that test fix got things sorted out. Let's watch CI and then merge.

PR #734, opened 2023-05-24
Allow `pathlib.Path` filenames for token storage adapters

kurtmckee:
No new tests were added because mypy linting confirms that the type is coerced to `str` during assignment to `self.filename`.

----
:books: Documentation preview :books:: https://globus-sdk-python--734.org.readthedocs.build/en/734/


sirosen (approved):
I like it!

PR #735, opened 2023-05-24
Fix test with bad Content-Length

sirosen:
Summary: I figured it out.

---

Under urllib3 v2.0, a Content-Length longer than the response content (leading to an incomplete read) is no longer silently ignored, but raises an error. This surfaced in an apparently unrelated test in which the body was 131 bytes, but Content-Length was specified at 138.

We could correct the Content-Length, but it is not essential to the test at all. Remove it.

----
:books: Documentation preview :books:: https://globus-sdk-python--735.org.readthedocs.build/en/735/


kurtmckee (approved):

kurtmckee:
I'll rebase my update branch on this work.

PR #736, opened 2023-05-30
transfer client local_user support

aaschaer:
Adds local user selection support to transfer, delete, ls, mkdir, and rename, and _testing support for mkdir and rename

SDK/CLI story: https://app.shortcut.com/globus/story/20504
Corresponding Transfer docs PR: https://github.com/globusonline/docs.globus.org/pull/1322

Opening as draft for a few reasons:
1. Transfer support is not live yet
2. Due to some bad session error behavior in some cases its not clear how loudly the Transfer support will be announced yet, and I'm not sure if/how that will effect the SDK/CLI
3. I have not yet written a changelog fragment due to reason 2 so this should fail checks

----
:books: Documentation preview :books:: https://globus-sdk-python--736.org.readthedocs.build/en/736/

### DIFF
--- a/src/globus_sdk/services/transfer/client.py
+++ b/src/globus_sdk/services/transfer/client.py
@@ -1236,9 +1236,9 @@ class TransferClient(client.BaseClient):
             If a list is supplied, it is passed as multiple params.
             See :ref:`filter formatting <transfer_filter_formatting>` for details.
         :type filter: str or dict, optional
-        :param local_user: Optional value passed through to identity mapping specifying
-            which local user account to map to. Only usable with Globus Connect Server
-            v5 mapped collections.
+        :param local_user: Optional value passed to identity mapping specifying which
+            local user account to map to. Only usable with Globus Connect Server v5
+            mapped collections.
         :type local_user: str, optional
         :param query_params: Additional passthrough query parameters
         :type query_params: dict, optional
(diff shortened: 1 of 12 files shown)


src/globus_sdk/services/transfer/client.py
```diff
@@ -1302,13 +1309,17 @@ def operation_mkdir(
...
         query_params: dict[str, t.Any] | None = None,
     ) -> response.GlobusHTTPResponse:
         """
         :param endpoint_id: The ID of the endpoint on which to create a directory
         :type endpoint_id: str or UUID
         :param path: Path to the new directory to create
         :type path: str
+        :param local_user: Optional value passed through to identity mapping specifying
+            which local user account to map to. Only usable with Globus Connect Server
+            v5 mapped collections.
```
sirosen, line 1322: Nit:
```suggestion
            v5 mapped collections.
        :type local_user: str, optional
```

sirosen (commented):
The feature looks pretty straightforward from the client side, the Transfer side of it and complexity around it notwithstanding.

If we expect this to take a while, it might be nice to get your new mkdir and rename tests into `main` with `local_user` removed. I'd fast-track review for that if you want to do the work that way.

I think we could have a very minimal changelog for the SDK which would be suitable no matter what happens. Something like

> Several `TransferClient` methods, `TransferData`, and `DeleteData` now support the `local_user`, `source_local_user`, and `destination_local_user` parameters

But if this is still being discussed, we don't need to rush it.

aaschaer:
This should be good to go now.

We still aren't loudly advertising this feature while better Transfer/GCS errors are in progress, but SDK support with a minimal changelog should be fine. We might want to make the CLI options hidden since that's aimed at a broader audience that might need better errors more, but that's a future PR.

I also added a change to `TransferRequestsTransport` to not retry on Transfer `EndpointErrors` that came up while testing this. I don't know of any transient `EndpointErrors` and I can report its not a great experience to try a local user option in the CLI and then wait several seconds to learn your GCS version isn't high enough

PR #739, opened 2023-06-02
Adding pagination for credential_list and user_credential_list

tylern4:
This is similar to https://github.com/globus/globus-sdk-python/pull/738 from the ticket I opened and @derek-globus fixed.

My next goal was to list the collections and found it wasn't paginated so I could only get the first page. I added in similar changes to the other pull request to make these two methods paginated and tested `get_collection_list` like before.

```python
data = client.paginated.get_collection_list(page_size=1)
for idx, page in enumerate(data.pages()):
    print(f"Page {idx}: {len(page['data'])} item")

Page 0: 1 item
Page 1: 1 item
```

----
:books: Documentation preview :books:: https://globus-sdk-python--739.org.readthedocs.build/en/739/

### DIFF
--- a/src/globus_sdk/services/gcs/client.py
+++ b/src/globus_sdk/services/gcs/client.py
@@ -116,6 +116,10 @@ class GCSClient(client.BaseClient):
     # collection methods
     #
 
+    @paging.has_paginator(
+        paging.MarkerPaginator,
+        items_key="data",
+    )
     def get_collection_list(
         self,
         *,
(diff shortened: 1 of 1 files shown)


derek-globus:
Thanks for the contribution, Nick.

I've double checked the iteration on both of those routes & initiated our automated testing.
Assuming that passes without issue, I'll merge this & ensure we get it into v3.20.0

PR #749, opened 2023-06-13
Expand sdk error testing & support AuthAPIError.request_id

sirosen:
## Changelog

* ``AuthAPIError`` will now parse a unique ``id`` found in the error subdocuments as the ``request_id`` attribute (:pr:`NUMBER`)

## Why

Following the 3.20 release and some reports of changes behavior, I worked on adding more tests for GCS and Auth error parsing to the testsuite. These don't reveal any bugs with the refined parsing behavior.

However, one thing which surfaced is that Globus Auth errors have an `id` attribute which we can and should surface in the `request_id` if possible. Since the `id` is found in the `errors` subdocument array, the same logic as our current convention for `code` is used.

----
:books: Documentation preview :books:: https://globus-sdk-python--749.org.readthedocs.build/en/749/


ada-globus (approved):
👍

kurtmckee (approved):

PR #750, opened 2023-06-13
git-ignore intermediary coverage files

kurtmckee:
These were showing up in my git manager, which watches the directory and provides live updates.

----
:books: Documentation preview :books:: https://globus-sdk-python--750.org.readthedocs.build/en/750/


sirosen (approved):
Yep!

PR #752, opened 2023-06-15
Move all scope parsing to 'experimental'

sirosen:
This is a large body of work which I've had sitting in a branch for some time.
I haven't been able to fully review it although I rebased it, cleaned it up, and got tests working against it.

We also haven't discussed the pros/cons of this being available in `globus_sdk.experimental`.
My summary of what I was able to pull together follows, but there may be important details missing. Given that this is meant for `experimental`, I think we can accept some gaps in terms of direction (e.g. Should it go through the Graph IR at all? I'm not fully sure).

---

Scope parsing is now exposed via `Scope.parse`, which is only available in `globus_sdk.experimental.scope_parser` as a method of parsing scopes.

The scope parsing passes through two IRs -- first a tree, then a graph -- which reject cycles, normalize any optional/non-optional redundancies, and ensure that the tree is built in the same way that it will be during authorization flows in Globus Auth.

This replaces our "shadow" implementation of MutableScope, which had been left in `scopes.scope_definition` under the name `Scope` but was never exported or documented. All of the features of `Scope` under the old location are now integrated into
`globus_sdk.experimental.scope_parser.Scope`.

Scope parsing is tested independently of other scope tests, moving and expanding these tests into `tests/unit/experimental/`.

----
:books: Documentation preview :books:: https://globus-sdk-python--752.org.readthedocs.build/en/752/


changelog.d/20230719_164657_sirosen_add_experimental_scope_parsing.rst
```diff
@@ -0,0 +1,3 @@
+* Introduce an experimental Globus AUth scope parser in
```
kurtmckee, line 1:
```suggestion
* Introduce an experimental Globus Auth scope parser in
```

docs/experimental/scope_parser.rst
```diff
@@ -0,0 +1,35 @@
...
+
+This component defines a Scope object and exposes use of a parser which can
+parse scope strings into trees of Scope objects.
+It should be regarded as very similar to the existing ``MutableScope`` object in
+``globus_sdk.scopes``. Key differences from ``MutableScope``:
+
+- ``Scope.parse`` is available, for parsing scopes from strings
+- The Globus SDK does not support using ``Scope`` in all of the locations where
+  ``MutableScope`` is supported -- Scope objects must be stringified for use
+- ``Scope`` objects define a ``_contains__`` method, allowing one to check if one scope
```
kurtmckee, line 16:
```suggestion
- ``Scope`` objects define a ``__contains__`` method, allowing one to check if one scope
```

docs/experimental/scope_parser.rst
```diff
@@ -0,0 +1,35 @@
...
+  properly contains another
+
+.. warning::
+
+    This module is experimental due to the limitations of scope strings as a
+    representation of consent trees. The scope trees produced by parsing do
+    not necessarily describe real consent structures in Globus Auth.
```
kurtmckee, line 23: > The scope trees produced by parsing do not necessarily describe real consent structures

I don't like this. I think this has been discussed before, but I'm not keen to introduce something that doesn't match the actual Globus Auth implementation.

sirosen: The one thing that I want to have very clearly stated is that it is not possible to represent the consent structures in Auth as a string. Scope string parsing can only match the consent structure produced by a single authentication and consent flow.

The implementation _does_ match what Auth does for a scope string used in a single consent flow. We're just bound by the limitations of the string representation vis-a-vis Auth's more elaborate internal state.

sirosen (commented):
Thanks for the typo fixes (applied)!

I do want us to discuss this more, and I'm sure we'll have a productive conversation when we find time for this.
I've left one inline note about not matching the Auth consent trees, and there is one other thing I'd like to share up-front: I'm motivated pretty strongly to take this implementation -- most of which exists but is not exposed in the project today -- and move it out of the "normal" parts of the source tree.

ada-globus (approved):
I feel comfortable with the correctness given the advisories about usage in the code.

I have a couple things that felt a little odd to me as I was reading, but none that make me want to block this from heading to `experimental` (but might be worth considering if we want to promote this out of experimental at some point):
- In thinking about what we're exposing on the `ScopeGraph`, I'm wondering if we want to lock down the public interface a bit more strictly as it seems possible to fairly easily break the representation by manipulating the public attributes (e.g., naively as a user I might think it safe to call `self.edges.add(foo)` w/o realizing that there is a private attribute that needs to track changes here)
- It's surprising that when I access `ScopeGraph.nodes` I'm not working with `ScopeTreeNodes` (I get that one is `ScopeTree` and one is `ScopeGraph` but we do build a `ScopeGraph` using `ScopeTreeNodes` so it all gets kind of 😵‍💫).

While I haven't thought about it _super_ deeply, I'm wondering if it might make sense to add something like a `ScopeGraph.add_scope()` method that will accept strings or `ScopeTreeNodes` and allow that to manage the properties of the graph so we don't need to leak some of these details, which I think would let us make `nodes` and `edges` private, and make `top_level_scopes` read-only...? But very possible I haven't thought through the downstream consequences sufficiently.

Regardless, this is non-blocking feedback IMO, so I am simultaneously hitting approve.

sirosen (commented):
I'm going to adjust in response to one bit of inline feedback but I'm also going to move the entire parsing implementation into a privately named module, `_parser`.

Neither `ScopeTreeNode` nor `ScopeGraph` are, in their current implementations, sanitized enough to be user consumable. I tend to think that the absence of docstrings is a deal-breaker on that in its own right. This is currently not as clear as it could be, and the presence of an underscore in the name should serve as a stronger warning sign (for users) and a better disambiguator (for contributors).

sirosen:
I've just pushed through some relatively significant refinements to the cycle detection used to find and reject `foo[bar[baz]] baz[bar]`, which I think merit new and fresh review.

Basically what happened was this:
1. Found a flaw in cycle detection while introducing some test cases for it
2. Did some research and implemented a classic algorithm for cycle detection (Tarjan's, details in the commits), which was fun and interesting but required a recursive implementation
3. Did some tests on absurdly large scope strings and found the recursive call performance to be unacceptable
4. Looked for any information from reliable sources on iterative unrolling of the original recursive algorithm and found none. In fact, I found *numerous* incorrect implementations in SO, Rosetta Code, etc. But nothing even remotely usable or cite-able.
5. Rewrote back to the "simplest possible" cycle detector taking O(n^2) time.
6. Final rewrite to a space-inefficient but time-efficient BFS-based cycle detector. Include a performance test in the testsuite (with appropriate warnings about slow hardware and flakiness).
7. The "maybe flaky" perf test fails in GitHub Actions. Adjust threshold up and amend.

All in all it was a fun adventure. I would tend to keep the commit history in this case. The simpler (correct) cycle detector and the full implementation of Tarjan's could be useful to us in the future if we need to do more work on this.

The "time efficient" approach in the end does *several* list copying operations which are, of course, O(n) in their inputs. Those inputs are typically lengths of paths in the graph, so that's some flavor of `log(n)` amortized. My fun graph theory research told me that the theoretical bounds are linear in the size of the set of vertices and edges -- O(V+E). I think we probably have something like O(V*log(E)) here in the average case, but it's a bit beyond me to figure out the exact _theoretical_ complexity. The wall time is more important to me.

ada-globus (approved):

joshbryan-globus:
Reading through the conversation here I want to leave a couple of notes for consideration.  None of this should be considered blocking critique, but rather just food for thought as you discuss.  By an large I'm fine with the PR.

Regarding "it is not possible to represent the consent structures in Auth as a string", while true from perspective of representing them as a single `scope` string that can passed to `/authorize`, I think some of the disagreement may be arguing semantics.  Going from a scope string to a tree representation of the scope dependencies is _always_ possible (auth itself does this),  the slightly surprising bit is just that the parse trees are a subset of the consent trees that are created by that string (e.g. `a[b] b[c]`  creates a consent tree that includes an `a->b->c` even though the parse trees only have the `a->b` tree).  Going from a consent tree as stored in auth to a single `scope` string isn't possible, but you can always create a sequence of scope strings that would generate that consent structure.  E.g. if the trees `a->b` and `b->c` existed in auth, then `a[b]` followed by `b[c]` would generate that structure.

In reality, neither the scope mini-dsl used for dynamic dependencies nor our internal data structure in auth for storing consents represent them explicitly as a graph.  The "graph" nature comes from the fact that "static dependencies" do impose a graph structure on scopes and the semantics of the scope mini-dsl say "for the purposes of this auth call assume static dependencies exist where edges between scopes exist in the parse tree of this string".  I think some of the confusion in the description come from a desire to conflate a consent tree and a scope parse tree.  It might even be worth being explicit about it.  E.g. rather than turn a scope string into a ScopeTree into a ScopeGraph, perhaps the final representation should be a ConsentTree (either via a ScopeGraph or directly).

Finally, a quick note on cycle detection (mostly to dig the rabbit hole stephen went down a bit deeper).  Tarjan's algorithm is "efficient" if you want to find all cycles, but, as you discovered complex to make iterative.  Since we only want to abort when "some" cycle is detected, BFS and DFS are options (just abort the first time you add a node to a path already in the path).  Both BFS and DFS should have the same worst case time complexity, but if you care about space complexity, DFS will win as you only need to keep the "current" path in memory.  It can be a little less intuitive to unroll into a non-recursive form, but the key difference between the two algorithms is that BFS will use a queue (`paths_to_explorer.append`) and DFS will use a stack (`.push`).

sirosen:
I think our meeting today synchronized us on most elements of this.
I mostly wanted to clarify that because there's no single-string for an arbitrary Consent Forest, scope parsing produces only a specific subset of Consent Forests (notably, those which can be created via a single authentication flow). I think by being overly attentive to this detail, I may have unfortunately confused people a bit and possibly sent some misleading signals (e.g. the documentation warning, now downgraded to a note and clarified).

> Finally, a quick note on cycle detection (mostly to dig the rabbit hole stephen went down a bit deeper). Tarjan's algorithm is "efficient" if you want to find all cycles, but, as you discovered complex to make iterative. Since we only want to abort when "some" cycle is detected, BFS and DFS are options (just abort the first time you add a node to a path already in the path). Both BFS and DFS should have the same worst case time complexity, but if you care about space complexity, DFS will win as you only need to keep the "current" path in memory. It can be a little less intuitive to unroll into a non-recursive form, but the key difference between the two algorithms is that BFS will use a queue (`paths_to_explorer.append` / shift) and DFS will use a stack (`.pop`).

Yeah, I think I got a full half-hour into iterative Tarjan's before realizing that I wasn't sure it was correct anymore. "Complex" almost understates it!
The funny thing about that algorithm is that it relies heavily on the recursive steps happening in the middle of a series of other operations. I almost built a small state machine to do "manual recursion". :grimacing: 

I'm looking at the cycle checker right now to reconsider the potential merit of switching it DFS. I think it's the most efficient approach I'm willing to implement for these purposes, but I still can't quite make the implementation as simple and "obvious" as the current BFS one. If I find a phrasing of it which is clear, I think I'll be satisfied.

sirosen:
I couldn't help it. Somehow it bothered me too much that I couldn't reshape that BFS to DFS cleanly, when my past experience has usually been that DFS is faster and simpler. Turns out, it was just a matter of phrasing it the right way.
Since I was in there, I started playing performance optimization games (I almost started putting `__slots__` on things!), and got the implementation ~25% faster on the 1000-nested-scopes test. It seemed a shame not to commit the results, and here we are...

joshbryan-globus:
> I couldn't help it. Somehow it bothered me too much that I couldn't reshape that BFS to DFS cleanly, when my past experience has usually been that DFS is faster and simpler. Turns out, it was just a matter of phrasing it the right way. Since I was in there, I started playing performance optimization games (I almost started putting `__slots__` on things!), and got the implementation ~25% faster on the 1000-nested-scopes test. It seemed a shame not to commit the results, and here we are...

Ha ... you beat me to it by 20 minutes ... I had just slapped together this general idea example and then came to post it and see you already got there ... not to let a good nerd sniping go to waste, here it is:

```
g = {
    'a': ['b', 'c', 'd'],
    'b': ['d', 'c'],
    'c': ['e'],
    'd': ['e'],
    'e': ['d'],
}

def has_cycle(g, start_node):
    stack = [({start_node} , g[start_node])]
    
    while stack:
        print(stack)
        path_set, (next_child, *remaining_children) = stack.pop()
        
        if next_child in path_set:
            # we've visited this node, abort!
            return True
        
        if remaining_children:
            # if there are more children to explorer, re-insert the current path on the stack to visit again
            stack.append((path_set, remaining_children))
        
        if g[next_child]:
            # if the next node has children then we need to push it on the stack and explorer
            stack.append((path_set | {next_child}, g[next_child]))
            
    return False

has_cycle(g, 'a')
```

Anyway, I agree today's discussion was useful, and I'm glad to see we landed on a phrasing we are all happy with.

tests/unit/experimental/test_scope_parser.py
```diff
@@ -0,0 +1,321 @@
...
+    Observed times on a test system were <20ms, and in CI were <60ms.
+
+    Although checking the speed in this way is not ideal, we want to avoid high
+    time-complexity in the cycle detection. This test offers good protection against any
+    major performance regression.
+    """
+    scope_string = ""
+    for i in range(1000):
+        scope_string += f"foo{i}["
+    scope_string += " foo10"
```
kurtmckee, line 162: Is it significant that there's a space, or that this points backwards to `foo10` instead of some other number?

tests/unit/experimental/test_scope_parser.py
```diff
@@ -0,0 +1,321 @@
...
+    with pytest.raises(ValueError):
+        Scope.deserialize("  ")
+
+
+def test_scope_deserialize_fails_on_multiple_top_level_scopes():
+    with pytest.raises(ValueError):
+        Scope.deserialize("foo bar")
+
+
+@pytest.mark.parametrize("scope_str", ("*foo", "foo[bar]", "foo[", "foo]", "foo bar"))
```
kurtmckee, line 206: `"foo bar"` doesn't appear to relate to special character tests, and appears to be tested immediately above...although now I need to go back and review the difference between `Scope.deserialize(s)` and `Scope(s)`.

src/globus_sdk/experimental/scope_parser/__main__.py
```diff
@@ -0,0 +1,62 @@
...
+        print("    python -m globus_sdk.experimental.scope_parser --timeit")
+        sys.exit(0)
+
+    if sys.argv[1] == "--timeit":
+        timeit_test()
+    else:
+        parse_test(sys.argv[1])
+
+
+main()
```
kurtmckee, line 62: Should this be:

```suggestion
if __name__ == "__main__":
    main()
```
```

File: src/globus_sdk/experimental/scope_parser/__main__.py [hunk 7]
```diff
@@ -0,0 +1,62 @@
+import sys
+import timeit
+import warnings
+
+from ._parser import parse_scope_graph
```
kurtmckee, line 5: I'm not able to run this file locally. This line appears to be the offender.

kurtmckee (approved):
I'm approving, but I had wanted to test timings on my PC as well and couldn't. Would you double-check that the `from...import` line is still working on your machine?

tests/unit/experimental/test_scope_parser.py
sirosen: I think the space is a typo. Either that, or I wrote this test in some way that needed it (maybe generating `[foo{i}`?) and then refactored. It doesn't do anything, so I'll remove it.

`foo10` is pretty arbitrary, I just wanted a long chain that looked like this:
```
a -> b -> c -> d -> e -> f -> g -> ... -> z
     ^                                   /
      \                                 /
       `-------------------------------'
```

I'll add a comment to note that it's an arbitrary choice of where to introduce the cycle.

src/globus_sdk/experimental/scope_parser/__main__.py
sirosen: I don't think so for a `__main__.py`.
Although I'm a little unsure about what's considered conventional for these files, my grasp is that such a wrapper would be unusual because `python -m` usage is the only supported interface.

I poked around in the stdlib for a minute to see if I could find an example, and the only package with a `__main__.py` I could find was `venv`, but it does match the style I'm using of "write it like a script":
https://github.com/python/cpython/blob/4f65f03f2d597881e1a805e91579d10889299a3d/Lib/venv/__main__.py#L6

It also is very explicit about exiting, which is interesting to me. I wouldn't think that would be necessary -- and it doesn't appear so in my usage of `-m`.

src/globus_sdk/experimental/scope_parser/__main__.py
sirosen: Can you share exactly what's going wrong? I switched to a `__main__.py` from having an `if __name__ == "__main__":` block in the parser module to facilitate easier execution of these tests.
If it's not working for you, then I want to resolve that. The testsuite is all well and good, but I wanted a very clear entry point into this code for us to troubleshoot issues.

In a venv with this branch installed, I can use it like so, including the warning lines in the output:
```
 $ python -m globus_sdk.experimental.scope_parser 'foo[bar[*baz]]'
/home/sirosen/dev/sdk-and-cli/sdk/src/globus_sdk/experimental/scope_parser/__main__.py:7: UserWarning: This is a test module for scope parsing. It is not intended for public use.
  warnings.warn(
top level scopes: foo
digraph scopes {
  rankdir="LR";

  foo

  foo -> bar;
  bar -> baz [ label = "optional" ];

}
```

What happens when you try that same usage?

tests/unit/experimental/test_scope_parser.py
sirosen: The behaviors tested are slightly different. Here are the three relevant methods at a high level:

1. `Scope.parse` does a full parse and provides a list of `Scope`s as the result.
2. `Scope.deserialize` runs `Scope.parse` and then asserts that the length of the resulting list is exactly 1 -- and it unwraps that value.
3. `Scope.__init__` does not do any parsing. It takes any input string and assigns it to the `Scope.scope_string` attribute, after checking that it was a valid value to use

`__init__` does no parsing and rejects these strings because this is a simpler design than some alternatives for breaking the potential for a `parse -> __init__ -> parse` cycle. (I considered the path of a true alternate constructor which does not call `__init__`, but I thought that wasn't worth it in the short term.)

So this test ensures that `Scope("foo bar")` is rejected "because you can't initialize a scope whose string contains a space anywhere".

By contrast, the above test is checking that "if you deserialize a scope string, it must parse such that there is only one top-level or 'root' scope". Spaces are allowed in cases like `Scope.deserialize("foo[bar baz]")`.

kurtmckee (approved):

PR #757, opened 2023-06-21
Add support for AuhtClient.get_identity_providers

sirosen:
Background / context:
At some point, the AuthClient and its subclasses started to slip behind the APIs which are documented at docs.globus.org . To heal this, start filling in the gaps.

---

This has been tested with multiple authentication types and supports more than just client creds. Therefore, it is defined on the base class, alongside `get_identities`.

`get_identity_providers` is a relatively small GET API which was missing from the SDK's clients.


src/globus_sdk/services/auth/client/base.py
```diff
@@ -203,6 +197,106 @@ def _convert_listarg(
...
+    def get_identity_providers(
+        self,
+        *,
+        domains: t.Iterable[str] | str | None = None,
+        ids: t.Iterable[UUIDLike] | UUIDLike | None = None,
+        query_params: dict[str, t.Any] | None = None,
+    ) -> GetIdentityProvidersResponse:
+        r"""
+        Given ``domains=...`` or (exclusive) ``ids=...``,
+        look up information for the matching set of identity providers.
```
kurtmckee, line 209: I think this first line could be clearer. That `domains` and `ids` are mutually exclusive is covered by their parameter documentation.

```suggestion
        Look up information about identity providers by domains or by IDs.
```

src/globus_sdk/services/auth/client/base.py
```diff
@@ -203,6 +197,106 @@ def _convert_listarg(
...
+        # letting us consume args whose `__str__` methods produce "the right
+        # thing"
+        elif domains:
+            query_params["domains"] = _commasep(domains)
+        elif ids:
+            query_params["ids"] = _commasep(ids)
+        else:
+            log.warning(
+                "neither 'domains' nor 'ids' provided to get_identity_providers(). "
+                "This can only succeed if 'query_params' were given."
```
kurtmckee, line 292: This is technically not accurate -- reaching this point in the code indicates that one *was provided* but was an empty value.

```suggestion
                "'domains' or 'ids' was an empty value in get_identity_providers(). "
                "The call can only succeed if 'query_params' were given."
```

tests/functional/services/auth/base/test_get_identity_providers.py
```diff
@@ -0,0 +1,30 @@
+import pytest
+
+import globus_sdk
+from globus_sdk._testing import get_last_request, load_response
+
+
+def test_get_identity_providers(client):
```
kurtmckee, line 7: I'd like to see a test that verifies we can get IDPs by query parameters as well. I'm recommending renaming this so there's symmetry across the test names.

```suggestion
def test_get_identity_providers_by_domains(client):
```

kurtmckee (changes requested):

src/globus_sdk/services/auth/client/base.py
```diff
@@ -577,3 +671,11 @@ def get_jwk(
             )
             log.debug("JWK PEM decoding finished successfully")
             return jwk_as_pem
+
+
+def _commasep(val: t.Iterable[str | UUIDLike] | str | UUIDLike) -> str:
```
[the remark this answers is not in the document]
sirosen: Ah, good call! I was rewriting from something which allowed a union of types and didn't quite do it correctly.

src/globus_sdk/services/auth/client/base.py
sirosen: Yep, I like this a lot better. (Aside: let's revisit the `get_identities` docstring, which is similarly stated.)

src/globus_sdk/services/auth/client/base.py
sirosen: My intent was to handle the case in which neither `domains` nor `ids` has been passed.
e.g.
```python
client.get_identity_providers(query_params={"other": "paramval"})
```

I see that you're calling out `ids=[]` as a possible usage which would still reach this.

Before I take this suggestion, which is one possible fix, what do we think of being more pedantic in our matching by switching the above to this?
```python
elif domains is not None: ...
elif ids is not None: ...
else: ...
```

It would mean that `ids=[]` gets passed through as a query param, but I think that's fine.

sirosen: I've gone ahead with a change to tighten these checks to `is not None`.
Let me know if you like (or dislike!) the results, sans any change to the warning log.

tests/functional/services/auth/base/test_get_identity_providers.py
kurtmckee: I'm reopening because I'd like to see test(s) that exercise passing query parameters.

sirosen: I'll get that added in two shakes of a dog's tail.

sirosen: R(er)esolving this with the test now in place.

kurtmckee (approved):

PR #760, opened 2023-06-21
Add SearchClient.batch_delete_by_subject

sirosen:
Exactly what the title says. :slightly_smiling_face: 

----
:books: Documentation preview :books:: https://globus-sdk-python--760.org.readthedocs.build/en/760/

### DIFF
--- a/src/globus_sdk/services/search/client.py
+++ b/src/globus_sdk/services/search/client.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 import typing as t
 
-from globus_sdk import client, paging, response
+from globus_sdk import client, paging, response, utils
 from globus_sdk._types import UUIDLike
 from globus_sdk.exc.warnings import warn_deprecated
 from globus_sdk.scopes import SearchScopes
@@ -398,7 +398,10 @@ class SearchClient(client.BaseClient):
         return self.post(f"/v1/index/{index_id}/delete_by_query", data=data)
 
     def batch_delete_by_subject(
-        self, index_id: UUIDLike, data: dict[str, t.Any]
+        self,
+        index_id: UUIDLike,
+        subjects: t.Iterable[str],
+        additional_params: dict[str, t.Any] | None = None,
     ) -> response.GlobusHTTPResponse:
         """
         Delete data in a Search index as an asynchronous task, deleting multiple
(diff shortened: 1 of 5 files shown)


src/globus_sdk/services/search/client.py
```diff
@@ -397,6 +397,62 @@ def delete_by_query(
...
+                    sc = globus_sdk.SearchClient(...)
+                    query_data = {
+                        "q": "user query",
+                        "filters": [
+                            {
+                                "type": "range",
+                                "field_name": "path.to.date",
+                                "values": [{"from": "*", "to": "2014-11-07"}],
+                            }
+                        ],
```
derek-globus, line 429: Is this relevant to the usage example?

src/globus_sdk/services/search/client.py
```diff
@@ -397,6 +397,62 @@ def delete_by_query(
...
+        """
+        Delete data in a Search index as an asynchronous task, deleting multiple
+        documents based on their ``subject`` values.
+
+        A ``task_id`` value will be included in the response.
+
+        :param index_id: The index in which to delete data
+        :type index_id: str or UUID
+        :param data: a batch-delete-by-subject document listing documents to delete
+        :type data: dict
```
derek-globus, line 412: I think we should accept a list of subjects (list of strings) explicitly here (with additional data params to keep in line with other operations).

No need to force people to look at additional docs to figure out what the post data object needs to look like for something so straightforward.

derek-globus (changes requested):

src/globus_sdk/services/search/client.py
sirosen: I just was about to comment. I pushed an amend to remove this. It was a copy-pasta mistake.

derek-globus: Any idea where I submit a request to have github include a pasta emoji in their list of reaction emojis?

src/globus_sdk/services/search/client.py
sirosen: Yeah, I'm okay with that as an interface change.

I'll do that and also expand the test to make an assertion about the payload sent.

sirosen: I think you will like the new tests. It's nothing to `snork` home about, but it gets the job done.

I'm going to mark this as resolved, since the basic thrust of it (allowing `subjects=...` as usage) is covered.

derek-globus (approved):

PR #764, opened 2023-06-23
Tighten up the AuthClient init signatures

sirosen:
This change serves three purposes:
- improve type annotations
- forbid the use of `authorizer` merely by omitting it from the signatures
- add tests which ensure that `authorizer` is rejected (previously missing)

The change is backwards compatible under most definitions of compatibility.

The new tests cover runtime (TypeError) and typing-time.

----
:books: Documentation preview :books:: https://globus-sdk-python--764.org.readthedocs.build/en/764/


ada-globus (approved):
This looks good. To be sure: "for most definitions of compatible" == "it will break if you were relying on being able to pass an unsupported argument, which we previously ignored"?

sirosen:
It's even more pedantic than that!

"It will break if you were relying on being able to pass an unsupported argument and have that raise a `ValueError` subclass, specifically."

Also, "It will break if you were using `inspect.get_signature` to check the initializer signatures and doing awful logic based on the presence of `**kwargs`."

PR #768, opened 2023-06-29
Add support for Globus Auth Requirements Error responses

ada-globus:
GARE
globus-auth-requirements-error

# Context
In addition to what is included in the documentation in this PR, additional context surrounding this change is provided [here](https://app.shortcut.com/globus/write/IkRvYyI6I3V1aWQgIjY0ODlmMjU4LTNhNmUtNGU2MS1hZDVjLTM1Y2E0NzMxZjMxYiI=) and will be added to the docs site in a subsequent PR.

# Notes
- There are several viable options for how we handle forward/backward compatibility concerns, and this represents the approach that felt to me like it struck the best balance between ease of adoption and clarity. I'd welcome additional conversation around this if you think another approach might be wiser here.
- I expect more formats to be added to our variants as we broaden the conversation. These are the ones I felt I could define well offhand.
- I had initially thought to remove the `session_` prefixes when assigning to attributes, but in conversation with Josh realized that I might be insufficiently worrying about a future where this could cause confusion or a collision between new fields. I've instead opted for an approach that closely matches the shape of the response format itself, leaning generally toward the principle of least astonishment.
- I think that variant formats should not be exported by the subpackage and instead only used by the utility functions, and those are the primary mechanism of coercing legacy responses to Globus Auth Requirements Error responses. Curious if there is a reasonable use case that I haven't considered.
- Docs maybe could use additional love that I am not able to provide because I am too embedded in the problem (referring both to the docs in the PR as well as the Shortcut doc. Would greatly appreciate any notes if you find things confusing or spot obvious shortcomings.

----
:books: Documentation preview :books:: https://globus-sdk-python--768.org.readthedocs.build/en/768/


kurtmckee:
Is this in draft because you don't have the right PR image picked out? :wink:

ada-globus:
> Is this in draft because you don't have the right PR image picked out? 😉

Ha!! I ran out of steam before figuring out the right image (but I am thinking on it now). Was left as a draft because I still had some mypy errors to fix :)

ada-globus:
Turns out if you force-push enough times, _dreams really do come true!_

src/globus_sdk/experimental/session_error/variants.py
```diff
@@ -0,0 +1,290 @@
+import typing as t
+
+from .session_error import GlobusSessionError, GlobusSessionErrorAuthorizationParameters
+
+T = t.TypeVar("T", bound="LegacySessionErrorVariant")
+
+
+class LegacySessionErrorVariant:
```
derek-globus, line 8: Is there a reason that this couldn't extend `GlobusSessionError` (adding new instance variables for each variant) rather than being composable into a `GlobusSessionError`?

Ideally the to_session_error call would be a thing that we could hide from a user as much as possible so that they can simply receive the error and check whether it is a `GlobusSessionError` instead of checking whether it is a `LegacySessionErrorVariant` which allows for conversion into a `GlobusSessionError`.

ada-globus: In theory a typical user shouldn't be handling these directly at all, they should be either creating a `GlobusSessionError` directly if they're emitting an error response or they should be using the `utils` methods if they're parsing an error response to see if it's a session error (in which case they get back a proper `GlobusSessionError` or `None` if it doesn't match).

I don't think I can loosen the types in descendant classes, right? Thinking this would be a problem particularly with `authorization_parameters` where the child class would have to support a union of `str` and `list[str]` for several fields.

src/globus_sdk/experimental/session_error/session_error.py
```diff
@@ -0,0 +1,195 @@
...
+        :type error_dict: dict
+        """
+
+        if "code" not in error_dict:
+            raise ValueError("GlobusSessionError must have a code")
+
+        # Enforce that authorization_parameters is in the error_dict and
+        # contains at least one of the fields we expect
+        if "authorization_parameters" not in error_dict:
+            raise ValueError("GlobusSessionError must have authorization_parameters")
```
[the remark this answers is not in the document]
ada-globus: I don't think these are doing quite the same work. A bit contrived but, for instance, if we remove these checks and you called `.from_dict({"code": None, "authorization_parameters": None})` in the context of parsing an error response received from a service that would not raise an error during runtime, so you'd potentially instantiate an invalid session error (and should probably attempt the legacy variants). This is stricter than the constructor.

The other concern is if these fields are absent entirely, the error that is raised would be a TypeError. While we could ignore ValueError and TypeError while trying to parse, I'm a little concerned that bulk catching TypeErrors as well as ValueErrors downstream could cause us to possibly cast too wide of a net. But willing to be convinced otherwise!

src/globus_sdk/experimental/session_error/session_error.py
```diff
@@ -0,0 +1,195 @@
...
+                "Must include at least one supported authorization parameter: "
+                ", ".join(cls.SUPPORTED_FIELDS.keys())
+            )
+
+        # Enforce the field types
+        for field_name, field_type in cls.SUPPORTED_FIELDS.items():
+            if field_name in param_dict and not isinstance(
+                param_dict[field_name], field_type
+            ):
+                raise ValueError(f"{field_name} must be of type {field_type.__name__}")
```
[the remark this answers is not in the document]
ada-globus: I actually implemented it in the constructor initially and then thought twice about it and changed my mind. I _think_ I could still be convinced otherwise, but there are potentially at least a couple issues.

Issue 1: If we just pass these to `__init__()` then calling `from_dict()` could raise either a `ValueError` or a `TypeError`, which means that we're potentially going to have to catch more types of errors in the utils methods, which means there's less precision and a greater chance we accidentally mask a programming error.

Issue 2: As a programmer manually instantiating this, you may want/need more flexibility, but there is higher trust here that you "know what you are doing" and it's not really the constructor's job to prevent you from instantiating at all (e.g., you might continue to modify these fields after instantiation and make it valid). If you are calling `from_dict` the assumption is that you're parsing something you've received from another service and trying to see if it's a proper session error, and in this case we act as strictly as possible.

I definitely feel less strongly about #2.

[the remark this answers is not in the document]
kurtmckee: > If you are calling `from_dict` the assumption is that you're parsing something you've received from another service

In that case, it seems like this logic should live in the session error utilities package. As is, `.from_dict()` and `to_session_error()` share similar intentions.

src/globus_sdk/experimental/session_error/utils.py
```diff
@@ -0,0 +1,118 @@
...
+
+def to_session_error(
+    error: t.Union[GlobusAPIError, ErrorSubdocument, t.Dict[str, t.Any]]
+) -> t.Optional[GlobusSessionError]:
+    """
+    Converts a GlobusAPIError, ErrorSubdocument, or dict into a GlobusSessionError by
+    attempting to match to GlobusSessionError (preferred) or legacy variants.
+
+    .. note::
+        a GlobusAPIError may contain multiple errors, and in this case only a single
```
kurtmckee, line 23:
```suggestion
    .. note::

        a GlobusAPIError may contain multiple errors, and in this case only a single
```

src/globus_sdk/experimental/session_error/utils.py
```diff
@@ -0,0 +1,118 @@
...
+def to_session_errors(
+    errors: t.List[t.Union[GlobusAPIError, ErrorSubdocument, t.Dict[str, t.Any]]]
+) -> t.List[GlobusSessionError]:
+    """
+    Converts a list of GlobusAPIErrors, ErrorSubdocuments, or dicts into a list of
+    GlobusSessionErrors by attempting to match each error to
+    GlobusSessionError (preferred) or legacy variants.
+
+    .. note::
+        A GlobusAPIError may contain multiple errors, so the result
```
kurtmckee, line 74:
```suggestion
    .. note::

        A GlobusAPIError may contain multiple errors, so the result
```

kurtmckee (changes requested):

src/globus_sdk/experimental/session_error/session_error.py
```diff
@@ -0,0 +1,195 @@
...
+    """
+
+    SUPPORTED_FIELDS = {
+        "session_message": str,
+        "session_required_identities": list,
+        "session_required_policies": list,
+        "session_required_single_domain": list,
+        "session_required_mfa": bool,
+        "session_required_scopes": list,
+    }
```
sirosen, line 45: I know it's tempting to build something which looks pretty and generic, but I think this loses valuable precision. Specifically: `list[str]` can't be validated fully (as containing only strings) based on this logic.

I think we can build ourselves something only _slightly_ less pretty and get a good result. What about the following alternative structure?
```python
# imagine `globus_sdk.experimental.session_error.validators` is a module
_FIELD_VALIDATORS = {
    "session_message": validators.opt_str,
    "session_required_identities": validators.opt_str_list,
    ...
}
```

src/globus_sdk/experimental/session_error/session_error.py
```diff
@@ -0,0 +1,195 @@
...
+    def __init__(
+        self,
+        session_message: t.Optional[str] = None,
+        session_required_identities: t.Optional[t.List[str]] = None,
+        session_required_policies: t.Optional[t.List[str]] = None,
+        session_required_single_domain: t.Optional[t.List[str]] = None,
+        session_required_mfa: t.Optional[bool] = None,
+        session_required_scopes: t.Optional[t.List[str]] = None,
+        **kwargs: t.Any,
+    ):
```
sirosen, line 56: There are two changes I'd like us to make here -- consistent with the `**kwargs` and keyword-only argument notes recently added to the contrib doc.
```suggestion
        self,
        *,
        session_message: t.Optional[str] = None,
        session_required_identities: t.Optional[t.List[str]] = None,
        session_required_policies: t.Optional[t.List[str]] = None,
        session_required_single_domain: t.Optional[t.List[str]] = None,
        session_required_mfa: t.Optional[bool] = None,
        session_required_scopes: t.Optional[t.List[str]] = None,
        extra: dict[str, t.Any],
    ):
```

src/globus_sdk/experimental/session_error/session_error.py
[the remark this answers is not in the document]
sirosen: I was on the fence about the concept of `__init__` doing no validation while `from_dict` does validation until reading these comments.

This is pushing me towards the opinion that allowing the unvalidated path is going to cause confusion -- as it already has.

I vote that we put all the validation into the initializers and that `from_dict` therefore inherits that validation logic. I'll return to the other thread above to offer a thought on how to handle things.

src/globus_sdk/experimental/session_error/session_error.py
```diff
@@ -0,0 +1,195 @@
...
+    :ivar extra_fields: A dictionary of additional fields that were provided. May
+        be used for forward/backward compatibility.
+    :vartype extra_fields: dict
+    """
+
+    def __init__(
+        self,
+        code: str,
+        authorization_parameters: GlobusSessionErrorAuthorizationParameters,
+        **kwargs: t.Any,
```
sirosen, line 140: Ibid of my `**kwargs` note. Mostly putting this here to make sure we don't miss it.

```suggestion
        self,
        code: str,
        authorization_parameters: GlobusSessionErrorAuthorizationParameters,
        *,
        extra: dict[str, t.Any],
```

sirosen (changes requested):
I've only gotten partway through review so far, focused almost solely on the module which provides the GlobusSessionError. However, I'm dropping the feedback now because I think some of the changes I'm requesting will -- if taken -- have ripple effects across other modules.

src/globus_sdk/experimental/session_error/session_error.py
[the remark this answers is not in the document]
derek-globus: > Issue 1: If we just pass these to __init__() then calling from_dict() could raise either a ValueError or a TypeError, which means that we're potentially going to have to catch more types of errors in the utils methods, which means there's less precision and a greater chance we accidentally mask a programming error.

Hmm, so it's about working around the fact that python treats a missing constructor argument as a TypeError (paired with the desire to have utils suppress errors on init). I hadn't thought about this line.

> Issue 2: As a programmer manually instantiating this, you may want/need more flexibility, but there is higher trust here that you "know what you are doing" and it's not really the constructor's job to prevent you from instantiating at all (e.g., you might continue to modify these fields after instantiation and make it valid). If you are calling from_dict the assumption is that you're parsing something you've received from another service and trying to see if it's a proper session error, and in this case we act as strictly as possible.

I'm less sympathetic to this line of thinking.
Reducing the verification done in construction reduces the guarantee of anyone interacting with the object.

If we want to allow users to build up an object which they finalize into a materialized object (which I'm not sure if we do), we should keep verification in the constructor but expose a builder pattern to explicitly defer verificiation until a certain point, not make our constructor a faux-builder while dropping it's verification:

```python
GlobusSessionError.builder()\
   .code("MyCoolError")\
   .session_authorization_parameter(parameter)\
   .build()
```
(this model/naming is ripped from the aws java sdk v2 [ref](https://github.com/awsdocs/aws-doc-sdk-examples/blob/main/javav2/example_code/dynamodb/src/main/java/com/example/dynamodb/GetItem.java#L58-L61))

---

Overall my two concerns are:
1. The one Stephen mentioned above - multiple paths to reason about lead to orphaned code (we update one path but not the other)
2. Callers not receiving the verification that they should because they used the standard constructor route.

What if we keep the key existence check in `from_dict(...)` for the sake of `TypeError` prevention but upstream any other checks into the constructor?

src/globus_sdk/experimental/session_error/variants.py
derek-globus: Discussed offline but copying that same note here for historical record:

* I didn't realize there was that type of a distinction between underlying authorization parameters.
* Given that these shouldn't be handled by typical users directly at all, I'd argue that we should make it more explicit in naming that they're internal artifacts. If we ensure it doesn't complicate the external interface of these session errors, I'm happy.

ada-globus:
Thanks all for the feedback! It feels like we've got momentum in a few directions. I think I'm prepared to adopt just about everything mentioned. Will draw up some changes to suit and then will decide if there's anything left to haggle over :)

src/globus_sdk/experimental/session_error/session_error.py
```diff
@@ -0,0 +1,195 @@
...
+
+        if "code" not in error_dict:
+            raise ValueError("GlobusSessionError must have a code")
+
+        # Enforce that authorization_parameters is in the error_dict and
+        # contains at least one of the fields we expect
+        if "authorization_parameters" not in error_dict:
+            raise ValueError("GlobusSessionError must have authorization_parameters")
+
+        kwargs = copy.deepcopy(error_dict)
```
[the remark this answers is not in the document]
kurtmckee: I had the same feedback but I thought that there might be a possibility that a shallow copy might have knock-on effects. I don't know that it's possible, but I appreciate that you brought this up.

src/globus_sdk/experimental/session_error/session_error.py
```diff
@@ -0,0 +1,195 @@
...
+    :vartype extra_fields: dict
+    """
+
+    SUPPORTED_FIELDS = {
+        "session_message": str,
+        "session_required_identities": list,
+        "session_required_policies": list,
+        "session_required_single_domain": list,
+        "session_required_mfa": bool,
+        "session_required_scopes": list,
```
joshbryan-globus, line 44: The original thinking behind the `authorization_parameters` field was that that would be a dictionary that could be directly mapped to a query string (e.g. using `urlencode`) that would be passed the `/v2/oauth2/authorize` to get a valid token.  `session_requires_scopes` is not a recognized query param by auth, and since this is one specified be oauth2, we don't have the luxury of changing it.   If we are going to change the names parameters in the error format from what is passed into the `authorize` request, I'd suggest we also provide a method for constructing a valid authorize request from the error itself.

Additionally, while `required_scopes` is more descriptive than `scope`, adding `session` to it is a bit misleading since scopes have no relationship to sessions at all.  The consent can be granted in a different session and still satisfy the original token's authorization needs.

joshbryan-globus (commented):
Overall this looks like a good direction.  I left one comment regarding naming of the scopes parameter.  Additionally, I am still not sold on the naming of "Globus Session Error".  While I don't have a wonderful alternative, since scopes are handled by this, it is not strictly related to "sessions".  "Globus Authorization Error" would be my best recommendation, but I do get that may be confusing with other kinds of authorization errors.  I'll continue to mull this, but given sessions are already confusing enough to people I'm little hesitant to further muddy the waters.

ada-globus:
This should be ready for consideration!

While I tried to keep the commit history well-organized, I think that unfortunately the sheer volume of changes to the implementation warrant a clean review, and simply recommend looking at the current state of the branch anew.

## Changes
- Fixes to docs, per code review and otherwise
- Adopt the `construct_error` class that was introduced in the interim, as an improvement to the tests
- Add debug logging as a developer affordance to clarify why GARE and/or variants are not matched during a parse
- Make the variants module non-public, as we don't intend users to instantiate variants directly
- Update circa 3.7 annotation conventions to contemporary style
- Move the validation from the `.from_dict` methods to the class constructors
- Rename from Session Error to Globus Auth Requirements Error
- Add and use a "validators" module to support stricter validation
- Change `session_required_scopes` to `required_scopes` to better reflect that this is not specifically scoped to a particular session

src/globus_sdk/experimental/auth_requirements_error/_variants.py
```diff
@@ -0,0 +1,312 @@
...
+        self,
+        *,
+        code: str | None,
+        required_scopes: list[str] | None,
+        extra: dict[str, t.Any] | None = None,
+    ):  # pylint: disable=unused-argument
+        # Validate and assign supported fields
+        for field_name, validator in self.SUPPORTED_FIELDS.items():
+            try:
+                field_value = validator(locals()[field_name])
```
kurtmckee, line 67: Neat!

src/globus_sdk/experimental/auth_requirements_error/_variants.py
```diff
@@ -0,0 +1,312 @@
...
+            setattr(self, field_name, field_value)
+
+        # Retain any additional fields
+        self.extra_fields = extra or {}
+
+        # Enforce that the error contains at least one of the fields we expect
+        if not any(
+            (getattr(self, field_name) is not None)
+            for field_name in self.SUPPORTED_FIELDS.keys()
+        ):
```
kurtmckee, line 199: Could I recommend this, to avoid double negatives?

```suggestion
        if all(
            getattr(self, field_name) is None
            for field_name in self.SUPPORTED_FIELDS
        ):
```

src/globus_sdk/experimental/auth_requirements_error/auth_requirements_error.py
```diff
@@ -0,0 +1,224 @@
...
+
+            setattr(self, field_name, field_value)
+
+        self.extra_fields = extra or {}
+
+        # Enforce that the error contains at least one of the fields we expect
+        if not any(
+            (getattr(self, field_name) is not None)
+            for field_name in self.SUPPORTED_FIELDS.keys()
+        ):
```
kurtmckee, line 84: As before:

```suggestion
        if all(
            getattr(self, field_name) is None
            for field_name in self.SUPPORTED_FIELDS
        ):
```

src/globus_sdk/experimental/auth_requirements_error/utils.py
```diff
@@ -0,0 +1,132 @@
...
+    error: GlobusAPIError | ErrorSubdocument | dict[str, t.Any]
+) -> GlobusAuthRequirementsError | None:
+    """
+    Converts a GlobusAPIError, ErrorSubdocument, or dict into a
+    GlobusAuthRequirementsError by attempting to match to
+    GlobusAuthRequirementsError (preferred) or legacy variants.
+
+    .. note::
+
+        a GlobusAPIError may contain multiple errors, and in this case only a single
```
kurtmckee, line 29:
```suggestion
        A GlobusAPIError may contain multiple errors, and in this case only a single
```

kurtmckee (changes requested):

sirosen (commented):
I've only gotten part-way through the doc so far, but it's looking good. Some inline comments have details.

We had a conversation about tinkering with the way the validators are assigned to field names and called which leads to some tinkering I'm going to go try and PR against this.

src/globus_sdk/experimental/session_error/session_error.py
ada-globus: Follow up on this: We've had some conversation about where to implement this (function/method to construct an `authorize` query string from a G.A.R.E.), and we're planning to start by implementing in the CLI and then see if we can use that to as an exercise to design something generic that makes sense to live in the SDK.

src/globus_sdk/experimental/auth_requirements_error/_variants.py
ada-globus: Nice catch! 👍

docs/experimental/auth_requirements_errors.rst
```diff
@@ -0,0 +1,136 @@
+Auth Requirements Errors
+========================
+
+Globus Auth Requirements Error is a response format that conveys to a client any
+modifications to a session (i.e., "boosting") that will be required
+to complete a particular request.
+
+The ``auth_requirements_error`` module provides a number of tools to make it easier to
```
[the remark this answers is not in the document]
ada-globus: Good call, I agree! Helpful for context.

docs/experimental/auth_requirements_errors.rst
```diff
@@ -0,0 +1,136 @@
...
+Auth Requirements Error responses. The shape of an instance closely
+matches that of the JSON response, such that in order to access a
+response's required_scopes one could use, e.g.,:
+
+.. code-block:: python
+
+    from globus_sdk.experimental import auth_requirements_error
+
+    error = auth_requirements_error.GlobusAuthRequirementsError(response)
+    error.authorization_parameters.required_scopes
```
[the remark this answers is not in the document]
ada-globus: That's a good point, and actually I think this is just flat out incorrect and was missed during a refactor. This example was ultimately more marketing and less instruction anyway, so I'm happy to let brevity and clarity prevail!

docs/experimental/auth_requirements_errors.rst
```diff
@@ -0,0 +1,136 @@
...
+
+``GlobusAuthRequirementsError`` enforces types strictly when parsing a Globus
+Auth Requirements Error response dictionary, and will raise a ``ValueError`` if a
+supported field is supplied with a value of the wrong type.
+``GlobusAuthRequirementsError`` does not, however, attempt to mimic or itself enforce
+any logic specific to the Globus Auth service with regard to what represents a valid
+combination of fields (e.g., ``session_required_mfa`` requires either
+``session_required_identities`` or ``session_required_single_domain``
+in order to be properly handled).
```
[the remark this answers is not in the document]
ada-globus: I think this is a great idea. I'll ruminate on whether there are enough additional errata that it warrants a full section. If not, I still think your point is very valid about breaking up the flow with details. In that case, perhaps we move to an end-of-section note so that it's not interrupting the narrative? Will follow up in a commit either way.

docs/experimental/auth_requirements_errors.rst
```diff
@@ -0,0 +1,136 @@
...
+
+If you are writing a service that needs to respond with a Globus Auth Requirements Error, you can
+instantiate the ``GlobusAuthRequirementsError`` class directly to emit auth requirements errors
+in your application, e.g.:
```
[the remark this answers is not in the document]
ada-globus: Love it!

kurtmckee (approved):

sirosen:
I've spent a lot of time staring as the field assignment and validation paradigm, grappling with some abstract discomfort and then reifying those thoughts into a pair of PRs against this branch (the former of which is closed in favor of the latter).

#796 is my latest take on the matter, and the underlying branch name I have used, `spartan-gare`, indicates the tack it takes.
However, that PR puts forth a concrete solution to the problem I see, which I would like to state here.

This is the current definition of init, with its requisite metadata, for the ConsentRequired format used by Transfer, all extraneous details removed:
```python
class LegacyConsentRequiredTransferError:
    code: str
    required_scopes: list[str]
    extra_fields: dict[str, t.Any]

    SUPPORTED_FIELDS = {
        "code": validators.StringLiteral("ConsentRequired"),
        "required_scopes": validators.ListOfStrings,
    }

    def __init__(
        self,
        *,
        code: str,
        required_scopes: list[str],
        extra: dict[str, t.Any] | None = None,
    ):
        for field_name, validator in self.SUPPORTED_FIELDS.items():
            try:
                field_value = validator(locals()[field_name])
            except ValueError as e:
                raise ValueError(f"Error validating field '{field_name}': {e}") from e

            setattr(self, field_name, field_value)

        self.extra_fields = extra or {}
```

This reads, to my eye, as very complicated for a data container with two fields plus an extra blob.
I understand the desire for a framework for writing these which makes them simpler. I pursued such solutions (e.g. #795), but ultimately came back to it being overkill for what we need.

I would like to see the above declaration of a data container look like this instead (again, extraneous detail omitted):
```python
class LegacyConsentRequiredTransferError:
    def __init__(
        self,
        *,
        code: Literal["ConsentRequired"],
        required_scopes: list[str],
        extra: dict[str, t.Any] | None = None,
    ):
        self.code = _check_code(code)
        self.required_scopes = _check_required_scopes(required_scopes)
        self.extra = extra or {}
```

My rationale follows.

For every field, we have the following three criteria, declared in the following order in the current solution:
1. a type checker knows its type on an instance
2. it has the right value at runtime and it is validated correctly at runtime
3. a type checker knows its (potentially different) type as a parameter to `__init__`

The current solution declares (1) first, using a class-level annotation. Then it declares (2) in the form of a name-to-validator map. Then it declares (3).

But (1) and (2) are really the same problem -- any validator for `it must be a string` should be typed as `(Any) -> str`. That's the natural type for that validator as a kind of TypeGuard (I have not used `typing.TypeGuard` in my solutions only because it does not provide any particular utility to us in this case), and it is very simple both for type checkers and developers to understand. `x = must_be_str(y)` logically implies `x: str` both for the human and the machine.

(3) could always be declared to match (1). But that actually doesn't allow for some important usages, like `str -split-> list[str]` as a well-defined part of the interface.

The fact that (1) and (3) can _intentionally_ mismatch makes every incidental possibility of an accidental mismatch more confusing. There's no rewrite which eliminates the possibility of an incorrect declaration, but I'm not convinced that we're deriving enough positive value from the `SUPPORTED_FIELDS`-framework to offset its cost.

---

Whether or not this is all convincing, there are some particulars in the PR which I think we should address. I'll leave inline comments for those items.

src/globus_sdk/experimental/auth_requirements_error/_variants.py
```diff
@@ -0,0 +1,309 @@
...
+        extras = {k: v for k, v in error_dict.items() if k not in cls.SUPPORTED_FIELDS}
+        kwargs: dict[str, t.Any] = {"extra": extras}
+        # Ensure required fields are supplied
+        for field_name in cls.SUPPORTED_FIELDS.keys():
+            kwargs[field_name] = error_dict.get(field_name)
+
+        return cls(**kwargs)
+
+    def to_auth_requirements_error(self) -> GlobusAuthRequirementsError:
+        raise NotImplementedError()
```
sirosen, line 40: I would like `LegacyAuthRequirementsErrorVariant` to be a protocol class which the other classes implement, but without providing any concrete methods.
Because `from_dict` is used in the `utils.py` module, it needs to be declared on the protocol, but the following is a suitable definition:

```python
class LegacyAuthRequirementsErrorVariant(Protocol):
    @classmethod
    def from_dict(cls: t.Type[T], error_dict: dict[str, t.Any]) -> T:
        ...

    def to_auth_requirements_error(self) -> GlobusAuthRequirementsError:
        ...
```

The main reason to prefer this declaration is that no inheritance is used, but type checkers will require that the type is matched in any assignment. For example, the following assignment requires (under type checkers) that the protocol is satisfied by the implementing class:
```python
x: type[LegacyAuthRequirementsErrorVariant] = LegacyConsentRequiredTransferError
```

This also separates out a question which I think should be answered differently:

I think we should make the implementations of `from_dict` and `to_dict` common for all of our classes. (I recognize that the classes in this module don't define `to_dict`, but I would have them inherit it.)
Rather than try to describe in detail, I'll just point to what I've done with `Serializable` in #796. Other names are fine, but I think sharing a common base for all of these helper objects which is defined _as_ a common base and is non-abstract leads to a better model.

src/globus_sdk/experimental/auth_requirements_error/_variants.py
```diff
@@ -0,0 +1,309 @@
...
+        # Validate and assign supported fields
+        for field_name, validator in self.SUPPORTED_FIELDS.items():
+            try:
+                field_value = validator(locals()[field_name])
+            except ValueError as e:
+                raise ValueError(f"Error validating field '{field_name}': {e}") from e
+
+            setattr(self, field_name, field_value)
+
+        self.extra_fields = extra or {}
```
sirosen, line 73: I would very strongly advocate for not having this attribute name mismatch with the constructor argument name.

In my own take on this, I went with `self.extra = extra or {}`. Equally valid would be `self.extra_fields = extra_fields or {}`. But the change is effectively the same either way. I'd like to see these named the same.

src/globus_sdk/experimental/auth_requirements_error/auth_requirements_error.py
```diff
@@ -0,0 +1,221 @@
...
+                error_dict[field] = getattr(self, field)
+
+        # Set any extra fields
+        if include_extra:
+            error_dict.update(self.extra_fields)
+
+        return error_dict
+
+
+class GlobusAuthRequirementsError(GlobusError):
```
sirosen, line 129: Is it good for this to be an Exception class? I do not think so.

An Exception supports `raise`, and in that context its repr is significant.
I think this is an API which we don't need to support here and we shouldn't be going out of our way to declare.

If an implementer wants to define their own
```python
class ServiceGARError(Exception):
    def __init__(self, content: GlobusAuthRequirementsError) -> None:
        self.content = content
```
they can do so and raise that.

Doing so separates the data de/serialization and normalization concerns (globus-sdk) from exception heirarchies and behaviors (web frameworks, clients, etc).

The current client-side use case looks like
```python
try:
    do_thing()
except GlobusAPIError as e:
    gare = decode_gare(e)
    if gare:
        handle_gare(gare)
```

The exception you get from a synchronous service invocation is _not_ one of these errors. That's part of why I think making this an error is actually _less_ helpful -- it means there's another class of error in the SDK but we never declare where and how it's raised (and in fact it _isn't_ raised!).

src/globus_sdk/experimental/auth_requirements_error/utils.py
```diff
@@ -0,0 +1,132 @@
+from __future__ import annotations
```
sirosen, line 1: I would like us to not call this module `utils` because I think it's non-descriptive and implies that it's okay for us to have some loosely organized utilities which we then expose to users.
It also enters the same territory as `globus_sdk.utils` which is very explicitly declared as non-public.

In #796 I went with `functional_interface`, but I'm also okay with `handlers`, `helpers`, etc.

I just want to avoid this _particular_ name because it's a name we already are using in a broader context for a different purpose.

(I prefer my own suggestion of `functional_interface` because it declares that these methods, in the very module name, are specifically public-facing.)

sirosen (approved):

PR #769, opened 2023-06-29
Update authorization params parsing: support list[str] for policies and more type safety

sirosen:
The changelog pretty much says it all in terms of what is changing.
The implementation was refactored a bit to support these changes.

# Changelog

## Changed

- ``session_required_policies`` parsing in ``AuthorizationParameterInfo`` now supports the policies being returned as a ``list[str]`` in addition to supporting ``str``

## Fixed

- ``AuthorizationParameterInfo`` is now more type-safe, and will not return parsed data from a response without checking that the data has correct types

----
:books: Documentation preview :books:: https://globus-sdk-python--769.org.readthedocs.build/en/769/

### DIFF
--- a/src/globus_sdk/exc/err_info.py
+++ b/src/globus_sdk/exc/err_info.py
@@ -71,16 +71,20 @@ class AuthorizationParameterInfo(ErrorInfo):
             data.get("session_required_single_domain"),
         )
 
-        # get str|None and parse as appropriate
+        # get str|list[str]|None and parse as appropriate
         self.session_required_policies: list[str] | None = None
         session_required_policies = data.get("session_required_policies")
         if isinstance(session_required_policies, str):
             self.session_required_policies = session_required_policies.split(",")
+        elif isinstance(session_required_policies, list):
+            self.session_required_policies = [
+                p for p in session_required_policies if isinstance(p, str)
+            ]
         elif session_required_policies is not None:
             log.warning(
                 "During ErrorInfo instantiation, got unexpected type for "
                 "'session_required_policies'. "
-                f"Expected 'str', but got '{type(session_required_policies)}'"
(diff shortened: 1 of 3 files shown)


kurtmckee:
mypy is unhappy with this.

sirosen:
As it should be! I had some incorrect return type annotations as `-> None`. The corrected types are _amend+push_ now live.

src/globus_sdk/exc/err_info.py
```diff
@@ -62,26 +66,75 @@ def __init__(self, error_data: dict[str, t.Any]):
...
+        elif session_message is not None:
+            self._warn_type("session_message", "str", session_message)
+        return None
+
+    def _parse_session_required_identities(
+        self, data: dict[str, t.Any]
+    ) -> list[str] | None:
+        session_required_identities = data.get("session_required_identities")
+        if _is_list_of_strs(session_required_identities):
+            return session_required_identities
```
kurtmckee, line 96: I am astounded that mypy understands this type based on a flagging function. Well done, mypy!

kurtmckee (approved):

src/globus_sdk/exc/err_info.py
sirosen: Oops, you're right! It has no idea what the type is!

With `reveal_type(session_required_identities)`:
```
src/globus_sdk/exc/err_info.py:110: note: Revealed type is "Union[Any, None]"
```

Fixing it now by making that method a type guard.

sirosen: With my latest change to use TypeGuard:
```
src/globus_sdk/exc/err_info.py:116: note: Revealed type is "builtins.list[builtins.str]"
```

All better!

PR #770, opened 2023-06-30
Add `construct_error` to `_testing` subpackage

sirosen:
These changes were inspired by a conversation about wanting the `make_response` helper in an SDK-dependent codebase using `_testing`.
I felt that it was better to expose this as an error builder than to directly expose the helper as it was built within the testsuite.

---

This helper replaces several "make_response" fixtures in error testing, not only making things more uniform and explicit within the SDK tests, but also making it possible for `_testing` users to leverage this art.

`construct_error` replaces an entire error-testing `conftest.py` file and is generally a simplification. Several tests become more verbose as they now contain more of the test data.

----
:books: Documentation preview :books:: https://globus-sdk-python--770.org.readthedocs.build/en/770/


tests/unit/errors/test_common_functionality.py
```diff
@@ -14,104 +15,127 @@ def _strmatch_any_order(inputstr, prefix, midfixes, suffix, sep=", "):
...
-def test_raw_json_works(default_json_response):
-    err = GlobusAPIError(default_json_response.r)
-    assert err.raw_json == default_json_response.data
+def test_raw_json_works():
+    data = {
+        "code": "Json Error",
+        "errors": [
+            {
+                "message": "json error message",
+                "title": "json error message",
```
kurtmckee, line 24:
```suggestion
                "title": "json error title",
```

tests/unit/errors/test_common_functionality.py
```diff
@@ -14,104 +15,127 @@ def _strmatch_any_order(inputstr, prefix, midfixes, suffix, sep=", "):
...
+    )
     assert err.raw_json is None
 
 
-def test_text_property_works(default_json_response, default_text_response):
-    err = GlobusAPIError(default_json_response.r)
-    assert err.text == json.dumps(default_json_response.data)
-    err = GlobusAPIError(default_text_response.r)
-    assert err.text == default_text_response.data
+@pytest.mark.parametrize("body", ("text_data", {"foo": "bar"}))
```
kurtmckee, line 50: `body` can be bytes, too.

tests/unit/errors/test_common_functionality.py
```diff
@@ -599,14 +595,8 @@ def test_non_jsonapi_parsing_uses_root_as_errors_array_by_default(
...
-def test_non_jsonapi_parsing_uses_errors_array_if_present(make_response, error_doc):
-    res = make_response(
-        error_doc,
-        422,
-        data_transform=json.dumps,
-        headers={"Content-Type": "application/json"},
-    )
-    err = GlobusAPIError(res.r)
+def test_non_jsonapi_parsing_uses_errors_array_if_present(error_doc):
+    err = construct_error(error_class=GlobusAPIError, body=error_doc, http_status=422)
```
kurtmckee, line 599: I haven't found a place yet where `error_class` does not duplicate the existing default class value. Would you add testing to affirm the default class is `GlobusAPIError` when not specified?

There's no need to remove all places where `error_class` specifies the same default value, but I would prefer it if the default wasn't explicitly overridden, like:

```suggestion
    err = construct_error(body=error_doc, http_status=422)
```

kurtmckee (changes requested):

tests/unit/errors/test_common_functionality.py
sirosen: Let me add tests to the meta-tests for `_testing` which cover this. I might remove all of the specifications of the default value here too.

The interface changed while I was working on both parts -- originally, `error_class` was required -- and I didn't do a cleanup pass here.

tests/unit/errors/test_common_functionality.py
sirosen: Good catch! Let me add that.

sirosen: In my refactor, this actually folded into the suite of tests which are meant to exercise `construct_error`.
In this file, you'll see this trimmed back to a much simpler test which ensures that `GlobusAPIError.text` returns the string which was passed as the `body`.

tests/unit/errors/test_common_functionality.py
sirosen: I think I like the results of testing this independently, so I'm going to mark it resolved as well.

kurtmckee (approved):

ada-globus (approved):
Tests aren't quite happy yet, but I'm enthusiastic for this change! Will update the session error PR to adopt this as soon as it merges.

sirosen:
This branch had some conflicts because there were other changes to the same tests in recent work.

Unfortunately, there were a couple of minor errors in my rebase, which I've just fixed. It should pass CI now.

PR #772, opened 2023-07-05
Implement `AuthClient.create_project`

sirosen:
I was sitting on this as part of a larger update, but there's no reason to hold it back.

----
:books: Documentation preview :books:: https://globus-sdk-python--772.org.readthedocs.build/en/772/

### DIFF
--- a/src/globus_sdk/utils.py
+++ b/src/globus_sdk/utils.py
@@ -45,13 +45,12 @@ def slash_join(a: str, b: str | None) -> str:
 
 
 def safe_strseq_iter(
-    value: t.Iterable[t.Any] | bytes | str | uuid.UUID,
+    value: t.Iterable[t.Any] | str | uuid.UUID,
 ) -> t.Iterator[str]:
     """
     Given an Iterable (typically of strings), produce an iterator over it of strings.
     This is a passthrough with some caveats:
     - if the value is a solitary string, yield only that value
-    - if the value is a bytestring, yield the utf-8 decoded string
     - if the value is a solitary UUID, yield only that value (as a string)
     - str values in the iterable which are not strings
 
@@ -61,8 +60,6 @@ def safe_strseq_iter(
     """
     if isinstance(value, str):
         yield value
-    elif isinstance(value, bytes):
-        yield value.decode()
     elif isinstance(value, uuid.UUID):
         yield str(value)
     else:
--- a/src/globus_sdk/utils.py
+++ b/src/globus_sdk/utils.py
@@ -6,6 +6,7 @@ import hashlib
 import os
 import sys
 import typing as t
+import uuid
 from base64 import b64encode
 from enum import Enum
 
@@ -44,13 +45,14 @@ def slash_join(a: str, b: str | None) -> str:
 
 
 def safe_strseq_iter(
-    value: t.Iterable[t.Any],
+    value: t.Iterable[t.Any] | str | uuid.UUID,
 ) -> t.Generator[str, None, None]:
     """
     Given an Iterable (typically of strings), produce an iterator over it of strings.
     This is a passthrough with two caveats:
     - if the value is a solitary string, yield only that value
-    - str value in the iterable which is not a string
+    - if the value is a solitary UUID, yield only that value (as a string)
+    - str values in the iterable which are not strings
 
     This helps handle cases where a string is passed to a function expecting an iterable
     of strings, as well as cases where an iterable of UUID objects is accepted for a
@@ -58,6 +60,8 @@ def safe_strseq_iter(
     """
     if isinstance(value, str):
         yield value
+    elif isinstance(value, uuid.UUID):
+        yield str(value)
     else:
         for x in value:
             yield str(x)
--- a/src/globus_sdk/utils.py
+++ b/src/globus_sdk/utils.py
@@ -45,12 +45,13 @@ def slash_join(a: str, b: str | None) -> str:
 
 
 def safe_strseq_iter(
-    value: t.Iterable[t.Any] | str | uuid.UUID,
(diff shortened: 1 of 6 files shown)


sirosen:
Something went amiss with mypy in my rebase. I'm fixing it now.

src/globus_sdk/utils.py
```diff
@@ -44,20 +45,23 @@ def slash_join(a: str, b: str | None) -> str:
...
+    - str values in the iterable which are not strings
 
     This helps handle cases where a string is passed to a function expecting an iterable
     of strings, as well as cases where an iterable of UUID objects is accepted for a
     list of IDs, or something similar.
     """
     if isinstance(value, str):
         yield value
+    elif isinstance(value, uuid.UUID):
+        yield str(value)
```
kurtmckee, line 64: Is this function unit tested, or is it implicitly tested by other code?

If unit tests exist for this (and I hope they do!) then I'd like to have them updated to confirm UUIDs are handled correctly.

sirosen: I think that -- and this is surprising to me -- we don't have dedicated unit tests for this, only the implicit runs of it via functional tests.
I'm adding a couple of quick tests now, so we'll get this included in the PR.

src/globus_sdk/utils.py
```diff
@@ -44,20 +45,26 @@ def slash_join(a: str, b: str | None) -> str:
...
+    - str values in the iterable which are not strings
 
     This helps handle cases where a string is passed to a function expecting an iterable
     of strings, as well as cases where an iterable of UUID objects is accepted for a
     list of IDs, or something similar.
     """
     if isinstance(value, str):
         yield value
+    elif isinstance(value, bytes):
+        yield value.decode()
```
kurtmckee, line 65: This branch didn't exist before, and now exists only to satisfy the unit tests.

I'd like this branch to be removed unless/until it's needed...although it seems that if code already knows it's dealing with `bytes`, there's no reason for it to call this function in the first place.

```suggestion
```

tests/unit/test_utils.py
```diff
@@ -68,3 +70,17 @@ def y(cls):
             return cls.x["x"]
 
     assert Foo.y == 1
+
+
+@pytest.mark.parametrize(
+    "value, expected_result",
+    (
+        ("foo", ["foo"]),
+        (b"foo", ["foo"]),
```
kurtmckee, line 79:
```suggestion
```

kurtmckee (changes requested):

src/globus_sdk/utils.py
sirosen: I think this is attributing the ordering incorrectly because I was _prompted_ to add this by taking a look at testing. The branch isn't here to satisfy the test -- the test is added to exercise the newly added branch.

I added it because in the course of writing up test cases for this, I realized that it's possible for someone to pass `bytes` regardless of what our type information specifies and that the failure mode will confuse them.
Without adding this branch, if someone writes `admin_ids=b"foo"`, the result will currently be that we pass (to the API) `["f", "o", "o"]`. I thought that was particularly nasty -- they'll get an error potentially telling them something like

> ['f is not a valid admin ID', 'b is not a valid admin ID', 'b is not a valid admin ID', 'f is not a valid admin ID', 'e is not a valid admin ID', '4 is not a valid admin ID', '4 is not a valid admin ID', 'e is not a valid admin ID', '- is not a valid admin ID', '1 is not a valid admin ID', 'f is not a valid admin ID', 'a is not a valid admin ID', 'f is not a valid admin ID', '- is not a valid admin ID', '4 is not a valid admin ID', 'e is not a valid admin ID', '5 is not a valid admin ID', 'f is not a valid admin ID', '- is not a valid admin ID', '9 is not a valid admin ID', 'f is not a valid admin ID', '0 is not a valid admin ID', 'd is not a valid admin ID', '- is not a valid admin ID', '0 is not a valid admin ID', 'f is not a valid admin ID', '6 is not a valid admin ID', '0 is not a valid admin ID', 'e is not a valid admin ID', '7 is not a valid admin ID', 'c is not a valid admin ID', 'f is not a valid admin ID', 'f is not a valid admin ID', 'a is not a valid admin ID', '5 is not a valid admin ID', '7 is not a valid admin ID']

(that's just `[f"{x} is not a valid admin ID" for x in "uuid-here"]`)

when their mistake was to pass bytes.

Maybe this is an unnecessary check to add, but I also think it's trivial to bury it in here in a utility and have the benefit that we'll pass bytes to APIs with fewer transformations accidentally applied.

In that light, are you still strongly :-1: on slipping in this addition?

kurtmckee: > it's possible for someone to pass `bytes`

My concern here is that the SDK doesn't do that so we're not in danger of bumping up against this possibility.

If this is a branch that someone could realistically reach when writing their own code that uses the SDK, then let's account for this. It didn't seem like a realistic possibility at first blush, which is why I piped up about it.

sirosen: All that's necessary is for a user to pass `bytes` to one of our methods which uses this helper to handle the data. It's definitely  possible -- although I'm quickly waning in how likely I think it really is.

In one of our past migrations, we stripped out a lot of `bytes` handling with the thinking being that it's not part of our documented contract with users. (If you pass us bytes we are free to fail in nasty ways, etc.)
I softened -- one could say I experienced a moment of weakness! -- when I realized that this helper does something really unintuitive for naive users if they pass bytes.

I think that I'm going to backtrack and remove it. Passing non-string, non-UUID data violates the user<->library contract and isn't covered by our docs. If we were to handle it anywhere like this, the better thing to do for the user would be to raise a TypeError or ValueError.

kurtmckee (approved):

PR #774, opened 2023-07-05
Implement AuthClient.update_project

sirosen:
Additionally, fix `create_project` in a minor way, to include a `*` marker in its signature for keyword-only arguments.

----
:books: Documentation preview :books:: https://globus-sdk-python--774.org.readthedocs.build/en/774/


sirosen:
Oh, oops. I omitted a test case for this. Change coming momentarily...

kurtmckee (approved):

PR #777, opened 2023-07-06
Add Auth Manage Projects example scripts

sirosen:
This doc was written in tandem with development of the AuthClient methods which are used here to implement the script functionality. Basically, the script I was using for testing is here elaborated a bit, simplified a bit, and generally reshaped to be documentation-grade.

---

This doc PR ~is~ was a draft for two reasons:
1. ~It is based on #776~ (done) 
2. ~It is blocked by a minor bug in update_project (see https://github.com/globus/globus-sdk-python/commit/9f59561b3577f7e31fefcb865c7c3086fd23bcef for details)~ We agreed to remove `update` functionality for now

----
:books: Documentation preview :books:: https://globus-sdk-python--777.org.readthedocs.build/en/777/

### DIFF
--- a/docs/examples/auth_manage_projects/list_and_create_projects.py
+++ b/docs/examples/auth_manage_projects/list_and_create_projects.py
@@ -48,7 +48,7 @@ def create_project(args):
     )
 
 
-def list_projects(args):
+def list_projects():
     auth_client = get_auth_client()
     for project in auth_client.get_projects():
         print(f"name: {project['display_name']}")
@@ -71,7 +71,7 @@ def execute(parser, args):
             parser.error("create requires --name")
         create_project(args)
     elif args.action == "list":
-        list_projects(args)
+        list_projects()
     else:
         raise NotImplementedError()
 
--- a/docs/examples/auth_manage_projects/list_and_create_projects.py
+++ b/docs/examples/auth_manage_projects/list_and_create_projects.py
@@ -1,0 +1,80 @@
+#!/usr/bin/env python
+
+import argparse
+import os
+
+import globus_sdk
+from globus_sdk.tokenstorage import SimpleJSONFileAdapter
+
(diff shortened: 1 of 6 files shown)


docs/examples/auth_manage_projects/list_and_create_projects.py
```diff
@@ -0,0 +1,80 @@
...
+    print(
+        auth_client.create_project(
+            args.name,
+            contact_email=userinfo["email"],
+            admin_ids=userinfo["sub"],
+        )
+    )
+
+
+def list_projects(args):
```
kurtmckee, line 51: Is `args` included as a parameter for easier extensibility? It doesn't make sense to me but could perhaps be a jumping-off point for new developers.

kurtmckee (approved):

docs/examples/auth_manage_projects/list_and_create_projects.py
sirosen: Extensibility and uniformity (since the other methods also consume `args`), but I could happily drop it if you think it's more likely to cause confusion than dispel it.

My goal here is less to write a truly exemplary script than it is to make something which is "good enough" and which will be approachable to the majority of our (python-inclined) users. The Auth logic is sufficiently intense to teach/learn/read that I think _anything_ which distracts from that should be removed.

So is it better to pass `args` to each method, even though it's useless in this case, or to omit it in this case but pass it in the others? Since you commented, it sounds to me like you're inclined towards removing it. I'll make that change, but let me know if you think it should be shaped differently or if there are other opportunities to simplify the example.

kurtmckee: I do lean toward removing it. If I opened the file (or copied/pasted it) in a competent IDE and saw code lint warnings I would wonder about the quality and utility of the code. Thanks for addressing this!

kurtmckee (approved):

PR #778, opened 2023-07-06
Fix the `FlowsClient.get_run()` `include_flow_description` parameter

kurtmckee:
Fixed
-----

-   Adjust the ``FlowsClient.get_run()`` ``include_flow_description`` parameter
    so it is sent as a lowercase string (``true`` or ``false``).

----
:books: Documentation preview :books:: https://globus-sdk-python--778.org.readthedocs.build/en/778/

### DIFF
--- a/src/globus_sdk/services/flows/client.py
+++ b/src/globus_sdk/services/flows/client.py
@@ -549,7 +549,7 @@ class FlowsClient(client.BaseClient):
         self,
         run_id: UUIDLike,
         *,
-        include_flow_description: bool = False,
+        include_flow_description: bool | None = None,
         query_params: dict[str, t.Any] | None = None,
     ) -> GlobusHTTPResponse:
         """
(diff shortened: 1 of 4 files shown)


src/globus_sdk/services/flows/client.py
```diff
@@ -587,14 +587,14 @@ def get_run(
...
+            consolidated_query_params["include_flow_description"] = value
 
         return self.get(
             f"/runs/{run_id}",
-            query_params={
-                "include_flow_description": include_flow_description,
-                **additional_query_params,
-            },
+            query_params=consolidated_query_params,
         )
```
sirosen, line 598: Is there a reason not to simply call it `query_params`?
i.e.
```python
if query_params is None:
    query_params = {}
```

(or inline that if line count is the enemy :wink: )

sirosen (commented):
The change looks good for what it is, but is fixing this client-side the correct approach?

I would prefer that we allow for the "normal python encoding" (`requests`) of `bool`s for any bool parameters in our APIs. I know that's conforming our services to a particular client, but a pretty important one for our usages.

If we fix the API, will we then revert this change in order to simplify the client?

src/globus_sdk/services/flows/client.py
kurtmckee: > is fixing this client-side the correct approach

Yes, because `requests` does not produce the documented values, and `responses` won't match unless the capitalization is correct.

sirosen (approved):

derek-globus (approved):
Yikes

sirosen (approved):
Thanks for making the changes!
We'll evolve things to a place where we're both happy. :smile:

PR #782, opened 2023-07-14
Implement FlowsClient.list_runs

sirosen:
----
:books: Documentation preview :books:: https://globus-sdk-python--782.org.readthedocs.build/en/782/

### DIFF
--- a/src/globus_sdk/_testing/data/flows/_common.py
+++ b/src/globus_sdk/_testing/data/flows/_common.py
@@ -1,9 +1,17 @@
 from __future__ import annotations
 
 import typing as t
+import uuid
 
 TWO_HOP_TRANSFER_FLOW_ID = "24bc4997-b483-4c25-a19c-64b0afc00743"
 TWO_HOP_TRANSFER_FLOW_OWNER_ID = "b44bddda-d274-11e5-978a-9f15789a8150"
+
+FLOW_ID = str(uuid.uuid1())
+RUN_ID = str(uuid.uuid1())
+USER1 = f"urn:globus:auth:identity:{uuid.uuid1()}"
+USER2 = f"urn:globus:auth:identity:{uuid.uuid1()}"
+GROUP = f"urn:globus:groups:id:{uuid.uuid1()}"
+
 TWO_HOP_TRANSFER_FLOW_USER_SCOPE = (
     "https://auth.globus.org/scopes/"
     + TWO_HOP_TRANSFER_FLOW_ID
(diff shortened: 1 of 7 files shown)


src/globus_sdk/services/flows/client.py
```diff
@@ -487,6 +487,47 @@ def delete_flow(
 
         return self.delete(f"/flows/{flow_id}", query_params=query_params)
 
+    @paging.has_paginator(paging.MarkerPaginator, items_key="runs")
+    def list_runs(
+        self,
+        *,
+        marker: str | None = None,
+        query_params: dict[str, t.Any] | None = None,
+    ) -> IterableRunsResponse:
```
derek-globus, line 496: There's a lot of filter query params (`filter_status`, `filter_roles`, `filter_role`, `filter_label` ,`filter_flow_id`, `filter_completion_time`, `filter_start_time`, `filter_flow_title`) allowed for this route.

I'm concerned about omitting the existence of these filters from all forms of sdk documentation. A lot of them are useful & users will only discover them if they navigate through the sdk docs, click on the api info tab, and follow the link to a separate website which documents the api info.

That's a really high level of effort required for a user to discover that they could filter runs by, for instance, the time the runs were started. Especially when we have customers (APS) who do have 1000s of runs that I could very much envision them wanting to filter through.

src/globus_sdk/_testing/data/flows/_common.py
```diff
@@ -1,9 +1,17 @@
...
+import uuid
 
 TWO_HOP_TRANSFER_FLOW_ID = "24bc4997-b483-4c25-a19c-64b0afc00743"
 TWO_HOP_TRANSFER_FLOW_OWNER_ID = "b44bddda-d274-11e5-978a-9f15789a8150"
+
+FLOW_ID = str(uuid.uuid1())
+RUN_ID = str(uuid.uuid1())
+USER1 = f"urn:globus:auth:identity:{uuid.uuid1()}"
+USER2 = f"urn:globus:auth:identity:{uuid.uuid1()}"
+GROUP = f"urn:globus:groups:id:{uuid.uuid1()}"
```
derek-globus, line 13: Pure curiosity: why use uuid v1?

src/globus_sdk/services/flows/client.py
sirosen: Just to put a bow on this: per our conversation, we're going to pull in only `filter_flow_id`.
It's the only one of these which we feel _somewhat_ confident _as of today_ offers value to keep even in a future in which we push users to query in Search more actively.

src/globus_sdk/_testing/data/flows/_common.py
sirosen: I generally only use v1 and v4 (because I never have needed the special namespacing features of v5).
Between v1 and v4 I don't really have much preference, but v1 is very slightly faster to generate (fewer random bits) so I tend towards it absent any particular reason to use v4 for test data, etc.

src/globus_sdk/services/flows/client.py
sirosen: I've updated with the addition of `filter_flow_id`, and took the opportunity to try a slightly different approach to testing the parameter passing behavior from what I've done in the past.

Rather than checking the request parameters directly, the test actually asserts that the filters have the desired effect on an orchestrated `ResponseList` with a few appropriately filter-able responses.

derek-globus (approved):

PR #783, opened 2023-07-14
Proof of concept: reject invalid inherited methods

sirosen:
AuthClient is gaining a number of methods related to authenticated usages which require token auth. These methods will not be functional for NativeAppAuthClient and ConfidentialAppAuthClient.

Because it is difficult to make a backwards-compatible change which alters the inheritance relationship for AuthClient and the other client types, it may be worth exploring other methods for ensuring that these "never functional" methods provide clear user-feedback that they do not work.
Possibly -- arguably -- the inheritance relationship that produces `isinstance(NativeAppAuthClient("foo"), AuthClient)` could be considered outside of our public interfaces. That would leave us free to make all three client types inherit from a new base (`BaseAuthClient`?) and attach the new methods only to `AuthClient`. However, if we consider this part of the public interface, then an alternative solution is needed.

This changeset introduces a class decorator
(`globus_sdk.utils.replace_notimplemented_methods`) which replaces named inherited methods with stubs which raise `NotImplementedError` with a generated message. The result is a slightly awkward trace, but a clear error on bad usage:

    >>> nc.get_identities(ids="foo")
      File ".../foo/bar/globus_sdk/utils.py", line 34, in injected_method
        raise NotImplementedError(
    NotImplementedError: NativeAppAuthClient does not support get_identities

If the methodology is considered sound, we can decorate NativeAppAuthClient and ConfidentialAppAuthClient with known-unsupported method names for SDK v3.x and plan to change the inheritance structure in v4.0 . A full enumeration of supported and unsupported usages would be needed to make this feasible, along with a detailed changelog enumerating the now NotImplementedError-raising methods.

----
:books: Documentation preview :books:: https://globus-sdk-python--783.org.readthedocs.build/en/783/

### DIFF
--- a/src/globus_sdk/services/auth/client/native_client.py
+++ b/src/globus_sdk/services/auth/client/native_client.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import logging
 import typing as t
 
+from globus_sdk import utils
 from globus_sdk._types import ScopeCollectionType, UUIDLike
 from globus_sdk.authorizers import NullAuthorizer
 
@@ -13,6 +14,9 @@ from .base import AuthClient
 log = logging.getLogger(__name__)
 
 
+@utils.replace_notimplemented_methods(
+    "get_identities",
+)
 class NativeAppAuthClient(AuthClient):
     """
     This type of ``AuthClient`` is used to represent a Native App's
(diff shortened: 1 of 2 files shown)


sirosen:
Although this utility might have independent life in the future (not sure I believe that?), this is part of the series of discussions and PRs which led to #849. I don't see benefit in keeping it open.

PR #785, opened 2023-07-17
Add search index lifecycle methods: create_index, delete_index, reopen_index

sirosen:
This work is structured in 3 commits for easy review:
1. Add the methods
2. Add test data for these methods
3. Add tests using that test data

----
:books: Documentation preview :books:: https://globus-sdk-python--785.org.readthedocs.build/en/785/

### DIFF
--- a/src/globus_sdk/_testing/data/search/reopen_index.py
+++ b/src/globus_sdk/_testing/data/search/reopen_index.py
@@ -15,5 +15,23 @@ RESPONSES = ResponseSet(
             "acknowledged": True,
         },
         metadata={"index_id": INDEX_ID},
-    )
+    ),
+    already_open=RegisteredResponse(
+        service="search",
+        method="POST",
+        path=f"/v1/index/{INDEX_ID}/reopen",
+        status=409,
+        json={
+            "code": "Conflict.IncompatibleIndexStatus",
+            "request_id": "e1ad6822156dea372027eee48c16e150",
+            "@datatype": "GError",
+            "message": (
+                "Index status (open) did not match required status for "
+                "this operation: delete_pending"
+            ),
+            "@version": "2017-09-01",
+            "status": 409,
+        },
+        metadata={"index_id": INDEX_ID},
+    ),
 )
--- a/src/globus_sdk/_testing/data/search/reopen_index.py
+++ b/src/globus_sdk/_testing/data/search/reopen_index.py
@@ -1,0 +1,19 @@
+import uuid
+
+from globus_sdk._testing.models import RegisteredResponse, ResponseSet
+
+INDEX_ID = str(uuid.uuid4())
+
+
+RESPONSES = ResponseSet(
+    default=RegisteredResponse(
+        service="search",
+        method="POST",
+        path=f"/v1/index/{INDEX_ID}/reopen",
+        json={
+            "index_id": INDEX_ID,
+            "acknowledged": True,
+        },
+        metadata={"index_id": INDEX_ID},
+    )
+)
--- a/src/globus_sdk/services/search/client.py
+++ b/src/globus_sdk/services/search/client.py
@@ -39,6 +39,119 @@ class SearchClient(client.BaseClient):
     # Index Management
     #
 
+    def create_index(
+        self, display_name: str, description: str
+    ) -> response.GlobusHTTPResponse:
+        """
+        Create a new index.
+
+        :param display_name: the name of the index
+        :type display_name: str
+        :param description: a description of the index
+        :type description: str
+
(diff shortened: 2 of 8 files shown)


src/globus_sdk/services/search/client.py
```diff
@@ -39,6 +39,119 @@ class SearchClient(client.BaseClient):
...
+
+        Globus Search does not immediately delete indices. Instead, this API sets the
+        index status to ``"delete-pending"``.
+        Search will move pending tasks on the index to the ``CANCELLED`` state and will
+        eventually delete the index.
+
+        If the index is a trial index, it will be deleted a few minutes after being
+        marked for deletion.
+        If the index is non-trial, it will be kept for 30 days and will be eligible for
+        use with the ``reopen`` API (see :meth:`~.reopen_index`) during that time.
```
kurtmckee, line 97: This seems like the wrong place to document the Search service backend implementation and limits (like "30 days").

src/globus_sdk/services/search/client.py
```diff
@@ -39,6 +39,119 @@ class SearchClient(client.BaseClient):
...
+
+                .. extdoclink:: Index Delete
+                    :ref: search/reference/index_delete/
+        """
+        log.info(f"SearchClient.delete_index({index_id!r}, ...)")
+        return self.delete(f"/v1/index/{index_id}")
+
+    def reopen_index(self, index_id: UUIDLike) -> response.GlobusHTTPResponse:
+        """
+        Reopen an index that has been marked for deletion, cancelling the deletion.
```
kurtmckee, line 127: I think that `cancelling` is the British spelling (it's how I learned it in school but the times they have a-changed apparently). I'm fine either way but wanted to offer this suggestion.

```suggestion
        Reopen an index that has been marked for deletion, canceling the deletion.
```

src/globus_sdk/_testing/data/search/reopen_index.py
```diff
@@ -0,0 +1,19 @@
...
+        service="search",
+        method="POST",
+        path=f"/v1/index/{INDEX_ID}/reopen",
+        json={
+            "index_id": INDEX_ID,
+            "acknowledged": True,
+        },
+        metadata={"index_id": INDEX_ID},
+    )
+)
```
kurtmckee, line 19: Is there a different response that might be returned for these scenarios:

* The index has not been marked for deletion
* The index is trial and within the brief window where it hasn't been deleted yet
* The index has already been deleted

src/globus_sdk/services/search/client.py
sirosen: Then perhaps only something shorter? I wanted to pull in something from the doc because `reopen` only makes sense with the broader context that `delete` may leave your index hanging out _but only for non-trial_.

I am a bit concerned that completely omitting the period (30 days) might make the doc here more confusing to read because it makes us describe it as something indefinite, when we do know the value. What about "several weeks"?

This is my crack at a rephrase:

> If the index is a trial index, it will be deleted shortly, but if it is non-trial, it will be kept in the ``"delete-pending"`` state for several weeks.
> Non-trial indices in the ``"delete-pending"`` state are eligible for use with the ``reopen`` API (see :meth:`~.reopen_index`) during that time.

src/globus_sdk/_testing/data/search/reopen_index.py
sirosen: Yeah, there are a few failure modes here, for all of these APIs.
Index create can hit your trial limit, delete can 404, etc.

I can add some failures into the dataset.

src/globus_sdk/services/search/client.py
kurtmckee: Hmm...is it possible to link to the documented constraints?

I like the specificity of "30 days" _et al_, but documenting the behavior and limits here seems duplicative, and worries me that it might age poorly. We can keep this as-is but I wanted to voice that concern.

Oh! I just realized I unlocked an achievement!

image

kurtmckee (approved):

src/globus_sdk/services/search/client.py
sirosen: The API reference doc link provides the requisite context, and since it's at the top of that page, there isn't really a better link to use than that one.

As of right now, I'm gently inclined to keep this as-is unless we have some alternative phrasing which still draws the connection that, for non-trial indices, `delete->reopen` is possible (for a time).

sirosen: I just saw that you approved, so I went ahead with the merge. But I'm happy to edit this further if there's a rephrase we can work out.

PR #786, opened 2023-07-17
Allow base_url without service_name in clients

sirosen:
base_url can now be used to effectively skip the check for `service_name == "_base"`, which has two significant impacts.

The first is simply that it is now possible to instantiate a BaseClient directly, which may prove useful for ad-hoc scripting. Although not generally a recommended approach (e.g. there is no benefit to using a BaseClient in lieu of a TransferClient), the possibility only makes the SDK more flexible.

The second effect is that a class inheriting from BaseClient no longer needs to set `service_name` if the setting has no appropriate value. This is relevant for 3rd party clients with nonstandard URL schemes (e.g. FuncX before the Compute rebranding) and for clients to an array of potential hosts, like GCSClient and any future
ActionProviderClient.

The only new test ensures that it is possible to instantiate a BaseClient directly, and the GCSClient.service_name value, although unchanged, is marked for a v4.0 update.

----
:books: Documentation preview :books:: https://globus-sdk-python--786.org.readthedocs.build/en/786/

### DIFF
--- a/changelog.d/20230717_162222_sirosen_improve_url_enforcement_logic.rst
+++ b/changelog.d/20230717_162222_sirosen_improve_url_enforcement_logic.rst
@@ -1,0 +1,13 @@
+Changed
+~~~~~~~
+
+- The enforcement logic for URLs in ``BaseClient`` instantiation has been
+  improved to only require that ``service_name`` be set if ``base_url`` is not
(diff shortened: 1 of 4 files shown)


ada-globus (approved):
This looks good and makes sense in light of our conversation yesterday!

PR #788, opened 2023-07-18
Add BaseAuthClient to separate methods

sirosen:
This should be viewed as my alternative pitch to #783 . Although #783 was more fun to implement, I believe this is a more correct path forward.

While testing I caught a surprise in that `ConfidentialAppAuthClient.get_identities` needs to be supported, which also serves to highlight how this separation is beneficial in terms of clarity.
For now, I've made no change to `NativeAppAuthClient.get_identities`, though we could remove that in a future version (at the very least, we should make sure it disappears in v4).

`git` and GitHub don't do a great job with this diff. Despite the noisy diff, client methods were not changed at all. They were mostly rearranged without even modified docstrings or comments.

---

Introduce a new base class for the Auth client hierarchy in order to split the developer API methods from the previously existing "base" methods. The important exception is `get_identities`, which supports client credentials authorization and therefore needs to stay on the base.

The changelog explains this change without drawing undue attention to the ways in which it could, in theory, be compatibility breaking for users relying on the inheritance relationship (e.g. via `isinstance` or type annotations).

No new tests are added and no tests are broken by this change.

----
:books: Documentation preview :books:: https://globus-sdk-python--788.org.readthedocs.build/en/788/

### DIFF
--- a/changelog.d/20230718_140424_sirosen_separate_auth_base.rst
+++ b/changelog.d/20230718_140424_sirosen_separate_auth_base.rst
@@ -1,0 +1,14 @@
+Changed
+~~~~~~~
+
+- The inheritance hierarchy for Globus Auth client classes has changed. (:pr:`NUMBER`)
+
+  - A new class, ``BaseAuthClient`` is used as the base for the other three
+    existing Globus Auth clients
+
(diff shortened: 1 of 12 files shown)


sirosen:
This work evolved into #849 ; closing as superseded.

PR #793, opened 2023-07-26
Set SpecificFlowClient default scope object and adjust utils.classproperty

sirosen:
There are two changes here.
Primarily, this changeset drives at improving the behavior at typing time and runtime for users of SpecificFlowClient.
1. `mypy` should allow `scope: str = myclient.scopes.user` -- for this, we need to manipulate the `scopes` classvar a little.
2. `SpecificFlowClient.scopes.user` should still error, but with a useful message which points at instance-only usage

In service of these changes, it turns out that our internal classproperty descriptor needs a minor adjustment. Although it is only used to implement the `BaseClient.resource_server` classproperty today, some pains are taken here to ensure that the behavior is kept unsurprising for future uses.

----
:books: Documentation preview :books:: https://globus-sdk-python--793.org.readthedocs.build/en/793/

### DIFF
--- a/changelog.d/20230726_115057_sirosen_specific_flow_client_default_scope_object.rst
+++ b/changelog.d/20230726_115057_sirosen_specific_flow_client_default_scope_object.rst
@@ -1,0 +1,8 @@
+Changed
+~~~~~~~
+
+- The ``scopes`` class attribute of ``SpecificFlowClient`` is now specialized
+  to ensure that type checkers will allow access to ``SpecificFlowClient``
+  scopes and ``resource_server`` values without ``cast``\ing. The value used is
+  a specialized stub which raises useful errors when class-based access is
+  performed. The ``scopes`` instance attribute is unchanged. (:pr:`NUMBER`)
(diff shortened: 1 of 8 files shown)


kurtmckee (approved):
This is fine, but I have a feeling that the amount of code to support mypy, code to overcome mypy constraints and limitations, code to fix Sphinx documentation, comments that change linter behavior, and comments that explain why weird things exist, are -- taken together -- contorting the codebase with dubious payoff.

sirosen:
I tend to agree WRT `classproperty`. It is probably something I never would have implemented if it didn't exist in 3.9 and seem like it was going to become part of the language in all future versions.
Now that it's here, I don't really want to drop it or introduce a metaclass just to replace it. Maybe in SDK v4 we quietly change this, but I'm not exactly sure how yet.

Sans `classproperty`, I think inserting the stub object is useful, in that `SpecificFlowClient` is _unlike_ most other clients, having no `scopes` at the class level but having them at the instance level, and this is a way of making that clearer both to mypy and to users.

Maybe there's a more refined approach to doing this, but again, I think v4.0 would be a good time to think about it. Perhaps we can reshape the BaseClient behavior and expectations to better match `SpecificFlowClient` (and `GCSClient`, which is similar).

sirosen:
I had to push another change to pacify pylint, but I'm going to move forward with the merge based on the prior review.

PR #796, opened 2023-08-06
Make GARE validation as simple as possible

sirosen:
This changeset completely replaces #795. I'm closing that as I think this is many times superior.

---

Rather than trying to build a framework, this is the "no framework" solution, in which we define very concrete helpers for very concrete purposes, including a dedicated one for the "ConsentRequired" Literal.

All of the GARE classes now inherit from an internal helper class, Serializable, which defines from_dict, to_dict, and _supported_fields in a generic way. The last of these *does* use some signature inspection magic, but nothing too abstruse.

The basic transformation is to replace each combination of a class-level annotation + a `SUPPORTED_FIELDS` entry with relevant assignment in `__init__`.

Some tangentially related simplifications and minor improvements are included:

- `extra_fields -> extra`

- Removal of unnecessary str splitting (after object is initialized)

- `isinstance` checking can also handle deserialization of dicts

- checking for non-null values will not accept `session_message`-only GARE data -- at least one of the semantic fields is required by the rewritten check

- GlobusAuthRequirementsError does not inherit from `GlobusError` or `Exception` -- it is not clear that this inheritance is useful or instructive to any user, since it mixes Exception hierarchies with data-representing hierarchies

- replace direct `ValueError` usage with a custom ValidationError class -- this avoids any messy scenarios in which a ValueError is accidentally introduced and incorrectly caught (e.g. from a stdlib call)

There are also some more significant improvements included. Most notably:

- Annotations explicitly do not accept `None` unless it is a valid value (i.e. annotations align with validation requirements)

- to_dict can now look for a concrete type (`Serializable`) and therefore can automatically invoke `to_dict` down a tree of objects

Although brevity is a non-goal of this changeset -- more verbose but clearer would be acceptable -- the result is almost 200 LOC lighter in the `src/` tree. The primary ways in which things became shorter appear to be:

- explicit version is often much shorter than the framework-ized version (e.g. `LegacyConsentRequiredAPError`), and even where the two are close, the explicit version is shorter

- `to_dict` and `from_dict` centralization

----
:books: Documentation preview :books:: https://globus-sdk-python--796.org.readthedocs.build/en/796/


src/globus_sdk/experimental/auth_requirements_error/_serializable.py
```diff
@@ -0,0 +1,60 @@
...
+
+class Serializable:
+    _EXLUDE_VARS: t.ClassVar[tuple[str, ...]] = ("self", "extra_fields", "extra")
+    extra: dict[str, t.Any]
+
+    @classmethod
+    def _supported_fields(cls) -> list[str]:
+        signature = inspect.signature(cls.__init__)
+        return [
+            name for name in signature.parameters.keys() if name not in cls._EXLUDE_VARS
```
ada-globus, line 17:
```suggestion
    _EXCLUDE_VARS: t.ClassVar[tuple[str, ...]] = ("self", "extra_fields", "extra")
    extra: dict[str, t.Any]

    @classmethod
    def _supported_fields(cls) -> list[str]:
        signature = inspect.signature(cls.__init__)
        return [
            name for name in signature.parameters.keys() if name not in cls._EXCLUDE_VARS
```

sirosen: At first, I read this and thought "oh no, how did this pass tests"?

And then I saw that I made the same typo twice! Thanks for spotting this.

ada-globus (commented):
Can we make sure this updates the docs to reflect changes to the interfaces? (e.g., https://github.com/globus/globus-sdk-python/blob/14887761bf5b6fc2f361153ed0436856707498fc/docs/experimental/auth_requirements_errors.rst#L65)

src/globus_sdk/experimental/auth_requirements_error/_validators.py
```diff
@@ -0,0 +1,77 @@
...
+        return None
+    if not isinstance(value, str):
+        raise ValidationError(f"'{name}' must be a string")
+    return value
+
+
+def consent_required_literal(name: str, value: t.Any) -> Literal["ConsentRequired"]:
+    if not isinstance(value, str) or value != "ConsentRequired":
+        raise ValidationError(f"'{name}' must be the string 'ConsentRequired'")
+    return t.cast(Literal["ConsentRequired"], value)
```
ada-globus, line 37: This feels to me a little bit like underlying details of specific variants are leaking into the validators, which makes this feel a little muddy. Not a hill I'm going to de on, and wouldn't want to spend more than a few minutes on it, but would prefer it if `ConsentRequired` was limited to the variants module.

sirosen: I don't have a strong stake in this being here vs there. Either way it's an implementation detail.

I'll happily move it to the end of `_variants.py`.

sirosen: Done! :smile:

src/globus_sdk/experimental/auth_requirements_error/_validators.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class ValidationError(ValueError):
+    pass
+
+
+def str_(name: str, value: t.Any) -> str:
+    if not isinstance(value, str):
+        raise ValidationError(f"'{name}' must be a string")
+    return value
```
kurtmckee, line 17: I recommend ending on `ValidationError`s throughout, like:

```suggestion
    if isinstance(value, str):
        return value
    raise ValidationError(f"'{name}' must be a string")
```

src/globus_sdk/experimental/auth_requirements_error/_validators.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def opt_str(name: str, value: t.Any) -> str | None:
+    if value is None:
+        return None
+    if not isinstance(value, str):
+        raise ValidationError(f"'{name}' must be a string")
```
kurtmckee, line 24: As above, I recommend ending on `ValidationError`.

In addition, should this note that the field could be `None` or `null` as well?

src/globus_sdk/experimental/auth_requirements_error/_variants.py
```diff
@@ -305,5 +194,15 @@ def to_auth_requirements_error(self) -> GlobusAuthRequirementsError:
...
+
+
+def _validate_consent_required_literal(
+    name: str, value: t.Any
+) -> Literal["ConsentRequired"]:
+    if not isinstance(value, str) or value != "ConsentRequired":
+        raise _validators.ValidationError(
+            f"'{name}' must be the string 'ConsentRequired'"
         )
+    return t.cast(Literal["ConsentRequired"], value)
```
kurtmckee, line 208: Could this simply be:

```suggestion
    return "ConsentRequired"
```

src/globus_sdk/experimental/auth_requirements_error/utils.py
```diff
@@ -56,18 +57,18 @@ def to_auth_requirements_error(
     # Prefer a proper auth requirements error, if possible
     try:
         return GlobusAuthRequirementsError.from_dict(error_dict)
-    except ValueError as err:
+    except _validators.ValidationError as err:
```
kurtmckee, line 60: `ValidationError` inherits from `ValueError`. Can the base `ValueError` be raised by `.from_dict()`?

src/globus_sdk/experimental/auth_requirements_error/_validators.py
sirosen: Sure thing; will apply momentarily as I catch up on other notes.
(The rest of these I'll just :+1: as I go.)

src/globus_sdk/experimental/auth_requirements_error/_variants.py
sirosen: :facepalm: 
I would be surprised if this didn't work. Will confirm momentarily.

src/globus_sdk/experimental/auth_requirements_error/_validators.py
```diff
@@ -0,0 +1,65 @@
...
+    return value
+
+
+def opt_str_list_or_commasep(name: str, value: t.Any) -> list[str] | None:
+    if value is None:
+        return None
+    if isinstance(value, str):
+        value = value.split(",")
+    if not (isinstance(value, list) and all(isinstance(s, str) for s in value)):
+        raise ValidationError(f"'{name}' must be a list of strings")
```
ada-globus, line 56: This is inaccurate for LegacyAuthorizationParameters and could be misleading if looking at the debug logs to determine why the parse failed.

```suggestion
        raise ValidationError(
            f"'{name}' must be a list of strings "
            "or a string of comma-delimited values."
        )
```

src/globus_sdk/experimental/auth_requirements_error/utils.py
sirosen: My take is that I'd like us to have a specific error class which means "validation failed".
Any case of a base `ValueError` _should_, in that line of thinking, mean that something which we didn't control failed.

Since the `ValidationError` could be raised into a caller-visible context, I also think I need to add it to the exported names in `__init__.py` and move it out of a purely-internal module (optional, but IMO cleaner).

src/globus_sdk/experimental/auth_requirements_error/_validators.py
kurtmckee: Similar questions for the other `opt_*` validators as well.

src/globus_sdk/experimental/auth_requirements_error/_validators.py
sirosen: Yeah, I waffled on this but can easily change. I believe that a test case assertion will change as well though. Let me cruise through Kurt's feedback and then I'll tackle this one.

src/globus_sdk/experimental/auth_requirements_error/utils.py
sirosen: Scratch that. I'm having trouble naming the module. I'm just going to expose it in `__init__` from the current location.

kurtmckee (approved):

PR #797, opened 2023-08-08
Add `IterableRunLogsResponse`

ada-globus:
Discovered in the course of working on the `globus flows run get-logs` command that the existing iterator doesn't point to the "entries" key, ~causing issues when using `Paginator.wrap()` which relies on the value of the `default_iter_key`.~

Update: Subsequently discovered the issue I was encountering arose due to a different issue. I still believe this is correct, but I am reverting to draft while I sort it out.

Update 2: Okay, I can see this is being overridden by the `paging.has_paginator` setting. I think this is still a _very_ modest improvement to correctness but don't expect a meaningful functional change either. Promoting back to PR.

----
:books: Documentation preview :books:: https://globus-sdk-python--797.org.readthedocs.build/en/797/

### DIFF
--- a/changelog.d/20230808_111757_ada_add_iterable_run_logs_response.rst
+++ b/changelog.d/20230808_111757_ada_add_iterable_run_logs_response.rst
@@ -1,0 +1,4 @@
+Changed
+~~~~~~~
+
+- ``FlowsClient.get_run_logs()`` now uses an ``IterableRunLogsResponse``. (:pr:`NUMBER`)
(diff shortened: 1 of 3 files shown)


kurtmckee (approved):

sirosen (approved):
Yep, this makes sense.

I can't remember how exactly we ended up with `default_iter_key` _and_ `items_key`, but I think these developed semi-independently.
I would be happy to work at some point on unifying these a little bit, but for now applying both does the trick.

PR #798, opened 2023-08-08
Introduce an internal TypeGuard library, globus_sdk._guards

sirosen:
This is a very modest internal improvement, but it does lead to cleaner and clearer internal code which is slightly stricter in certain edge-cases.

Some of the improvements are more obvious than others. For example, here's a usage site which reads more easily now:
```diff
-        if isinstance(self._dict_data.get("errors"), list) and all(
-            isinstance(subdoc, dict) for subdoc in self._dict_data["errors"]
-        ):
+        if _guards.is_list_of(self._dict_data.get("errors"), dict):
```

Is this _shorter_ / leaner / smaller? Not really.
Limiting ourselves to the `src/` tree, `_guards.py` itself is 62 LOC, and we have +110,-68.
So if we discount the `_guards.py` module itself, it's _marginally_ shorter. That's not really the point though.

This is more of a marginal improvement around the edges to how we do things, especially checking of `list` contents.
For example, in the TimerAPIError class, we have these changes:
```diff
-        elif isinstance(self._dict_data.get("detail"), list):
+        elif _guards.is_list_of(self._dict_data.get("detail"), dict):
 ... # context ...
-            self.errors = [
-                ErrorSubdocument(d)
-                for d in self._dict_data["detail"]
-                if isinstance(d, dict)
-            ]
+            self.errors = [ErrorSubdocument(d) for d in self._dict_data["detail"]]
```

These are not quite the same.
Because the only tool ready-to-hand before was `isinstance(X, list)`, that was used, and then we had to handle, later, the contents of that list. Putting `_guards.is_list_of(X, dict)` in easy reach eliminates that potential point of divergence between various components.

In the first 2 commits, I went after the "easy" version of this. It works nicely.
In the second 2 commits, I went after an aspirational goal to make the GARE validators part of a broader suite of behaviors we have within the SDK. They add a bit of complexity (`reduce / _Reducer`), but I think the result is nice.

----
:books: Documentation preview :books:: https://globus-sdk-python--798.org.readthedocs.build/en/798/


sirosen:
I've reverted and removed the fancy `_Reducer`. Per discussion, although it's cool, it's not intuitive enough to offset the cost of having to make sure that everyone understands it, or the potential maintenance burden if `mypy` or other type checkers start to treat it differently.

kurtmckee (approved):

PR #804, opened 2023-08-14
Add a `README.rst` content type to resolve a warning

kurtmckee:
This resolves a warning thrown while a GitHub workflow publishes to PyPI. [[recent example](https://github.com/globus/globus-sdk-python/actions/runs/5857734295/job/15880231600#step:7:24)]

> WARNING  `long_description_content_type` missing. defaulting to `text/x-rst`.

Development
-----------

- Add a ``README.rst`` content type to resolve a warning during publication to PyPI.

----
:books: Documentation preview :books:: https://globus-sdk-python--804.org.readthedocs.build/en/804/

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,6 +6,7 @@ author = Globus Team
 author_email = support@globus.org
 description = Globus SDK for Python
 long_description = file: README.rst
+long_description_content_type = text/x-rst
 url = https://github.com/globus/globus-sdk-python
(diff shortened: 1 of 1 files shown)


sirosen (approved):
Huh, I thought that was a safe default to rely upon.

It's surprising to me because I would have thought that the content-type was being set by `setuptools` during the build. Good to fix!

PR #805, opened 2023-08-14
Use Trusted Publishing

kurtmckee:
This PR migrates the SDK GitHub workflows from PyPI token secrets to Trusted Publishing.

The projects on TestPyPI and PyPI have already been updated, and the GitHub environments have already been created.

In addition, this has been tested by creating a `3.27.1b1` tag on this branch and pushing it to GitHub to trigger the TestPyPI workflow.

----
:books: Documentation preview :books:: https://globus-sdk-python--805.org.readthedocs.build/en/805/

### DIFF
--- a/.github/workflows/publish_to_pypi.yaml
+++ b/.github/workflows/publish_to_pypi.yaml
@@ -7,6 +7,9 @@ on:
 jobs:
   publish:
     runs-on: ubuntu-latest
+    environment: publish-pypi
+    permissions:
+      id-token: write
 
     steps:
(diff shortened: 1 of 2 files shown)


kurtmckee:
> Are there any details which were necessary when you set these up, or are they basically empty "labels" for now?

They're empty labels. No configuration was required to make it work, but I did find that any attempt to restrict what branch the TestPyPI environment ran against failed miserably...perhaps because it triggers on new tags, not branch pushes? So currently the environments simply exist.

PR #808, opened 2023-08-15
Add support for python3.12

sirosen:
- Test on 3.12.0rc1
- Add a changelog entry for this (the phrasing is a little bit "early", since 3.12.0 hasn't released yet, but this is the changeset which adds support for everything expected in 3.12.0)
- Update test dependencies to get the latest `coverage` version, for 3.12.0rc1 support
- Add py312 to the tox env list
- One minor tweak in the testsuite: `datetime.datetime.utcnow()` is deprecated because it produces a naive timestamp. Swap it to `datetime.datetime.now()` which is similar, still naive, and shouldn't trip things up much or at all.

----
:books: Documentation preview :books:: https://globus-sdk-python--808.org.readthedocs.build/en/808/


sirosen:
Two issues right now:
- need to specify as `3.12.0-rc1` for the github action (annoying but fine)
- the latest versions of some transitive dependencies in `mindeps` dropped py3.7 (it was going to happen sooner or later)

I'll probably do a separate body of work for the `mindeps` issue, possibly start on dropping 3.7 ourselves, etc.
Once I sort that out, I'll rebase and fix the py312 specifier for github.

.github/workflows/build.yaml
```diff
@@ -39,6 +39,8 @@ jobs:
             python-version: "3.11"
           - os: macos-latest
             python-version: "3.11"
+          - os: ubuntu-latest
+            python-version: "3.12.0rc1"
```
kurtmckee, line 43: You can fold this directly into the `python-version` list above as simply `"3.12"`.

Then, in the `with:` property of the `setup-python` action, you can add:

```yaml
allow-prereleases: true
```

These lines are not in scope for this diff so I cannot leave comments or suggestions. :cry:

changelog.d/20230815_161655_sirosen_test_on_py3_12.rst
```diff
@@ -0,0 +1,4 @@
+Python Support
+~~~~~~~~~~~~~~
+
+- Add support for python3.12 (:pr:`NUMBER`)
```
kurtmckee, line 4:
```suggestion
- Add support for Python 3.12. (:pr:`NUMBER`)
```

kurtmckee (changes requested):
It may be possible to add constraints for the dependencies that hinge on Python 3.7...? I don't recall if we have deprecation plans of our own for Python 3.7 support yet.

.github/workflows/build.yaml
sirosen: Thanks for the pointers!

My intent was to give us only one 3.12 build -- and my linux-first bias was at play.
But I don't mind setting up all 3 platforms, since it might catch things, so I'll go with your suggestion.

I'll also set `allow-prereleases` and we'll see if that does the trick for a nice clean build.

sirosen:
> It may be possible to add constraints for the dependencies that hinge on Python 3.7...? I don't recall if we have deprecation plans of our own for Python 3.7 support yet.

Yeah, we don't have a plan yet, but we should. I want to try to whip something up which will read our `python_requires` from package data and validate it... _against our CI config_.
Plus, use that same data source when running the `freezedeps` script to get a `pip-compile` run on the correct python.

We'll see if I commit anything -- I give myself a 50/50 shot at writing something non-disastrous.

.github/workflows/build.yaml
kurtmckee: Putting `"3.12"` into the list of versions above would only add it for Linux, wouldn't it?

sirosen: _Looks at the build matrix again._

Yes. Yes it would.

...

Some projects have bigger default matrices, but we're just using `include` for our non-linux builds. I forgot.

Well, anyway! We can add more builds to the matrix whenever we want, so we can always debate more builds in the future. :smile:

kurtmckee (changes requested):
The pip-compile command is being run with Python 3.11.

Can it be run with Python 3.7 to resolve the Python 3.7 pinning problems? Alternatively, I cooked up a Python-by-Python requirements generator from within tox that can be used to generate the requirements files for all supported versions of Python.

kurtmckee (approved):
I'll leave it to you to tackle the Python 3.7 compatibility.

This PR gets a big ol' :+1:!

sirosen:
I had to do a little bit of tinkering post-rebase, mostly to get the 3.12 requirements regenerated.
Now that I've sorted that out, I'm going to let CI run and merge based on the prior approval when it passes.

PR #809, opened 2023-08-16
Validate Requires-Python metadata against the "mindeps" build in CI config

sirosen:
# Rationale

This is my first crack at solving a problem I've dealt with in this and other projects in the past:

Suppose you drop support for a python version (e.g. 3.7, which is EOL and which we should drop soon).
There are numerous fields and locations which need updating. CI, tox, classifiers, etc.
The trouble with doing this is that `python_requires`/`Requires-Python` (one is the key in setuptools, the other is the field in PKG-INFO) has very particular semantics and _is the most important of all of these fields_.
Failing to set your new python version boundary correctly results in solvers *always* having some chance of picking up on your new version even on an unsupported python version.

Ideally updates are done perfectly and there's never any worry, but I've seen this go wrong and I've come close to doing it wrong myself.

Happily, there's a clear directionality at play: if you no longer test on a python version, you shouldn't say you support that version.
We have a specific build (mindeps) whose purpose is to check the lowest versions of our requirements, *including the meta-requirement that is the python version*.
So all we need to do is make sure that "mindeps" is testing our Requires-Python.

# Approach & Details

## Part I: `get_python_requires.py`

First, I wrote a script to build a virtualenv, install `build`, `python -m build --sdist`, and extract the `Requires-Python` data from the sdist. It keeps the virtualenv for `build` cached and puts the sdist in a temp dir so that we're always building the parts we need afresh but we're not doing a full virtualenv construction on every call.
This is structured as a standalone component which I might test and ship separately at some point as its own packaged application. (Doing so would allow use of `packaging` to do more sophisticated version parsing, etc.)

Now I can reliably pull `Requires-Python` from any PEP 517 package, including our own. It will work if we switch to `hatch` or `flit` or `poetry` or `distutils2023` (no, that's not real, don't worry :wink: ).

## Part II: `ensure_min_python_is_tested.py`

Now we need to know what version we're testing. There are several sources of info we _could_ check (e.g. `tox list`), but for this I have intentionally chosen *only* the mindeps CI build. We can expand it in the future.

To do this, I need a YAML parser. It was nice to make `get_python_requires.py` "pure" stdlib python which wraps its need for `build`, but that won't work here.
So this _will always_ be called via `tox`. Once the YAML parser is available, pulling out `test-mindeps` and the `python-version` it uses is pretty easy.
Check against `get_python_requires.py` and we're done!

## Part III: Cleanup

There are some minor changes to finalize.
- cache that `build` virtualenv inside of the `.tox` dir so that we don't dirty up everyone's homedirs
- add this test to CI tests

There are additional improvements we could allow. e.g. If `build` were installed in the surrounding env, `get_python_requires.py` would not need to build its own venv. I can pursue anything like this if it is important to anyone reviewing this.

----
:books: Documentation preview :books:: https://globus-sdk-python--809.org.readthedocs.build/en/809/


sirosen:
I've rebuilt the `get_python_requires.py` script into a public alpha package, the MetaData DJ (`mddj`).
As a result, this changeset now doesn't carry that project in tree at all.
We just install `mddj==0.0.3` and call it with `python -m mddj read python-requires --lower-bound`.

sirosen:
Now that `mddj` supports parsing tox factors from `tox --listenvs`, I've bumped the version of it to 0.0.6 and made it part of the checks.
Currently:
```
$ mddj read tox min-version
3.7
```

scripts/ensure_min_python_is_tested.py
```diff
@@ -43,3 +43,21 @@
             file=sys.stderr,
         )
         sys.exit(1)
+
+
+proc = subprocess.run(
+    ["python", "-m", "mddj", "read", "tox", "min-version"],
```
kurtmckee, line 49: Can mddj simply be imported? It strikes me as odd to run mddj instead of calling some function in it collect a value via STDOUT.

sirosen: The issue here is that I've only defined/documented CLI interfaces for `mddj`.

I could add library/functional interfaces as well, but I haven't done so yet.

If I'm going to do this, it would be a dedicated `mddj.api` subpackage which declares itself as the public library interface. Does that sound sensible? `mddj.api.read.tox.min_version()` as the callable to use? (Some classes or other namespace type to make that path reflect the CLI)

I'll do this if you think it's worth it, but I wouldn't want to start trying to use any existing functions from there.

kurtmckee: Let's not force the interface of mddj just so this feedback has an actionable outcome. Let's just mark this as resolved.

kurtmckee (approved):

scripts/ensure_min_python_is_tested.py
sirosen: Cool by me! I'm happy to circle back on this later though. I like the idea of trying to build an explicit functional interface to match the CLI -- seems like a fun side-project when I have some time to give it a shot.

PR #810, opened 2023-08-16
Freeze 'mindeps' requirements with py3.7

sirosen:
Rather than using the same (default) python for all of the requirements files in `requirements/*.in`, specify it more strictly thusly:

- if `--mindeps` is passed, the freeze script will target the files with `mindeps` in their names
- otherwise, the freeze script will skip files with `mindeps` in their names
- `tox.ini` is extended to make `freezedeps` a label which runs `freezedeps` and `freezedeps-mindeps`
- `freezedeps` has the same invocation path as today
- `freezedeps-mindeps` uses py3.7, holds back the `pip` version to work around a pip-tools incompatibility which isn't ported back to the py3.7 pip-tools releases, and passes `--mindeps` to the script

The end result is that we maintain *exactly* the same number of files as before, but now we use py3.7 for the 'mindeps' files and the user's current python (3.11 for me) for all of the others.

----
:books: Documentation preview :books:: https://globus-sdk-python--810.org.readthedocs.build/en/810/


sirosen:
_Nevermind._

I completely forgot that one of our requirements is to test python3.7 against the *non* mindeps build. I'll be replacing this, so I'm just going to close to provide a clearer solution.

PR #811, opened 2023-08-16
Freeze requirements data per python version

sirosen:
Rather than using the same (default) python for all of the requirements files in `requirements/*.in`, specify it more strictly by making each supported python version map out to a set of requirements files.

The mapping is that for any `requirements/foo.in`, we produce `requirements/py{major}.{minor}/foo.txt` for all python versions we support. `major.minor` aligns with the `{py_dot_ver}` available under `tox`, making it trivially easy to then refer to these built requirements from `tox.ini`.

There is one tweak to pin back the version of `pip` under py3.7 in order to support an older pip-tools version.

---

This PR replaces #810 

----
:books: Documentation preview :books:: https://globus-sdk-python--811.org.readthedocs.build/en/811/


scripts/freezedeps.py
kurtmckee, line 1: It looks like freezedeps could be migrated directly to `tox.ini` as a series of 4 commands.

requirements/py3.11/docs.txt
kurtmckee, line 1: The Read the Docs config needs to be updated to point to this file (and use Python 3.11!). :grin:

kurtmckee (changes requested):
Read the Docs will break; the config needs to be updated.

`freezedeps.py` can likely disappear into just four commands in `tox.ini`.

kurtmckee:
* Doc builds are now passing, and use Python 3.11.
* Doc requirements are only built for Python 3.11.
* Docs are built as a default environment in the test suite to catch regressions locally.
* There's now a compatibility bug between Furo and Sphinx and the SDK configuration, so the Sphinx version is pinned to 7.1.2.
* `freezedeps.py` is replaced with explicit commands and config in `tox.ini`.

https://readthedocs.org/projects/globus-sdk-python/builds/

kurtmckee (approved):

sirosen (commented):
Good catch about the RTD build.

A couple of notes in approval of your changes (I can't approve my own PR, but I can merge it!):
- I agree that this reads simpler than the globbing freezedeps script, and it more gracefully handles the py311-docs situation.
- We could maybe limit the `mindeps` requirements a little more, to only specific versions.
- :crossed_fingers: that Furo gets a path soon for the new sphinx version
- Let's keep an eye on pip-tools releases -- hopefully we won't need to pass `--resolver` or `--strip-extras` soon

PR #813, opened 2023-08-18
Add a `prompt` kwarg to `AuthClient.oauth2_get_authorize_url()`

kurtmckee:
Added
-----

-   Add a ``prompt`` keyword parameter to ``AuthClient.oauth2_get_authorize_url()``.

    Setting this parameter requires users to authenticate with an identity provider,
    even if they are already logged in. Doing so can help avoid errors caused by
    unexpected session required policies, which would otherwise require a second,
    follow-up login flow.

    ``prompt`` could previously only be set via the ``query_params`` keyword parameter.
    It is now more discoverable.

----
:books: Documentation preview :books:: https://globus-sdk-python--813.org.readthedocs.build/en/813/

### DIFF
--- a/tests/functional/services/auth/test_auth_client_flow.py
+++ b/tests/functional/services/auth/test_auth_client_flow.py
@@ -37,46 +37,50 @@ def confidential_client(no_retry_transport):
     )
 
 
-# build a partial matrix over
+# build a nearly-diagonal matrix over
 #
 #   domain: str | list[str]
 #   identities: uuid | str | list[uuid] | list[str] | list[str | uuid]
 #   policies: uuid | str | list[uuid] | list[str] | list[str | uuid]
+#   prompt: Literal["prompt"] | None
 #
-# start by seeding a small set of combinations of params to test, then run some loops to
-# fill in the rest
-_ALL_SESSION_PARAM_COMBINATIONS = [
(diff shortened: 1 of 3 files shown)


tests/functional/services/auth/test_auth_client_flow.py
```diff
@@ -37,46 +37,50 @@ class CustomAuthClient(globus_sdk.ConfidentialAppAuthClient):
...
+# There must be 4 parameters so parametrized tuple unpacking works in the test.
+_ALL_SESSION_PARAM_COMBINATIONS = [(None,) * 4]
+for idx, options in enumerate(
+    (domain_options, identity_options, policy_options, prompt_options)
 ):
-    _ALL_SESSION_PARAM_COMBINATIONS.append((None, None, policy_option))
+    for option in options:
+        parameters = [None] * 4
+        parameters[idx] = option
+        _ALL_SESSION_PARAM_COMBINATIONS.append(tuple(parameters))
```
sirosen, line 76: Since the authorize URL builder doesn't do any logic to combine any of these, I think this is a very neat solution.

We can always manually append things if there's some combination introduced in the future with special rules/behavior.

sirosen (approved):

PR #819, opened 2023-08-18
Tighten requirements generation and testing

kurtmckee:
This PR introduces these changes:

* `test-mindeps` requirements are only generated for Python 3.7
* `docs` and `typing` requirements are only generated for Python 3.11
* CI explicitly uses Python 3.11, not _any Python 3 version_
* `typing-mindeps` no longer exists and will not be tested
* The Sphinx version restriction is lifted because furo theme was patched in the last 20 hours

This should close most (if not all) Dependabot PRs that were opened against the testing requirements files.

----
:books: Documentation preview :books:: https://globus-sdk-python--819.org.readthedocs.build/en/819/


sirosen (approved):
I checked our other "3.x" builds to see if anything else looks like it will be impacted, and I think we'll be fine. We can always tweak CI things as needed, but this seems like the right spec in terms of our `requirements x versions x builds`. So if there's any change needed, it will be minor.

Also, for the public record: we're pretty sure that typing-mindeps isn't an effective test. I'm under the impression that type checkers may bundle/vendor `typing-extensions` rather than using the installed version.

PR #827, opened 2023-08-22
Add support for pausing and resuming timers

ada-globus:
image

### DIFF
--- a/changelog.d/20230822_155752_ada.rst
+++ b/changelog.d/20230822_155752_ada.rst
@@ -1,0 +1,4 @@
+Added
+~~~~~
+
+- Add support for pausing and resuming timers. (:pr:`NUMBER`)
--- a/src/globus_sdk/services/timer/client.py
+++ b/src/globus_sdk/services/timer/client.py
@@ -116,3 +116,38 @@ class TimerClient(client.BaseClient):
         """
         log.info(f"TimerClient.delete_job({job_id})")
         return self.delete(f"/jobs/{job_id}")
+
+    def pause_job(
+        self,
+        job_id: UUIDLike,
+    ) -> response.GlobusHTTPResponse:
+        """
+        ``POST /jobs/<job_id>/pause``
+
+        **Examples**
+
+        >>> timer_client = globus_sdk.TimerClient(...)
+        >>> timer_client.pause_job(job_id)
+        """
+        log.info(f"TimerClient.pause_job({job_id})")
+        return self.post(f"/jobs/{job_id}/pause")
+
+    def resume_job(
+        self,
+        job_id: UUIDLike,
+        *,
+        update_credentials: bool | None = None,
+    ) -> response.GlobusHTTPResponse:
+        """
+        ``POST /jobs/<job_id>/resume``
+
+        **Examples**
+
+        >>> timer_client = globus_sdk.TimerClient(...)
+        >>> timer_client.resume_job(job_id)
+        """
+        log.info(f"TimerClient.resume_job({job_id})")
+        data = {}
+        if update_credentials is not None:
+            data["update_credentials"] = update_credentials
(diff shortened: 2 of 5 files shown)


tests/functional/services/timer/test_jobs.py
```diff
@@ -99,3 +99,17 @@ def test_delete_job(timer_client):
...
+
+
+def test_pause_job(timer_client):
+    meta = load_response(timer_client.pause_job).metadata
+    response = timer_client.pause_job(meta["job_id"])
+    assert response.http_status == 200
+    assert "Successfully paused" in response.data["message"]
+
+
+def test_resume_job(timer_client):
```
kurtmckee, line 111: I'd prefer to see testing for `update_credentials` getting passed correctly (getting left out entirely, or being set to `true/false` in JSON.

kurtmckee (changes requested):

kurtmckee (changes requested):

changelog.d/20230822_155752_ada.rst
```diff
@@ -0,0 +1,4 @@
+Added
+~~~~~
+
+- Add support for pausing and resuming timers. (:pr:`NUMBER`)
```
sirosen, line 4: I like for the SDK changelogs to call out the method names when we add things. This is both for users and for us -- if we want to find the version in which `pause_job` was added, it's nice to be able to search the changelog for `pause_job`.

Exact phrasing is up to you, but I'd do it something like
> - Add ``TimerClient.pause_job`` and ``TimerClient.resume_job`` for pausing and resuming timers.

src/globus_sdk/services/timer/client.py
```diff
@@ -116,3 +116,38 @@ def delete_job(
...
         log.info(f"TimerClient.delete_job({job_id})")
         return self.delete(f"/jobs/{job_id}")
+
+    def pause_job(
+        self,
+        job_id: UUIDLike,
+    ) -> response.GlobusHTTPResponse:
+        """
+        ``POST /jobs/<job_id>/pause``
+
```
sirosen, line 126: We need to make sure we have parameter docs on all of our methods.
I notice that we're missing them for all of the other TimerClient methods (!). I'll work up a quick PR to fix that.
But we should add it for the new methods to make sure we don't add to the problem.

```suggestion

        :param job_id: The ID of the timer to pause
        :type job_id: str or UUID
```

Aside: I thought we had a pylint check which catches undocumented parameters. I'll need to investigate what's wrong with that.

src/globus_sdk/services/timer/client.py
```diff
@@ -116,3 +116,38 @@ def delete_job(
...
+    def resume_job(
+        self,
+        job_id: UUIDLike,
+        *,
+        update_credentials: bool | None = None,
+    ) -> response.GlobusHTTPResponse:
+        """
+        ``POST /jobs/<job_id>/resume``
+
```
sirosen, line 143: As above, param docs. Suggested content as a starting point for this:

```suggestion

        :param job_id: The ID of the timer to resume
        :type job_id: str or UUID
        :param update_credentials: When true, replace the credentials for the timer based
            on the credentials for this resume call. This can be used to resolve session
            and consent errors on timers, but it also could introduce session and consent
            errors, depending on the credentials being used to resume.
        :type update_credentials: bool, optional
```

sirosen (changes requested):
There are some minor changes I'd like us to apply to the changelog and docstrings. No functional changes though! :+1:

src/globus_sdk/services/timer/client.py
ada-globus: Yeah, I'd just matched the pattern in the module—should be better now! Additionally updated the descriptions while I was at it (replaced the route with a description of the actual functionality, which should be more helpful).

kurtmckee (approved):

PR #829, opened 2023-08-25
Move pylint config into setup.cfg and cleanup

sirosen:
Mostly a lift-and-shift of config, but also the disable list comments can be made clearer.

Leave an important note about accept-no-param-doc, but leave the setting commented out. Currently we have 68 failures, too many for a quick fix, so this will be a process to resolve.

----
:books: Documentation preview :books:: https://globus-sdk-python--829.org.readthedocs.build/en/829/


kurtmckee (approved):
Is there anything in our toolchain that explicitly tells pylint to use `.pylintrc` as its config file? If not, then I think we're good to go!

sirosen:
I'm pretty sure the only time that pylint is invoked via any part of the repo is via `tox`. If it turns out that I'm wrong, I'd treat that as a bug that should be corrected to use `tox r -e pylint` unless we have some other need.

But I really think we are safe, so I'm going forward with the merge. After this, I'll start toggling that docstring flag so that I can heal our missing param docs.

PR #833, opened 2023-08-25
Suppress pylint docparams warnings in client.py

sirosen:
In many other cases missing parameters from the docstring represents a
problem we need to fix. In this case, however, we have everything
documented in one place, and then other methods refer to it ("see
`request` params"). We could expand out the docs in all of them, but
it makes the experience reading the docs worse, rather than better.
In this unusual case, suppress the warning.

For our `resource_server` classproperty, a slightly different warning
is appropriate to suppress. pylint cannot understand the decorator as
a descriptor, so it expects documentation of an implicit parameter
(which, as with the other case, would be detrimental to the docs).

----
:books: Documentation preview :books:: https://globus-sdk-python--833.org.readthedocs.build/en/833/

### DIFF
--- a/src/globus_sdk/client.py
+++ b/src/globus_sdk/client.py
@@ -114,7 +114,9 @@ class BaseClient:
         self._app_name = self.transport.user_agent = value
 
     @utils.classproperty
-    def resource_server(self_or_cls) -> str | None:
+    def resource_server(  # pylint: disable=missing-any-param-doc
+        self_or_cls,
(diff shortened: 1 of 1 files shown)


sirosen:
Slightly odd formatting from running
```
$ gh pr create --label no-news-is-good-news --fill
```

But I do love not having to jump around the web GUI so much for a series of small PRs like this. Expect to see others like this.

kurtmckee (approved):

PR #836, opened 2023-08-27
Fix failing docparams in transport/

sirosen:
----
:books: Documentation preview :books:: https://globus-sdk-python--836.org.readthedocs.build/en/836/

### DIFF
--- a/src/globus_sdk/transport/requests.py
+++ b/src/globus_sdk/transport/requests.py
@@ -262,7 +262,7 @@ class RequestsTransport:
         ``max_sleep``.
 
         :param ctx: The context object which describes the state of the request and the
-            retries which may already have been attempted
+            retries which may already have been attempted.
         :type ctx: RetryContext
         """
         sleep_period = min(self.retry_backoff(ctx), self.max_sleep)
@@ -396,7 +396,7 @@ class RequestsTransport:
         Check if a network error was encountered
 
         :param ctx: The context object which describes the state of the request and the
-            retries which may already have been attempted
+            retries which may already have been attempted.
         :type ctx: RetryContext
         """
         if ctx.exception and isinstance(ctx.exception, requests.RequestException):
@@ -408,7 +408,7 @@ class RequestsTransport:
         Check for a retry-after header if the response had a matching status
 
         :param ctx: The context object which describes the state of the request and the
-            retries which may already have been attempted
+            retries which may already have been attempted.
         :type ctx: RetryContext
         """
         if (
@@ -427,7 +427,7 @@ class RequestsTransport:
         the request
 
         :param ctx: The context object which describes the state of the request and the
-            retries which may already have been attempted
+            retries which may already have been attempted.
         :type ctx: RetryContext
         """
         if ctx.response is not None and (
@@ -448,7 +448,7 @@ class RequestsTransport:
         The check is flagged to only run once per request.
 
         :param ctx: The context object which describes the state of the request and the
-            retries which may already have been attempted
+            retries which may already have been attempted.
         :type ctx: RetryContext
         """
         if (  # is the current check applicable?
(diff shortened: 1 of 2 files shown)


src/globus_sdk/transport/requests.py
```diff
@@ -260,6 +260,10 @@ def _retry_sleep(self, ctx: RetryContext) -> None:
         Given a retry context, compute the amount of time to sleep and sleep that much
         This is always the minimum of the backoff (run on the context) and the
         ``max_sleep``.
+
+        :param ctx: The context object which describes the state of the request and the
+            retries which may already have been attempted
```
kurtmckee, line 265: There is a fairly consistent inconsistency with closing periods throughout the docstrings.

In most cases, the last sentence (which might be the only sentence) doesn't have a closing period. However, as highlighted above, some sentences run on directly into the next sentence without a closing period.

I prefer that all sentences have a closing period, whether they are the only sentence or are the final sentence in a docstring or parameter definition. Developing this habit will prevent run-on sentences, though calling it out when missing will initially result in more friction in PR reviews.

kurtmckee (changes requested):

src/globus_sdk/transport/requests.py
sirosen: Just to share some reading I've just done and how and why I have this somewhat inconsistent behavior as an author:

I don't like to put a period on a short, non-sentence phrase, e.g. `ctx: a Context object`
On a longer phrase like the above it becomes unclear to me whether or not it should be punctuated. It's a definition, not normal sentence, but it's longer and has more grammatical structure.

However! I'm fixing that behavior starting today. I just looked to a dictionary for guidance and found that OED puts a period after all definitions, short or long.
I'm tempted a little to follow that precedent even though I don't think it's the norm for programmers to put periods after very short phrases.

At the very least this removes, for me, any discomfort about punctuating a sentence fragment. I'll add periods here for sure, and I now have some new context to inform how I do this.

I'm still not sure how to handle very small phrases. OED puts a period on all definitions, no matter how short. Perhaps we should do the same, but it feels like an edge case...

kurtmckee: I generally follow "If it's a sentence, it needs a period. If it's a fragment, it doesn't necessarily need a period."

Docstrings are always sentences. Comments are almost always sentences.

Not rules I live and die by, just guidelines that make things more readable for me.

sirosen: I can generally get behind that line of thinking, but with a caveat:

The concrete case above is a fragment.
> The context object which describes the state of the request and the retries which may already have been attempted

is just a very long noun. In a sentence, the descriptive phrase probably becomes an appositive, with commas improving readability and clarity:
> The context object, which describes the state of the request and the retries which may already have been attempted, is passed to this method.

But I think this case should be given a trailing period so that there is less opportunity for disagreement between different bits of doc. If we read a colon in a definition as implicitly constructing a sentence with a being verb, then a trailing period fits. i.e.
```
foo: bar.
```
Can be read as
```
"foo" is a "bar".
```

---

Anyway, I've added periods here and I'm going to try to start including periods for
- trailing sentences where absent
- long fragments

For very short fragments (< 5 words), I'll probably exercise some judgement and usually omit a trailing period.

kurtmckee (approved):

src/globus_sdk/transport/requests.py
kurtmckee: I just realized that what I was trying to draw attention to -- what prompted me to call out periods in the first place -- might not have been clear.

It's line 260 into line 261 that caught my eye and prompted these comments.

sirosen: > It's line 260 into line 261 that caught my eye and prompted these comments.

Aha! Yes, that's very much "just wrong" and not subject to as much interesting discussion. :sweat_smile: 

I'll swing back around with a PR to address that too.

PR #849, opened 2023-08-30
Refactor the AuthClient inheritance tree

sirosen:
- introduce AuthLoginClient and AuthServiceClient as the dividing line between each part of the tree
- AuthClient becomes an alias for AuthLoginClient
- the "app client" classes inherit from login client
- a new base class, AuthBaseClient, provides OIDC and JWK methods, base data like service name, scopes, and error class, but no other methods

The calls provided on AuthBaseClient are those which are simple GET methods (and a decorated version thereof) which do not require authentication. As such, they make sense to expose equally to all client types.

----
:books: Documentation preview :books:: https://globus-sdk-python--849.org.readthedocs.build/en/849/


kurtmckee (changes requested):

tests/unit/test_auth_clients.py
```diff
@@ -3,28 +3,60 @@
...
+    client = globus_sdk.AuthClient()
+    # accessing the attribute warns, but provides None
+    with pytest.warns(globus_sdk.RemovedInV4Warning):
+        assert client.client_id is None
+
+
+@pytest.mark.parametrize("pass_value_as", ("str", "uuid"))
+def test_service_client_allows_client_id_but_warns(pass_value_as):
+    pass_value = CLIENT_ID_UUID if pass_value_as == "uuid" else CLIENT_ID_STR
+
```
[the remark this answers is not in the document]
sirosen: I think you'll like the way I fixed this.
IIRC, this was something I did to give the tests nice IDs from pytest -- but, of course, `ids=...` is a better way to do that.

changelog.d/20230829_122956_sirosen_loginclient.rst
```diff
@@ -0,0 +1,28 @@
...
+  (:pr:`NUMBER`)
+
+  - A new class, ``AuthLoginClient``, is the base for ``NativeAppAuthClient``
+    and ``ConfidentialAppAuthClient``. These classes no longer inherit from
+    ``AuthClient``.
+
+  - ``AuthClient`` is now the only class which provides functionality
+    for accessing Globus Auth APIs. Previously, some of the API accessing
+    methods were inherited by the ``AuthClient`` subclasses, but would never
+    work.
```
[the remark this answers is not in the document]
sirosen: This is a good callout -- I did a marginally more extensive change to include a third bullet-point to clarify the rearrangement further.

src/globus_sdk/services/auth/client/base_login_client.py
```diff
@@ -73,6 +73,10 @@ def __init__(
             f"client_id='{client_id}', type(authorizer)={type(authorizer)}"
         )
 
+    # FYI: this get_openid_configuration method is duplicated in AuthClient
+    # if this code is modified, please update that copy as well
+    # this will be resolved in SDK v4 by removing this copy and restructuring the
+    # LoginClients to contain AuthClient instances
```
kurtmckee, line 79: It's possible that this comment will prove incorrect when we launch v4. Perhaps just these two lines could be removed.

```suggestion
```

src/globus_sdk/services/auth/client/base_login_client.py
```diff
@@ -99,6 +103,10 @@ def get_jwk(
     ) -> dict[str, t.Any]:
         ...
 
+    # FYI: this get_jwk method is duplicated in AuthClient
+    # if this code is modified, please update that copy as well
+    # this will be resolved in SDK v4 by removing this copy and restructuring the
+    # LoginClients to contain AuthClient instances
```
kurtmckee, line 109: #same

```suggestion
```

src/globus_sdk/services/auth/client/service_client.py
```diff
@@ -130,6 +133,9 @@ def get_jwk(
     ) -> dict[str, t.Any]:
         ...
 
+    # FYI: this get_jwk method is duplicated in AuthLoginBaseClient
+    # if this code is modified, please update that copy as well
+    # this will be resolved in SDK v4 by making this the only copy
```
kurtmckee, line 138: #same

```suggestion
```

src/globus_sdk/services/auth/client/service_client.py
```diff
@@ -104,6 +104,9 @@ def client_id(self, value: UUIDLike) -> None:
         )
         self._client_id = str(value) if value is not None else None
 
+    # FYI: this get_openid_configuration method is duplicated in AuthLoginBaseClient
+    # if this code is modified, please update that copy as well
+    # this will be resolved in SDK v4 by making this the only copy
```
kurtmckee, line 109: #same

```suggestion
```

kurtmckee (changes requested):

kurtmckee:
This looks great! I have but one last bit of feedback, but if you push back then I'm going to approve this PR. 👍

src/globus_sdk/services/auth/client/base_login_client.py
sirosen: I see how you're coming at this, and yes, these could be made inaccurate if we release v4 without changing things in this way.

However, I would really like to have some kind of expressed opinion in the code about which one is meant to stay in place and which should be removed, since that's tightly related to my decision to duplicate the code between the two classes (rather than trying to find some abstraction for it). And I think that declaration belongs in the code -- if we only have it in issue tracking, the information is disassociated from the files which really need this added context.

Therefore, I'd like to find a phrasing which is more open and doesn't tie us to v4 specifically, but which still expresses the same idea. Maybe...

```suggestion
    # we would like to restructure code using this method to be calling the matching
    # AuthClient method
    # for example, a future SDK version may make an AuthLoginClient contain
    # an AuthClient which it uses
```

Does that seem more on-target?

kurtmckee: That works for me.

kurtmckee (approved):
I'm approving this, whether the final suggestions get incorporated or not. I don't have a strong opinion about it. 😀

src/globus_sdk/services/auth/client/base_login_client.py
sirosen: I went ahead with this, nearly verbatim. If it mattered enough to comment, then IMO it matters enough to push a fix -- it's trivial to change it like this anyway.

I'll wait for CI and then probably merge on the strength of the existing approval.

kurtmckee (approved):

PR #865, opened 2023-10-06
Fix location for userinfo call; deprecate old name

sirosen:
The `oauth2_userinfo` method was accidentally placed on the `AuthLoginClient` class in #849 rather than the `AuthClient` class.

The mistake was made because it has a name which suggests that it is an OAuth2 login-related method, when that is not the case.

In order to more clearly delineate that this is an API-client method, the `oauth2_` prefix is here stripped and deprecated.
A compatibility shim method is included for now, but it emits a deprecation warning.

The new name is `AuthClient.userinfo`.

Data in `_testing` has also been shimmed temporarily.
This is an arguably unnecessary step, but it is nearly trivial and saves us potential migration and rollout pain at the time of our next release.
Because `_testing` has a weaker compatibility contract, the shim there can be removed after a single release.

----
:books: Documentation preview :books:: https://globus-sdk-python--865.org.readthedocs.build/en/865/


ada-globus (approved):
Looks right—I think the new name is definitely better.

PR #868, opened 2023-10-16
Add support for limit and offset to ls

sirosen:
- Add 'limit' and 'offset' params to 'ls'
- Convert operation_ls tests to newer style
- Add limit+offset to ls param tests

----
:books: Documentation preview :books:: https://globus-sdk-python--868.org.readthedocs.build/en/868/


src/globus_sdk/services/transfer/client.py
```diff
@@ -1243,6 +1250,17 @@ def operation_ls(
         :param query_params: Additional passthrough query parameters
         :type query_params: dict, optional
 
+        .. note::
+
+            Pagination is not supported by the GridFTP protocol, and therefore
+            limit+offset pagination will result in the Transfer service repeatedly
+            fetching an entire directory listing from the server and filtering it before
+            retuning it to the client.
```
kurtmckee, line 1258:
```suggestion
            returning it to the client.
```

src/globus_sdk/services/transfer/client.py
```diff
@@ -1243,6 +1250,17 @@ def operation_ls(
...
+
+            Such usage may still be more efficient than asking for a very large directory
+            listing, for latency-sensitive applications, as it reduces the size of the payload
```
kurtmckee, line 1261:
```suggestion
            listing for latency-sensitive applications, as it reduces the size of the payload
```

src/globus_sdk/services/transfer/client.py
```diff
@@ -1226,6 +1228,11 @@ def operation_ls(
         :param show_hidden: Show hidden files (names beginning in dot).
             Defaults to true.
         :type show_hidden: bool, optional
+        :param limit: Limit the number of results returned. Defaults to 100,000 ,
```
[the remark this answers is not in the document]
sirosen: I wasn't sure if it would impact the rendering, but I thought this is slightly easier to read because it puts some separation between the "literal" and adjacent punctuation. I don't care much, but unless it causes a rendering issue, I'd _tend_ to keep the space.

src/globus_sdk/services/transfer/client.py
kurtmckee: Alternatively, the "for latency-sensitive applications" line could be the opening to the sentence, but that was ugly to rewrite as a suggestion in this interface.

> For latency-sensitive applications, such usage may still ...

sirosen: I was hesitating on this one because it felt awkward once I was looking at it.
I agree with the `For ...`-first rewrite -- it reads best. I'll apply that momentarily.

kurtmckee (approved):

PR #884, opened 2023-10-24
Added support for additional Auth API

JasonAlt:
* AuthClient has support for projects, policies, clients, scopes and user credentials
* NativeAppAuthClient has support for creating native app instances
* ConfidentialAppAuthClient has support for creating child clients

[[sc-28003](https://app.shortcut.com/globus/story/28003/add-additional-support-to-the-sdk-for-auth-endpoints)]

This is ready for feedback. I have left some TODO's throughout to denote some concerns that I think warrant specific attention. Some of these are merely style conflicts (what to name something, where to define it) and some may be due to my lack of understanding of the project (like validating generated docs).

* For client creation, Auth handles the type of client created by the type of authorization used by the caller. With the SDK's recent refactoring of the Auth client, it makes this a little clunky since the only obvious location for some of this was to put it into the AuthLoginClient subclasses. I created:
  * `NativeAppAuthClient.create_native_app_instance()`
  * `ConfidentialAppAuthClient.create_child_client()`
  * `AuthClient.create_client`

  It feels wrong to add service-client type stuff to a login client. It's also a little clunky since two of those functions have to switch authorization types in order to manage the client that was just created. I wonder if anyone has thoughts on how to better organize this.

* Some Auth update endpoints accept None and possibly even [] in addition to None which makes detecting which fields to update a little kludgy. I added a sentinel _DefaultNull which feels a little dirty and at the very least likely needs to be defined elsewhere. If it's of interest, these update calls make use of it:
    * update_policy():
       - domain_constraints_include=[] != domain_constriants_include=None
    * update_client():
       - terms_and_conditions=None
       - privacy_policy=None
       - required_idp=None
       - preselect_idp=None

* Scope creation and updating required a means to specify dependent scope information. I choose to wrap this into a NamtedTuple called DependentScope, but I'm open to suggestions.

* `AuthClient.add_fqdn()` is present, without docstring and without testing because it is currently busted in Auth for token authorization: https://app.shortcut.com/globus/story/27992/register-fqdn-should-support-bearer-tokens. If it is fixed before this PR is merged, I'll finish it.

* My unfamiliarity with the doc building makes it a little tough to proof them myself. I do worry about how the examples turn out. I think at the very least I likely missed the 'optional' param value in docstrings.

----
:books: Documentation preview :books:: https://globus-sdk-python--884.org.readthedocs.build/en/884/

### DIFF
--- a/src/globus_sdk/services/auth/client/confidential_client.py
+++ b/src/globus_sdk/services/auth/client/confidential_client.py
@@ -294,3 +294,163 @@ class ConfidentialAppAuthClient(AuthLoginClient):
         if include is not None:
             body["include"] = include
         return self.post("/v2/oauth2/token/introspect", data=body, encoding="form")
+
+    def create_child_client(
+        self,
+        name: str,
+        *,
+        public_client: bool | utils.MissingType = utils.MISSING,
+        visibility: t.Literal["public", "private"] | utils.MissingType = utils.MISSING,
+        redirect_uris: t.Iterable[str] | utils.MissingType = utils.MISSING,
+        terms_and_conditions: str | utils.MissingType = utils.MISSING,
+        privacy_policy: str | utils.MissingType = utils.MISSING,
+        required_idp: UUIDLike | utils.MissingType = utils.MISSING,
+        preselect_idp: UUIDLike | utils.MissingType = utils.MISSING,
+        client_type: (
+            t.Literal[
+                "client_identity",
+                "confidential_client",
+                "globus_connect_server",
+                "public_installed_client",
+                "hybrid_confidential_client_resource_server",
+                "resource_server",
+            ]
+            | utils.MissingType
+        ) = utils.MISSING,
+        additional_fields: dict[str, t.Any] | utils.MissingType = utils.MISSING,
+    ) -> GlobusHTTPResponse:
+        """
+        Create a new client. Requires the ``manage_projects`` scope.
+
+        :param name: The display name shown to users on consents. May not contain
+            linebreaks.
+        :type name: str
+        :param public_client: This is used to infer which OAuth grant_types the client
+            will be able to use. Should be false if the client is capable of keeping
+            secret credentials (such as clients running on a server) and true if it is
+            not (such as native apps). After creation this value is immutable. This
+            option is mutually exclusive with ``client_type``, exactly one must be
+            given.
+        :type public_client: bool, optional
+        :param visibility: If set to "public", any authenticated entity can view it.
+            When set to "private", only entities in the same project as the client can
+            view it.
+        :type visibility: str, optional
+        :param redirect_uris: list of URIs that may be used in OAuth authorization
+            flows.
+        :type redirect_uris: iterable of str, optional
+        :param terms_and_conditions: URL of client's terms and conditions.
+        :type terms_and_conditions: str, optional
+        :param privacy_policy: URL of client's privacy policy.
+        :type privacy_policy: str, optional
+        :param required_idp: In order to use this client a user must have an identity
+            from this IdP in their identity set.
+        :type required_idp: str or uuid, optional
+        :param preselect_idp: This pre-selects the given IdP on the Globus Auth login
+            page if the user is not already authenticated.
+        :type preselect_idp: str or uuid, optional
+        :param additional_fields: Any additional parameters to be passed through.
+        :type additional_fields: dict, optional
+        :param client_type: Defines the type of client that will be created. This
+            option is mutually exclusive with ``public_client``, exactly one must
+            be given.
+        :type client_type: str, optional
+
+        ``client_type`` must be one of the following values:
+
+        "confidential_client": Applications that are OAuth confidential clients, and can
+        manage a client secret and requests for user consent.
+
+        "public_installed_client" : Applications that are OAuth public clients or native
+        applications that are distributed to users, and thus cannot manage a client
+        secret.
+
+        "client_identity": Applications that authenticate and act as the application
+        itself. These applications are used for automation and as service or community
+        accounts, and do NOT act on behalf of other users. Also known as a "Service
+        Account".
+
+        "resource_server":  A RESTful API (OAuth resource server) that uses Globus Auth
+        tokens for authentication. Users accessing the service login via Globus and
+        consent for the client to use your API.
+
+        "globus_connect_server": Create a client that will service as a Globus Connect
+        Server endpoint.
+
+        "hybrid_confidential_client_resource_server": Register any client with Globus
+        Auth - an application (confidential or public client), service account, or
+        service API.
+
+
+        .. tab-set::
+
+            .. tab-item:: Example Usage
+
+                .. code-block:: pycon
+
+                    >>> ac = globus_sdk.AuthClient(...)
+                    >>> project_id = ...
+                    >>> r = ac.create_child_client(
+                    ...     "My client",
+                    ...     True,
+                    ...     True,
+                    ... )
+                    >>> client_id = r["client"]["id"]
+
+            .. tab-item:: Example Response Data
+
+                .. expandtestfixture:: auth.create_child_client
+
+            .. tab-item:: API Info
+
+                ``POST /v2/api/clients``
+
+                .. extdoclink:: Create Client
+                    :ref: auth/reference/#create_client
+        """
+        body: dict[str, t.Any] = {
+            "name": name,
+            "visibility": visibility,
+        }
+        if not isinstance(redirect_uris, utils.MissingType):
+            body["redirect_uris"] = list(utils.safe_strseq_iter(redirect_uris))
+        if required_idp is not utils.MISSING:
+            body["required_idp"] = str(required_idp)
+        if preselect_idp is not utils.MISSING:
+            body["preselect_idp"] = str(preselect_idp)
+
+        # Must specify exactly one of public_client or client_type
+        if public_client is not utils.MISSING and client_type is not utils.MISSING:
+            raise exc.GlobusSDKUsageError(
+                "AuthClient.create_client does not take both "
+                "'public_client' and 'client_type'. These are mutually exclusive."
+            )
+
+        if public_client is utils.MISSING and client_type is utils.MISSING:
+            raise exc.GlobusSDKUsageError(
+                "AuthClient.create_client requires either 'public_client' or "
+                "'client_type'."
+            )
+        if public_client is not utils.MISSING:
+            body["public_client"] = public_client
+        if client_type is not utils.MISSING:
+            body["client_type"] = client_type
+
+        # terms_and_conditions and privacy_policy must both be set or unset
+        if bool(terms_and_conditions) ^ bool(privacy_policy):
+            raise exc.GlobusSDKUsageError(
+                "terms_and_conditions and privacy_policy must both be set or unset"
+            )
+        links: dict[str, str] = {}
+        if not isinstance(terms_and_conditions, utils.MissingType):
+            links["terms_and_conditions"] = terms_and_conditions
+        if not isinstance(privacy_policy, utils.MissingType):
+            links["privacy_policy"] = privacy_policy
+
+        if links:
+            body["links"] = links
+
+        if not isinstance(additional_fields, utils.MissingType):
+            body.update(additional_fields)
+
+        return self.post("/v2/api/clients", data={"client": body})
--- a/src/globus_sdk/services/auth/client/service_client.py
+++ b/src/globus_sdk/services/auth/client/service_client.py
@@ -13,11 +13,16 @@ from globus_sdk.response import GlobusHTTPResponse, IterableResponse
 from globus_sdk.scopes import AuthScopes
 
 from .._common import get_jwk_data, pem_decode_jwk_data
+from ..data import DependentScopeSpec
 from ..errors import AuthAPIError
 from ..response import (
+    GetClientCredentialsResponse,
+    GetClientsResponse,
     GetIdentitiesResponse,
     GetIdentityProvidersResponse,
+    GetPoliciesResponse,
     GetProjectsResponse,
+    GetScopesResponse,
 )
 
 if sys.version_info >= (3, 8):
@@ -352,13 +357,13 @@ class AuthClient(client.BaseClient):
                     >>> r = ac.get_identity_providers(ids="41143743-f3c8-4d60-bbdb-eeecaba85bd9")
                     >>> r.data
                     {
-                      "identity_providers": [
+                      'identity_providers': [
                         {
-                          "alternative_names": [],
-                          "name": "Globus ID",
-                          "domains": ["globusid.org"],
-                          "id": "41143743-f3c8-4d60-bbdb-eeecaba85bd9",
-                          "short_name": "globusid"
+                          'alternative_names': [],
+                          'name': 'Globus ID',
+                          'domains': ['globusid.org'],
+                          'id': '41143743-f3c8-4d60-bbdb-eeecaba85bd9',
+                          'short_name': 'globusid'
                         }
                       ]
                     }
@@ -424,6 +429,50 @@ class AuthClient(client.BaseClient):
     # Developer APIs
     #
 
+    def get_project(self, project_id: UUIDLike) -> GlobusHTTPResponse:
+        """
+        Look up a project. Requires the ``manage_projects`` scope.
+
+        :param project_id: The ID of the project to lookup
+        :type project_id: str or uuid
+
+        .. tab-set::
+
+            .. tab-item:: Example Usage
+
+                .. code-block:: pycon
+
+                    >>> ac = globus_sdk.AuthClient(...)
+                    >>> r = ac.get_project("927d7238-f917-4eb2-9ace-c523fa9ba34e")
+                    >>> r.data
+                    {
+                      'project': {
+                        'admin_ids': ['41143743-f3c8-4d60-bbdb-eeecaba85bd9']
+                        'contact_email': 'support@globus.org',
+                        'display_name': 'Globus SDK Demo Project',
+                        'admin_group_ids': None,
+                        'id': '927d7238-f917-4eb2-9ace-c523fa9ba34e',
+                        'project_name': 'Globus SDK Demo Project',
+                        'admins': {
+                          'identities': ['41143743-f3c8-4d60-bbdb-eeecaba85bd9'],
+                          'groups': [],
+                        },
+                      }
+                    }
+
+            .. tab-item:: Example Response Data
+
+                .. expandtestfixture:: auth.get_project
+
+            .. tab-item:: API Info
+
+                ``GET /v2/api/projects/{project_id}``
+
+                .. extdoclink:: Get Projects
+                    :ref: auth/reference/#get_projects
+        """
+        return self.get(f"/v2/api/projects/{project_id}")
+
     def get_projects(self) -> IterableResponse:
         """
         Look up projects on which the authenticated user is an admin.
@@ -439,18 +488,19 @@ class AuthClient(client.BaseClient):
                     >>> r = ac.get_projects()
                     >>> r.data
                     {
-                      'projects": [
+                      'projects': [
                         {
-                          'admin_ids": ["41143743-f3c8-4d60-bbdb-eeecaba85bd9"]
-                          'contact_email": "support@globus.org",
-                          'display_name": "Globus SDK Demo Project",
-                          'admin_group_ids": None,
-                          'id": "927d7238-f917-4eb2-9ace-c523fa9ba34e",
-                          'project_name": "Globus SDK Demo Project",
-                          'admins": {
-                            'identities": ["41143743-f3c8-4d60-bbdb-eeecaba85bd9"],
-                            'groups": [],
+                          'admin_ids': ['41143743-f3c8-4d60-bbdb-eeecaba85bd9']
+                          'contact_email': 'support@globus.org',
+                          'display_name': 'Globus SDK Demo Project',
+                          'admin_group_ids': None,
+                          'id': '927d7238-f917-4eb2-9ace-c523fa9ba34e',
+                          'project_name': 'Globus SDK Demo Project',
+                          'admins': {
+                            'identities': ['41143743-f3c8-4d60-bbdb-eeecaba85bd9'],
+                            'groups': [],
                           },
+                        }
                       ]
                     }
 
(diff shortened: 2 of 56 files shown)


JasonAlt:
[sc-28003]

JasonAlt:
Looks like the changelog is the only broken part (as far as the automation goes). I'm not familiar with it, I hope someone else can fill in the gaps.

sirosen:
> Looks like the changelog is the only broken part (as far as the automation goes). I'm not familiar with it, I hope someone else can fill in the gaps.

I've put the magic label on this PR which turns off the changelog check. I'll write up a changelog for this separately so that we don't have to worry about it in the scope of this PR.

---

I'm excited to start digging in here! My initial scan tells me this is almost or exactly as I would want it, which is really promising. I'll make sure to do a full review though, just to play it safe.

We had this work listed as a set of something like 30 tickets (all meant to be bite-sized) hanging out in the backlog for us to chip away at. I can't tell you how jazzed I am to bulk select those and archive them all! :partying_face:

src/globus_sdk/services/auth/client/service_client.py
```diff
@@ -631,3 +680,1187 @@ def delete_project(self, project_id: UUIDLike) -> GlobusHTTPResponse:
...
+        if required_idp is not None:
+            body["required_idp"] = str(required_idp)
+        if preselect_idp is not None:
+            body["preselect_idp"] = str(preselect_idp)
+
+        # terms_and_conditions and privacy_policy must both be set or unset
+        if type(terms_and_conditions) is not type(privacy_policy):
+            raise exc.GlobusSDKUsageError(
+                "terms_and_conditions and privacy policy must both be set or unset"
+            )
```
kurtmckee, line 1225: This requirement isn't documented in the Auth API as far as I can see. Is this correct?

If so, should we note this in a PR against the API docs?

src/globus_sdk/services/auth/client/service_client.py
```diff
@@ -631,3 +680,1187 @@ def delete_project(self, project_id: UUIDLike) -> GlobusHTTPResponse:
...
+                When creating a project, your account is not necessarily included as an
+                admin. The following snippet uses the ``manage_projects`` scope as well
+                as the ``openid`` and ``email`` scopes to get the current user ID and
+                email address and use those data to setup the project.
+
+                .. code-block:: pycon
+
+                    >>> ac = globus_sdk.AuthClient(...)
+                    >>> client_id = ...
+                    >>> r = ac.create_upadte(client_id, name="Foo Utility")
```
kurtmckee, line 1294:
```suggestion
                    >>> r = ac.create_update(client_id, name="Foo Utility")
```

src/globus_sdk/services/auth/client/service_client.py
```diff
@@ -631,3 +680,1187 @@ def delete_project(self, project_id: UUIDLike) -> GlobusHTTPResponse:
...
+            query_params["ids"] = utils.commajoin(ids)
+
+        return GetScopesResponse(self.get("/v2/api/scopes", query_params=query_params))
+
+    # TODO: Should we relocate this? Would we prefer a different way to pass this
+    # information?
+    class DependentScope(t.NamedTuple):
+        scope: UUIDLike
+        optional: bool
+        requires_refresh_token: bool
```
kurtmckee, line 1651: I do think that this needs to exist outside of the class definition; otherwise it will be accessed as a class attribute, like:

```python
dep_scope = AuthClient.DependentScope(...)
```

It may be sufficient to move it to the top of this file and make sure that it's imported all the way up into the `globus_sdk` namespace.

src/globus_sdk/services/auth/client/service_client.py
```diff
@@ -631,3 +680,1187 @@ def delete_project(self, project_id: UUIDLike) -> GlobusHTTPResponse:
...
+    # report for accepting bearer tokens. We could leave this out for now or move it
+    # to confidential_client.py. I'd prefer to leave it out.
+    # https://app.shortcut.com/globus/story/27992/register-fqdn-should-support-bearer-tokens
+    def add_fqdn(
+        self,
+        client_id: UUIDLike,
+        fqdn: str,
+    ) -> GlobusHTTPResponse:
+        # TODO: Add docstring
+        return self.post(f"/v2/api/clients/{client_id}/fqdns", data={"fqdn": fqdn})
```
kurtmckee, line 1866: Let's leave it out for now.

src/globus_sdk/services/auth/client/confidential_client.py
```diff
@@ -294,3 +294,107 @@ def oauth2_token_introspect(
         if include is not None:
             body["include"] = include
         return self.post("/v2/oauth2/token/introspect", data=body, encoding="form")
+
+    # TODO: It is a little strange to place this in this class but Auth treats
+    # client creation different based on how the client was authenticated.
```
kurtmckee, line 299: This doesn't strike me as a TODO. I recommend removing this comment here, and in the Native App code.

JasonAlt:
I am looking forward to hearing opinions on where the `create_child_client()` and `create_native_app_instance()` should go. Their current locations kind of violates 3.29.0's claim:

> AuthClient is now the only class which provides functionality for accessing Globus Auth APIs.

I can think of some variations of AuthClient based on authorizer used but that gets pretty muddy and I didn't want to PR that.

src/globus_sdk/services/auth/client/confidential_client.py
```diff
@@ -294,3 +294,107 @@ def oauth2_token_introspect(
...
+        if required_idp is not None:
+            body["required_idp"] = str(required_idp)
+        if preselect_idp is not None:
+            body["preselect_idp"] = str(preselect_idp)
+
+        # terms_and_conditions and privacy_policy must both be set or unset
+        if type(terms_and_conditions) is not type(privacy_policy):
+            raise exc.GlobusSDKUsageError(
+                "terms_and_conditions and privacy policy must both be set or unset"
+            )
```
kurtmckee, line 387: As noted elsewhere: this requirement isn't documented in the API docs. If it's accurate, we should target updating the Auth API docs to match the Auth behavior, too!

JasonAlt:
FYI:
> I have left some TODO's throughout to denote some concerns that I think warrant specific attention.

The TODO is "to do discussion". All of that will be removed before merge. I just didn't want those items to go under the radar.

sirosen:
> I am looking forward to hearing opinions on where the `create_child_client()` and `create_native_app_instance()` should go. Their current locations kind of violates 3.29.0's claim:
> 
> > AuthClient is now the only class which provides functionality for accessing Globus Auth APIs.
> 
> I can think of some variations of AuthClient based on authorizer used but that gets pretty muddy and I didn't want to PR that.

I took a look at those actually, as part of my initial quick pass. I'm okay with it, and I'm not bothered about the past changelog note.
The point of breaking apart the client class structure was to make each client class able to evolve independently -- the changelog entry is more of an explainer for the conceptual breakdown than a strict rule for us to follow in our work.

The real difference is:
- if it's token-authenticated, it probably goes on `AuthClient`
- if it's client-creds authenticated, it probably goes on `ConfidentialAppAuthClient`

The odd thing / outlier is having methods on `NativeAppAuthClient`, which is inherently unauthenticated. But I think there's no better place for a method, so :+1: to where you put it.

`create_child_client` is "a thing which confidential clients do" and `create_native_app_instance` is "a thing which native apps do", so putting those on the relevant clients is a good structure.

---

There was one other thing I noticed inline that I'll comment on before I have to shift away from this thread (for now).

src/globus_sdk/services/auth/client/service_client.py
```diff
@@ -631,3 +680,1187 @@ def delete_project(self, project_id: UUIDLike) -> GlobusHTTPResponse:
...
+    #   - domain_constraints_include=[] != domain_constriants_include=None
+    # - update_client():
+    #   - terms_and_conditions=None
+    #   - privacy_policy=None
+    #   - required_idp=None
+    #   - preselect_idp=None
+    class _DefaultNull:
+        ...
+
+    _DEFAULT_NULL = _DefaultNull()
```
sirosen, line 865: I actually have a solution I want to put into the SDK for this problem of "null vs missing in python", which we need elsewhere as well. We're going to add a `globus_sdk.MISSING` sentinel value, which can be used just like this sentinel.

I'm going to ask that we move to using that -- since that means that it blocks this PR, I'll reprioritize it for later today.

---

Aside: my plan for globus-sdk v4 includes making all defaults use `MISSING`, never `None`, so that _any_ explicit call with `foo=None` can be interpreted as `"foo": null` properly, anywhere in the SDK.

sirosen:
Oh, I'm just now seeing Kurt's review. As a fast-follow before I context-switch away again:

> 1. Some of the Auth API parameters have been reinterpreted in this code (for example, the conversion from `visibility` as a string to `publicly_visible` as a bool). I haven't checked every API call but I'd like to know if any other API parameters have been reinterpreted from what's documented.

:+1: to this. Any cases in which an input parameter is being translated from one name to another deserves scrutiny. I'd like SDK to match the underlying API as much as possible unless there's a very strong reason for the mismatch.

> 2. Some of the Auth API parameters have requirements placed on them that don't appear to match the Auth API documentation (for example, `terms_of_service` and `privacy_policy` must both be set or unset). If these are known requirements of the Auth API, it may be helpful to update the Auth API documentation to reflect these requirements.

Although I agree, I think even attempting to fix or improve those docs should be firmly out of scope for this work.
We're not responsible for upstream docs here, and attempting to take on that responsibility has bitten me _badly_ in the past.

But! I think we are responsible for informing the Auth team about any doc issues we've found.
I think the ideal would be to write up some notes of any discrepancies or missing doc which you found in the course of this work. At least, those which you can identify; some things may slip through. Whatever works best for that note-taking seems fine to me -- a text file is sufficient.

JasonAlt:
@sirosen 
What do you think about API calls like `DELETE /v2/api/clients/{client_id}` which can be called with a token or with client credentials. Should that be available in ConfidentialAppAuthClient as well? Initially I thought 'no' because it's possible to use AuthClient with the client credentials authorizer. But I can see that being a pain for developers, for example, in CLI's when you delete the client instance in `logout` operations.

- [ ] Actually, we need to verify that token-based access works for that scenario.

JasonAlt:
> > 2. Some of the Auth API parameters have requirements placed on them that don't appear to match the Auth API documentation (for example, `terms_of_service` and `privacy_policy` must both be set or unset). If these are known requirements of the Auth API, it may be helpful to update the Auth API documentation to reflect these requirements.
> 
> Although I agree, I think even attempting to fix or improve those docs should be firmly out of scope for this work. We're not responsible for upstream docs here, and attempting to take on that responsibility has bitten me _badly_ in the past.
> 
> But! I think we are responsible for informing the Auth team about any doc issues we've found. I think the ideal would be to write up some notes of any discrepancies or missing doc which you found in the course of this work. At least, those which you can identify; some things may slip through. Whatever works best for that note-taking seems fine to me -- a text file is sufficient.

I mentioned these issues in slack [here](https://globus.slack.com/archives/C03N0A8J7/p1697930174897869) and [here](https://globus.slack.com/archives/C03N0A8J7/p1697986937197199). I think it's a bit lazy of me to not to create stories or PRs, but I needed to draw the line somewhere.

sirosen:
> @sirosen What do you think about API calls like `DELETE /v2/api/clients/{client_id}` which can be called with a token or with client credentials. Should that be available in ConfidentialAppAuthClient as well? Initially I thought 'no' because it's possible to use AuthClient with the client credentials authorizer. But I can see that being a pain for developers, for example, in CLI's when you delete the client instance in `logout` operations.
> 
>   * [ ]  Actually, we need to verify that token-based access works for that scenario.

Assuming that it works with a token *and is equivalent to using client creds*, my preference would be to only define these on `AuthClient` and to leave out the ConfidentialAppAuthClient variant.

Auth has stated in the past that they'd like to scope down use of client creds to _only_ doing the oauth2 flows and getting tokens, and that all of the various APIs like `DELETE /v2/api/clients/{client_id}` and `GET /v2/api/identities` should at some point stop supporting client credentials. I'm trying to align us with that even though it might slightly restrict the SDK in some places. I think it will lead to better alignment between SDK and docs, and is less likely to leave us with broken methods if Auth ever drops support for these activities.

`ConfidentialAppAuthClient.get_identities` is a slightly special case because it _predated_ the change in client class hierarchy, so we had to define it to remain backwards compatible.

If the token-auth'ed and the client-creds-auth'ed versions of an API are different and documented, however, I think we need to consider adding methods to both classes, but we should check with the Auth team on any of these.

> I asked about this [...] but no follow up yet. I wonder what other API endpoints might work with client credentials.

I wouldn't block on hearing back, but we should maybe follow-up next week.
If there are APIs which are known to need both versions, that would be worth capturing.

> Maybe move these types of functions to a mixin?

Let's not try to solve the code sharing problem in this current PR.
I'm still toying with how to handle these cross-client-type methods in my head.
The first version of the client inheritance refactor used a mixin, but it didn't work out well in that case so we went with the current version (no mixins, some code duplication) instead.

My current idea is that in SDK v4, we'll change things like so:
```python
class ConfidentialApp:
    def __init__(self, client_id, client_secret, ...):
        self.client = AuthClient(authorizer=BasicAuthorizer(client_id, client_secret))
```
And then you'll be able to do
```python
app = ConfidentialApp(...)
app.client.delete_client(child_client_id)
```

It means that `ConfidentialAppAuthClient` and `NativeAppAuthClient` "contain a client" rather than "being a client", which is interesting and might make this all easier to handle.

But I don't know if it will really work. I need to have some more concrete examples to look at.

> I think it's a bit lazy of me to not to create stories or PRs, but I needed to draw the line somewhere.

No, I don't think it's lazy. IMO, this is responsible behavior.
You've informed the upstream team about the doc issues -- that's where our responsibility (insofar as we're working on / maintaining the SDK) ends.

src/globus_sdk/services/auth/client/service_client.py
sirosen: I agree -- let's move it out to be separate from the class. Does losing the namespacing from the class require a rename for clarity? `DependentScopeSpec`? (We should discuss the name in any case.)

I think my preference would be to put it into a new module at `src/globus_sdk/services/auth/data.py` as a dedicated class. That's the (informal) convention we have for locating our payload-building helpers.

I would also adjust it to use `globus_sdk.utils.PayloadWrapper` as its base (aside: that class name is _slightly_ amiss at this point, but I don't have a clear plan to rename it yet), which is a dict-like (UserDict). e.g.
```python
from globus_sdk import utils
from globus_sdk._types import UUIDLike

class DependentTokenSpec(utils.PayloadWrapper):
    def __init__(self, *, scope: UUIDLike, optional: bool, requires_refresh_token: bool) -> None:
        super().__init__()
        self["scope"] = str(scope)
        self["optional"] = optional
        self["requires_refresh_token"] = requires_refresh_token
```

The advantage is that with some recent changes in `main` (the same ones which handle `MISSING`), you can then include it in a payload and we'll serialize it before sending. e.g.
```python
self.post(
    "/foo",
    data={"bar": DependentTokenSpec(scope="baz", optional=False, requires_refresh_token=False)}
)
# sends a JSON payload of
# {"bar": {"scope": "baz", "optional": false, "requires_refresh_token": false}}
```

src/globus_sdk/services/auth/client/service_client.py
```diff
@@ -631,3 +680,1187 @@ def delete_project(self, project_id: UUIDLike) -> GlobusHTTPResponse:
...
+                        'id": "6336437e-37e8-4559-82a8-674390c1fd2e",
+                        'prompt_for_named_grant": False,
+                        'fqdns": ["globus.org"],
+                        'project": "da84e531-1afb-43cb-8c87-135ab580516a",
+                        'client_type": "client_identity",
+                        'visibility": "private",
+                        'parent_client": None,
+                        'userinfo_from_effective_identity": True,
+                        'preselect_idp": None,
+                        'public_client": False
```
sirosen, line 1032: Just spot-checking, I see this one has `'public_client": False` -- note the mismatched quote chars. Other keys fit that pattern too.
I'm guessing that this came from some python with double-quotes, and then it went through `s/"/'/` when you meant `s/"/'/g`. "Classic" sed typo. :smile: 

Anyway, I'm calling this one out for early feedback, but we should double-check the others.

src/globus_sdk/services/auth/client/confidential_client.py
```diff
@@ -294,3 +294,107 @@ def oauth2_token_introspect(
...
+        name: str,
+        public_client: bool,
+        publicly_visible: bool,
+        *,
+        redirect_uris: t.Iterable[str] | None = None,
+        terms_and_conditions: str | None = None,
+        privacy_policy: str | None = None,
+        required_idp: UUIDLike | None = None,
+        preselect_idp: UUIDLike | None = None,
+        create_params: dict[str, t.Any] | None = None,
```
sirosen, line 312: I'm trying to be consistent in new methods in calling any "add-in" dict like this `additional_fields` -- let's follow that in this case.

If you `git grep 'additional_fields'`, you'll see it's all over helpers and methods for various services.
We have some past inconsistency about this, calling it things like `additional_params`, and there are some other one-off names. But my goal is that you can learn `additional_fields` as a name for this when using the SDK and have it always mean the same thing service-to-service.

src/globus_sdk/services/auth/client/service_client.py
```diff
@@ -631,3 +680,1187 @@ def delete_project(self, project_id: UUIDLike) -> GlobusHTTPResponse:
...
+        body: dict[str, t.Any] = {}
+
+        if project_id is not None:
+            body["project_id"] = str(project_id)
+        if authentication_assurance_timeout is not None:
+            body["authentication_assurance_timeout"] = authentication_assurance_timeout
+        if display_name is not None:
+            body["display_name"] = display_name
+        if description is not None:
+            body["description"] = description
```
sirosen, line 935: This is a good spot to call out -- there's a new way for you to use `MISSING` once you pick it up from `main`, and not just for the fields below with `_DEFAULT_NULL`. You can use it on these default-`None` fields too!

```python
from globus_sdk.utils import MissingType, MISSING

...

    def update_policy(
        ...
        authentication_assurance_timeout: int | MissingType = MISSING,
        display_name: str | MissingType = MISSING,
        description: str | MissingType = MISSING,
        ...
    ) -> GlobusHTTPResponse:
        ...
        body["authentication_assurance_timeout"] = authentication_assurance_timeout
        body["display_name"] = display_name
        body["description"] = description
```

_drops the mic_

I think this is easier on all of us to write and read, and the `MISSING` values will get auto-stripped out of the request.

---

I haven't decided if I think it's safe for us to go back and make a lot of `None`-type things default to `MISSING` instead. I think it might be a mild enough compatibility break in a lot of cases that it will be worth it, but maybe it's better to save that all for v4.

src/globus_sdk/services/auth/client/service_client.py
JasonAlt: 👍 I actually saw this in `get_projects()` and assumed that the ' followed by " was required by some doc generator. I'll change this throughout the file.

sirosen: Oh, so it's _my_ typo! Sorry about that. Yeah, we should fix them all, but you only have to worry about your new ones -- I'll circle back for anything pre-existing.

JasonAlt:
Sorry for the delay. I believe all the changes are complete. My main concerns now are with the documentation. I think some of the function signatures are too busy and thus not readable, and I can't see how to included any description for now to build a `DependentScopeSpec`.

sirosen:
> I think some of the function signatures are too busy and thus not readable, and I can't see how to included any description for now to build a `DependentScopeSpec`.

Some of that is probably related a known issue with the way that sphinx renders signatures with annotations + the way that the furo theme lays them out. I really wish that we could get the `black`-like styling in our doc site, which would make the signature lines way more readable, but I've been told by the furo creator that it's difficult with the available HTML coming out of sphinx.

As for documenting `DependentScopeSpec`, I think you'll want to add another directive to the [RST file which defines the Auth SDK docs](https://github.com/globus/globus-sdk-python/blob/main/docs/services/auth.rst). I'm pretty sure another autoclass in the helpers section (next to the `IdentityMap` one) would be all we need.

sirosen:
Sorry about the close-open noise! I fat-fingered a Tab+Enter combo -- first time that's ever happened to me as far as I can recall - which immediately closed this.
All I was going to add to the above is that I'm about to start a, hopefully final, review pass now.

src/globus_sdk/services/auth/client/confidential_client.py
```diff
@@ -294,3 +294,106 @@ def oauth2_token_introspect(
...
+        :type name: str
+        :param public_client: This is used to infer which OAuth grant_types the client
+            will be able to use. Should be false if the client is capable of keeping
+            secret credentials (such as clients running on a server) and true if it is
+            not (such as native apps). After creation this value is immutable.
+        :type public_client: bool
+        :param visibility: If set to "public", any authenticated entity can view it.
+            When set to "private", only entities in the same project as the client can
+            view it.
+        :type visibility: str
```
sirosen, line 325: nit:
```suggestion
        :type visibility: str, optional
```

src/globus_sdk/services/auth/client/confidential_client.py
```diff
@@ -294,3 +294,106 @@ def oauth2_token_introspect(
...
+            body["visibility"] = visibility
+        if redirect_uris is not None:
+            body["redirect_uris"] = list(utils.safe_strseq_iter(redirect_uris))
+        if required_idp is not None:
+            body["required_idp"] = str(required_idp)
+        if preselect_idp is not None:
+            body["preselect_idp"] = str(preselect_idp)
+
+        # terms_and_conditions and privacy_policy must both be set or unset
+        if type(terms_and_conditions) is not type(privacy_policy):
```
sirosen, line 383: Very minor, but I wonder if we consider this cleaner if, rather than `type(x) is not type(y)`, we use `bool` + X-or?
```
bool(x) ^ bool(y)
```

It's pretty similar, so I can be convinced, but I thought I should mention it.

src/globus_sdk/services/auth/client/service_client.py
```diff
@@ -631,3 +681,1111 @@ def delete_project(self, project_id: UUIDLike) -> GlobusHTTPResponse:
...
+        body: dict[str, t.Any] = {
+            "authentication_assurance_timeout": authentication_assurance_timeout,
+            "display_name": display_name,
+            "description": description,
+            "domain_constraints_include": domain_constraints_include,
+            "domain_constraints_exclude": domain_constraints_exclude,
+        }
+
+        if not isinstance(project_id, utils.MissingType):
+            body["project_id"] = str(project_id)
```
sirosen, line 923: I see what's going on here, with `UUID | str | MissingType` -- and of course you need to `str()` it if it's a `UUID`.

:+1: to this as a way of handling it for now.
I need to think more about what we can do. Maybe the SDK should just support stringifying `UUID`s when we prepare a payload and strip out the `MISSING` values. It's a really common type for us... Your input is welcome, but no change needed here!

src/globus_sdk/services/auth/client/service_client.py
```diff
@@ -631,3 +681,1111 @@ def delete_project(self, project_id: UUIDLike) -> GlobusHTTPResponse:
...
+        """  # noqa: E501
+        return GetClientsResponse(self.get("/v2/api/clients"))
+
+    def create_client(
+        self,
+        name: str,
+        public_client: bool,
+        project: UUIDLike,
+        visibility: t.Literal["public", "private"] | utils.MissingType = utils.MISSING,
+        *,
```
sirosen, line 1114: Similar to the other client class, I think this is better:
```suggestion
        *,
        visibility: t.Literal["public", "private"] | utils.MissingType = utils.MISSING,
```
but let me know if there's a reason to keep it as a positional-or-keyword arg.

src/globus_sdk/services/auth/client/service_client.py
```diff
@@ -631,3 +681,1111 @@ def delete_project(self, project_id: UUIDLike) -> GlobusHTTPResponse:
...
+            will be able to use. Should be false if the client is capable of keeping
+            secret credentials (such as clients running on a server) and true if it is
+            not (such as native apps). After creation this value is immutable.
+        :type public_client: bool
+        :param project: ID representing the project this client belongs to.
+        :type project: str or uuid
+        :param visibility: If set to "public", any authenticated entity can view it.
+            When set to "private", only entities in the same project as the client can
+            view it.
+        :type visibility: str
```
sirosen, line 1138: same nit as above:
```suggestion
        :type visibility: str, optional
```

src/globus_sdk/services/auth/client/service_client.py
```diff
@@ -631,3 +681,1111 @@ def delete_project(self, project_id: UUIDLike) -> GlobusHTTPResponse:
...
+                "terms_and_conditions and privacy_policy must both be set or unset"
+            )
+        links: dict[str, str] = {}
+        if not isinstance(terms_and_conditions, utils.MissingType):
+            links["terms_and_conditions"] = terms_and_conditions
+        if not isinstance(privacy_policy, utils.MissingType):
+            links["privacy_policy"] = privacy_policy
+
+        if links:
+            body["links"] = links
```
sirosen, line 1208: Okay, this is another one for me to chew on.
As above, no need to change and :+1: for this way of doing it right now.

But maybe a good future would be... this?
```python
links = {
    "terms_and_conditions": terms_and_conditions,
    "privacy_policy": privacy_policy,
}
if utils.MissingType.any_values(links):
    body["links"] = links
```
or maybe even it's
```python
body["links"] = utils.MissingType.subdict(
    {
        "terms_and_conditions": terms_and_conditions,
        "privacy_policy": privacy_policy,
    }
)
```
which returns `MISSING` if no values are non-`MISSING`?

Lots to think about here; and, again, as someone who just implemented this stuff I'd love your feedback or ideas.

src/globus_sdk/services/auth/client/service_client.py
```diff
@@ -631,3 +681,1111 @@ def delete_project(self, project_id: UUIDLike) -> GlobusHTTPResponse:
...
+        body: dict[str, t.Any] = {
+            "name": name,
+            "visibility": visibility,
+            "redirect_uris": redirect_uris,
+        }
+
+        if required_idp is not utils.MISSING:
+            body["required_idp"] = str(required_idp)
+        if preselect_idp is not utils.MISSING:
+            body["preselect_idp"] = str(preselect_idp)
```
sirosen, line 1292: I'm losing count but I think we're now at ~6 places where auto-stringifying UUIDs would make this smoother. And I'm sure it's only going to go up. So I'm definitely going to give that a look after this PR.

sirosen (commented):
There are a couple of items to discuss and one or two minor tweaks which I think we'll make, but I think this is trending towards an approval and merge very soon!

There are a number of cases in here in which the `MISSING` value doesn't help smooth things out quite as much as I expected it to.
I still think the changes to use it are good because it makes that the external interface for these methods -- and we just need to do more plumbing work to make it all fit together seamlessly. I think some plan of attack for subdicts of the form `body={"foo": {"bar": MISSING}}` is needed, and a minor enhancement to our UUID handling would go a long way. But I don't want to block this PR on that stuff, especially when it's all in my head and we haven't discussed it at all yet. I may open one or two PRs in parallel though.

src/globus_sdk/services/auth/client/confidential_client.py
```diff
@@ -294,3 +294,106 @@ def oauth2_token_introspect(
         if include is not None:
             body["include"] = include
         return self.post("/v2/oauth2/token/introspect", data=body, encoding="form")
+
+    def create_child_client(
+        self,
+        name: str,
+        public_client: bool,
+        visibility: t.Literal["public", "private"] | None = None,
+        *,
```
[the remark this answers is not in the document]
JasonAlt: fixed in upcoming commit

src/globus_sdk/services/auth/client/service_client.py
JasonAlt: fixed in upcoming commit

src/globus_sdk/services/auth/client/confidential_client.py
```diff
@@ -294,3 +294,106 @@ def oauth2_token_introspect(
...
+        redirect_uris: t.Iterable[str] | None = None,
+        terms_and_conditions: str | None = None,
+        privacy_policy: str | None = None,
+        required_idp: UUIDLike | None = None,
+        preselect_idp: UUIDLike | None = None,
```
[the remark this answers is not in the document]
JasonAlt: fixed in upcoming commit

src/globus_sdk/services/auth/client/confidential_client.py
JasonAlt: fixed in upcoming commit

src/globus_sdk/services/auth/client/confidential_client.py
JasonAlt: Using `bool` is fine. I think this was a holdover from `update_client()` where using `terms_and_conditions=None, privacy_policy=MISSING` would have passed `bool` but failed in Auth.

FWIW, I'm not sure if the SDK should check these anyways (1. Auth could remove the requirement, 2. SDK is supposed to be a thin layer). But I saw similar minor checks like this in other calls so I added it. I think those mostly dealt with query params though.

JasonAlt: fixed in upcoming commit

src/globus_sdk/services/auth/client/service_client.py
JasonAlt: fixed in upcoming commit

src/globus_sdk/services/auth/client/service_client.py
JasonAlt: It's a little goofy since you may come across APIs where you want the field to be included but empty: `{"links": {}}` and others where it shouldn't `{"links": MISSING}` (or others where it doesn't matter).

JasonAlt:
I believe all requested changes have been committed, rebased and pushed. A couple items:

1. I'm going to make use of the PR's doc preview to get `DependentScopeSpec` that way I'd like.  I don't see a good way to view it within my dev environment (windows/wsl2/docker/fedora 38). All my work is in a docker volume so I can't exactly open index.html locally with my browser. I do wish where was a `view` command like docs.globus.org.

2. There's some funniness with `AuthClient.create_client()` and `ConfidentialAppAuthClient.create_client()` that needs to be figured out before merge. There's an undocumented field `client_type`. I had asked Auth if it should remain hidden but with no response I decided to just let `additional_fields` handle it for now. But as it turns out, `POST /v2/api/clients` requires either `public_client` or `client_type` which could lead to `public_client` becoming an optional value. Making that change later may be a breaking change.

   FWIW, the values for `client_type` are currently: 

      confidential_client 
      public_installed_client
      client_identity
      resource_server
      globus_connect_server
      hybrid_confidential_client_resource_server
      public_webapp_client

  We could either:
   1. leave it as is and let it break later
   2. include `client_type`, make `public_client` optional and add logic to check that exactly one is set
   3. omit `client_type` and make `public_client=True` a keyword arg so that introducing `client_type` later may not break existing code

src/globus_sdk/services/auth/client/service_client.py
sirosen: FYI, we _just_ merged a change that makes SDK auto-`str()` any UUIDs.
I think these cases can change now to
```python
body["project_id"] = project_id
```

IMO any updates to this style are completely optional though. The current code works either way, which is the main thing.

---

As for `is MISSING`, I wish there were a nicer way to handle that for mypy. (`Literal` is wonky with custom types and doesn't work...)
I still haven't figured out what yet. Maybe we'll take a leaf out of sqlalchemy's book for naming things which look similar to language keywords, like this?
```python
if MISSING.is_(foo): ...
```
But I'm far from decided yet.

src/globus_sdk/services/auth/client/confidential_client.py
sirosen: Regarding safety checking in the client:
I'm fine with checks like these. I don't want us to do anything elaborate or expensive client-side, but if it's easy to add a guard like this against a "trap" where it looks like you can do something with an API which we know is invalid, I'm in favor.

We can always remove these guards _if_ there's a problem/bug with them which we can't trivially fix, because we aren't declaring it as part of the API surface -- function signatures are much tougher to modify later.

sirosen:
I've kickstarted a fresh discussion to try to get clarity around the `client_type` funkiness, and it seems that the doc omission is accidental and we're safe to include `client_type`.
I also confirmed that we're okay to advertise the full list of valid values -- a couple looked odd to me.

The web app is using the following mapping from these values to "nice names":
```
    Confidential Client Portal - confidential_client
    Command line / Thick Client / SPA - public_installed_client
    Service Account / credential only client - client_identity
    Resource Server / API - resource_server
    GCS - globus_connect_server
    Advanced - hybrid_confidential_client_resource_server
```

This might be nice to include in the docs in the future, but for now I'm posting it here so we don't lose it.

JasonAlt:
Added the updates, tests, docstrings for `create_client()` and `create_child_client()`. Just need `DependentScopeSpec`!

JasonAlt:
Done. 🤞

sirosen (approved):

PR #885, opened 2023-10-25
Introduce the MISSING sentinel and use it for HA timeout on GroupPolicies

sirosen:
The commits here reflect a few quick iterative steps I took to get `globus_sdk.MISSING` into place and use it.

- [Introduce the MISSING singleton](https://github.com/globus/globus-sdk-python/commit/de90985f668ca59213f2c01079cd190b20882f75): add the singleton and some unit tests, but no usage
- [Default to MISSING for GroupPolicies ha timeout](https://github.com/globus/globus-sdk-python/commit/53009d915f88043a5c6fcbe92ff35e7888598719): add a first-draft implementation in which exactly one class uses `MISSING` and is internally responsible for stripping it out before the result is passed to other SDK code
- [Automatically strip MISSING from payloads](https://github.com/globus/globus-sdk-python/commit/b1e9f58c16a666f5310c2442dc879cb2124f0a18): Refine the implementation to allow `MISSING` to be passed around, and strip it out when the payload is actually being prepared for use in the transport layer

This final change makes preparation of payload data the responsibility of the `PayloadWrapper` helper class, and also makes it responsible for recursively transforming any `PayloadWrapper`s to native dicts. Although I was initially concerned that this would introduce a noticeable slowdown, some quick benchmarking revealed this to be very fast on typical data sizes.

We could also push the handling of `MISSING` "down the stack" into the transport layer, where it would make sense. But for now, it's the client class' responsibility to prune out `MISSING`.

---

In the final state, we should now be able to write client methods like so:
```python
def create_foo(self, x: int | None | MissingType = MISSING):
    body = {}
    body["x"] = x
    return self.post("/foo", body)
```

Notably, there is not (yet) any support for use of `MISSING` in query params or other locations. That is intentionally omitted as future work, although it should presumably fit comfortably into this same paradigm.

----
:books: Documentation preview :books:: https://globus-sdk-python--885.org.readthedocs.build/en/885/


src/globus_sdk/services/groups/data.py
```diff
@@ -263,7 +263,9 @@ def __init__(
         group_members_visibility: _GROUP_MEMBER_VISIBILITY_T,
         join_requests: bool,
         signup_fields: t.Iterable[_GROUP_REQUIRED_SIGNUP_FIELDS_T],
-        authentication_assurance_timeout: int | None = None,
+        authentication_assurance_timeout: int
+        | None
+        | utils.MissingType = utils.MISSING,
```
kurtmckee, line 268: Oof. This style runs afoul of my similar preference for implicit string concatenation: in parentheses. I'm surprised that black doesn't modify this already, but I may have to adjust to this style.

src/globus_sdk/utils.py
```diff
@@ -22,6 +22,41 @@
...
+
+
+# a sentinel value for "missing" values which are distinguished from `None` (null)
+# this is the default used to indicate that a parameter was not passed, so that
+# method calls passing `None` can be distinguished from those which did not pass any
+# value
+# users should typically not use this value directly, but it is part of the public SDK
+# interfaces along with its type for annotation purposes
+#
+# *new after v3.29.0* (older methods do not use MISSING)
```
kurtmckee, line 56:
```suggestion
# *new after v3.29.0* (older versions do not use MISSING)
```

kurtmckee (approved):
I've left feedback but this is great as-is.

src/globus_sdk/services/groups/data.py
sirosen: It looks like `black` leaves this in the "user specified style".
I'll change this one -- and BTW I have a much stronger feeling that parenthesized is better here than with the strings.

I'll look into adding a `slyp` check for this later.

src/globus_sdk/utils.py
sirosen: I actually meant to indicate that methods which were introduced prior to MISSING don't use it (as opposed to those introduced after it). But on consideration, I think the whole parenthetical is just distracting, so I'm striking it.

kurtmckee (approved):

sirosen:
There's one additional change here, which is to apply "filter MISSING" behavior to query params and headers, along with a test.
It's quite small (last commit right now), but I think it warrants a refresh of the "review clock". :slightly_smiling_face:

kurtmckee (approved):

PR #889, opened 2023-10-27
Release v3.30.0

sirosen:
Standard release updates, e.g. changelog.
Also make a minor tweak to the "new in version" comment on the `MISSING` sentinel.

---

# Changelog

## Added

- `TransferClient.operation_ls` now supports the `limit` and `offset`
  parameters (#868)

- A new sentinel value, `globus_sdk.MISSING`, has been introduced.
  It is used for method calls which need to distinguish missing parameters from
  an explicit `None` used to signify `null` (#885)

  - `globus_sdk.MISSING` is now supported in payload data for all methods, and
    will be automatically removed from the payload before sending to the server

## Changed

- `GroupPolicies` objects now treat an explicit instantiation with
  `high_assurance_timeout=None` as setting the timeout to `null` (#885)

----
:books: Documentation preview :books:: https://globus-sdk-python--889.org.readthedocs.build/en/889/

### DIFF
--- a/changelog.d/20231016_114115_sirosen_add_ls_params.rst
+++ b/changelog.d/20231016_114115_sirosen_add_ls_params.rst
@@ -1,5 +1,0 @@
-Added
-~~~~~
-
-- ``TransferClient.operation_ls`` now supports the ``limit`` and ``offset``
-  parameters (:pr:`868`)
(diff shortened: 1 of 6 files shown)


ada-globus (approved):
👍

PR #894, opened 2023-11-03
Improve the way that enum members get stringified

sirosen:
The GroupsClient and related helpers have a number of enums of strings which are accepted as inputs. Previously, rendering these enums to their `.value` attributes (their values) was done via explicit calls in all of the locations where they are used.

Under this change, the serialization of any `Enum` member as `x.value` is codified as one of the characteristics of the primitive-value processing done in the transport layer's encoders, alongside UUID stringification.

The end result is a minor but useful simplification to any code which handles these enums, as the enum value can be passed to the underlying transport without a prior transformation. A minor helper function is eliminated (in favor of the addition to the encoders).

One test, which evaluated the correct stringification of an enum, has been updated to use `responses` to check the sent request, rather than a payload object prior to sending.

----
:books: Documentation preview :books:: https://globus-sdk-python--894.org.readthedocs.build/en/894/

### DIFF
--- a/src/globus_sdk/services/groups/data.py
+++ b/src/globus_sdk/services/groups/data.py
@@ -112,7 +112,7 @@ class BatchMembershipActions(utils.PayloadWrapper):
         :type role: str or :class:`~.GroupRole`
         """
         self.setdefault("add", []).extend(
-            {"identity_id": identity_id, "role": utils.render_enums_for_api(role)}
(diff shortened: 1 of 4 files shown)


kurtmckee (approved):
Nice! This lines up neatly with the changes to handle UUIDs. 🚀

PR #895, opened 2023-11-03
TransferClient relies on UUID encoding transport

sirosen:
Instead of explicitly stringifying values in TransferClient, allow UUIDs to be passed to the transport in all cases.

Additionally, make use of `utils.commajoin` as a replacement for custom and inlined behaviors which are identical in meaning.

The goal of this change is to be a small net-negative diff -- not only in lines, but also in some cases in complexity on the same number of lines.

----
:books: Documentation preview :books:: https://globus-sdk-python--895.org.readthedocs.build/en/895/

### DIFF
--- a/src/globus_sdk/services/transfer/client.py
+++ b/src/globus_sdk/services/transfer/client.py
@@ -3,9 +3,8 @@ from __future__ import annotations
 import logging
 import time
 import typing as t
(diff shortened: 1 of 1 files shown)


kurtmckee (approved):
🪨 🌟

PR #897, opened 2023-11-03
Service auth client trusts more to the transport

sirosen:
This is an analogous change to #896, making the service client rely more heavily on the fact that the transport will strip MISSING and translate UUIDs.

There are a number of cases here which are related to but not covered by the current toolchain for handling MISSING. In particular, checks for an "empty" subdocument (see `links` in client create/update) and handling of "commajoin" on a list which may be MISSING are both places where utility improvements may allow even further simplifications to the client code. However, given that there is no change possible today, no further adjustment is made to those parts of the client behavior.

----
:books: Documentation preview :books:: https://globus-sdk-python--897.org.readthedocs.build/en/897/

### DIFF
--- a/src/globus_sdk/services/auth/client/service_client.py
+++ b/src/globus_sdk/services/auth/client/service_client.py
@@ -842,7 +842,7 @@ class AuthClient(client.BaseClient):
                     :ref: auth/reference/#create_policy
         """
         body: dict[str, t.Any] = {
-            "project_id": str(project_id),
+            "project_id": project_id,
             "high_assurance": high_assurance,
             "authentication_assurance_timeout": authentication_assurance_timeout,
             "display_name": display_name,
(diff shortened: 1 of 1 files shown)


src/globus_sdk/services/auth/client/service_client.py
```diff
@@ -1222,47 +1219,38 @@ def create_client(
...
                 "'client_type'."
             )
-        if public_client is not utils.MISSING:
-            body["public_client"] = public_client
-        if client_type is not utils.MISSING:
-            body["client_type"] = client_type
 
+        body: dict[str, t.Any] = {
+            "name": name,
+            "project": project,
```
kurtmckee, line 1236: Hang on. Why is this "project" when several other APIs here use "project_id"? Is this correct?

kurtmckee (approved):
Just noticed that the `create_client` API uses `project` instead of `project_id` (which is what, say, the policy APIs use.

I double-checked that that's correct, so we're in the clear.

sirosen:
Yeah, I noticed that as well -- but it's a real API quirk, so it's our job to dutifully reflect it. Making these changes happened to be a good chance for me to go back over everything to make sure I hadn't missed anything in all of the excitement.

PR #898, opened 2023-11-03
Enhance docs for 'client_type'

sirosen:
1. Reorder the keyword-only parameters for client creation to put `client_type` adjacent to `public_client`

This has no semantic impact. It's done only to make the parameter order in code match the order in doc.

2. Reorder parameter docs to make sure `client_type` is adjacent to `public_client`

These are mutex and one is required, which makes me want them near each other and at the top of the parameter list.

3. Convert doc of `client_type` values to a list-table under a dropdown

I find this to be a good balance of information, vertical density for scrolling, and easy authorship. `list-table` is pretty simple and requires minimal clever ASCII-art arrangements (vs many other table systems for markup languages).

---

Doc preview:
expanded | collapsed
--- | ---
image | image

----
:books: Documentation preview :books:: https://globus-sdk-python--898.org.readthedocs.build/en/898/

### DIFF
--- a/src/globus_sdk/services/auth/client/confidential_client.py
+++ b/src/globus_sdk/services/auth/client/confidential_client.py
@@ -300,12 +300,6 @@ class ConfidentialAppAuthClient(AuthLoginClient):
         name: str,
         *,
         public_client: bool | utils.MissingType = utils.MISSING,
-        visibility: t.Literal["public", "private"] | utils.MissingType = utils.MISSING,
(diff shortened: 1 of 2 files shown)


sirosen:
cc @JasonAlt, just as a small FYI, I promised to share some nice table styling for these param docs -- this is what I had in mind.
There might be even _fancier_ things available, but this is the best thing I know how to do today.

ada-globus (approved):

PR #903, opened 2023-11-15
Support custom CA certificate bundles

kurtmckee:
Added
-----

* Support custom CA certificate bundles.

  Previously, SSL/TLS verification allowed only a boolean ``True`` or ``False`` value.
  It is now possible to specify a CA certificate bundle file
  using the existing ``verify_ssl`` parameter
  or ``GLOBUS_SDK_VERIFY_SSL`` environment variable.

  This may be useful for interacting with Globus through certain proxy firewalls.

----
:books: Documentation preview :books:: https://globus-sdk-python--903.org.readthedocs.build/en/903/

### DIFF
--- a/src/globus_sdk/config/env_vars.py
+++ b/src/globus_sdk/config/env_vars.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 
 import logging
 import os
+import pathlib
 import typing as t
 
 log = logging.getLogger(__name__)
@@ -17,16 +18,6 @@ T = t.TypeVar("T")
 ENVNAME_VAR = "GLOBUS_SDK_ENVIRONMENT"
 HTTP_TIMEOUT_VAR = "GLOBUS_SDK_HTTP_TIMEOUT"
 SSL_VERIFY_VAR = "GLOBUS_SDK_VERIFY_SSL"
-
-
-def _str2bool(val: str) -> bool:
-    val = val.lower()
-    if val in ("y", "yes", "t", "true", "on", "1"):
-        return True
-    elif val in ("n", "no", "f", "false", "off", "0"):
-        return False
-    else:
-        raise ValueError(f"invalid truth value: {val}")
 
 
 @t.overload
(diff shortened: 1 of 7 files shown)


src/globus_sdk/config/env_vars.py
```diff
@@ -70,12 +61,28 @@ def _load_var(
...
+        if value.lower() in {"n", "no", "f", "false", "off", "0"}:
+            return False
+        if os.path.isfile(value):
+            return value
+    if isinstance(value, pathlib.Path):
+        if value.is_file():
+            return str(value.absolute())
+    raise ValueError(
+        f"SSL verification {value} is not a valid boolean value "
+        "nor a path to a file that exists"
```
sirosen, line 84: Nit: not -> or, or neither -> nor; but not -> nor looks awkward to me. (I'm not certain about this being a correctness thing, but it reads in an unusual way.)

```suggestion
        "or a path to a file that exists"
```

src/globus_sdk/transport/requests.py
```diff
@@ -162,8 +163,9 @@ def tune(
         In particular, this can be used to temporarily adjust request-sending minutiae
         like the ``http_timeout`` used.
 
-        :param verify_ssl: Explicitly enable or disable SSL verification
-        :type verify_ssl: bool, optional
+        :param verify_ssl: Explicitly enable or disable SSL verification,
+            or configure the path to a CA certificate bundle to use for SSL verification
+        :type verify_ssl: bool, str, or pathlib.Path, optional
```
sirosen, line 168: This same doc update needs to be reflected in the `__init__` parameter docs in the class docstring.

src/globus_sdk/config/env_vars.py
```diff
@@ -70,12 +61,28 @@ def _load_var(
...
+    if isinstance(value, str):
+        if value.lower() in {"y", "yes", "t", "true", "on", "1"}:
+            return True
+        if value.lower() in {"n", "no", "f", "false", "off", "0"}:
+            return False
+        if os.path.isfile(value):
+            return value
+    if isinstance(value, pathlib.Path):
+        if value.is_file():
+            return str(value.absolute())
```
sirosen, line 81: :+1: or :-1: on this adjustment?
```suggestion
    if isinstance(value, pathlib.Path) and value.is_file():
        return str(value.absolute())
```

sirosen (changes requested):
Some very minor changes needed, but this looks good!

This has been an outstanding issue for a little while but I, for some reason, thought it would be more complex to implement support.

src/globus_sdk/config/env_vars.py
kurtmckee: "Not" and "nor" are correct, but I can rephrase to something like

> SSL verification value must be a boolean value or a path to a file that exists

src/globus_sdk/transport/requests.py
kurtmckee: Oh! It's in the class docstring! It seemed weird to me that `.__init__` was completely undocumented. 😆

src/globus_sdk/config/env_vars.py
kurtmckee: New rev coming

src/globus_sdk/transport/requests.py
kurtmckee: New rev coming

src/globus_sdk/config/env_vars.py
kurtmckee: New rev.

sirosen (approved):

PR #907, opened 2023-11-28
Remove the built-in ``/v2`` API route prefix from `GroupsClient`

kurtmckee:
Changed
-------

-   Remove the built-in ``/v2`` API route prefix from the Groups client. (:pr:`NUMBER`)

    This allows copying-and-pasting of Groups API routes directly from the documentation.

### DIFF
--- a/changelog.d/20231128_104558_kurtmckee_fix_groups_api.rst
+++ b/changelog.d/20231128_104558_kurtmckee_fix_groups_api.rst
@@ -1,0 +1,6 @@
+Changed
+~~~~~~~
+
+-   Remove the built-in ``/v2`` API route prefix from the Groups client. (:pr:`NUMBER`)
+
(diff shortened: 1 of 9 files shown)


sirosen:
I don't believe we can do this safely -- it's a breaking change.
I've long wanted to remove the very notion of base-paths from the client definitions, but consider it unsafe.

If a client is directly accessing an API (perhaps one for which we don't have a method yet), they will face an issue:
```python
client = get_groups_client()
r = client.get("/some/cool/api")
```

Does this call `/some/cool/api` or `/v2/some/cool/api`?

kurtmckee:
Argh, that's disappointing. That means that -- for Groups -- we can't copy-and-paste API routes to the command line.

I couldn't figure out what I was doing wrong until I enabled `-vvv` at the CLI and discovered that the route started with `/v2/v2/groups`.

I'll refile this against the CLI to address this issue.

PR #910, opened 2023-12-01
Address flaky scope parsing test

kurtmckee:
This PR introduces the following changes:

* Add the `flaky` pytest plugin to mark the scope parsing timing test as flaky.
  This has the effect that it will be re-run once if it fails ([and it does](https://github.com/globus/globus-sdk-python/actions/runs/7060037397/job/19218803435#step:5:320)).

* Disable the flaky report for successes.

* Run `tox -m freezedeps` to pick up the flaky pytest plugin.
  This had the effect that mypy was upgraded, too.

* Resolve newly-triggered mypy lint failures caught by mypy 1.7.1.

### DIFF
--- a/requirements/py3.10/test.txt
+++ b/requirements/py3.10/test.txt
@@ -4,17 +4,19 @@
 #
 #    tox p -m freezedeps
 #
-certifi==2023.7.22
+certifi==2023.11.17
     # via requests
-charset-normalizer==3.3.0
+charset-normalizer==3.3.2
     # via requests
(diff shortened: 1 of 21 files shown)


sirosen (approved):
We could also remove that perf test, move it outside of the testsuite, or change its parameters and/or bounds. I've never seen it fail, but I assume this work was prompted by a failure.

kurtmckee:
I linked to the failure.

PR #912, opened 2023-12-04
Fix annotation for `IdentityMap.__init__` to allow `ConfidentialAppAuthClient`

sirosen:
The annotation should allow `ConfidentialAppAuthClient`, as that client type supports `get_identities`.

Includes a typing-test for `IdentityMap`.

----
:books: Documentation preview :books:: https://globus-sdk-python--912.org.readthedocs.build/en/912/


kurtmckee (approved):
Thanks for the fast fix!

PR #918, opened 2023-12-05
Migrate from a CHANGELOG symlink to the RST `.. include` directive

kurtmckee:
This resolves an issue I've had where PyCharm transparently loads `changelog.rst` (the symlink) and claims there are no changes to the file, which confused me during the recent release cycles.

Development
-----------

- Migrate from a CHANGELOG symlink to the RST ``.. include`` directive.

----
:books: Documentation preview :books:: https://globus-sdk-python--918.org.readthedocs.build/en/918/


sirosen (approved):
In addition to your PyCharm-related rationale, I think this is just "the right way to do it". So there's that too. :smile:

PR #925, opened 2023-12-18
Fix AuthClient.create_policy signature

sirosen:
Fixing the signature is an inherently breaking change. Therefore, move to keyword-only arguments as a progressive upgrade which makes this easier to manage into the future.

---

Per our team discussion, although this is a violation of the semver backwards compatibility policy (and not in a marginal way, but very clearly we are breaking an API which we had previously declared), doing this does more good than harm.
The API is relatively newly released, and any fix here either has to encode these as optional positional args or else has to handle things in an awkward way.

Given the short lifetime to-date of the SDK version with this method available, we are making the decision that breaking the compatibility policy is better than strict adherence in this case.

----
:books: Documentation preview :books:: https://globus-sdk-python--925.org.readthedocs.build/en/925/

### DIFF
--- a/changelog.d/20231218_161848_sirosen_fix_create_policy.rst
+++ b/changelog.d/20231218_161848_sirosen_fix_create_policy.rst
@@ -1,0 +1,6 @@
+Changed
+~~~~~~~
+
+- The argument specification for ``AuthClient.create_policy`` was incorrect.
+  Fixing this required a breaking change to the signature, and all arguments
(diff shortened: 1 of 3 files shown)


sirosen:
Upon self-review, I do not feel comfortable with the backwards incompatible change.
Although we sometimes bend the rules on exactly what semver specifies (mostly around spacebar-heating-esque things), this is very clearly not allowed.

I'm closing this and will present a backwards compatible fix.

PR #926, opened 2023-12-19
Fix AuthClient.create_policy to make some fields optional in a backwards-compatible way

sirosen:
This is done in a very carefully backwards-compatible way, and also introduces new helpers which make it easier to apply this sort of fix to other methods in the future (if necessary).

In the fixed implementation -- which makes the HA fields optional -- backwards compatibility shims ensure that positional argument usage with all 5 expected positional arguments works as expected. When some subset of those positional arguments are used, the remainder may be specified as keyword arguments.

The following cases are carefully handled and treated as errors:
- too many positional arguments
- same argument multiple times (positional and keyword)
- required argument omitted (now harder to detect)
- too few arguments when only positionals were used

The arguments are checked in the order in which they are passed, one at a time. This is done so that they can be compared against provided keyword argument values.
This allows easy enforcement of errors on usages which repeat a parameter, as in `ac.create_policy("foo", project_id="bar")`.

In order to implement these checks, new helpers have been added to `utils` which examine the locals in their calling context and provide a relatively simple-to-apply compatibility shim. They are tested relatively exhaustively in the context of the changes to `create_policy`, but they also have dedicated unit tests so that they can be maintained independently from this usage site.

Using a protocol to declare the desired type for the method + a decorator to cast to that type (without name-related issues), the type of `create_policy` is declared to type checkers as distinct from the runtime type. As a result, type checkers can treat non-keyword argument usage as erroneous while still allowing for the positional argument usage at runtime.

As an end result, `*args` are unpacked, TypeErrors are emitted at runtime in all of the appropriate cases, type checkers enforce correct calling conventions, and in any case in which `*args` is unpacked, a deprecation warning is emitted.

----
:books: Documentation preview :books:: https://globus-sdk-python--926.org.readthedocs.build/en/926/

### DIFF
--- a/changelog.d/20231218_161848_sirosen_fix_create_policy.rst
+++ b/changelog.d/20231218_161848_sirosen_fix_create_policy.rst
@@ -1,0 +1,6 @@
+Changed
+~~~~~~~
+
+- The argument specification for ``AuthClient.create_policy`` was incorrect.
+  The corrected method will emit deprecation warnings if called with positional
+  arguments, as the corrected version uses keyword-only arguments. (:pr:`NUMBER`)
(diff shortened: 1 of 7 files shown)


sirosen:
Closing in favor of #936

PR #928, opened 2023-12-19
Improve sphinx documentation rendering to use autodoc type hints for params and not in signatures

sirosen:
This includes some minor additional fixes but is mostly a long procedural change from using explicit `:type:` documentation to using autodoc type hints as much as possible.
This reduces a great deal of duplication and visual noise, making signatures significantly more readable in our rendered docs.

In a couple of cases, we need to get slightly "clever" in order for the hints to render well.

One of the more obvious ones of these is customized rendering of `globus_sdk.MISSING` as the bare string `MISSING` in signatures.
That makes `MISSING` usage look highly ergonomic in signatures.

In the case of Groups, there was one case of a class, `GroupPolicies`, whose docstring did not render correctly under autodoc.
A decorator, local to the module where `GroupPolicies` is defined, makes a modification to the `__doc__` of that class to give sphinx the appropriate values.

In `globus_sdk.paging`, there was a minor fix as this changeset uncovered a discrepancy between the type annotations and documented type.
The documentation was more correct (splatted args are a tuple, not a list), so changes have been made to align with that.

In `AuthClient`, there are some unions of `Literal[...]` with `MissingType` which sphinx only renders correctly if `MissingType` is the first member of the union, so for simplicitly these are handled by changing the union order to put `MissingType` first.
(This looks to be a sphinx bug, which will need a MRE + bug report.)

Under Flows, several methods had constraints (e.g. string length) documented in the type information.
That doesn't fit within the auspices of the Python type system -- there's no definition for constrained types within the language -- so this information was preserved by moving it to the parameter documentation in each case.

In a few cases, `:rtype:` and `:type:` have been used to explicitly document overrides which replace the autodoc type hint behavior.

----
:books: Documentation preview :books:: https://globus-sdk-python--928.org.readthedocs.build/en/928/

-----

Here's a pair of before/after clips of AuthClient. First of some innocuous methods, then of a pathological one.

case | before | after
---|---|---
simple | image | image
pathological | image | image

Note how the return type is now always annotated. Although things might be slightly longer in the simple cases, in the pathological cases they are _dramatically_ improved.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -26,6 +26,9 @@ intersphinx_mapping = {
 # pull signature docs from type hints, into body
 # and keep the signatures concise
 autodoc_typehints = "description"
+# do not generate doc stubs for inherited parameters from a superclass,
+# merely because they are type annotated
+autodoc_typehints_description_target = "documented_params"
 
 
 # sphinx extensions (minimally, we want autodoc and viewcode to build the site)
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -23,6 +23,10 @@ intersphinx_mapping = {
     "python": ("https://docs.python.org/3", None),
 }
 
+# pull signature docs from type hints, into body
+# and keep the signatures concise
+autodoc_typehints = "description"
+
 
 # sphinx extensions (minimally, we want autodoc and viewcode to build the site)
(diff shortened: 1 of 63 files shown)


kurtmckee (approved):
Nice work. Very, very nice work!

PR #933, opened 2024-01-02
GCS get and update endpoint

derek-globus:
[sc-28586](https://app.shortcut.com/globus/story/28586)

## What?

Added 2 new operations to the `GCSClient` class:
  * `gcs_client.get_endpoint()`
  * `gcs_client.update_endpoint(EndpointDocument(...))`

## Testing
* Updated an endpoint in sandbox environment.
* Unit tested with _testing patched responses.
* Tested nulling an endpoint subscription on a sandbox gcs endpoint:
  ```python
  gcs.update_endpoint(EndpointDocument(subscription_id=None))
  ```

----
📚 Documentation preview 📚: https://globus-sdk-python--933.org.readthedocs.build/en/933/

### DIFF
--- a/src/globus_sdk/services/gcs/client.py
+++ b/src/globus_sdk/services/gcs/client.py
@@ -9,6 +9,7 @@ from globus_sdk.authorizers import GlobusAuthorizer
 
 from .data import (
     CollectionDocument,
+    EndpointDocument,
     GCSRoleDocument,
     StorageGatewayDocument,
     UserCredentialDocument,
@@ -114,6 +115,71 @@ class GCSClient(client.BaseClient):
             "7e3f3f5e-350c-4717-891a-2f451c24b0d4": "SpectraLogic BlackPearl",
         }
         return connector_dict.get(str(connector_id))
+
+    #
+    # endpoint methods
+    #
+
+    def get_endpoint(
+        self, query_params: dict[str, t.Any] | None = None
+    ) -> UnpackingGCSResponse:
+        """
+        Get the details of the Endpoint that this client is configured to talk to.
+
+        :param query_params: Additional passthrough query parameters
+
+        .. tab-set::
+
+            .. tab-item:: API Info
+
+                ``GET /endpoint``
+
+                .. extdoclink:: Get Endpoint
+                    :ref: openapi_Endpoint/#getEndpoint
+                    :service: gcs
+        """
+        return UnpackingGCSResponse(
+            self.get("/endpoint", query_params=query_params),
+            "endpoint",
+        )
+
+    def update_endpoint(
+        self,
+        endpoint_data: dict[str, t.Any] | EndpointDocument,
+        *,
+        include: list[t.Literal["endpoint"]] | None = None,
+        query_params: dict[str, t.Any] | None = None,
+    ) -> UnpackingGCSResponse:
+        """
+        Update a GCSv5 Endpoint
+
+        :param endpoint_data: The endpoint document for the modified endpoint
+        :param include: Optional list of document types to include in the response
+           (currently only supports the value ``["endpoint"]``)
+        :param query_params: Additional passthrough query parameters
+
+        .. tab-set::
+
+            .. tab-item:: API Info
+
+                ``PATCH /endpoint``
+
+                .. extdoclink:: Update Endpoint
+                    :ref: openapi_Endpoint/#patchEndpoint
+                    :service: gcs
+        """
+        query_params = query_params or {}
+        if include is not None:
+            query_params["include"] = ",".join(include)
+
+        return UnpackingGCSResponse(
+            self.patch(
+                "/endpoint",
+                data=endpoint_data,
+                query_params=query_params,
+            ),
+            "endpoint",
+        )
 
     #
     # collection methods
--- a/src/globus_sdk/services/gcs/client.py
+++ b/src/globus_sdk/services/gcs/client.py
@@ -147,7 +147,9 @@ class GCSClient(client.BaseClient):
         self,
         endpoint_data: dict[str, t.Any] | EndpointDocument,
         *,
-        include: list[t.Literal["endpoint"]] | None = None,
+        include: (
+            t.Iterable[t.Literal["endpoint"]] | t.Literal["endpoint"] | None
+        ) = None,
         query_params: dict[str, t.Any] | None = None,
     ) -> UnpackingGCSResponse:
         """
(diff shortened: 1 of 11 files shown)


src/globus_sdk/services/gcs/data/endpoint.py
```diff
@@ -0,0 +1,146 @@
...
+        keywords: t.Iterable[str] | None = None,
+        # bools
+        allow_udt: bool | None = None,
+        public: bool | None = None,
+        # ints
+        gridftp_control_channel_port: int | None = None,
+        max_concurrency: int | None = None,
+        max_parallelism: int | None = None,
+        preferred_concurrency: int | None = None,
+        preferred_parallelism: int | None = None,
```
sirosen, line 112: I've reached out to the GCS team to ask if any of these fields can be set explicitly to `null` in an update.
If there are fields which need to be nullable, we'll want to switch this to using `MISSING`.

src/globus_sdk/services/gcs/data/endpoint.py
```diff
@@ -0,0 +1,146 @@
...
+
+    DATATYPE_BASE = "endpoint"
+    DATATYPE_VERSION_IMPLICATIONS: dict[str, tuple[int, int, int]] = {
+        "gridftp_control_channel_port": (1, 1, 0),
+    }
+    DATATYPE_VERSION_CALLBACKS: tuple[DatatypeCallback, ...] = ()
+
+    # Note: The fields below represent the set of mutable endpoint fields in
+    #   an Endpoint#1.2.0 document.
+    #   https://docs.globus.org/globus-connect-server/v5.4/api/schemas/Endpoint_1_2_0_schema/
```
sirosen, line 81: I'm alert to the fact that I see this schema listed as 1.2.0, but the mapping above only includes 1.1.0.
Is that because 1.2.0 added read-only fields? I want to confirm that this isn't a mistake.

derek-globus: Yep; from that doc:
> Version 1.2.0 of the endpoint includes read-only earliest_last_access to put a limit on collections which are missing a last_access value.

There's no mutable fields included in the 1.1.0 -> 1.2.0 version bump.

src/globus_sdk/services/gcs/data/endpoint.py
derek-globus: Saw in the thread there are 2 fields that do allow nulling which is unfortunate. The existing PayloadWrapper `_set_opt<type>` methods don't allow for null setting. It'll expand the scope of this work, but I'll need to refactor those all I guess if we're going to support explicit nulls.

Have I mentioned how much I love API nulls!

derek-globus: I couldn't figure out a clean way to genericize this in the existing interfaces (specifically the batch `_set_opt<type>s` functions). This is because "opt" in those helper functions use "optional" to mean "omittable", not "nullable"). 

What I've done is make `PayloadWrapper` configurable so that extending classes can override the "omit this" sentinel & gone directly to "_set_value(...)" for nullable fields.

src/globus_sdk/utils.py
```diff
@@ -147,28 +147,70 @@ class PayloadWrapper(PayloadWrapperBase):
     # internal helpers for setting non-null values
     #
 
+    # Some payloads use None to represent "no value" and others use the MISSING sentinel
+    _MISSING_SENTINEL: None | MissingType = None
```
sirosen, line 151: I would prefer that we not introduce this as a mechanism for skipping over `MISSING` values when they are passed for a PayloadWrapper. In the normal case, any class which accepts `MISSING` can do so like so:
```python
def __init__(self, foo: int | MissingType = MISSING):
    super().__init__()
    self["foo"] = foo
```

Having a path for skipping MISSING therefore seems inappropriate, since one of the main goals in introducing it was to allow a caller to pass the value forward and rely on the request-time serialization code to remove it.

I don't believe it's necessary to alter the `_set_value` and `_set_opt*` helpers in order to handle `MISSING` fields, and I'd instead prefer that we simply stop using those helpers for new classes as appropriate.

I recognize that there are callbacks being invoked here, but I think we can generally avoid them as well. I'll need to work up a a PR against this PR to make the recommendation which I think is appropriate. It's a little too much for the GitHub suggestion UI.

sirosen:
I've offered an alternative (PR on this PR) here: https://github.com/derek-globus/globus-sdk-python/pull/1

I'd like us to discuss this more if we don't want to go down that path.

src/globus_sdk/services/gcs/data/endpoint.py
derek-globus: Merged your PR.

sirosen: Okay, I'm doing a re-review now.

src/globus_sdk/services/gcs/client.py
```diff
@@ -115,6 +116,71 @@ def connector_id_to_name(connector_id: UUIDLike) -> str | None:
...
+        return UnpackingGCSResponse(
+            self.get("/endpoint", query_params=query_params),
+            "endpoint",
+        )
+
+    def update_endpoint(
+        self,
+        endpoint_data: dict[str, t.Any] | EndpointDocument,
+        *,
+        include: list[t.Literal["endpoint"]] | None = None,
```
sirosen, line 150: May I suggest that we update this to `list[Literal["endpoint"]] | Literal["endpoint"] | None = None`, and then use [`utils.commajoin`](https://github.com/globus/globus-sdk-python/blob/0b593d848fe88ea7b0e9a1f877ffbd6c4a3315f4/src/globus_sdk/utils.py#L118)?

The reason is a common bug which creeps in with string handling, since strings are iterable.

As currently written, if someone calls `update_endpoint(..., include="endpoint")`, they'll send `?include=e,n,d,p,o,i,n,t`, and then have to cope with the error.

Because my operating assumption is that most users of the SDK are not running type checkers, the likelihood of such a mistake is pretty high, in my experience.

(You can see `get_collection_list` below doing this same kind of handling with `filter`, but in a way which predates the `commajoin` utility.)

sirosen (changes requested):
I spotted one last, minor thing which I'd like us to improve before approving+merging.

src/globus_sdk/services/gcs/client.py
derek-globus: Sure

sirosen (approved):

PR #938, opened 2024-01-25
update transfer api links

aaschaer:
SC: https://app.shortcut.com/globus/story/29602/sdk-update-links-to-transfer-api-docs

Updates links that were broken by transfer doc changes for v4 eol and switching to collection terminology. There may be more breaks on the way, but this should be the biggest set.

Many of the existing links were aspirational in their `#` subsections, I opened a docs PR to make those actually work: https://github.com/globusonline/docs.globus.org/pull/1533 and will go through and verify all the links work once that's merged.

Not sure about the "DEPRECATION WARNING" and "WARNING" sections, will need to see how those render at the very least and am open to suggestions on text.

I didn't mess with the descriptions beyond warnings for things that can't really be used, might be worthwhile to make another story to do some work to use collection terminology there as well, or maybe we should just wait until the next major version where we can change the method names too.

----
📚 Documentation preview 📚: https://globus-sdk-python--938.org.readthedocs.build/en/938/

### DIFF
--- a/src/globus_sdk/services/transfer/client.py
+++ b/src/globus_sdk/services/transfer/client.py
@@ -472,7 +472,7 @@ class TransferClient(client.BaseClient):
 
                 .. extdoclink:: Get my guest collection list
                     :ref: transfer/endpoints_and_collections/#get_my_guest_collection_list
-        """
+        """  # noqa: E501
         log.info(f"TransferClient.my_shared_endpoint_list({endpoint_id}, ...)")
         return IterableTransferResponse(
             self.get(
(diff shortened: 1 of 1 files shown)


sirosen:
I'm checking things over before merging -- so far everything looks good; I just want to have a handle on what I'm merging -- and I think we may need to discuss the naming issues and potentially what we do about things like `create_endpoint`.

Technically, that one is still a route you can use if you want to get a setup_key for GCP, but if there is or will be a new API for the SDK to expose, dedicated only to that, maybe we can make an adjustment. (I'd have to crack open the GCP setup code to see, but I _think_ there was a new API added specifically for GCP? So maybe we just need to update SDK+CLI.

sirosen:
I just merged, but I might try a PR to move those deprecation warnings into `.. warning:` blocks, to see how it looks.

aaschaer:
> but I think there was a new API added specifically for GCP?

Yes, although its under `/private` so I'm not sure if we want to expose it through the SDK or not.

> I just merged, but I might try a PR to move those deprecation warnings into .. warning: blocks, to see how it looks.

This also reminds me that I marked this as "no-news-is-good-news" but deprecating several helper methods probably warrants a changelog update

sirosen:
It's a docs-only deprecation. We can add a changelog if we start having them emit deprecation warnings.
(But I'm not sure that we should yet, if some have no suitable replacement.)

PR #939, opened 2024-01-26
Add session_required_mfa to some missing auth interfaces

derek-globus:
## What?
1. Add `session_required_mfa` to `globus_sdk.exc.AuthorizationParameterInfo`
2. Add `session_required_policies` to docstring of `globus_sdk.exc.AuthorizationParameterInfo` (attribute already existed)
3. Add `session_required_mfa` to `AuthLoginClient.oauth2_get_authorize_url`

## Why?
Bring auth python objects closer to parity with Auth specification:
* [Required Error Format for Services](https://docs.globus.org/api/auth/sessions/#required-error-format)
* [Authorize URL Params](https://docs.globus.org/api/auth/reference/#authorization_code_grant_preferred)

## Testing
Updated relevant unit tests to supply `session_required_mfa` values.

----
📚 Documentation preview 📚: https://globus-sdk-python--939.org.readthedocs.build/en/939/

### DIFF
--- a/tests/functional/services/auth/test_auth_client_flow.py
+++ b/tests/functional/services/auth/test_auth_client_flow.py
@@ -34,6 +34,7 @@ def confidential_client(no_retry_transport):
 #   domain: str | list[str]
 #   identities: uuid | str | list[uuid] | list[str] | list[str | uuid]
 #   policies: uuid | str | list[uuid] | list[str] | list[str | uuid]
+#   mfa: True | False
 #   prompt: Literal["prompt"] | None
 #
 # The order of these options is consequential.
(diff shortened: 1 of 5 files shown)


tests/functional/services/auth/test_auth_client_flow.py
```diff
@@ -55,23 +56,24 @@ class CustomAuthClient(globus_sdk.ConfidentialAppAuthClient):
     ["baz-id", "quux-id"],
     ["baz-id", uuid.UUID(int=5)],
 )
+mfa_options = (True, False)
 prompt_options = ("login",)
 # Seed an all-`None` option test, then use a loop to fill in the rest.
 # There must be 4 parameters so parametrized tuple unpacking works in the test.
```
sirosen, line 62:
```suggestion
# There must be 5 parameters so parametrized tuple unpacking works in the test.
```

I feel like there's a rephrase such that the number isn't repeated here... Maybe just this?

```suggestion
# The number of parameters here must match the test parameters.
```

tests/unit/errors/test_common_functionality.py
```diff
@@ -346,6 +355,27 @@ def test_authz_params_info_containing_malformed_session_required_policies():
             "session_required_identities=None",
             "session_required_single_domain=None",
             "session_required_policies=None",
+            "session_required_mfa=None",
+        ],
+        ")",
+    )
+
+
+def test_authz_params_info_containing_malformed_session_required_mfa():
```
sirosen, line 364:
```suggestion
def test_authz_params_info_containing_session_required_mfa():
```

I think some bad pasta got into your copypasta.

If I'm reading the test right, it's not a malformed test.

You could add a malformed test by putting in something like `{"session_required_mfa": {"foo": "bar"}}`.

sirosen (commented):
There are some very minor corrections which I'd like to see before I approve.

tests/unit/errors/test_common_functionality.py
derek-globus: Yeah, that guy wasn't quite al dente. Good catch.

sirosen (approved):

PR #948, opened 2024-02-06
Add `filter_endpoint_use` to endpoint_manager_task_list

sirosen:
Add support for this parameter on ``TransferClient.endpoint_manager_task_list``.

Also revamp some tests with a new test fixture, introduced here.

----
📚 Documentation preview 📚: https://globus-sdk-python--948.org.readthedocs.build/en/948/


src/globus_sdk/services/transfer/client.py
```diff
@@ -1963,6 +1969,7 @@ def endpoint_manager_task_list(
         filter_task_id: None | UUIDLike | t.Iterable[UUIDLike] = None,
         filter_owner_id: UUIDLike | None = None,
         filter_endpoint: UUIDLike | None = None,
+        filter_endpoint_use: Literal["source", "destination"] | None = None,
```
kurtmckee, line 1972: Should this be `MISSING` instead of `None`?

sirosen: I'm _very_ unsure of how we should handle the `MISSING | None` situation across existing APIs, where `None` is already in use with `MISSING`-like semantics. I'll note that I did think about this, and that using `None` was a conscious choice here (one which is open to debate).

So far, my thought has been that we should try to keep methods self-consistent. Either their params use `None` or `MISSING`, but not a mix. The neighboring params are using `None` for "missing", therefore, this one should also use `None`.

I want all of these parameters to use `MISSING` when we work on v4.0. So I am setting us up for "more changes" later by doing this. But I think it will be easier for us (and for users?) to adapt from "all parameters default to `None`" to "all parameters default to `MISSING`" for most cases, rather than a mix.

---

That's my current line of thinking on this -- more or less in its entirety -- so there's lots of room to convince me if you think that starting to use `MISSING` for any new parameters is a better path.

kurtmckee: Works for me!

kurtmckee (approved):

PR #949, opened 2024-02-08
Add subscription_id support to create_flow

sirosen:
- add the param
- add a test for the param
- adjust the test matcher to be more lenient

The test is just a param being optionally passed to the existing
create_flow test. This revealed a strict match in the test fixture,
which I've here loosened to be non-strict (matches on a subset of
parameters).

----
📚 Documentation preview 📚: https://globus-sdk-python--949.org.readthedocs.build/en/949/


derek-globus (approved):
LGTM

PR #951, opened 2024-02-13
Allow passing `bytes` for payload bodies

sirosen:
This change is only applied to the "raw" HTTP methods. Higher level methods, belonging to per-service clients, still operate in terms of strings and do not complicate their signatures with bytes.

This requires various type signatures to be updated in the transport layer. Most notably, a check for `isinstance(X, str)` has been expanded to `isinstance(X, (bytes, str))`, meaning that `bytes` are now treated similarly to `str` when choosing how data should be encoded as JSON, urlencoded, or passed as raw text.


src/globus_sdk/transport/encoders.py
```diff
@@ -31,6 +31,7 @@ def encode(
                 "Either manually encode the data or use `encoding=form|json` to "
                 "correctly format this data."
             )
+        data = self._prepare_data(data)
```
sirosen, line 34: I did a self-review (on my way to hitting merge) and this jumped out as an unwanted line. Let me fix to remove.
(I think I was going to pull this out into a separate block in my first steps, and then I realized it wasn't necessary. So this is just a stray edit.)

PR #955, opened 2024-02-23
Introduce GCS `ConnectorTable` as mapping-like

sirosen:
The `ConnectorTable` supports the following:

- class attribute access: `ConnectorTable.POSIX`
- classmethod lookup by name: `ConnectorTable.lookup("posix")`
- classmethod lookup by ID: `ConnectorTable.lookup_by_id("...")`
- classmethod iteration: `ConnectorTable.all_connectors()`

The name-based lookups are normalized per some basic rules to make
most intelligible strings match.

Autodoc is extended to cover the new `ConnectorTable` as well as the
dataclass, `GlobusConnectServerConnector`, which it uses to represent
connector information.

----
📚 Documentation preview 📚: https://globus-sdk-python--955.org.readthedocs.build/en/955/

### DIFF
--- a/src/globus_sdk/services/gcs/connector_table.py
+++ b/src/globus_sdk/services/gcs/connector_table.py
@@ -51,40 +51,40 @@ class ConnectorTable:
     """
 
     _connectors: tuple[tuple[str, str, str], ...] = (
-        ("POSIX", "POSIX", "145812c8-decc-41f1-83cf-bb2a85a2a70b"),
+        ("ACTIVESCALE", "ActiveScale", "7251f6c8-93c9-11eb-95ba-12704e0d6a4d"),
         ("AZURE_BLOB", "Azure Blob", "9436da0c-a444-11eb-af93-12704e0d6a4d"),
         ("BLACKPEARL", "BlackPearl", "7e3f3f5e-350c-4717-891a-2f451c24b0d4"),
         ("BOX", "Box", "7c100eae-40fe-11e9-95a3-9cb6d0d9fd63"),
         ("CEPH", "Ceph", "1b6374b0-f6a4-4cf7-a26f-f262d9c6ca72"),
         ("DROPBOX", "Dropbox", "49b00fd6-63f1-48ae-b27f-d8af4589f876"),
-        ("ONEDRIVE", "OneDrive", "28ef55da-1f97-11eb-bdfd-12704e0d6a4d"),
-        ("GOOGLE_DRIVE", "Google Drive", "976cf0cf-78c3-4aab-82d2-7c16adbcc281"),
         (
             "GOOGLE_CLOUD_STORAGE",
             "Google Cloud Storage",
             "56366b96-ac98-11e9-abac-9cb6d0d9fd63",
         ),
-        ("ACTIVESCALE", "ActiveScale", "7251f6c8-93c9-11eb-95ba-12704e0d6a4d"),
-        ("S3", "S3", "7643e831-5f6c-4b47-a07f-8ee90f401d23"),
-        ("POSIX_STAGING", "POSIX Staging", "052be037-7dda-4d20-b163-3077314dc3e6"),
-        ("IRODS", "iRODS", "e47b6920-ff57-11ea-8aaa-000c297ab3c2"),
+        ("GOOGLE_DRIVE", "Google Drive", "976cf0cf-78c3-4aab-82d2-7c16adbcc281"),
         ("HPSS", "HPSS", "fb656a17-0f69-4e59-95ff-d0a62ca7bdf5"),
+        ("IRODS", "iRODS", "e47b6920-ff57-11ea-8aaa-000c297ab3c2"),
+        ("POSIX", "POSIX", "145812c8-decc-41f1-83cf-bb2a85a2a70b"),
+        ("POSIX_STAGING", "POSIX Staging", "052be037-7dda-4d20-b163-3077314dc3e6"),
+        ("ONEDRIVE", "OneDrive", "28ef55da-1f97-11eb-bdfd-12704e0d6a4d"),
+        ("S3", "S3", "7643e831-5f6c-4b47-a07f-8ee90f401d23"),
     )
 
-    POSIX: t.ClassVar[GlobusConnectServerConnector]
+    ACTIVESCALE: t.ClassVar[GlobusConnectServerConnector]
     AZURE_BLOB: t.ClassVar[GlobusConnectServerConnector]
     BLACKPEARL: t.ClassVar[GlobusConnectServerConnector]
     BOX: t.ClassVar[GlobusConnectServerConnector]
     CEPH: t.ClassVar[GlobusConnectServerConnector]
     DROPBOX: t.ClassVar[GlobusConnectServerConnector]
-    ONEDRIVE: t.ClassVar[GlobusConnectServerConnector]
-    GOOGLE_DRIVE: t.ClassVar[GlobusConnectServerConnector]
     GOOGLE_CLOUD_STORAGE: t.ClassVar[GlobusConnectServerConnector]
-    ACTIVESCALE: t.ClassVar[GlobusConnectServerConnector]
-    S3: t.ClassVar[GlobusConnectServerConnector]
-    POSIX_STAGING: t.ClassVar[GlobusConnectServerConnector]
-    IRODS: t.ClassVar[GlobusConnectServerConnector]
+    GOOGLE_DRIVE: t.ClassVar[GlobusConnectServerConnector]
     HPSS: t.ClassVar[GlobusConnectServerConnector]
+    IRODS: t.ClassVar[GlobusConnectServerConnector]
+    ONEDRIVE: t.ClassVar[GlobusConnectServerConnector]
+    POSIX: t.ClassVar[GlobusConnectServerConnector]
+    POSIX_STAGING: t.ClassVar[GlobusConnectServerConnector]
+    S3: t.ClassVar[GlobusConnectServerConnector]
 
     @classmethod
     def all_connectors(cls) -> t.Iterable[GlobusConnectServerConnector]:
--- a/src/globus_sdk/services/gcs/connector_table.py
+++ b/src/globus_sdk/services/gcs/connector_table.py
@@ -136,7 +136,7 @@ class ConnectorTable:
     ) -> GlobusConnectServerConnector | None:
         """
         Convert a connector_id into a connector object (containing name and ID).
-        Returns None if the id is not recognized.
+        Returns None if the ID is not recognized.
 
         :param connector_id: The ID of the connector
         """
--- a/src/globus_sdk/services/gcs/connector_table.py
+++ b/src/globus_sdk/services/gcs/connector_table.py
@@ -29,14 +29,14 @@ class ConnectorTable:
     This class defines the known Globus Connect Server Connectors in a mapping
     structure.
 
-    It supports access by attribute or via helper methods for doing lookups.
+    It supports access by attribute or via a helper method for doing lookups.
     For example, all of the following three usages retrieve the Azure Blob connector:
 
     .. code-block:: pycon
 
         >>> ConnectorTable.AZURE_BLOB
         >>> ConnectorTable.lookup("Azure Blob")
-        >>> ConnectorTable.lookup_by_id("9436da0c-a444-11eb-af93-12704e0d6a4d")
+        >>> ConnectorTable.lookup("9436da0c-a444-11eb-af93-12704e0d6a4d")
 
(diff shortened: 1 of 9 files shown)


src/globus_sdk/services/gcs/connector_table.py
```diff
@@ -0,0 +1,147 @@
...
+                return item
+        return None
+
+    @classmethod
+    def lookup_by_id(
+        cls, connector_id: UUIDLike
+    ) -> GlobusConnectServerConnector | None:
+        """
+        Convert a connector_id into a connector object (containing name and ID).
+        Returns None if the id is not recognized.
```
ada-globus, line 139:
```suggestion
        Returns None if the ID is not recognized.
```

src/globus_sdk/services/gcs/connector_table.py
```diff
@@ -0,0 +1,147 @@
...
+import re
+import typing as t
+
+from globus_sdk._types import UUIDLike
+
+_NORMALIZATION_PATTERN = re.compile(r"[_\- ]+")
+
+
+def _normalize_name(name: str) -> str:
+    return _NORMALIZATION_PATTERN.sub("-", name.strip()).lower()
```
ada-globus, line 13: Admittedly ends up being a very "permissive" normalization (afa resulting in matching on some crazy strings) but I also don't see any reason to expect this will be harmful. Given that users might build apps that looks these up by crazy names (and thus rely on this normalization) are we comfortable committing to this relatively long term (in software lifecycle time)?

src/globus_sdk/services/gcs/connector_table.py
```diff
@@ -0,0 +1,147 @@
...
+    HPSS: t.ClassVar[GlobusConnectServerConnector] = GlobusConnectServerConnector(
+        name="HPSS", connector_id="fb656a17-0f69-4e59-95ff-d0a62ca7bdf5"
+    )
+
+    @classmethod
+    def all_connectors(cls) -> t.Iterable[GlobusConnectServerConnector]:
+        """
+        Return an iterator of all known connectors.
+        """
+        for item in vars(cls).values():
```
ada-globus, line 110: This feels like the sort of move I would make, so while I don't take any issue w/ this, I also support (per our conversation) getting @kurtmckee's input on whether this is potentially problematic or whether there is a better, more idiomatic alternative we should reach for.

ada-globus (changes requested):
I largely approve! One micro change.

src/globus_sdk/services/gcs/connector_table.py
sirosen: I am, in that the rule can be explained pretty simply as
- `_`, `-`, and spaces are all equivalent
- runs of those things of any non-zero length are equivalent

So I don't think we're tied to this particular implementation -- which is the main thing which would give me discomfort.

src/globus_sdk/services/gcs/connector_table.py
kurtmckee: I think that I would reach for this implementation, since I like to keep raw data as raw data:

```python
class ConnectorTable:
    AZURE_BLOB: t.ClassVar[GlobusConnectServerConnector]
    ACTIVESCALE: t.ClassVar[GlobusConnectServerConnector]
    ...

    _connectors = (
        ("AZURE_BLOB", "Azure Blob", "9436da0c-a444-11eb-af93-12704e0d6a4d"),
        ("ACTIVESCALE", "Active Scale", "7251f6c8-93c9-11eb-95ba-12704e0d6a4d"),
        ...
    )

    for _attribute, _name, _id in _connectors:
        locals()[_attribute] = GlobusConnectServerConnector(name=_name, connector_id=_id)
    del _attribute, _name, _id

    @classmethod
    def all_connectors(cls) -> ...:
        yield from (getattr(cls, attribute) for attribute, _, _ in cls._connectors)
```

This accommodates name aliases and name changes, and means that the underlying data don't have to be duplicated/synchronized elsewhere in the test suite.

src/globus_sdk/services/gcs/connector_table.py
```diff
@@ -0,0 +1,147 @@
...
+            if isinstance(item, GlobusConnectServerConnector):
+                yield item
+
+    @classmethod
+    def lookup(cls, name: str) -> GlobusConnectServerConnector | None:
```
kurtmckee, line 115: I'm not convinced that it's necessary to differentiate lookups by name vs by ID using separate lookup methods.

tests/unit/helpers/gcs/test_connector_table.py
```diff
@@ -0,0 +1,92 @@
...
+import uuid
+
+import pytest
+
+from globus_sdk import ConnectorTable, GCSClient, exc
+
+# tuple of attribute name, connector name, and connector id
+#
+# this is basically a copy of the ConnectorTable data, but more simply structured
+_CONNECTORS: tuple[tuple[str, str, str], ...] = (
```
kurtmckee, line 12: This is exactly what I would want in the `ConnectorTable`.

src/globus_sdk/services/gcs/connector_table.py
sirosen: How would you prefer to dispatch between the lookup styles?
That is, if the def is
```python
def lookup(cls, name_or_id: str | UUIDLike) -> ...:
```
do you expect the body of that method to be
```
if looks_like_id(name_or_id): ...
else: ...
```
or something else?

I don't think it's strictly necessary to have two interface methods, but I want to be clear about what combining them means before I make a change.

src/globus_sdk/services/gcs/connector_table.py
sirosen: I can do something like this -- I'll work on it now -- but I'll note that I'm not convinced that this is a net simplification.

I'm okay with the data being defined here as a tuple of tuples, but then I need to add some testing which validates that all of the declared classvars get assigned values and that all of the values have a corresponding classvar declaration.

If the real target here is the duplication in the testsuite, then I think this all makes sense to me. I'm not that bothered by having a replica of data in tests in order to validate a data container, but I'm open to eliminating it as long as we don't introduce new possibilities for discrepancies.

sirosen: https://github.com/globus/globus-sdk-python/pull/955/commits/cc83ef1cfcaffb7bde03f0bafb9781692282cb6f is my interpretation of this request.
I learned today about `inspect.get_annotations()`, which worked out to be much simpler for this kind of inspection than `typing.get_type_hints()`. So the main thing I was concerned about -- testsuite complexity -- is handled nicely.

Since this implementation is pretty close to what you requested, I'm marking this resolved. LMK if there are further tweaks on this front though.

src/globus_sdk/services/gcs/connector_table.py
kurtmckee: I figured it would be something along the lines of:

```python
normalized = _normalize(name_or_id)
for attribute, name, id_ in cls._connectors:
    if normalized in {_normalize(name), id_}:
        return getattr(cls, attribute)
return None
```

src/globus_sdk/services/gcs/connector_table.py
```diff
@@ -50,66 +50,50 @@ class ConnectorTable:
...
+    CEPH: t.ClassVar[GlobusConnectServerConnector]
+    DROPBOX: t.ClassVar[GlobusConnectServerConnector]
+    ONEDRIVE: t.ClassVar[GlobusConnectServerConnector]
+    GOOGLE_DRIVE: t.ClassVar[GlobusConnectServerConnector]
+    GOOGLE_CLOUD_STORAGE: t.ClassVar[GlobusConnectServerConnector]
+    ACTIVESCALE: t.ClassVar[GlobusConnectServerConnector]
+    S3: t.ClassVar[GlobusConnectServerConnector]
+    POSIX_STAGING: t.ClassVar[GlobusConnectServerConnector]
+    IRODS: t.ClassVar[GlobusConnectServerConnector]
+    HPSS: t.ClassVar[GlobusConnectServerConnector]
```
kurtmckee, line 87: Recommending alphabetization here.

```suggestion
    ACTIVESCALE: t.ClassVar[GlobusConnectServerConnector]
    AZURE_BLOB: t.ClassVar[GlobusConnectServerConnector]
    BLACKPEARL: t.ClassVar[GlobusConnectServerConnector]
    BOX: t.ClassVar[GlobusConnectServerConnector]
    CEPH: t.ClassVar[GlobusConnectServerConnector]
    DROPBOX: t.ClassVar[GlobusConnectServerConnector]
    GOOGLE_CLOUD_STORAGE: t.ClassVar[GlobusConnectServerConnector]
    GOOGLE_DRIVE: t.ClassVar[GlobusConnectServerConnector]
    HPSS: t.ClassVar[GlobusConnectServerConnector]
    IRODS: t.ClassVar[GlobusConnectServerConnector]
    ONEDRIVE: t.ClassVar[GlobusConnectServerConnector]
    POSIX: t.ClassVar[GlobusConnectServerConnector]
    POSIX_STAGING: t.ClassVar[GlobusConnectServerConnector]
    S3: t.ClassVar[GlobusConnectServerConnector]
```

src/globus_sdk/services/gcs/connector_table.py
```diff
@@ -50,66 +50,50 @@ class ConnectorTable:
...
+        (
+            "GOOGLE_CLOUD_STORAGE",
+            "Google Cloud Storage",
+            "56366b96-ac98-11e9-abac-9cb6d0d9fd63",
+        ),
+        ("ACTIVESCALE", "ActiveScale", "7251f6c8-93c9-11eb-95ba-12704e0d6a4d"),
+        ("S3", "S3", "7643e831-5f6c-4b47-a07f-8ee90f401d23"),
+        ("POSIX_STAGING", "POSIX Staging", "052be037-7dda-4d20-b163-3077314dc3e6"),
+        ("IRODS", "iRODS", "e47b6920-ff57-11ea-8aaa-000c297ab3c2"),
+        ("HPSS", "HPSS", "fb656a17-0f69-4e59-95ff-d0a62ca7bdf5"),
```
kurtmckee, line 71: Recommending alphabetization here.

```suggestion
        ("ACTIVESCALE", "ActiveScale", "7251f6c8-93c9-11eb-95ba-12704e0d6a4d"),
        ("AZURE_BLOB", "Azure Blob", "9436da0c-a444-11eb-af93-12704e0d6a4d"),
        ("BLACKPEARL", "BlackPearl", "7e3f3f5e-350c-4717-891a-2f451c24b0d4"),
        ("BOX", "Box", "7c100eae-40fe-11e9-95a3-9cb6d0d9fd63"),
        ("CEPH", "Ceph", "1b6374b0-f6a4-4cf7-a26f-f262d9c6ca72"),
        ("DROPBOX", "Dropbox", "49b00fd6-63f1-48ae-b27f-d8af4589f876"),
        (
            "GOOGLE_CLOUD_STORAGE",
            "Google Cloud Storage",
            "56366b96-ac98-11e9-abac-9cb6d0d9fd63",
        ),
        ("GOOGLE_DRIVE", "Google Drive", "976cf0cf-78c3-4aab-82d2-7c16adbcc281"),
        ("HPSS", "HPSS", "fb656a17-0f69-4e59-95ff-d0a62ca7bdf5"),
        ("IRODS", "iRODS", "e47b6920-ff57-11ea-8aaa-000c297ab3c2"),
        ("POSIX", "POSIX", "145812c8-decc-41f1-83cf-bb2a85a2a70b"),
        ("POSIX_STAGING", "POSIX Staging", "052be037-7dda-4d20-b163-3077314dc3e6"),
        ("ONEDRIVE", "OneDrive", "28ef55da-1f97-11eb-bdfd-12704e0d6a4d"),
        ("S3", "S3", "7643e831-5f6c-4b47-a07f-8ee90f401d23"),
```

src/globus_sdk/services/gcs/connector_table.py
sirosen: Yep, there's no other good ordering to use; let's fix it!

src/globus_sdk/services/gcs/connector_table.py
sirosen: I feel like this might not play out quite this simply -- the normalized name might not match the UUID string. I'll take a look and see if we can get this or something close though. New revision coming in just a few.

sirosen: Ah, no, it really should be this simple, since we normalize with lowercase and `-`.

sirosen: I decided to use the `all_connectors()` iterator, but it's so close to this implementation that I jotted it down as a co-authored commit.

kurtmckee (approved):

PR #956, opened 2024-02-23
Fix automodule docs and missing/undocumented public symbols

sirosen:
Originally, the focus of this work was on `automodule` directives which rendered using the source module name rather than `globus_sdk`.
To fix this, I switched from `automodule` to the component `autoclass` directives (which respect the `current_module` setting).
That presented an opportunity to cleanup the GCS docs a little bit.

Having done this, a problem became apparent: how do we know that everything in the SDK is documented?
How do we keep our docs in sync as new features (classes) are added?
For this, a quick-and-dirty lint check is added, which simply does substring checking of the public exports in `globus_sdk.__all__` against all of `docs/**/*.rst`.

Getting the linter to pass requires some further doc tweaks, and then it can be added to `pre-commit` with `always-run: true`.

---

- Replace automodule directives with autoclass
- Add script to check that exports are documented
- Add exported symbol checker to pre-commit

----
📚 Documentation preview 📚: https://globus-sdk-python--956.org.readthedocs.build/en/956/

### DIFF
--- a/src/globus_sdk/exc/warnings.py
+++ b/src/globus_sdk/exc/warnings.py
@@ -4,7 +4,13 @@ import warnings
 
 
 class RemovedInV4Warning(DeprecationWarning):
-    pass
+    """
+    This warning indicates that a feature or usage was detected which will be
+    unsupported in globus-sdk version 4.
+
+    Users are encouraged to resolve these warnings when possible, so that they will be
+    able to upgrade to version 4 without issue.
+    """
 
 
 def warn_deprecated(message: str, stacklevel: int = 2) -> None:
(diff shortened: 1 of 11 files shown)


src/globus_sdk/exc/warnings.py
```diff
@@ -4,7 +4,13 @@
 
 
 class RemovedInV4Warning(DeprecationWarning):
-    pass
+    """
+    This warning indicates that a feature or usage was detected which will be
+    unsupported in globus-sdk version 4.
+
+    Users are encouraged to resolve these warnings when possible, so that they will be
+    able to upgrade to version 4 without issue.
```
kurtmckee, line 12: I don't recommend guaranteeing that resolving usage warnings now will result in zero upgrade issues in v4.

```suggestion
    Users are encouraged to resolve these warnings when possible.
```

kurtmckee (changes requested):

kurtmckee (approved):

PR #961, opened 2024-03-01
add TransferClient.operation_stat

aaschaer:
SC: https://app.shortcut.com/globus/story/30451/sdk-cli-add-support-for-transfer-stat-operation

Transfer docs for stat: https://docs.globus.org/api/transfer/file_operations/#stat

I decided not to add special treatment for `NotFound` here to keep with our pattern of being a thin wrapper, but am planning on doing so in the CLI

----
📚 Documentation preview 📚: https://globus-sdk-python--961.org.readthedocs.build/en/961/

### DIFF
--- a/changelog.d/20240301_143526_aaschaer_stat.rst
+++ b/changelog.d/20240301_143526_aaschaer_stat.rst
@@ -1,0 +1,4 @@
+ Added
+ ~~~~~
+
+ - Added `TransferClient,operation_stat` helper method for getting the status of a path on a collection (:pr:`NUMBER`)
--- a/changelog.d/20240301_143526_aaschaer_stat.rst
+++ b/changelog.d/20240301_143526_aaschaer_stat.rst
@@ -1,4 +1,4 @@
  Added
  ~~~~~
 
- - Added `TransferClient,operation_stat` helper method for getting the status of a path on a collection (:pr:`NUMBER`)
+ - Added ``TransferClient.operation_stat`` helper method for getting the status of a path on a collection (:pr:`NUMBER`)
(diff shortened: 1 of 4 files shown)


changelog.d/20240301_143526_aaschaer_stat.rst
```diff
@@ -0,0 +1,4 @@
+ Added
+ ~~~~~
+
+ - Added `TransferClient,operation_stat` helper method for getting the status of a path on a collection (:pr:`NUMBER`)
```
sirosen, line 4: Some very mild typos to fix (the eternal rst doc issue of backticks):
```suggestion
 - Added ``TransferClient.operation_stat`` helper method for getting the status of a path on a collection (:pr:`NUMBER`)
```

src/globus_sdk/services/transfer/client.py
```diff
@@ -1235,6 +1235,40 @@ def operation_rename(
...
+        query_params: dict[str, t.Any] | None = None,
+    ) -> response.GlobusHTTPResponse:
+        """
+        :param endpoint_id: The ID of the collection on which to do the stat operation
+        :param path: Path on the collection to do the stat operation on
+        :param local_user: Optional value passed to identity mapping specifying which
+            local user account to map to. Only usable with Globus Connect Server v5
+            mapped collections.
+        :param query_params: Additional passthrough query parameters
+            .. tab-item:: API Info
```
sirosen, line 1253: This looks accidentally-truncated. You need the `.. tab-set::` directive above.

I'd also like it if we used the `_testing` data to generate an example response here. I just did this method, so it's what I'll use as an example:
https://github.com/globus/globus-sdk-python/blob/8e06b1b70202ca954c6a343b8bfca617c39d249e/src/globus_sdk/services/groups/client.py#L103-L105
(`expandtestfixture` does the `_testing` lookup and splats the result into a generated code-block)

Maybe another tab with example usage (like `operation_rename`, above) would be good to add as well.

sirosen (changes requested):
Minor tweaks needed, but I'm looking forward to the whole stack having CLI support, etc.
Let's get this over the finish line!

sirosen (approved):

PR #962, opened 2024-03-04
add additional _testing cases for stat

aaschaer:
Mostly for https://github.com/globus/globus-cli/pull/963, but since at least one other service (TransferAP) will have special cases around NotFound errors, these seemed like good things to have generally available in the SDK rather than just the CLI

----
📚 Documentation preview 📚: https://globus-sdk-python--962.org.readthedocs.build/en/962/


sirosen (approved):
Yep; good call!

PR #965, opened 2024-03-19
Add `scope` to `oauth2_get_dependent_tokens`

sirosen:
Also update the docs for this method to modernize and use newer tools.

Add the parameter
- it's a string or iterable of strings, defaults to MISSING
- it gets space-joined when non-MISSING

A test confirms a few cases are sent correctly (no value, string value,
tuple of strings).

Per the Auth docs, use of the parameter is recommended.

----
📚 Documentation preview 📚: https://globus-sdk-python--965.org.readthedocs.build/en/965/

### DIFF
--- a/src/globus_sdk/services/auth/client/confidential_client.py
+++ b/src/globus_sdk/services/auth/client/confidential_client.py
@@ -216,7 +216,8 @@ class ConfidentialAppAuthClient(AuthLoginClient):
             to access tokens. [Default: ``False``]
         :param scope: The scope or scopes of the dependent tokens which are being
             requested. Applications are recommended to provide this string to ensure
-            that they are receiving the tokens they expect.
+            that they are receiving the tokens they expect. If omitted, all available 
+            dependent tokens will be returned.
         :param additional_params: Additional parameters to include in the request body
 
         .. tab-set::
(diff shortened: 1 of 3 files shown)


src/globus_sdk/services/auth/client/confidential_client.py
```diff
@@ -211,11 +211,44 @@ def oauth2_get_dependent_tokens(
         their relationship). As long as that is the case, Service A can use
         this Grant to get those "Dependent" or "Downstream" tokens for Service B.
 
-        :param token: A Globus Access Token as a string
+        :param token: An access token as a string
         :param refresh_tokens: When True, request dependent refresh tokens in addition
             to access tokens. [Default: ``False``]
+        :param scope: The scope or scopes of the dependent tokens which are being
+            requested. Applications are recommended to provide this string to ensure
+            that they are receiving the tokens they expect.
```
derek-globus, line 219:
```suggestion
        :param scope: The scope or scopes of the dependent tokens which are being
            requested. Applications are recommended to provide this string to ensure
            that they are receiving the tokens they expect. If omitted, all available 
            dependent tokens will be returned.
```

sirosen: Good call!

derek-globus (approved):

PR #966, opened 2024-03-22
Non experimental scope parsing sc 30765

derek-globus:
## What?
1. Move Scope out of experimental
2. Remove Scope `contains` function
3. Remove Scope benchmarking script

## Why?
1. It's sat in experimental long enough for us to be comfortable that the interface won't change in impactful ways.
2. `contains` was a good objective but conveys too much semantically about what is being evaluated
3. This was already run against the `scope` construct. If we want to run it again, we can revive it from github history.

----
📚 Documentation preview 📚: https://globus-sdk-python--966.org.readthedocs.build/en/966/


changelog.d/20240322_122646_derek_non_experimental_scope_parsing_sc_30765.rst
```diff
@@ -0,0 +1,6 @@
+
+Added
+~~~~~
+
+- Moved Scope parsing out of experimental. The `Scope` construct is not importable from
```
sirosen, line 5: We could possibly add more here (e.g. example usage in the readme), but I think these minor adjustments are necessary:

```suggestion
- Moved scope parsing out of experimental. The ``Scope`` construct is now importable from
```

src/globus_sdk/__init__.py
```diff
@@ -133,6 +133,11 @@ def _force_eager_imports() -> None:
         "TransferClient",
         "TransferData",
     },
+    "scopes": {
+        "Scope",
+        "ScopeParseError",
+        "ScopeCycleError",
+    },
```
sirosen, line 140: Should we be adding these as top-level exports? Or should we keep it at `globus_sdk.scopes.Scope`, as with `MutableScope`?

On the one hand, being exported at the top level is "simpler" -- on the other hand, it's already a very crowded namespace.

sirosen (commented):
Before we move this forward, I think we need to make sure that `Scope` objects -- both singular and plural as contained in a list -- work as inputs to methods which take strings or `MutableScope` objects as their inputs.

Primarily, I'm thinking of the `requested_scopes=...` parameters for OAuth2 flows, but there might be other things I'm forgetting.

derek-globus:
>  Before we move this forward, I think we need to make sure that Scope objects -- both singular and plural as contained in a list -- work as inputs to methods which take strings or MutableScope objects as their inputs.
> Primarily, I'm thinking of the requested_scopes=... parameters for OAuth2 flows, but there might be other things I'm forgetting.

Agreed but splitting this off into it's own bucket of work.

sirosen:
Per our conversation, we're dropping any notion that _this_ PR is going to handle updating the `ScopeCollectionType` to include `Scope`. We'll do that in a fast-follow PR (which should be quite easy!).

sirosen (approved):

PR #970, opened 2024-04-01
Update pre-commit hook URLs

kurtmckee:
* Update hook URLs that redirect (`asottile/blacken-docs` -> `adamchainz/blacken-docs`)
* Use preferred URLs (2x faster compiled black hook URL)
* Run `upadup`
* Change the pre-commit auto-update schedule to 'quarterly'

----
📚 Documentation preview 📚: https://globus-sdk-python--970.org.readthedocs.build/en/970/

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,7 +33,7 @@ repos:
     - id: flake8
       name: "Lint python files"
       additional_dependencies:
-        - 'flake8-bugbear==23.9.16'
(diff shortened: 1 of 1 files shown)


sirosen (approved):
I'm not sure I agree with quarterly cadence, but I'm also not sure I disagree. I disagree with bundling it with these other changes, but it's not enough of an issue for me to demand the busywork of splitting it out.

PR #972, opened 2024-04-05
LoginFlowManager and CommandLineLoginFlowManager

aaschaer:
Shortcut: https://app.shortcut.com/globus/story/30763/sdk-loginmanagers-authrequirements

Two places of potential issue/improvement:

1. I'm using `GlobusAuthorizationParameters` as the input to `run_login_flow` instead of implementing `AuthRequirements` as originally spec'd. This mostly makes sense because that object has all the parameters needed for an authorization flow. It might be a bit fishy with types though. Making `prompt` a `Literal["login]` everywhere wasn't too painful, but currently any `Scope` objects will have to be serialized into strings despite that fact that they're just being passed through to `oauth2_start_flow` which accepts any `ScopeCollectionType` (which should have `Scope` added to it at some point). Making `GlobusAuthorizationParameters` accept `ScopeCollectionType` seemed to not line up well with the validation logic.

2. I've added `type: ignore` to my call to `self.login_client.oauth2_start_flow` as `AuthLoginClient` doesn't have a signature for `oauth2_start_flow` - presumably because it has different signatures across `NativeAppAuthClient` and `ConfidentialAppAuthClient`? Not sure if there is some abstract method magic that could be done here? Or maybe I should just make `login_client`'s type `NativeAppAuthClient | ConfidentialAppAuthClient`, but that also feels wrong.

### DIFF
--- a/src/globus_sdk/experimental/auth_requirements_error/_auth_requirements_error.py
+++ b/src/globus_sdk/experimental/auth_requirements_error/_auth_requirements_error.py
@@ -1,14 +1,24 @@
 from __future__ import annotations
 
+import sys
 import typing as t
+
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
 
 from . import _serializable, _validators
 
 
 class GlobusAuthorizationParameters(_serializable.Serializable):
     """
-    Represents authorization parameters that can be used to instruct a client
-    which additional authorizations are needed in order to complete a request.
+    Data class containing authorization parameters that can be passed during
+    an authentication flow to control how the user will authenticate.
+
+    When used with a GlobusAuthRequirementsError this represents the additional
+    authorization parameters needed in order to complete a request that had
+    insufficient authorization state.
 
     :ivar session_message: A message to be displayed to the user.
     :vartype session_message: str, optional
@@ -49,7 +59,7 @@ class GlobusAuthorizationParameters(_serializable.Serializable):
         session_required_single_domain: list[str] | None = None,
         session_required_mfa: bool | None = None,
         required_scopes: list[str] | None = None,
-        prompt: str | None = None,
+        prompt: Literal["login"] | None = None,
         extra: dict[str, t.Any] | None = None,
     ):
         self.session_message = _validators.opt_str("session_message", session_message)
@@ -68,7 +78,10 @@ class GlobusAuthorizationParameters(_serializable.Serializable):
         self.required_scopes = _validators.opt_str_list(
             "required_scopes", required_scopes
         )
-        self.prompt = _validators.opt_str("prompt", prompt)
+        if prompt in [None, "login"]:
+            self.prompt = prompt
+        else:
+            raise _validators.ValidationError("'prompt' must be 'login' or null")
         self.extra = extra or {}
 
         # Enforce that the error contains at least one of the fields we expect
--- a/src/globus_sdk/experimental/auth_requirements_error/_auth_requirements_error.py
+++ b/src/globus_sdk/experimental/auth_requirements_error/_auth_requirements_error.py
@@ -1,12 +1,6 @@
 from __future__ import annotations
 
-import sys
 import typing as t
-
-if sys.version_info >= (3, 8):
-    from typing import Literal
-else:
-    from typing_extensions import Literal
 
 from . import _serializable, _validators
 
@@ -59,7 +53,7 @@ class GlobusAuthorizationParameters(_serializable.Serializable):
         session_required_single_domain: list[str] | None = None,
         session_required_mfa: bool | None = None,
         required_scopes: list[str] | None = None,
-        prompt: Literal["login"] | None = None,
+        prompt: str | None = None,
         extra: dict[str, t.Any] | None = None,
     ):
         self.session_message = _validators.opt_str("session_message", session_message)
@@ -78,10 +72,7 @@ class GlobusAuthorizationParameters(_serializable.Serializable):
         self.required_scopes = _validators.opt_str_list(
             "required_scopes", required_scopes
         )
-        if prompt in [None, "login"]:
-            self.prompt = prompt
-        else:
-            raise _validators.ValidationError("'prompt' must be 'login' or null")
+        self.prompt = _validators.opt_str("prompt", prompt)
         self.extra = extra or {}
 
         # Enforce that the error contains at least one of the fields we expect
--- a/src/globus_sdk/experimental/login_flow_manager/command_line_login_flow_manager.py
+++ b/src/globus_sdk/experimental/login_flow_manager/command_line_login_flow_manager.py
@@ -10,6 +10,15 @@ class CommandLineLoginFlowManager(LoginFlowManager):
     """
     A ``CommandLineLoginFlowManager`` is a ``LoginFlowManager`` that uses the command
     line for interacting with the user during its interactive login flows.
+
+    Example usage:
+
+    >>> login_client = globus_sdk.NativeAppAuthClient(...)
+    >>> login_flow_manager = CommandLineLoginFlowManager(login_client)
+    >>> scopes = [globus_sdk.scopes.TransferScopes.all]
+    >>> auth_params = GlobusAuthorizationParameters(required_scopes=scopes)
+    >>> tokens = login_flow_manager.run_login_flow(auth_params)
+
     """
 
     def __init__(
--- a/src/globus_sdk/experimental/login_flow_manager/command_line_login_flow_manager.py
+++ b/src/globus_sdk/experimental/login_flow_manager/command_line_login_flow_manager.py
@@ -79,7 +79,7 @@ class CommandLineLoginFlowManager(LoginFlowManager):
                     ),
                     session_required_policies=auth_parameters.session_required_policies,
                     session_required_mfa=auth_parameters.session_required_mfa,
-                    prompt=auth_parameters.prompt,
+                    prompt=auth_parameters.prompt,  # type: ignore
                 ),
             )
         )
--- a/src/globus_sdk/experimental/login_flow_manager/command_line_login_flow_manager.py
+++ b/src/globus_sdk/experimental/login_flow_manager/command_line_login_flow_manager.py
@@ -1,0 +1,79 @@
+from globus_sdk import AuthLoginClient, OAuthTokenResponse
+from globus_sdk.experimental.auth_requirements_error import (
+    GlobusAuthorizationParameters,
+)
+from globus_sdk.experimental.login_flow_manager import LoginFlowManager
+
+
+class CommandLineLoginFlowManager(LoginFlowManager):
+    """
+    A ``CommandLineLoginFlowManager`` is a ``LoginFlowManager`` that uses the command
+    line for interacting with the user during its interactive login flows.
+    """
+
+    def __init__(
+        self,
+        login_client: AuthLoginClient,
+        *,
+        login_prompt: str = "Please authenticate with Globus here",
+        code_prompt: str = "Enter the resulting Authorization Code here",
+    ):
+        """
+        :param login_client: The ``AuthLoginClient`` that will be making the Globus
+            Auth API calls needed for the authentication flow.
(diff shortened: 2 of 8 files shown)


src/globus_sdk/experimental/login_flow_manager/command_line_login_flow_manager.py
```diff
@@ -0,0 +1,79 @@
...
+
+        :param auth_parameters: ``GlobusAuthorizationParameters`` passed through
+            to the authentication flow to control how the user will authenticate.
+        :param refresh_tokens: Whether or not refresh tokens will be requested.
+        """
+
+        # type is ignored here as AuthLoginClient does not provide a signature for
+        # oauth2_start_flow since it has different positional arguments between
+        # NativeAppAuthClient and ConfidentialAppAuthClient
+        self.login_client.oauth2_start_flow(  # type: ignore
```
kurtmckee, line 50: This looks fine to me. It's wholly self-contained in this method and looks like it works just fine.

src/globus_sdk/experimental/login_flow_manager/login_flow_manager.py
```diff
@@ -0,0 +1,35 @@
...
+        auth_parameters: GlobusAuthorizationParameters,
+        *,
+        refresh_tokens: bool = False,
+    ) -> OAuthTokenResponse:
+        """
+        Run an interactive login flow to get tokens for the user.
+
+        :param auth_parameters: ``GlobusAuthorizationParameters`` passed through
+            to the authentication flow to control how the user will authenticate.
+        :param refresh_tokens: Whether or not refresh tokens will be requested.
```
kurtmckee, line 34: Perhaps:

```suggestion
        :param refresh_tokens: Control whether refresh tokens will be requested.
```

src/globus_sdk/experimental/auth_requirements_error/_auth_requirements_error.py
```diff
@@ -68,7 +78,10 @@ def __init__(
         self.required_scopes = _validators.opt_str_list(
             "required_scopes", required_scopes
         )
-        self.prompt = _validators.opt_str("prompt", prompt)
+        if prompt in [None, "login"]:
+            self.prompt = prompt
+        else:
+            raise _validators.ValidationError("'prompt' must be 'login' or null")
```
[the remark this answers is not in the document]
aaschaer: I kinda assumed that this was in relation to json parsing because it used null instead of None, but I'm not sure I completely understand how this object is meant to be used (which is part of why I'm hesitant to mess around with it)

kurtmckee (approved):

src/globus_sdk/experimental/login_flow_manager/command_line_login_flow_manager.py
```diff
@@ -10,6 +10,15 @@ class CommandLineLoginFlowManager(LoginFlowManager):
...
     A ``CommandLineLoginFlowManager`` is a ``LoginFlowManager`` that uses the command
     line for interacting with the user during its interactive login flows.
+
+    Example usage:
+
+    >>> login_client = globus_sdk.NativeAppAuthClient(...)
+    >>> login_flow_manager = CommandLineLoginFlowManager(login_client)
+    >>> scopes = [globus_sdk.scopes.TransferScopes.all]
+    >>> auth_params = GlobusAuthorizationParameters(required_scopes=scopes)
+    >>> tokens = login_flow_manager.run_login_flow(auth_params)
```
kurtmckee, line 20: Is this extracted by a Sphinx extension? If so, it needs this formatting:

```suggestion
    ..  code-block:: pycon

        >>> login_client = globus_sdk.NativeAppAuthClient(...)
        >>> login_flow_manager = CommandLineLoginFlowManager(login_client)
        >>> scopes = [globus_sdk.scopes.TransferScopes.all]
        >>> auth_params = GlobusAuthorizationParameters(required_scopes=scopes)
        >>> tokens = login_flow_manager.run_login_flow(auth_params)
```

aaschaer: I don't think it is? I'm not seeing that formatting anywhere else in the SDK at least

src/globus_sdk/experimental/login_flow_manager/command_line_login_flow_manager.py
```diff
@@ -0,0 +1,92 @@
...
+        :param auth_parameters: ``GlobusAuthorizationParameters`` passed through
+            to the authentication flow to control how the user will authenticate.
+        :param refresh_tokens: Whether or not refresh tokens will be requested.
+        """
+
+        # type is ignored here as AuthLoginClient does not provide a signature for
+        # oauth2_start_flow since it has different positional arguments between
+        # NativeAppAuthClient and ConfidentialAppAuthClient
+        self.login_client.oauth2_start_flow(  # type: ignore
+            redirect_uri=self.login_client.base_url + "v2/web/auth-code",
```
derek-globus, line 64: Would it make sense to make this configurable (init kwarg)?

I'm not too familiar with how extensively custom uri redirection is actually used. Maybe something we add later if customer demand surfaces?

src/globus_sdk/experimental/login_flow_manager/command_line_login_flow_manager.py
```diff
@@ -0,0 +1,92 @@
...
+        """
+        self.login_prompt = login_prompt
+        self.code_prompt = code_prompt
+        super().__init__(login_client)
+
+    def run_login_flow(
+        self,
+        auth_parameters: GlobusAuthorizationParameters,
+        *,
+        refresh_tokens: bool = False,
```
derek-globus, line 50: Would this make more sense as a class init param?

(Thinking out loud more than making a suggestion)

I think that builds more in the direction of "I'm operating a script/tool which deals with refresh tokens so I configure my login manager to use refresh tokens".

src/globus_sdk/experimental/auth_requirements_error/_auth_requirements_error.py
```diff
@@ -49,7 +59,7 @@ def __init__(
         session_required_single_domain: list[str] | None = None,
         session_required_mfa: bool | None = None,
         required_scopes: list[str] | None = None,
-        prompt: str | None = None,
+        prompt: Literal["login"] | None = None,
```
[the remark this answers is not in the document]
aaschaer: @sirosen thoughts? I mostly did this to make mypy happy with the call to `oauth2_get_authorize_url` which calls for a literal. We could make that more permissive or just tell mypy not to worry about it.

I'm a little worried about the actual validation code that I changed to match, that would raise an actual runtime error if in the future people needed to pass other types of `prompt`

src/globus_sdk/experimental/login_flow_manager/command_line_login_flow_manager.py
aaschaer: I'm not aware of any use case for not using the Globus Auth provided uri for a command line copy and paste flow so I'm happy waiting for customer demand

src/globus_sdk/experimental/login_flow_manager/command_line_login_flow_manager.py
kurtmckee: Built it locally, and you're correct. Thanks!

src/globus_sdk/experimental/login_flow_manager/command_line_login_flow_manager.py
aaschaer: Hmm, yeah, I think that lines up better with `GlobusApp` having it configured at the app level rather than at individual calls to `run_login_flow` I'll move it!

src/globus_sdk/experimental/login_flow_manager/command_line_login_flow_manager.py
derek-globus: Sounds good

src/globus_sdk/experimental/auth_requirements_error/_auth_requirements_error.py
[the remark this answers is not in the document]
sirosen: I'm getting up to speed on this right now. My baseline reaction is that a Literal type-hint is "just a lint" and so we need not spill too much ink over it. If we're wrong in the annotation, but the runtime behavior is that we accept any `str`, then I think that's correct -- in theory, `prompt="consent"` could be added in Globus Auth and I would want our code to support such usage *at runtime*.

I need to read a bit more about what's changing here to see if there's a deeper issue for discussion.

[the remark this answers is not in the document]
sirosen: _reads literally the next comment_
Okay! I think there is a need for a small change here.

I do not believe that we should be doing runtime validation that `prompt in (None, 'login')`. Doing so forbids usages which we can foresee, even though they aren't supported service-side today.

I have no particularly strong opinion about the annotation here being `str | None` vs `Literal["login"] | None`. Either is acceptable to me. One provides a useful hint to developers interacting with this directly, but may fail spuriously. The other provides less "hinting" but is pedantically more true to the service and specs.
OIDC specifies several values for `prompt` beyond `"login"`, but Globus Auth only exposes special behavior for `"login"` (as far as I know).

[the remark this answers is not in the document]
aaschaer: My validation changes to make mypy happy would break at runtime, so I'll revert those

src/globus_sdk/experimental/login_flow_manager/command_line_login_flow_manager.py
sirosen: I agree that this seems fine -- for this code, as a user of the client classes.

It strikes me as a "code/design smell in the clients themselves". I'd love to resolve it, but I think it's safely out of scope for this PR to even try to fix that issue. I'm not sure what the fix really is. (A new method on the base accepting `*args` and raising notimplemented? :thinking: )

src/globus_sdk/experimental/auth_requirements_error/_auth_requirements_error.py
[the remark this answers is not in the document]
sirosen: Let the record show: I endorse a `type:ignore` comment in this case. :slightly_smiling_face:

sirosen:
Some high-level review comments:

> I'm using GlobusAuthorizationParameters as the input to run_login_flow instead of implementing AuthRequirements as originally spec'd.

This seems like a good "optimization" to me. Maintaining yet-another type which has borderline identical semantics is wasted effort if we can avoid it. And we have time before we promote this out of `experimental` to decide if it's causing us other harms.

> Making GlobusAuthorizationParameters accept ScopeCollectionType seemed to not line up well with the validation logic.

:+1: I would be against that, if avoidable. IMO those GARE-internal objects should be pretty close to the wire-format for the data after init. That keeps the mode of interaction consistent, in that you're expected to put JSON-serializable types in there. Allowing init to convert other types to JSON-serializable ones -- like `ScopeCollection -> str` -- is fair game.

> I've added type: ignore to my call to self.login_client.oauth2_start_flow as AuthLoginClient doesn't have a signature for oauth2_start_flow - presumably because it has different signatures across NativeAppAuthClient and ConfidentialAppAuthClient? Not sure if there is some abstract method magic that could be done here?

Let's consider this noted but non-blocking for now. I agree, some :magic_wand: magic :sparkles: can probably fix this, but let's treat it separately. I'm not sure what kind of fix might be best, and it will probably involve some level of gamesmanship around backwards compatibility, typing, and abstract classes. Probably we'll do something that falls into that classic category of "backwards incompatible, but only if you're really strict about it".

src/globus_sdk/experimental/auth_requirements_error/_auth_requirements_error.py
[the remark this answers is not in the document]
aaschaer: I went ahead and also reverted the annotation to `str | None` since that felt like it lined up better and was covered by the same `type: ignore`

derek-globus (approved):

PR #973, opened 2024-04-05
Fix a test suite concurrency problem

kurtmckee:
TIL isort writes `.isorted` files...and that when those are written to the test directory while running tox environments in parallel, pytest's test finder can experience a race condition where it can list a `.isorted` temporary file but then cannot read it. Neat!

[relevant isort code](https://github.com/PyCQA/isort/blob/7de182933fd50e04a7c47cc8be75a6547754b19c/isort/api.py#L349-L350)
```python
def _tmp_file(source_file: File) -> Path:
    return source_file.path.with_suffix(source_file.path.suffix + ".isorted")
```

This PR:

* Fixes a missing `py312` environment dependency, likely overlooked due to inconsistent sorting of the `py{37,38,...}` vs `py{312,311,...}` sorting of tox environment dependencies.
* Add `lint` as a dependency for running `py{XY}` environments.

### DIFF
--- a/changelog.d/20240405_154552_kurtmckee_prevent_isorted_files_causing_test_failures.rst
+++ b/changelog.d/20240405_154552_kurtmckee_prevent_isorted_files_causing_test_failures.rst
@@ -1,0 +1,5 @@
+Development
+~~~~~~~~~~~
+
+- Fix concurrency problems in the test suite
+  caused by isort's `.isorted` temporary files. (:pr:`NUMBER`)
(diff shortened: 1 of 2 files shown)


sirosen (approved):
Good catch on `py312` and :+1: for the resort in tox.ini .
As for `isort`... I didn't see that coming but now that I know about it I can't help but think that it's not-unreasonable. (I'm not sure I'd implement it that way, but who am I to judge, right?)

derek-globus (approved):

PR #977, opened 2024-04-16
LocalServerLoginFlowManager

aaschaer:
Shortcut: https://app.shortcut.com/globus/story/30774/sdk-localwebserverloginflowmanager

Mostly the same as in the CLI. Notable change is the ability to specify a `server_address` or `html_template` when instantiating a `LocalServerLoginFlowManager` for customization purposes.

~One thing notably missing from this that's present in various levels across the CLI/GCP/FAIR login is the ability to fallback to a link login in known failure cases around the browser/server. We could maybe detect those cases here and raise an error, but for an end user that's probably not very helpful? I'm wondering if we need something like a `FlexibleLoginFlowManager` that tries to use a browser and server if it can, but falls back to a link and copy and paste if it can't.~
I've added LocalServerLoginFlowError to be raised in these cases, so hopefully the CLI can do something like
```
try:
    login_manager = LocalServerLoginFlowManager(...)
    login_manager.run_login_flow(...)
except LocalServerLoginFLowError:
    login_manager = CommandLineLoginFlowManager(...)
    login_manager.run_login_flow(...)
```

There is also a bit of a testing gap around the use of `webbrowser`, but I think that's fine.

----
📚 Documentation preview 📚: https://globus-sdk-python--977.org.readthedocs.build/en/977/

### DIFF
--- a/src/globus_sdk/experimental/login_flow_manager/_local_server.py
+++ b/src/globus_sdk/experimental/login_flow_manager/_local_server.py
@@ -15,47 +15,17 @@ from http.server import BaseHTTPRequestHandler, HTTPServer
 from string import Template
 from urllib.parse import parse_qsl, urlparse
 
-DEFAULT_HTML_TEMPLATE = Template(
-    """
-<!DOCTYPE html>
-<html lang="en-US">
-<head>
-  <meta charset="utf-8">
-  <meta http-equiv="x-ua-compatible" content="ie=edge">
-  <title>Globus Login</title>
-  <style type="text/css" media="screen">
-    html { font: 75% "Helvetica Neue","Arial","Helvetica",sans-serif }
-    html, body { display: block; margin: 0; padding: 0 }
-    a { color: #5783a6; text-decoration: none; }
-    a img { border: none; }
-    header { background: #2e5793; }
-    main { padding: 25px 0 50px; }
-    main h1 { border-bottom: solid 1px #aaa; font-size: 233.33%;
-              font-weight: normal; }
-    main img { display: block; margin: 0 auto; max-width: 100%; height: auto; }
-    main p { color: #333; font-size: 116.67%; max-width: 560px;
-             margin: 1em auto; line-height: 150%; }
-    header > div, main, footer { display: block; max-width: 980px;
-                                 margin: 0 auto; }
-  </style>
-</head>
-<body>
-  <header><div><a href="https://www.globus.org" title="Go to Globus.org Home">
-    <img alt="Globus" width="215" height="64"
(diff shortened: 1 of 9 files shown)


src/globus_sdk/experimental/login_flow_manager/_local_server.py
```diff
@@ -0,0 +1,140 @@
...
+        self._auth_code_queue.put(excval)
+
+    def return_code(self, code: str | BaseException) -> None:
+        self._auth_code_queue.put_nowait(code)
+
+    def wait_for_code(self) -> str | BaseException:
+        # timeout is set as a workaround for handling control-c interrupt on Windows
+        # https://docs.python.org/3/library/queue.html#queue.Queue.get
+        try:
+            return self._auth_code_queue.get(block=True, timeout=3600)
```
kurtmckee, line 102: I _think_ that this means that `CTRL+C` will not be processed for 1 hour on Windows platforms, which is not a desirable outcome.

There may be other mechanisms that can block without ignoring `CTRL+C` on Windows. I'm specifically considering `threading.Event().wait()` as a possibility for the main thread, so I'll test that on Windows 10 shortly with a few Python versions.

kurtmckee (changes requested):

src/globus_sdk/experimental/login_flow_manager/_local_server.py
kurtmckee: Python 3.12 on Windows 10 blocks until the full `timeout` period has passed. _Then_ it processes the `CTRL+C` it received.

Testing alternative implementations now.

src/globus_sdk/experimental/login_flow_manager/_local_server.py
```diff
@@ -0,0 +1,140 @@
...
+import socket
+import sys
+import typing as t
+from http.server import BaseHTTPRequestHandler, HTTPServer
+from string import Template
+from urllib.parse import parse_qsl, urlparse
+
+DEFAULT_HTML_TEMPLATE = Template(
+    """
+<!DOCTYPE html>
```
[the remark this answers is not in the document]
aaschaer: Looks like this adds a new dependency, but only for specific versions of python, how do I set that up correctly?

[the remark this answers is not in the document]
kurtmckee: In `setup.cfg`, line 35, you'll find an example of the syntax. Just add a line for `importlib_resources` that relies on Python < 3.9. :+1:

src/globus_sdk/experimental/login_flow_manager/_local_server.py
```diff
@@ -0,0 +1,140 @@
...
+        self.html_template = html_template
+        self._auth_code_queue: queue.Queue[str | BaseException] = queue.Queue()
+
+    def handle_error(
+        self,
+        request: socket.socket | tuple[bytes, socket.socket],
+        client_address: t.Any,
+    ) -> None:
+        _, excval, _ = sys.exc_info()
+        assert excval is not None
```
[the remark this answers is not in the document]
aaschaer: Yes, removing it gives
Argument 1 to "put" of "Queue" has incompatible type "BaseException | None"; expected "str | BaseException"

src/globus_sdk/experimental/login_flow_manager/_local_server.py
sirosen: I spoke with Kurt about this to get some info and learn more -- I wrote some of the early versions of this for the CLI and have definitely touched on this Windows behavior; and I think I've learned that my understanding was not quite correct.

I recommend that we follow the above suggestion to loop, doing short (e.g., 1 second) sleeps.

We discussed and also have the following suggestions:
1. Gate this to Windows only. That way, Linux, macOS, etc get good behavior (a long block on the queue with working Ctrl+C), but Windows users get the best behavior we can offer them. This will require two codepaths where there was only one before, but c'est la vie.
2. Do not trust `time.sleep` to be precise. Calculate a "deadline" and wait until the deadline is reached, rather than doing a fixed count of `time.sleep` calls.

Regarding (2), something like this would work:
```python
deadline = time.time() + 3600
while time.time() < deadline: ...
```

src/globus_sdk/experimental/tokenstorage_v2
kurtmckee, line 1: I think this may not have been intended for inclusion.

aaschaer: whoops! thanks

src/globus_sdk/experimental/login_flow_manager/_local_server.py
```diff
@@ -0,0 +1,122 @@
...
+                try:
+                    return self._auth_code_queue.get()
+                except queue.Empty:
+                    time.sleep(1)
+            raise LocalServerError("Login timed out. Please try again.")
+        else:
+            try:
+                return self._auth_code_queue.get(block=True, timeout=3600)
+            except queue.Empty as exc:
+                raise LocalServerError("Login timed out. Please try again.") from exc
```
kurtmckee, line 86: I think this gives Windows users a much better experience for `CTRL+C` behavior! Thank you!

My additional suggestion here is basically "`raise` from the same location, regardless of platform".

```suggestion
        else:
            try:
                return self._auth_code_queue.get(block=True, timeout=3600)
            except queue.Empty as exc:
                pass
        raise LocalServerError("Login timed out. Please try again.")
```

kurtmckee (approved):

kurtmckee (approved):

PR #978, opened 2024-04-22
Validating Storage Adapter

derek-globus:
## What?

1. A new experimental storage adapter (``ValidatingStorageAdapter``) which validates that
  identity is maintained and scope requirements are met on token storage/retrieval.
  (:pr:`NUMBER`)
1. Move `tests/common.py` file to `tests/common/` directory & added some common consent-based constructs.

## Why?

1. Isolates token validation into the StorageAdapter interface so that upcoming GlobusApp constructs don't need to be concerned with validating tokens.
1. Better isolation of re-usable functionaly 

----
📚 Documentation preview 📚: https://globus-sdk-python--978.org.readthedocs.build/en/978/


src/globus_sdk/experimental/globus_app/_validating_storage_adapater.py
```diff
@@ -0,0 +1,213 @@
...
+            requirements to validate on token storage/retrieval.
+        :param consent_client: An AuthClient to be used for consent polling. If omitted,
+            dependent scope requirements are ignored during validation.
+        """
+        self._storage_adapter = storage_adapter
+        self.scope_requirements = scope_requirements
+        self._consent_client = consent_client
+
+        self.identity_id = self._lookup_stored_identity_id()
+        self._cached_consent_forest = self._poll_and_cache_consents()
```
aaschaer, line 60: Would it make sense to store the cached consent forest similarly to the identity id? Otherwise I think the CLI would be making a `get_consents` call for every command that has dependent scope requirements

derek-globus: Talked offline; yes but not in this unit of work.

aaschaer (approved):

tests/common.py
```diff
@@ -1,150 +0,0 @@
-"""
```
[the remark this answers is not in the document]
sirosen: Discussed; things are complicated, not worrying much.

PR #979, opened 2024-04-22
Add `validate_flow` operation to FlowsClient

ada-globus:
I added a quick copyedit commit to normalize capitalization of "flow" and "run", which appeared to be inconsistent. It's not per se part of the scope, but was very fast and relevant to matching the new docstring for `validate_flow`. If it's a significant barrier to getting this merged, I can break it into a separate PR.

----
📚 Documentation preview 📚: https://globus-sdk-python--979.org.readthedocs.build/en/979/

### DIFF
--- a/src/globus_sdk/services/flows/client.py
+++ b/src/globus_sdk/services/flows/client.py
@@ -46,14 +46,14 @@ class FlowsClient(client.BaseClient):
         additional_fields: dict[str, t.Any] | None = None,
     ) -> GlobusHTTPResponse:
         """
-        Create a Flow
+        Create a flow
 
         :param title: A non-unique, human-friendly name used for displaying the
             flow to end users. (1 - 128 characters)
         :param definition: JSON object specifying flows states and execution order. For
             a more detailed explanation of the flow definition, see
             `Authoring Flows <https://docs.globus.org/api/flows/authoring-flows>`_
-        :param input_schema: A JSON Schema to which Flow Invocation input must conform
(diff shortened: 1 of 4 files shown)


src/globus_sdk/services/flows/client.py
```diff
@@ -46,14 +46,14 @@ def create_flow(
         additional_fields: dict[str, t.Any] | None = None,
     ) -> GlobusHTTPResponse:
         """
-        Create a Flow
+        Create a flow
```
derek-globus, line 49: 😬

derek-globus (approved):

src/globus_sdk/services/flows/client.py
sirosen: We need a style-guide for this. We started using emphasis (boldface, IIRC?) in some of the docs for `flow`, `run`, `timer`, and other proper nouns which are easily confused with service names.

sirosen (commented):
I'd like us to hold this change until there's a Flows service release to put that validate route out there. When the SDK or CLI "run ahead", it produces a slightly strange situation if we release.

src/globus_sdk/services/flows/client.py
ada-globus: Yeah, I think the way I'd phrase it is roughly:

> Entities (e.g., flow, run, action, action provider) are lowercase and first appearance is set off as a "term" in whatever paradigm a given document uses; for the docs site, we use bold

Is there such a paradigm—or related prior-art—in the SDK docs? If so, happy to make consistent!

PR #980, opened 2024-04-25
TokenStorage (v2)

aaschaer:
Shortcut: https://app.shortcut.com/globus/story/33154/sdk-storageadapterv2

Change Summary (all in experimental):
* Adds `TokenStorage` which expands upon `StorageAdapter` by adding `namespace` support and requires implementing classes to support `remove_tokens_for_resource_server`. It also allows passing token data as a `dict` in addition to a `OAuthTokenResponse` and removes `on_refresh` as that was only ever an alias for `store`
* Adds `TokenData` which is a data class for tokens and metadata, notably including `identity_id` which is needed by `ValidatingTokenStorage`
* Adds `FileTokenStorage` which is nearly much identical to `FileAdapter` beyond implementing `StorageAdapterV2` instead of `StorageAdapterV2`
* Adds `MemoryTokenStorage` which remains fairly simple.
* Adds `JSONTokenStorage` which now stores data under namespaced keys and has logic for migrating storage from a `SimpleJSONFileAdapter`
* Adds `SQLiteTokenStorage` which is very similar `SQLiteAdapter` beyond implementing `StorageAdapterV2`. Both classes should even be able to use the same storage without causing issues
* Changes `ValidatingStorageAdapater` to `ValidatingTokenStorage` which now implements `TokenStorage` instead of `StorageAdapter`
* Removes `IdentifiedOAuthTokenResponse` as its functionality is now covered by `TokenData`

Notes:
- the default namespace is "DEFAULT" instead of `None` as originally spec'd this was to not require data migration for `SQLiteTokenStorage` to use storage previously used by `SQLiteAdapter`
- `namespace` is defined at class initialization rather than passed to each function as originally spec'd. This was to minimize needed changes for `SQLiteAdapter` and better conform with `GlobusApp` which will also have one `namespace` per instance.
-  Its not clear to me that adding namespace support to `MemoryTokenStorage` changed anything since initializing another `MemoryAdapter` already partitioned the data in memory. Maybe that value should just be ignored?

----
📚 Documentation preview 📚: https://globus-sdk-python--980.org.readthedocs.build/en/980/


src/globus_sdk/experimental/tokenstorage_v2/base.py
```diff
@@ -0,0 +1,129 @@
...
+
+import abc
+import contextlib
+import os
+import typing as t
+
+from globus_sdk.services.auth import OAuthTokenResponse
+
+
+class StorageAdapterV2(metaclass=abc.ABCMeta):
```
derek-globus, line 11: Would it make sense to rename it to something more specific with this change?

`TokenStorageAdapterV2` or `TokenStorage` or something?

The interface `StorageAdapter` is was more generic in name than the actual class's scope. It implies with such a generic name that we're not gonna be storing anything else (e.g. consents)

src/globus_sdk/experimental/tokenstorage_v2/base.py
```diff
@@ -0,0 +1,129 @@
...
+
+        Either returns a dict with the access token, refresh token (optional), and
+        expiration time, or returns ``None``, indicating that there was no data for that
+        resource server.
+
+        :param resource_server: The resource_server string to get tokens for
+        """
+
+    @abc.abstractmethod
+    def remove_tokens_for_resource_server(self, resource_server: str) -> bool:
```
derek-globus, line 54: I know this name is coming from the current sqliteadapater but I wonder, could we actually just change this? It's way more verbose than the other method names.

* "store" doesn't mention tokens at all
* "get" mentions tokens but not that it's for a resource server
* "remove" mentions tokens and that it's for a resource server

I guess store isn't bounded to a particular resource server and get/remove are so maybe `remove_token_data(resource_server: str)` to mirror `get_token_data(resource_server: str)`?

src/globus_sdk/experimental/tokenstorage_v2/base.py
```diff
@@ -0,0 +1,129 @@
...
+        """
+        Remove all tokens for a resource server from the underlying storage under the
+        current namespace.
+
+        Returns True if token data was deleted, False if none was found to delete.
+
+        :param resource_server: The resource server string to remove tokens for
+        """
+
+    def normalize_token_response(
```
derek-globus, line 64: This doesn't feel like it should be exposed on the public interface for every storage adapter.

It's a convenient thing for implementing adapters to have access to, but maybe as `_normalize_token_response` or even as a separate method which can be imported?

src/globus_sdk/experimental/tokenstorage_v2/file_adapters.py
```diff
@@ -0,0 +1,171 @@
...
+
+        # 1.0 data was stored under a "by_rs" key without namespaces, to upgrade we
+        # move everything under the "DEFAULT" key and remove the "by_rs" key.
+        if format_version == "1.0":
+            read_data = {
+                "data": {
+                    "DEFAULT": read_data["by_rs"],
+                },
+                "format_version": self.format_version,
+                "globus-sdk.version": __version__,
```
derek-globus, line 67: Why the difference in delimiter here? `_` vs `-`

src/globus_sdk/experimental/tokenstorage_v2/sqlite_adapter.py
```diff
@@ -0,0 +1,265 @@
...
+    resource_server VARCHAR NOT NULL,
+    token_data_json VARCHAR NOT NULL,
+    PRIMARY KEY (namespace, resource_server)
+);
+CREATE TABLE sdk_storage_adapter_internal (
+    attribute VARCHAR NOT NULL,
+    value VARCHAR NOT NULL,
+    PRIMARY KEY (attribute)
+);
+            """
```
derek-globus, line 79:
```suggestion
              textwrap.dedent(
                """
                CREATE TABLE config_storage (
                    namespace VARCHAR NOT NULL,
                    config_name VARCHAR NOT NULL,
                    config_data_json VARCHAR NOT NULL,
                    PRIMARY KEY (namespace, config_name)
                );
                CREATE TABLE token_storage (
                    namespace VARCHAR NOT NULL,
                    resource_server VARCHAR NOT NULL,
                    token_data_json VARCHAR NOT NULL,
                    PRIMARY KEY (namespace, resource_server)
                );
                CREATE TABLE sdk_storage_adapter_internal (
                    attribute VARCHAR NOT NULL,
                    value VARCHAR NOT NULL,
                    PRIMARY KEY (attribute)
                );
                """
              )
```
```

File: src/globus_sdk/experimental/tokenstorage_v2/base.py (hunk 2 above)
Comment (sirosen) [reply src/globus_sdk/experimental/tokenstorage_v2/base.py:54]: I would like us to reevaluate all of the names in play, for the record. That goes for the class name as well, per one of your other comments.

File: src/globus_sdk/experimental/tokenstorage_v2/file_adapters.py (hunk 4 above)
Comment (aaschaer) [reply src/globus_sdk/experimental/tokenstorage_v2/file_adapters.py]: I think this is the only comment that didn't get addressed in my last commit/rebase.

Essentially I just wanted to keep the value used by `SimpleJSONStorageAdapter` so there was one less thing for users to worry about when migrating.

As for why it was like this originally :shrug:

File: src/globus_sdk/experimental/tokenstorage_v2/base.py (hunk 1 above)
Comment (aaschaer) [reply src/globus_sdk/experimental/tokenstorage_v2/base.py]: `TokenStorage` seemed the most accurate to me

File: src/globus_sdk/experimental/tokenstorage_v2/base.py (hunk 2 above)
Comment (aaschaer) [reply src/globus_sdk/experimental/tokenstorage_v2/base.py]: I went with

`store_response`
`store_token_data_by_resource_server`
`get_token_data_by_resource_server`
`get_token_data`
`remove_token_data`

Open to suggestions for changes, but it should at least be more uniform now

File: src/globus_sdk/experimental/tokenstorage_v2/base.py (hunk 3 above)
Comment (aaschaer) [reply src/globus_sdk/experimental/tokenstorage_v2/base.py]: This has been removed

**Review by derek-globus — COMMENTED** [review]
Haven't finished reviewing; posting comments now though for conversation.

File: src/globus_sdk/experimental/tokenstorage_v2/base.py [hunk 6]
```diff
@@ -0,0 +1,139 @@
...
+        :param token_response: An ``OAuthTokenResponse`` from an authentication flow
+        """
+        token_data_by_resource_server = {}
+
+        # get identity_id from id_token if available
+        if token_response.get("id_token"):
+            decoded_id_token = token_response.decode_id_token()
+            identity_id = decoded_id_token["sub"]
+        else:
+            identity_id = None
```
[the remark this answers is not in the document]
aaschaer: I like changing it to `store_token_response`

derek-globus (approved):
I like the place this ended up.

Thanks again for iterating on it to refine it to this point!

PR #985, opened 2024-05-17
add AuthorizerFactory and implementations

aaschaer:
Shortcut: https://app.shortcut.com/globus/story/30767/sdk-authorizerfactory

Change Summary (all in experimental):
- adds ``AuthorizerFactory`` an interface for getting authorizers from a ``ValidatingTokenStorage``
- adds ``AccessTokenAuthorizerFactory``, ``RefreshTokenAuthorizerFactory``, and ``ClientCredentialsAuthorizerFactory`` which implement ``AuthorizerFactory`` and create ``AccessTokenAuthorizer`` ``RefreshTokenAuthorizer`` and ``ClientCredentialsAuthorizer`` respectively

----
📚 Documentation preview 📚: https://globus-sdk-python--985.org.readthedocs.build/en/985/

### DIFF
--- a/src/globus_sdk/experimental/globus_app/authorizer_factory.py
+++ b/src/globus_sdk/experimental/globus_app/authorizer_factory.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import abc
+import typing as t
 
 from globus_sdk import AuthLoginClient, ConfidentialAppAuthClient
 from globus_sdk.authorizers import (
@@ -15,8 +16,13 @@ from globus_sdk.services.auth import OAuthTokenResponse
 from ._validating_token_storage import ValidatingTokenStorage
 from .errors import MissingTokensError
 
+GA = t.TypeVar("GA", bound=GlobusAuthorizer)
 
-class AuthorizerFactory(metaclass=abc.ABCMeta):
+
+class AuthorizerFactory(
+    t.Generic[GA],
+    metaclass=abc.ABCMeta,
+):
     """
     An ``AuthorizerFactory`` is an interface for getting some class of
     ``GlobusAuthorizer`` from a ``ValidatingTokenStorage`` that meets the
@@ -33,7 +39,7 @@ class AuthorizerFactory(metaclass=abc.ABCMeta):
         constructed authorizers will meet and accessing underlying token storage
         """
         self.token_storage = token_storage
-        self._authorizer_cache: dict[str, GlobusAuthorizer] = {}
+        self._authorizer_cache: dict[str, GA] = {}
 
     def _get_token_data_or_error(self, resource_server: str) -> TokenData:
         token_data = self.token_storage.get_token_data(resource_server)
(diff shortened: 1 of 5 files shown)


src/globus_sdk/experimental/globus_app/authorizer_factory.py
```diff
@@ -0,0 +1,198 @@
...
+        """
+        if resource_server in self._authorizer_cache:
+            return self._authorizer_cache[resource_server]
+
+        new_authorizer = self._make_authorizer(resource_server)
+        self._authorizer_cache[resource_server] = new_authorizer
+        return new_authorizer
+
+    @abc.abstractmethod
+    def _make_authorizer(self, resource_server: str) -> GlobusAuthorizer:
```
sirosen, line 75: I don't think it's essential, but this interface being the abstract part makes me think that this should be a generic class wherein `GlobusAuthorizer` is replaced with a bounded type var.

The construction would come out something like...
```python
GA = t.TypeVar("GA", bound=GlobusAuthorizer)

class AuthorizerFactory(t.Generic[GA]):
    def get_authorizer(self, resource_server: str) -> GA: ...
    def _make_authorizer(self, resource_server: str) -> GA: ...

class AccessTokenAuthorizerFactory(AuthorizerFactory[AccessTokenAuthorizer]): ...
```

I don't think it makes a big difference because the authorizer classes provide a uniform interface, but I think it's a closer match for your intent, in which a subclass is declared for each subtype.

src/globus_sdk/experimental/globus_app/authorizer_factory.py
```diff
@@ -0,0 +1,198 @@
...
+        self._authorizer_cache: dict[str, GlobusAuthorizer] = {}
+
+    def _get_token_data_or_error(self, resource_server: str) -> TokenData:
+        token_data = self.token_storage.get_token_data(resource_server)
+        if token_data is None:
+            raise MissingTokensError(f"No token data for {resource_server}")
+
+        return token_data
+
+    def store_token_response(self, token_res: OAuthTokenResponse) -> None:
```
sirosen, line 45: What are the callers for this method?

I'm noticing that the renewing authorizers bypass it (presumably, so that they don't blow the cache), so I'm guessing this is used as part of login flows.

Can we distinguish the name from the token storage method, so that the two are not likely to be confused?

aaschaer: This will probably only be called by a `GlobusApp` after its had a user complete a login flow (or someone trying to do `GlobusApp` like things on their own for some reason)

`store_token_response_and_clear_cache` seems reasonable to me?

src/globus_sdk/experimental/globus_app/authorizer_factory.py
aaschaer: Yeah that's more what I was going for, thanks!

sirosen (approved):

PR #986, opened 2024-05-30
Globus app

aaschaer:
Shortcut: https://app.shortcut.com/globus/story/30768/sdk-globusapp-constructs

Change Summary (all in experimental:
* Adds abstract base class `GlobusApp` which defines `add_scope_requirements`, `run_login_flow`, and `get_authorizer` as an interface for getting tokens that meet scope requirements
* Adds `UserApp` which implements `GlobusApp` and has a `LoginFlowManager` for running interactive login flows.
* Adds `ClientApp` which implements `GlobusApp`
* Adds a first pass `GlobusAppConfig`, which has its own story https://app.shortcut.com/globus/story/30772/sdk-globusappconfig. I'm not sure to what extent we'll want to expand this so I'm leaving that story as is for now. I mostly added it now because `refresh_tokens` significantly changes behavior of `UserApp`, and being able to configure a `MemoryTokenStorage` is useful for testing.
* Adds a `Scope.merge_scopes` method for scope de-duplication. This isn't as comprehensive as what was originally specc'd (https://app.shortcut.com/globus/story/33823/scope-de-duplication) but avoids the issue of how to hash non-static Scopes

Notes:
* evidently I only didn't hit typing issues with generic abstract base classes because I hadn't rebased as recently as I thought. I added a few `type: ignore` comments in places I feel should be correct but are still raising mypy errors, but if anyone knows a better way to handle that lmk
* I chose the filename for the default `JSONTokenStorage` somewhat arbitrarily, happy to change it if desired

----
📚 Documentation preview 📚: https://globus-sdk-python--986.org.readthedocs.build/en/986/


src/globus_sdk/scopes/representation.py
```diff
@@ -72,6 +73,39 @@ def parse(scope_string: str) -> list[Scope]:
...
+        for scope_string, scopes in scope_data.items():
+            combined_scope = Scope(scope_string)
+            combined_scope.optional = reduce(
+                lambda x, y: x and y, [s.optional for s in scopes]
+            )
+            combined_scope.dependencies = reduce(
+                Scope.merge_scopes, [s.dependencies for s in scopes]
+            )
+            ret.append(combined_scope)
+        return ret
```
sirosen, line 107: I've had a couple of short conversations about this, but I'll drop it here to consolidate:

I think we should swap this out for stringifying each list, joining into a one string with `" ".join(...)`, and then re-parsing the result with `Scope.parse()`. There are three main cases I have in mind, which are:

---

1. Cycles. `foo[bar]` combined with `bar[foo]`
2. `optional` and deduplication. `foo[bar]` combined with `foo[*bar]`
3. Chains. `foo[bar]` combined with `bar[baz]`.

Cycles are invalid, so catching them is nice (but not essential).

`optional` handling with deduplication needs to be handled in both code locations. So I prefer to have only one implementation.

Chains are interesting because `foo[bar] bar[baz]` is equivalent to a variety of other strings, including `foo[bar[baz]] bar`. I believe we would render it out as `foo[bar[baz]] bar[baz]`, which is the most verbose form but also clearest / most explicit.

---

Of these three, only (2) provides a really compelling case for taking this approach, but I also think it's simpler to maintain, which is a separate benefit.

aaschaer: For now I'm treating whatever `parse` does as correct. I'm not sure that is in fact the case, but I don't even understand what auth is doing to compare it to, so I'm just going to let it be for now.

src/globus_sdk/experimental/globus_app/globus_app.py
```diff
@@ -0,0 +1,429 @@
...
+        auth_rs = str(AuthClient.resource_server)
+        openid_scope = Scope(AuthScopes.openid)
+        if scope_requirements:
+            if auth_rs not in scope_requirements:
+                scope_requirements[auth_rs] = [openid_scope]
+            else:
+                scope_requirements[auth_rs] = Scope.merge_scopes(
+                    scope_requirements[auth_rs], [openid_scope]
+                )
+            self._scope_requirements = scope_requirements
```
derek-globus, line 138: Could we make scope_requirements public?

Keeping it private removes the ability for users to read what scope requirements have actually been defined on the App.

derek-globus: In a similar vein, I'd recommend that we do scope merging on scope read (when initiating a login flow) not write (when a scope requirement is being added).

Scope deduplication (not adding the same exact scope twice) could make sense; but merging has the potential to mutate the change data structure from under the user unexpected ways.

src/globus_sdk/experimental/globus_app/globus_app.py
```diff
@@ -0,0 +1,429 @@
...
+        """
+
+    def _auth_params_with_required_scopes(
+        self,
+        auth_params: GlobusAuthorizationParameters | None = None,
+    ) -> GlobusAuthorizationParameters:
+        """
+        Either make a new GlobusAuthorizationParameters with this app's required scopes
+        or combine this app's required scopes with given auth_params.
+        """
```
derek-globus, line 192: Ooh! I like this shared functionality line a lot!

src/globus_sdk/experimental/globus_app/globus_app.py
```diff
@@ -0,0 +1,429 @@
...
+                self._scope_requirements[resource_server] = scopes
+
+            else:
+                self._scope_requirements[resource_server] = Scope.merge_scopes(
+                    self._scope_requirements[resource_server], scopes
+                )
+
+        # update the app's ValidatingTokenStorage's scope requirements to enforce the
+        # updated scope requirements
+        self._validating_token_storage.scope_requirements = self._scope_requirements
```
derek-globus, line 242: This shouldn't be necessary. Scope requirements are passed by reference; so an update to them here will mean an update to them there.

src/globus_sdk/experimental/globus_app/globus_app.py
```diff
@@ -0,0 +1,429 @@
...
+        AccessTokenAuthorizerFactory | RefreshTokenAuthorizerFactory
+    )
+
+    def __init__(
+        self,
+        app_name: str,
+        *,
+        scope_requirements: dict[str, list[Scope]] | None = None,
+        client_id: UUIDLike | None = None,
+        client_secret: str | None = None,
```
derek-globus, line 275: We should also accept a pre-instantiated login client here (mutexed with these client id/secret values) to make it easier for users to plug this into their existing applications that either (1) re-use the login client in multiple places or (2) extend and customize a particular login client.

(Comment applies to ClientApp as well)

src/globus_sdk/experimental/globus_app/globus_app.py
```diff
@@ -0,0 +1,429 @@
...
+    """
+
+    _login_client: ConfidentialAppAuthClient
+    _authorizer_factory: ClientCredentialsAuthorizerFactory  # type:ignore
+
+    def __init__(
+        self,
+        app_name: str,
+        client_id: UUIDLike,
+        client_secret: str,
```
derek-globus, line 377: Given that these could become optional in the future (with environment variables) can we move them to the kwargs & validate their presence in runtime code instead?

src/globus_sdk/experimental/globus_app/globus_app.py
```diff
@@ -0,0 +1,429 @@
...
+
+        # either get config's LoginFlowManager
+        # or make a default CommandLineLoginFlowManager
+        if self.config.login_flow_manager:
+            self._login_flow_manager = self.config.login_flow_manager
+        else:
+            self._login_flow_manager = CommandLineLoginFlowManager(
+                self._login_client,
+                refresh_tokens=self.config.refresh_tokens,
+            )
```
derek-globus, line 309: The config interface for login flow manager doesn't really expose the ability to override default behavior in a somewhat confusing way:

* A GlobusApp will always construct a login client (which it holds privately).
* If `config.login_flow_manager` was omitted, that login client will be passed to the default login flow manager (CommandLineLoginFlowManager)
* `config.login_flow_manager` can be overridden but cannot have access to the login client created by the app.

So, for instance, if I write my own custom `CarrierPigeonLoginFlowManager` which communicates information back and forth by carrier pigeons there's no way to have the login client associated my globusapp associated with that loginflowmanager.

Not sure if there's a simple fix to align this overrideability (I also see that refreh_token configuration comes through only for the CommandLineLoginFlowManager). Maybe we could:
* Accept a class instead of an actual loginflowmanager in the config?
* Accept a provider which takes a pre-canned set of a parameters & create a loginflowmanager?
* Make the login client association be something that happens post-instantiation for a `LoginFlowManager`?

derek-globus (changes requested):

src/globus_sdk/experimental/globus_app/globus_app.py
aaschaer: For some reason I thought client environment variables would only be for a `UserApp`, but not sure where that idea came from

src/globus_sdk/experimental/globus_app/globus_app.py
aaschaer: I'll take a pass at improving this, but might need to discuss further.

I also want to note that `CarrierPigeonLoginFlowManager` made me smile ^^

src/globus_sdk/experimental/globus_app/globus_app.py
```diff
@@ -0,0 +1,429 @@
...
+    """
+    if sys.platform == "win32":
+        # try to get the app data dir, preferring the local appdata
+        datadir = os.getenv("LOCALAPPDATA", os.getenv("APPDATA"))
+        if not datadir:
+            home = os.path.expanduser("~")
+            datadir = os.path.join(home, "AppData", "Local")
+        return os.path.join(datadir, "globus", app_name, "tokens.json")
+    else:
+        return os.path.expanduser(f"~/.globus/{app_name}/tokens.json")
```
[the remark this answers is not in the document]
derek-globus: I'm happy with that (and similar change for the windows variety)

src/globus_sdk/experimental/globus_app/globus_app.py
derek-globus: Sure, I'm happy to jump on a huddle if it'd be easier to try to find a path through this synchronously; just ping me. 

`CarrierPigeonLoginFlowManager` is great for those times when logins as happening just too dang fast!

src/globus_sdk/experimental/globus_app/globus_app.py
```diff
@@ -0,0 +1,460 @@
...
+
+    def __init__(
+        self,
+        app_name: str,
+        *,
+        login_client: AuthLoginClient | None = None,
+        client_id: UUIDLike | None = None,
+        client_secret: str | None = None,
+        scope_requirements: dict[str, list[Scope]] | None = None,
+        config: GlobusAppConfig | None = None,
```
derek-globus, line 106: With config being immutable now, we should be able to set the default in the constructor.

```suggestion
        config: GlobusAppConfig  = GlobusAppConfig()
```

src/globus_sdk/experimental/globus_app/globus_app.py
```diff
@@ -0,0 +1,460 @@
...
+        ``request_refresh_tokens``.
+        If not given the default behavior will depend on the type of ``GlobusApp``.
+    :param token_storage: a ``TokenStorage`` instance or class that will
+        be used for storing token data. If not passed a ``JSONTokenStorage``
+        will be used.
+    :param request_refresh_tokens: If True, the ``GlobusApp`` will request refresh
+        tokens for long lived access.
+    """
+
+    login_flow_manager: LoginFlowManager | type[LoginFlowManager] | None = None
```
derek-globus, line 75: This is an elegant solution to my initial thread.

I'm very slightly hesitant about having the constructor interface be mandated like this but I can live with it. If a user is subclassing their own LoginFlowManager, they're 90% going to be pre-instantiating it notgiving it to us to instantiate it.

derek-globus (approved):
2 minor comments about default location & type hints but overall looks very good!

derek-globus (approved):

PR #988, opened 2024-06-06
Document how to manage Globus SDK warnings

kurtmckee:
Documentation
-------------

* Document how to manage Globus SDK warnings.

----
📚 Documentation preview 📚: https://globus-sdk-python--988.org.readthedocs.build/en/988/

### DIFF
--- a/changelog.d/20240606_181744_kurtmckee_howto_disable_v4_warnings_sc_29682.rst
+++ b/changelog.d/20240606_181744_kurtmckee_howto_disable_v4_warnings_sc_29682.rst
@@ -1,0 +1,4 @@
+Documentation
+~~~~~~~~~~~~~
+
+- Document how to manage Globus SDK warnings. (:pr:`NUMBER`)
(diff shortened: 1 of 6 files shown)


kurtmckee:
I figured testing docs were more appropriate since that's where developers would encounter it, and I anticipated that a see-also link would get developers where they needed to be. It could go in the exceptions docs, though.

Let me know if there's a change request for this. :grin:

sirosen:
I think that we should move it, yes.

Keep in mind that the testing docs are these:
https://globus-sdk-python.readthedocs.io/en/stable/testing/

So the section title is `Globus SDK _testing`, and we have this listed in the sidebar as unstable, alongside `experimental`, to indicate that this is part of a space of components with APIs that do not necessarily follow our normal backwards-compatibility policy.
(If `experimental` had predated `_testing` and the docs thereof, we probably would have structured this sectioning slightly differently.)

I don't feel strongly about putting it in with the exceptions or which other section we choose, but it's not part of `globus_sdk._testing`.

kurtmckee:
@sirosen Take a gander again when you have a chance!

sirosen (approved):
C'est parfait!

PR #990, opened 2024-06-21
Implement the X-Globus-Client-Info header as a feature of transport objects

sirosen:
## Why

The goal of this PR is to start emitting a `X-Globus-Client-Info` header
which encodes the SDK version, and can easily be expanded to include other
product name/version combinations. (e.g. compute SDK, globus-cli, etc)

`X-Globus-Client-Info` is a recent spec we've put together in collaboration
with our frontend dev team and supersedes the use of a custom User-Agent to
encode app information (`app_name`). We needed some new art in this space as
setting a custom User-Agent is not an appropriate solution for the browser
context.

The `app_name` value and `User-Agent` behavior is not deprecated at this
time. In future SDK versions, we may choose to deprecate that behavior if we
feel it is in our interest to only provide this data via the new header.

## What

- Add a new class to implement basic encoding of the header,
  `GlobusClientInfo`. This class is documented and provides the public
  interface for this component. However, it is only exported at the
  `globus_sdk.transport` layer, as most users will not need to interact with
  it directly.

- Add a private, custom subclass, `DefaultGlobusClientInfo`, which seeds
  `product=python-sdk,version={globus_sdk.__version__}` on init.

- `RequestsTransport` has been enhanced to include
  `self.globus_clientinfo = DefaultGlobusClientInfo()` on init, meaning that
  it seeds and stores this information in a publicly accessible attribute.
  When `RequestsTransport` builds headers for a request, it now also includes
  `self.globus_clientinfo`.

## Notes

- `DefaultGlobusClientInfo` behavior could simply be baked into the base
  `GlobusClientInfo` -- input and debate welcome. There are various
  equally-valid ways of seeding this information. The goals are to ensure that
  this data is set for all SDK calls, that the implementation is easy to
  understand, and that it is easy to extend, modify, and even disable.

- A clear target use-case here is ``globus-cli``. The intent here is that the
  CLI can update such that each client builder adds a call to

    ```python
    newclient.transport.globus_clientinfo.add(
        {"product": "cli", "version": globus_cli.__version__}
    )
    ```

  A better/easier hooking interface could be added in the future, if one can be
  designed based on our experience in that context.

----
📚 Documentation preview 📚: https://globus-sdk-python--990.org.readthedocs.build/en/990/

### DIFF
--- a/src/globus_sdk/transport/_clientinfo.py
+++ b/src/globus_sdk/transport/_clientinfo.py
@@ -2,21 +2,8 @@
 This module models a read-write object representation of the
 X-Globus-ClientInfo header.
 
-The spec for X-Globus-ClientInfo is as follows:
-
-Header Name: X-Globus-Client-Info
-Header Value
-  - A semicolon (;) separated list of client information.
-  - Client information is a comma-separated list of `=` delimited key-value pairs.
-    Well-known values for client-information are:
-        product - A unique identifier of the product.
-        version - Relevant version information for the product.
-  - Based on the above, the characters `;,=` should be considered reserved and
-    should NOT be included in client information values to ensure proper parsing.
-
-Examples:
-  X-Globus-ClientInfo: product=python-sdk,version=3.32.1
-  X-Globus-ClientInfo: product=python-sdk,version=3.32.1;product=cli,version=4.0.0a1
+The spec for X-Globus-ClientInfo is documented, in brief, in the GlobusClientInfo class
+docstring.
 """
 
 from __future__ import annotations
@@ -31,8 +18,41 @@ _RESERVED_CHARS = ";,="
 
 class GlobusClientInfo:
     """
-    An implementation of X-Globus-Client-Info as an object model.
-    """
+    An implementation of X-Globus-Client-Info as an object. This header encodes a
+    mapping of multiple products to versions and potentially other information.
+
+    Values can be added to a clientinfo object via the ``add()`` method.
+
+    .. rubric:: ``X-Globus-Client-Info`` Specification
+
+    Header Name: ``X-Globus-Client-Info``
+
+    Header Value:
+
+    - A semicolon (``;``) separated list of client information.
+
+    - Client information is a comma-separated list of ``=`` delimited key-value pairs.
+      Well-known values for client-information are:
+
+      - ``product``: A unique identifier of the product.
+      - ``version``: Relevant version information for the product.
+
+    - Based on the above, the characters ``;,=`` should be considered reserved and
+      should NOT be included in client information values to ensure proper parsing.
+
+    .. rubric:: Example Headers
+
+    .. code-block:: none
+
+        X-Globus-ClientInfo: product=python-sdk,version=3.32.1
+        X-Globus-ClientInfo: product=python-sdk,version=3.32.1;product=cli,version=4.0.0a1
+
+    .. note::
+
+        The ``GlobusClientInfo`` object is not guaranteed to reject all invalid usages.
+        For example, ``product`` is required to be unique per header, and users are
+        expected to enforce this in their usage.
+    """  # noqa: E501
 
     def __init__(self) -> None:
         self.infos: list[str] = []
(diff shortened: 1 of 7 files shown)


kurtmckee:
CI is failing.

sirosen:
I made a change (as a fixup, so you can't see it in the history) to switch from `._globus_clientinfo` to `.globus_clientinfo` and didn't catch one location. :facepalm: 

I've just fixup'ed my fixup and pushed, so the state _should_ be clean now.

src/globus_sdk/transport/_clientinfo.py
```diff
@@ -0,0 +1,115 @@
...
+    A variant of GlobusClientInfo which always initializes itself to start with
+
+        product=python-sdk,version=...
+
+    using the current package version information.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.add({"product": "python-sdk", "version": __version__})
```
kurtmckee, line 99: It seems like this information should _always_ be tacked on at serialization time.

That is to say: devs who are using the SDK will always have critical SDK client information present.

Therefore I think that there should be no `DefaultGlobusClientInfo` class, but instead the SDK client information should be hard-coded to serialize into the HTTP header (and, probably, be the _first_ client info in the serialized header, in case the remainder of the header is b0rked for any reason and cannot be deserialized).

src/globus_sdk/transport/requests.py
```diff
@@ -114,6 +115,7 @@ def __init__(
         self.verify_ssl = config.get_ssl_verify(verify_ssl)
         self.http_timeout = config.get_http_timeout(http_timeout)
         self._user_agent = self.BASE_USER_AGENT
+        self.globus_clientinfo: GlobusClientInfo = DefaultGlobusClientInfo()
```
kurtmckee, line 118: I would prefer to call this `.globus_client_info` so spell checkers won't argue with us.

src/globus_sdk/transport/requests.py
```diff
@@ -135,7 +137,10 @@ def user_agent(self, value: str) -> None:
 
     @property
     def _headers(self) -> dict[str, str]:
-        return {"Accept": "application/json", "User-Agent": self.user_agent}
+        headers = {"Accept": "application/json", "User-Agent": self.user_agent}
+        if self.globus_clientinfo:
+            headers["X-Globus-ClientInfo"] = self.globus_clientinfo.format()
```
kurtmckee, line 142: I thought this was "X-Globus-Client-Info"?

tests/unit/transport/test_clientinfo.py
```diff
@@ -0,0 +1,105 @@
...
+from globus_sdk import __version__
+from globus_sdk.transport._clientinfo import DefaultGlobusClientInfo, GlobusClientInfo
+
+
+def parse_clientinfo(header):
+    """
+    Sample parser.
+
+    Including this in the testsuite not only validates the mechanical implementation of
+    X-Globus-Client-Info, but also acts as a safety check that we've thought through the
```
kurtmckee, line 12: "X-Globus-Client-Info" here, "X-Globus-ClientInfo" elsewhere. This location reflects my preferred spelling.

kurtmckee (changes requested):

src/globus_sdk/transport/_clientinfo.py
sirosen: Okay, let's merge it into the superclass so that it's part of the base implementation here. I buy a lot of the thinking here.

The place where I want to exercise some caution is around baking it in so deeply that removing, inspecting, or manipulating it requires subclassing the transport class itself. I'd therefore prefer to keep it as structured data, but structured data which is populated by default.

src/globus_sdk/transport/requests.py
sirosen: What if I want to argue with spell checkers?! :stuck_out_tongue_winking_eye: 

That's an easy change for me to get behind.

src/globus_sdk/transport/requests.py
sirosen: Oy vey; good catch! The initial spec was `X-Globus-ClientInfo` but then we found that some browser behaviors are based around the expectation that headers are Title-Kebab-Case, so it changed.

kurtmckee (approved):

PR #993, opened 2024-06-26
Globus app integration

aaschaer:
This PR covers https://app.shortcut.com/globus/story/30906/sdk-support-globusapp-in-the-baseclient, https://app.shortcut.com/globus/story/30771/sdk-support-globusapp-in-all-but-one-sdk-service-clients, and https://app.shortcut.com/globus/story/32979/sdk-support-globusapp-in-gcsclient. This is mostly because linting was happier once all subclasses implemented `default_scope_requirements`

Change summary (all in experimental):
* Add `app` as an initialization parameter to BaseClient and clients things that implement base client (notable exception being `AuthLoginClient`s) and logic for automatically setting up the app's required scopes with the client's default scope requirements and getting an authorizer from the app when making requests
* Add `add_app_scope` and a `TransferClientSpecific` `add_app_data_access_scope` for adding scope requirements to a client's app after initialization.
* `GCSClient` supports a `get_gcs_info` helper that makes an unauthenticated call to `/info` needed for resolving a endpoint domain to a endpint client id. This required adding `ignore_authorizer` option to all `BaseClient` request methods.

Notes:
* I didn't implement `AppRegistrationParams` as it didn't handle the dependent data access use case, and once `add_app_scope` existed, its scope functionality seemed redundant.
* Per conversation with the GCS team, the `client_id` of an endpoint's gcs manager client is the canonical value for scopes and resource server. Previous code asks for the `endpoint_id` so we might want to consider changing that name for v4, although for most users `endpoint_id` is both correct and possibly more clear

----
📚 Documentation preview 📚: https://globus-sdk-python--993.org.readthedocs.build/en/993/


src/globus_sdk/client.py
```diff
@@ -91,16 +102,86 @@ def __init__(
...
+                raise ValueError(
+                    "Unable to use an 'app' with a client with no "
+                    "'resource_server' defined."
+                )
+
+            self._app.add_scope_requirements(
+                {self.resource_server: self.default_scope_requirements}
+            )
+
+    def add_app_scope(self, scope_collection: ScopeCollectionType) -> None:
```
derek-globus, line 150: Since AppRegistrationParams were omitted, could this scope adding function return the client instead of None for chained function calls?

I'd recommend we rename to  to `with_app_scope(...)` to indicate this but thinking about usage like.

```
flows = FlowsClient(app=app)\
   .with_app_scope(FlowsScopes.manage_flows)\
   .with_app_scope(FlowsScopes.run_manage)
```

src/globus_sdk/client.py
```diff
@@ -129,14 +212,21 @@ def get(  # pylint: disable=missing-param-doc
         *,
         query_params: dict[str, t.Any] | None = None,
         headers: dict[str, str] | None = None,
+        ignore_authorizer: bool = False,
```
derek-globus, line 215: Hmm, I can get behind this approach (I assume this is for GCS client id lookup) but I don't love the name. It references an SDK specific concept of an `authorizer` where I think it could convey the same information to a broader audience if instead it used the more generic term `authorization` (referring to both the concept and the oauth2 header).

How do you feel about `skip_authorization`, `omit_authorization`, or even `include_authorization` if we wanted to invert the default to be `True`?

src/globus_sdk/services/transfer/client.py
```diff
@@ -110,6 +110,39 @@ class TransferClient(client.BaseClient):
     transport_class: type[TransferRequestsTransport] = TransferRequestsTransport
     error_class = TransferAPIError
     scopes = TransferScopes
+    default_scope_requirements = [Scope(TransferScopes.all)]
+
+    def add_app_data_access_scope(self, collection_id: UUIDLike) -> None:
```
derek-globus, line 115: Ooh, cool idea. I like it!

src/globus_sdk/services/transfer/client.py
```diff
@@ -110,6 +110,39 @@ class TransferClient(client.BaseClient):
...
+        """
+        Add a dependent ``data_access`` scope for a given ``collection_id`` to this
+        client's ``GlobusApp``. Useful for resolving ``ConsentRequired`` errors
+        when using standard Globus Connect Server mapped collections.
+
+        Raises ``GlobusSDKUsageError`` if this client was not initialized with an app.
+
+        :param collection_id: UUID for a standard Globus Connect Server mapped
+            collection. Other values will likely cause unknown scope errors on
+            app driven authentication flows.
```
derek-globus, line 125: Can we pull this note about raising unknown scope errors up further into the docstring; it's relevant enough to know for authors who use this function that I'd advocate that it merits its own non-param warning section.

derek-globus (commented):
This looks great!

I have a couple stylistic & documentation notes to discuss but overall really happy with this!

src/globus_sdk/client.py
aaschaer: Yeah this is for GCS client id lookup.

Part of the reason I went with this name was for a hypothetical use case of passing an explicit `Authorization` header in `headers` that you don't want overwritten by the `authorizer`. Calling it `skip_authorization` or `omit_authorization` might imply that such an `Authorization` header might still be omitted.

Maybe something like `automatic_authorization` with a default of `True`?

src/globus_sdk/client.py
aaschaer: Definitely like chaining it! I'll plan to do the same to `TransferClient.add_app_data_access_scope`
Think I prefer keeping the name as is though.

As a chain `add_app_scope` still reads fine to me:
```
flows = FlowsClient(app=app)\
   .add_app_scope(FlowsScopes.manage_flows)\
   .add_app_scope(FlowsScopes.run_manage)
```

but when not chained, `with_app_scope` seems kinda strange to me:
```
try:
    client.do_the_thing()
except ConsentRequired as consent_error:
    client.with_app_scope(consent_error.required_scope)
```

src/globus_sdk/client.py
derek-globus: Okay, I hadn't considered that possibility.

I can be happy with either ignore_authorizer as you initially had it or automatic_authorization in that case. Not much of a preference in either direction.

sirosen: I have an alternative to offer on this, which is to make it a capability of the `RequestsTransport` which can be accessed via `tune()`.
e.g., in GCSClient:
```python
with self.transport.tune(automatic_authorization=False):
    return self.get(...)
```

The advantage I see in this is that it avoids exposing an underlying detail of behavior in a very prominent position (the `BaseClient.get` method), and instead nests it in a way that makes it more of an "advanced" option.

derek-globus: I don't love the contextmanager interface for this but preference aside, I also don't think it actually would workwork (at least in the current structure). 

Disabling `automatic_authorization` must cause GlobusApp to skip *creation* of an authorizer, not just the use of it. That's handled by the BaseClient, not the transport layer & pushing that into the transport layer would mean attaching both the resource server & globus app to that request interface.

sirosen: I put together a quick sample of how this would work here:
https://github.com/sirosen/globus-sdk-python/tree/allow-suppress-automatic-authz

aaschaer: I like the idea of putting this in the Transport layer, but we would still need some way to skip trying to create the `authorizer` at the client level for this to be useful for `GCSClient` so I don't think we'd be saving much interface wise

derek-globus: Alternate way uglier, but less impactful idea.

What if `get_gcs_info` just didn't use the standard `BaseClient.get` and  by the transitive property `BaseClient.request` methods?

We could either copy/paste the logic inside of request or shove it into a function which accepts a transport + request params & don't expose that publicly.

derek-globus: Bear in mind though that we do have one more use case which requires this style of "retrieve client/scope info without any authorization" which is the to-be-built ActionProviderClient.

derek-globus:
```python

class BaseClient:
   ...
   def request(
        self,
        method: str,
        path: str,
        *,
        data: _DataParamType = None,
        encoding: str | None = None,
        allow_redirects: bool = True,
        stream: bool = False,
    ) -> GlobusHTTPResponse:
        return self._request(method, path, query_params=query_params, data=data, headers=headers, encoding=encoding, allow_redirects=allow_redirects, stream=stream)

    def _request(
        self,
        method: str,
        path: str,
        *,
        data: _DataParamType = None,
        encoding: str | None = None,
        allow_redirects: bool = True,
        stream: bool = False,
        automatic_authorization: bool = True,
    ) -> GlobusHTTPResponse:
        ... regular transport layer invocation ...
```

then `get_gcs_info` would just call the internal route

```python
class GCSClient:
  def get_gcs_info(self) -> UnpackingGCSResponse:
      return self._request(...)
```

aaschaer: I'm personally not too bothered by the impact, anyone working at the `BaseClient.get` level is already a fairly advanced user, and I think it makes sense to have this available at the same interface for providing custom headers

derek-globus: Kk, I'm happy with the way it's structured as well.

I'm clicking approve on this PR; but will leave this thread unresolved for Stephen to continue the discussion if he'd like to.

derek-globus (approved):

src/globus_sdk/client.py
sirosen: With the rename to `automatic_authorization` I'm less bothered by this. I feel like `request()` and its siblings have a lot of options, but... is that bad? It seems fine, on a second read.

Thanks for indulging me in some thinking and speculation! Approving!

sirosen (approved):

PR #994, opened 2024-06-28
Added overridable auto-login to GlobusApp

derek-globus:
Added overridable auto-login to GlobusApp

----
📚 Documentation preview 📚: https://globus-sdk-python--994.org.readthedocs.build/en/994/

### DIFF
--- a/src/globus_sdk/experimental/globus_app/_validating_token_storage.py
+++ b/src/globus_sdk/experimental/globus_app/_validating_token_storage.py
@@ -1,16 +1,14 @@
 from __future__ import annotations
 
-import time
-
 from globus_sdk import AuthClient, Scope
 from globus_sdk.experimental.consents import ConsentForest
 from globus_sdk.experimental.tokenstorage import TokenData, TokenStorage
 
 from ..._types import UUIDLike
 from .errors import (
-    ExpiredTokenError,
     IdentityMismatchError,
     MissingIdentityError,
+    MissingTokensError,
     UnmetScopeRequirementsError,
 )
 
@@ -99,6 +97,14 @@ class ValidatingTokenStorage(TokenStorage):
     def store_token_data_by_resource_server(
         self, token_data_by_resource_server: dict[str, TokenData]
     ) -> None:
+        """
+        :param token_data_by_resource_server: A dict of TokenData objects indexed by
+            their resource server
+
+        :raises: :exc:`IdentityValidationError` to signal a problem with the identity
+            info aspect of the token data.
+        :raises :exc:`TokenValidationError` to signal a problem with token data.
+        """
 
         self._validate_token_data_by_resource_server(token_data_by_resource_server)
         self._token_storage.store_token_data_by_resource_server(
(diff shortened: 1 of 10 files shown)


src/globus_sdk/experimental/globus_app/_validating_token_storage.py
```diff
@@ -108,31 +114,31 @@ def store_token_data_by_resource_server(
...
-    def get_token_data(self, resource_server: str) -> TokenData | None:
+    def get_token_data(self, resource_server: str) -> TokenData:
         """
         :param resource_server: A resource server with cached token data.
-        :returns: The token data for the given resource server, or None if no token data
-            is present in the attached storage adapter.
-        :raises: :exc:`TokenValidationError` if the token has expired or does not meet
-            the attached scope requirements.
+        :returns: The token data for the given resource server.
+        :raises: :exc:`TokenValidationError` to signal a problem with stored token data.
```
kurtmckee, line 134:
```suggestion
        :raises: :exc:`MissingTokensError` to signal a problem with stored token data.
```

src/globus_sdk/experimental/globus_app/authorizer_factory.py
```diff
@@ -141,9 +153,10 @@ def _make_authorizer(self, resource_server: str) -> RefreshTokenAuthorizer:
         :param resource_server: The resource server the authorizer will produce
             authentication for
         """
-        token_data = self._get_token_data_or_error(resource_server)
+        token_data = self.token_storage.get_token_data(resource_server)
         if token_data.refresh_token is None:
-            raise MissingTokensError(f"No refresh_token for {resource_server}")
+            msg = f"No refresh_token for {resource_server}"
+            raise MissingTokensError(msg, resource_server=resource_server)
```
kurtmckee, line 159: The docstring isn't implicated in this change so I can't highlight it, but for consistency with other changes, could the narrative "Raises blah blah blah" convert to autodoc-compatible markup, like `:raises: :exc:...`?

kurtmckee (changes requested):

src/globus_sdk/experimental/globus_app/_validating_token_storage.py
derek-globus: `TokenValidationError` is the base subclass of errors raised by this method (of which `MissingTokensError` is one) and multiple different `TokenValidationError` types can be raised from this method.

The basic error heirarchy structure is
  * "TokenValidationError" - category of errors
     * `MissingTokensError` - No token exists for resource server.
     * `ExpiredTokenError` - Token exists but is expired.
     * `UnmetScopeRequirementsError` - Token exists but root or dependent scope requirements are unmet.
     * "IdentityMismatchError" - category of errors
        * `MissingIdentityError` - Token response does not include identity info.
        * `IdentityMismatchError` - Token response includes identity info but it differs from the stored identity.

With the latest rewrite, `get_token_error` does only raise 2 of these error types: [`MissingTokensError`, `UnmetScopeRequirementsError`] (`ExpiredTokenError` was pulled based on a discussion Aaron & I had about how RefreshAuthorizerFactory should function in the case of an expired access token) so I suppose I could scope the docstring down to just those two?

src/globus_sdk/experimental/globus_app/authorizer_factory.py
derek-globus: Sure, updated in new rev

src/globus_sdk/experimental/globus_app/_validating_token_storage.py
derek-globus: Updated this & other docstrings to define more concretely which errors they raise.

kurtmckee (commented):
GitHub is indicating that all 10 files changed, perhaps due to a rebase or rewrite. I'm not able to track what changed since I left comments.

kurtmckee (approved):

PR #998, opened 2024-07-09
Automatically create the directory path for a JSONTokenStorage

derek-globus:
## What?
* Automatically create the directory path for a JSONTokenStorage

## Why?
* The first usage of GlobusApp with default configuration raised an error on my machine (`~/.globus/app` dir didn't exist).

----
📚 Documentation preview 📚: https://globus-sdk-python--998.org.readthedocs.build/en/998/

### DIFF
--- a/changelog.d/20240709_111157_derek_auto_create_json_token_path.rst
+++ b/changelog.d/20240709_111157_derek_auto_create_json_token_path.rst
@@ -1,0 +1,6 @@
+
+Fixed
+~~~~~
+
+- When a JSONTokenStorage is used, the containing directory will be automatically be
+  created if it doesn't exist instead of erroring (on Mac).
--- a/src/globus_sdk/experimental/tokenstorage/json.py
+++ b/src/globus_sdk/experimental/tokenstorage/json.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+import os
 import pathlib
 import typing as t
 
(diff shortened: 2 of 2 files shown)


changelog.d/20240709_111157_derek_auto_create_json_token_path.rst
```diff
@@ -0,0 +1,6 @@
+
+Fixed
+~~~~~
+
+- When a JSONTokenStorage is used, the containing directory will be automatically be
+  created if it doesn't exist instead of erroring (on Mac).
```
sirosen, line 6: Since it's not mac-specific, I don't think we should note the platform. Since I was suggesting an edit, I also went ahead and restored the pr link text from the template (<code>:pr:\`998\`</code> also works).

```suggestion
  created if it doesn't exist instead of erroring. (:pr:`NUMBER`)
```

kurtmckee (approved):

src/globus_sdk/experimental/tokenstorage/json.py
```diff
@@ -40,8 +41,15 @@ def __init__(self, filename: pathlib.Path | str, *, namespace: str = "DEFAULT"):
         :param namespace: A user-supplied namespace for partitioning token data
         """
         self.filename = str(filename)
+        self._ensure_containing_dir_exists()
```
sirosen, line 44: If we're adding this, shouldn't we do it on the parent class, not here?

sirosen (changes requested):

derek-globus:
Yep, sorry didn't realize you were reviewing & expected this change  to be small & uncontroversial enough to not wait around in case someone else was reviewing like I normally try to.

changelog.d/20240709_111157_derek_auto_create_json_token_path.rst
derek-globus: I'm not sure if it's mac exclusive or not. I only know it doesn't work on mac.

src/globus_sdk/experimental/tokenstorage/json.py
derek-globus: Sure, that's a fair question. Would you like to open that PR or me?

sirosen: Sure, I'll work that up!

(Also, no worries about merge/review order. I noticed it, and we'll sort out now. :smile: )

PR #1000, opened 2024-07-10
Support dependent token evaluation in GlobusApp

derek-globus:
- `GlobusApp.add_scope_requirements` now has the side effect of clearing the authorizer
  cache for any referenced resource servers.
- `GlobusAuthorizer.scope_requirements` was made private and a new method for
  accessing scope requirements was added at `GlobusAuthorizer.get_scope_requirements`.
- A `GlobusApp` will now auto-create an Auth consent client for dependent scope
  evaluation against consents as a part of instantiation.

----
📚 Documentation preview 📚: https://globus-sdk-python--1000.org.readthedocs.build/en/1000/

### DIFF
--- a/src/globus_sdk/experimental/globus_app/authorizer_factory.py
+++ b/src/globus_sdk/experimental/globus_app/authorizer_factory.py
@@ -55,7 +55,21 @@ class AuthorizerFactory(
             in the underlying ``ValidatingTokenStorage``.
         """
         self.token_storage.store_token_response(token_res)
-        self._authorizer_cache = {}
+        self.clear_cache()
+
+    def clear_cache(self, *resource_servers: str) -> None:
+        """
+        Clear the authorizer cache for the given resource servers.
+        If no resource servers are given, clear the entire cache.
+
+        :param resource_servers: The resource servers for which to clear the cache
+        """
+        if not resource_servers:
+            self._authorizer_cache = {}
+        else:
+            for resource_server in resource_servers:
+                if resource_server in self._authorizer_cache:
+                    del self._authorizer_cache[resource_server]
 
     def get_authorizer(self, resource_server: str) -> GA:
         """
@@ -102,6 +116,16 @@ class AccessTokenAuthorizerFactory(AuthorizerFactory[AccessTokenAuthorizer]):
     ) -> None:
         super().store_token_response_and_clear_cache(token_res)
         self._cached_authorizer_expiration = {}
+
+    def clear_cache(self, *resource_servers: str) -> None:
+        if not resource_servers:
+            self._cached_authorizer_expiration = {}
+        else:
+            for resource_server in resource_servers:
+                if resource_server in self._cached_authorizer_expiration:
+                    del self._cached_authorizer_expiration[resource_server]
+
+        super().clear_cache(*resource_servers)
 
     def get_authorizer(self, resource_server: str) -> AccessTokenAuthorizer:
         """
--- a/src/globus_sdk/experimental/globus_app/globus_app.py
+++ b/src/globus_sdk/experimental/globus_app/globus_app.py
@@ -173,18 +173,7 @@ class GlobusApp(metaclass=abc.ABCMeta):
             self.client_id = client_id
             self._initialize_login_client(client_secret)
 
-        # get our initial scope requirements make sure we at least have "openid"
-        # to get identity information for token validation
-        auth_rs = str(AuthClient.resource_server)
-        openid_scope = Scope(AuthScopes.openid)
-        if scope_requirements:
-            if auth_rs not in scope_requirements:
-                scope_requirements[auth_rs] = [openid_scope]
-            else:
-                scope_requirements[auth_rs].append(openid_scope)
-            self.scope_requirements = scope_requirements
-        else:
-            self.scope_requirements = {auth_rs: [openid_scope]}
+        self._scope_requirements = scope_requirements or {}
 
         # either get config's TokenStorage, or make the default JSONTokenStorage
         if self.config.token_storage:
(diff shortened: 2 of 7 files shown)


src/globus_sdk/experimental/globus_app/authorizer_factory.py
```diff
@@ -103,6 +117,16 @@ def store_token_response_and_clear_cache(
...
 
+    def clear_cache(self, *resource_servers: str) -> None:
+        if not resource_servers:
+            self._cached_authorizer_expiration = {}
+        else:
+            for resource_server in resource_servers:
+                if resource_server in self._cached_authorizer_expiration:
+                    del self._cached_authorizer_expiration[resource_server]
+
+        super().clear_cache(*resource_servers)
```
sirosen, line 128: I see this matching the superclass structure, and I can't help but think that there's some abstraction or interface here we could add to merge this.
What about this?
```python
    if not resource_servers:
        self._cached_authorizer_expiration = {}
    super().clear_cache(*resource_servers)

...

def _clear_resource_server_from_cache(self, resource_server: str) -> None:
    self._cached_authorizer_expiration.pop(resource_server, None)
    super()._clear_resource_server_from_cache(resource_server)
```

and the base class uses that method. Not worth it?

sirosen (commented):
I'm a bit concerned that this changeset removes the injection of the `openid` scope into users' requests, but I'm also not sure that I'm correct that it does this.

src/globus_sdk/experimental/globus_app/globus_app.py
```diff
@@ -198,12 +187,19 @@ def __init__(
...
 
         # initialize our authorizer factory
         self._initialize_authorizer_factory()
 
+        # create a consent client for token validation
+        # reducing the scope requirements to barebones openid (user identification)
+        # additionally, this will ensure that openid scope requirement is always
+        # registered (it's required for token identity validation).
+        consent_client = AuthClient(app=self, app_scopes=[Scope(AuthScopes.openid)])
+        self._validating_token_storage.set_consent_client(consent_client)
```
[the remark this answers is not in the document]
derek-globus: Nope, the client creation registers openid automatically. We don't do any deduping, so with the old block, openid would be registered twice.

Demo:
```python
>>> from globus_sdk.experimental.globus_app import UserApp
>>>
>>> app = UserApp("my-cool-app", client_id="61338d24-54d5-408f-a10d-66c06b59f6d2")
>>> app.get_scope_requirements("auth.globus.org")
(Scope('openid'),)
```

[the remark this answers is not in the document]
sirosen: Oh, excellent! That was my main concern. I need to reread and see where or how that happens, and evaluate the PR in that light.

src/globus_sdk/experimental/globus_app/authorizer_factory.py
derek-globus: We could, I think it's slightly over-engineered to genericize that touch point though & would prefer to keep it the way it is for now.

src/globus_sdk/experimental/globus_app/globus_app.py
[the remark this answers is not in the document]
derek-globus: It happens because of the `app_scopes` kwarg in the code section you highlighted above.

Code ref: [client.py#L157-L162](https://github.com/globus/globus-sdk-python/blob/main/src/globus_sdk/client.py#L157-L162)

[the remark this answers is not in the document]
sirosen: I was confused and reread everything, but I finally understood when I realized that this client is created with `app=self`. Somehow I missed that -- now I get it.

Marking this thread resolved.

sirosen (approved):
My only worry has turned out to be a non-issue, and we've considered the thing that I wanted to suggest. So we're good!

tests/unit/experimental/globus_app/test_globus_app.py
```diff
@@ -211,6 +218,29 @@ def test_user_app_get_authorizer():
...
+    memory_storage.store_token_data_by_resource_server(_mock_token_data_by_rs())
+    config = GlobusAppConfig(token_storage=memory_storage)
+    user_app = UserApp("test-app", client_id=client_id, config=config)
+
+    initial_authorizer = user_app.get_authorizer("auth.globus.org")
+    assert isinstance(initial_authorizer, AccessTokenAuthorizer)
+    assert initial_authorizer.access_token == "mock_access_token"
+
+    # We should've cached the authorizer from the first call
+    assert user_app.get_authorizer("auth.globus.org") == initial_authorizer
```
kurtmckee, line 233: When checking for cached instances, it seems like this should be an identity check.

```suggestion
    assert user_app.get_authorizer("auth.globus.org") is initial_authorizer
```

tests/unit/experimental/globus_app/test_globus_app.py
```diff
@@ -211,6 +218,29 @@ def test_user_app_get_authorizer():
...
+
+    user_app.add_scope_requirements({"auth.globus.org": [Scope("openid")]})
+
+    # The cache should've been cleared
+    updated_authorizer = user_app.get_authorizer("auth.globus.org")
+    assert initial_authorizer != updated_authorizer
```
kurtmckee, line 239:
```suggestion
    assert initial_authorizer is not updated_authorizer
```

tests/unit/experimental/globus_app/test_globus_app.py
derek-globus: Sure, good callout.

PR #1003, opened 2024-07-12
Fix LocalServer on MacOS py3.10 and earlier

derek-globus:
## What?
* Added a special handling case when evaluating the browser name.

## Why?
* The webbrowser class `MacOSXOSAScript` only implements a public `name` attribute starting in python3.11
  * https://github.com/python/cpython/issues/82828
* It's an odd edge case, but without this fix, all MacOS usages of LocalServerLoginFlowManager on python3.10 and earlier fail.

## Testing
* built the code & tested locally on my mac using python versions: [py38, py310, py311, py312]
* Observed successful login flows for all versions

----
📚 Documentation preview 📚: https://globus-sdk-python--1003.org.readthedocs.build/en/1003/

### DIFF
--- a/changelog.d/20240712_133910_derek_fix_local_server_login_on_mac.rst
+++ b/changelog.d/20240712_133910_derek_fix_local_server_login_on_mac.rst
@@ -1,0 +1,6 @@
+
+Fixed
+~~~~~
+
+- Fixed a GlobusApp bug which would cause LocalServerLoginFlowManager to error on
+  MacOS when versions earlier than python 3.11. (:pr:`NUMBER`)
--- a/changelog.d/20240712_133910_derek_fix_local_server_login_on_mac.rst
+++ b/changelog.d/20240712_133910_derek_fix_local_server_login_on_mac.rst
@@ -3,4 +3,4 @@ Fixed
 ~~~~~
 
 - Fixed a GlobusApp bug which would cause LocalServerLoginFlowManager to error on
-  MacOS when versions earlier than python 3.11. (:pr:`NUMBER`)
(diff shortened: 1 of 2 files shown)


changelog.d/20240712_133910_derek_fix_local_server_login_on_mac.rst
```diff
@@ -0,0 +1,6 @@
+
+Fixed
+~~~~~
+
+- Fixed a GlobusApp bug which would cause LocalServerLoginFlowManager to error on
+  MacOS when versions earlier than python 3.11. (:pr:`NUMBER`)
```
kurtmckee, line 6:
```suggestion
  MacOS when versions earlier than Python 3.11. (:pr:`NUMBER`)
```

src/globus_sdk/experimental/login_flow_manager/local_server_login_flow_manager.py
```diff
@@ -56,14 +56,23 @@ def _open_webbrowser(url: str) -> None:
...
+        if browser_name in BROWSER_BLACKLIST:
             raise LocalServerLoginFlowError(
                 "Cannot use LocalServerLoginFlowManager with "
-                f"text-only browser '{browser.name}'"
+                f"text-only browser '{browser_name}'"
             )
 
         if not browser.open(url, new=1):
-            raise LocalServerLoginFlowError("Failed to open browser '{browser.name}'")
+            raise LocalServerLoginFlowError(f"Failed to open browser '{browser_name}'")
```
kurtmckee, line 75: Very nice that this also fixed a quiet bug!

kurtmckee (changes requested):

kurtmckee (approved):

PR #1004, opened 2024-07-12
Refactor experimental SQLiteTokenStorage ("v2") to better accord with its parent class

sirosen:
These changes are all related, although not necessarily in a perfectly obvious way.

The story runs thusly:
- First, we agreed that we wanted to move the "ensure dir" dynamics to the parent class, FileTokenStorage.
- Second, we discovered that this would conflict with use of `:memory:` databases. (Probably solvable, but a red flag about being a "FileTokenStorage" but not having a file.)
- Third, we considered whether or not such usages could simply use MemoryTokenStorage. But there are concerns around use of the 'config_storage' area making SQLiteTokenStorage the "only production storage mechanism".
- Fourth, we agreed upon a resolution trajectory in which, at least until we see evidence to the contrary, we will remove the 'config_storage' table and make these classes more uniform.
- Fifth, this PR.

One notable addition to the above:
- This changeset sets the "schema version" to `2` for the database. It's not yet certain, IMO, that this field will be useful, but we could use it to trivially check if `config_storage` is expected to be present and to manage data migrations.

I have already started to think about what a data migration for globus-cli would look like, and it's definitely nontrivial.
We will need to develop some kind of compatibility layer which migrates data and presents a uniform interface to its callers, but doing so in a way which doesn't break the upgrade-downgrade path will require careful thought.
I'm not certain that this will turn out to be worth the pain, but we could, *at the very least*, provide a distinct (non-TokenStorage) interface for accessing the `config_storage` table found in the sqlite DB. That is: the remodeling of classes done here is robust to the possibility of our deciding against a migration of the `config_storage` table to some other source of truth.

----
📚 Documentation preview 📚: https://globus-sdk-python--1004.org.readthedocs.build/en/1004/

### DIFF
--- a/src/globus_sdk/experimental/tokenstorage/sqlite.py
+++ b/src/globus_sdk/experimental/tokenstorage/sqlite.py
@@ -6,6 +6,7 @@ import sqlite3
 import textwrap
 import typing as t
 
+from globus_sdk import exc
 from globus_sdk.experimental.tokenstorage.base import FileTokenStorage
 from globus_sdk.version import __version__
 
@@ -28,39 +29,37 @@ class SQLiteTokenStorage(FileTokenStorage):
 
     def __init__(
         self,
-        dbname: pathlib.Path | str,
+        filename: pathlib.Path | str,
         *,
         connect_params: dict[str, t.Any] | None = None,
         namespace: str = "DEFAULT",
     ):
         """
-        :param dbname: The name of the DB file to write to and read from. If the string
-            ":memory:" is used, an in-memory database will be used instead.
+        :param filename: The name of the DB file to write to and read from.
         :param connect_params: A pass-through dictionary for fine-tuning the SQLite
              connection.
         :param namespace: A user-supplied namespace for partitioning token data.
         """
-        self.filename = self.dbname = str(dbname)
+        if filename == ":memory:":
+            raise exc.GlobusSDKUsageError(
+                "SQLiteTokenStorage cannot be used with a ':memory:' database. "
+                "If you want memory-backed TokenStorage, use MemoryTokenStorage "
+                "instead."
+            )
+        self.filename = str(filename)
         self._connection = self._init_and_connect(connect_params)
         super().__init__(namespace=namespace)
-
-    def _is_memory_db(self) -> bool:
-        return self.dbname == ":memory:"
 
     def _init_and_connect(
         self,
         connect_params: dict[str, t.Any] | None,
     ) -> sqlite3.Connection:
-        init_tables = self._is_memory_db() or not self.file_exists()
         connect_params = connect_params or {}
-        if init_tables and not self._is_memory_db():  # real file needs to be created
+        if not self.file_exists():
             with self.user_only_umask():
                 conn: sqlite3.Connection = sqlite3.connect(
-                    self.dbname, **connect_params
+                    self.filename, **connect_params
                 )
-        else:
-            conn = sqlite3.connect(self.dbname, **connect_params)
-        if init_tables:
             conn.executescript(
                 textwrap.dedent(
                     """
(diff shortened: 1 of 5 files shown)


sirosen:
The failure in CI seems to happen consistently (I tried a rerun) and is Windows-specific. Here's what build output shows:
```
.tox\py\Lib\site-packages\flaky\flaky_pytest_plugin.py:146: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests\functional\tokenstorage_v2\conftest.py:15: in tempdir
    shutil.rmtree(d)
C:\hostedtoolcache\windows\Python\3.11.9\x64\Lib\shutil.py:787: in rmtree
    return _rmtree_unsafe(path, onerror)
C:\hostedtoolcache\windows\Python\3.11.9\x64\Lib\shutil.py:634: in _rmtree_unsafe
    onerror(os.unlink, fullname, sys.exc_info())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

path = 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmpk168spxq'
onerror = <function rmtree.<locals>.onerror at 0x0000028B89AC2520>

    def _rmtree_unsafe(path, onerror):
        try:
            with os.scandir(path) as scandir_it:
                entries = list(scandir_it)
        except OSError:
            onerror(os.scandir, path, sys.exc_info())
            entries = []
        for entry in entries:
            fullname = entry.path
            if _rmtree_isdir(entry):
                try:
                    if entry.is_symlink():
                        # This can only happen if someone replaces
                        # a directory with a symlink after the call to
                        # os.scandir or entry.is_dir above.
                        raise OSError("Cannot call rmtree on a symbolic link")
                except OSError:
                    onerror(os.path.islink, fullname, sys.exc_info())
                    continue
                _rmtree_unsafe(fullname, onerror)
            else:
                try:
>                   os.unlink(fullname)
E                   PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmpk168spxq\\test.db'
```

So I think this may be a difference in behavior between the stdlib's `sqlite` on older and newer versions, and apparently by platform. My current _guess_ is that this is basically a resource-leak. We never `close()` a tokenstorage which is pointed at the temporary DB, and as a result it consistently fails. I'm going to try right now to see if I can find a relevant snippet in these tests, and then I'll check if pushing it here resolves the CI failure.

sirosen:
The above guess is _not_ borne out by the diff. I'm wondering if this has been a standing issue which is somehow now exposed by the slight changes to the testsuite.

sirosen:
I've rebased onto #1007, and dropped the added commit here which explored that solution for the `rmtree` issue we had in CI.

src/globus_sdk/experimental/tokenstorage/sqlite.py
```diff
@@ -28,48 +29,40 @@ class SQLiteTokenStorage(FileTokenStorage):
...
         :param namespace: A user-supplied namespace for partitioning token data.
         """
-        self.filename = self.dbname = str(dbname)
-        self._connection = self._init_and_connect(connect_params)
-        super().__init__(namespace=namespace)
+        if filename == ":memory:":
+            raise exc.GlobusSDKUsageError(
+                "SQLiteTokenStorage cannot be used with a ':memory:' database. "
+                "If you want memory-backed TokenStorage, use MemoryTokenStorage "
+                "instead."
```
kurtmckee, line 47:
```suggestion
                "If you want to store tokens in memory, use MemoryTokenStorage instead."
```

kurtmckee (changes requested):

kurtmckee (approved):

PR #1011, opened 2024-07-19
Add 'orderby' to TransferClient.task_list

sirosen:
I noticed this parameter was missing during some globus-cli work.
Adding it here will allow us to avoid relying on `query_params` to send it.

The existing test data for `task_list` was stored in a pretty old testing format, so I've split this work into three commits rather than one:

- Add 'orderby' as to 'TransferClient.task_list'
- Move task_list tests to new paradigm
- Add a test for the 'orderby' task_list param

----
📚 Documentation preview 📚: https://globus-sdk-python--1011.org.readthedocs.build/en/1011/


src/globus_sdk/services/transfer/client.py
```diff
@@ -1567,6 +1571,11 @@ def task_list(
             query_params["limit"] = limit
         if offset is not None:
             query_params["offset"] = offset
+        if orderby is not None:
+            if isinstance(orderby, str):
+                query_params["orderby"] = orderby
+            else:
+                query_params["orderby"] = ",".join(orderby)
```
jakeglobus, line 1578: I haven't looked up the exact syntax, but the else branch here joins on a comma instead of a space which is specified in the `:param` doc above

sirosen: I think this is right -- but if it's wrong we probably have to fix other method implementations, since I 100% copied this from elsewhere in the module.

I think Transfer accepts multiple ordering criteria comma-separated, like `created ASC,terminated DESC`. I'll double-check the docs and run a live test or two to see if this is right.

If it is though, it maybe needs slightly more docstring, since I think this is a potential point of confusion with `["created ASC", "terminated DESC"]` vs `["created", "ASC"]`. :face_with_spiral_eyes:

sirosen: Just live-blogging my process here a little bit, I'm reaching for `globus api` to check this really easily.
This fails (good!):
```
$ globus api transfer GET /task_list -Q 'orderby=status,DESC'
```
and this works:
```
$ globus api transfer GET /task_list -Q 'orderby=status DESC'
```

So the space is correct. And then we get to some other things which work, like
```
$ globus api transfer GET /task_list -Q 'orderby=status DESC,label'

$ globus api transfer GET /task_list -Q 'orderby=status DESC,label ASC'
```

So I'm pretty sure I have the usage here correct ( :relieved: ).

But now I think this accepting a list is potentially confusing, so I think this needs enhanced doc (i.e. at least one complex example).

kurtmckee: I'm with Jake that the docstring can be updated to help clarify this behavior.

sirosen: I've added an example usage because I'm not sure what the full scope of the documentation here should be. I don't want to overwhelm a reader, but I want to point them in the right direction.

It's in a fresh commit; let me know if it needs further refinement!

src/globus_sdk/services/transfer/client.py
```diff
@@ -1553,6 +1553,20 @@ def task_list(
                             f"{task['source_endpoint']} -> {task['destination_endpoint']}"
                         )
 
+                Fetch the last 5 failed tasks by submission time:
+
+                    tc = TransferClient(...)
```
kurtmckee, line 1558:
```suggestion
                Fetch the last 5 failed tasks by submission time:

                ..  code-block:: python

                    tc = TransferClient(...)
```

kurtmckee (changes requested):

src/globus_sdk/services/transfer/client.py
sirosen: Good catch!

PR #1013, opened 2024-07-25
Using a GlobusApp example doc

derek-globus:
Added a doc explaining how to use a GlobusApp.

Not Included
  * Custom configuration of a globus app (custom token storage, custom login managers, etc.)

New Example doc link: https://globus-sdk-python--1013.org.readthedocs.build/en/1013/experimental/examples/oauth2/globus_app.html

----
📚 Documentation preview 📚: https://globus-sdk-python--1013.org.readthedocs.build/en/1013/

### DIFF
--- a/docs/experimental/examples/oauth2/globus_app.rst
+++ b/docs/experimental/examples/oauth2/globus_app.rst
@@ -166,9 +166,10 @@ MFA token or rendering with a specific message:
 Manually Defining Scope Requirements
 ------------------------------------
 
-Globus service client classes all maintain internal list of default scope requirements
-to be attached to any bound app. These scopes represent an approximation of a
-"standard set" for each service. This list however is not sufficient for all use cases.
+Globus service client classes all maintain an internal list of default scope
+requirements to be attached to any bound app. These scopes represent an approximation of
+a "standard set" for each service. This list however is not sufficient for all use
+cases.
 
 For example, the FlowsClient defines its default scopes as ``flows:view_flows`` and
 ``flows:run_status`` (read-only access). These scopes will not be sufficient for a
@@ -193,7 +194,7 @@ This can be done in one of two ways:
 
     ..  code-block:: python
 
-        from globus_sdk import Scope, FlowsClient
+        from globus_sdk import FlowsClient
 
         flows_client = FlowsClient(app=my_app)
         flows_client.add_app_scope(FlowsClient.scopes.manage_flows)
--- a/docs/experimental/examples/oauth2/globus_app.rst
+++ b/docs/experimental/examples/oauth2/globus_app.rst
@@ -9,14 +9,14 @@ Using a GlobusApp
 Programmatic communication with Globus services relies on the authorization of requests.
 Management and resolution of this authorization can become an arduous task, especially
 when a script needs to interact with different services each carrying an individual set
-of complex auth requirements. To assist with this task, this library provides a utility
-construct called a GlobusApp.
+of complex authentication and authorization requirements. To assist with this task, this
+library provides a utility construct called a GlobusApp.
 
-A :py:class:`~GlobusApp` is a distinct object which will manage auth requirements
-(i.e. **scopes**) identified by their associated service (i.e. **resource server**).
-In addition to storing these requirements, a GlobusApp provides a mechanism to resolve
-unmet ones through browser and api-based auth flows, supplying the resulting tokens to
-bound clients as requested.
+A :py:class:`~GlobusApp` is a distinct object which will manage Globus Auth requirements
+(i.e., **scopes**) identified by their associated service (i.e., **resource server**).
+In addition to storing them, a GlobusApp provides a mechanism to resolve unmet
+requirements through browser- and API-based authorization flows, supplying the resulting
+tokens to bound clients as requested.
 
 There are two flavors of GlobusApp:
 
@@ -30,10 +30,10 @@ There are two flavors of GlobusApp:
 Setup
 -----
 
-A GlobusApp is heavily configurable object. For common scripting usage however,
+A GlobusApp is a heavily configurable object. For common scripting usage however,
 instantiation only requires two parameters:
 
-#.  **App Name** - A human readable slug to identify your app in http requests and token
+#.  **App Name** - A human readable slug to identify your app in HTTP requests and token
     caches.
 
 #.  **Client Info** - either a *Native Client's* ID or a *Confidential Client's* ID and
@@ -42,7 +42,7 @@ instantiation only requires two parameters:
     *   There are important distinctions to consider when choosing your client type; see
         `Developing an Application Using Globus Auth <https://docs.globus.org/api/auth/developer-guide/#developing-apps>`_.
 
-        For a simplified heuristic however:
+        A simplified heuristic to help choose the client type however is:
 
         *   Use a *Confidential Client* when your client needs to own cloud resources
             itself and will be used in a trusted environment where you can securely
@@ -50,32 +50,32 @@ instantiation only requires two parameters:
 
         *   Use a *Native Client* when your client will be facilitating interactions
             between a user and a service, particularly if it is bundled within a
-            script or cli tool to be distributed to end-user's machines.
+            script or cli tool to be distributed to end-users' machines.
 
 
-..  Note::
+    ..  Note::
 
-    Even in the context of a UserApp, client info is required to interact with a
-    service. Those interactions will be performed on behalf of a user instead of the
-    client itself, but the client is necessary to facilitate the interaction.
+        Both UserApps and ClientApps require a client to operate. In a UserApp those
+        interactions are done on behalf of a user but the client is still the actor
+        requesting them.
 
 
-Once instantiated, an app can be passed to any service client using the init
-``app`` kwarg (e.g. ``TransferClient(app=my_app)``). Doing this will bind the app to the
-client, registering a default set of scopes requirements for the service client's
-resource server and configuring the app as the service client's auth provider.
+Once instantiated, a GlobusApp can be passed to any service client using the init
+``app`` keyword argument (e.g. ``TransferClient(app=my_app)``). Doing this will bind the
+app to the client, registering a default set of scopes requirements for the service
+client's resource server and configuring the app as the service client's auth provider.
 
 
 ..  tab-set::
 
     ..  tab-item:: UserApp
 
-        Construct a UserApp then bind it to a transfer client and a flows client.
+        Construct a UserApp then bind it to a Transfer client and a Flows client.
 
         ..  Note::
 
-            ``UserApp.__init__(...)`` also accepts a `client_secret` kwarg which must be
-            supplied for confidential clients.
+            ``UserApp.__init__(...)`` also accepts a `client_secret` keyword argument
+            which must be supplied for confidential clients.
 
         ..  code-block:: python
 
@@ -91,12 +91,12 @@ resource server and configuring the app as the service client's auth provider.
 
     .. tab-item:: ClientApp
 
-        Construct a ClientApp then bind it to a transfer client and flows_client.
+        Construct a ClientApp, then bind it to a Transfer client and a Flows client.
 
         ..  Note::
 
-            ``ClientApp.__init__(...)`` does not allow omission of the `client_secret`
-            kwarg as native clients are not allowed.
+            ``ClientApp.__init__(...)`` requires the `client_secret` keyword argument.
+            Native clients, which lack secrets, are not allowed.
 
         ..  code-block:: python
 
@@ -114,8 +114,8 @@ resource server and configuring the app as the service client's auth provider.
 Usage
 -----
 
-From this point, the app manages scope validation, token caching, and authorizer
-supply for any the clients it is bound to.
+From this point, the app manages scope validation, token caching and routing for any
+bound clients.
 
 In the above example, listing a client's or user's flows becomes as simple as:
 
@@ -123,11 +123,12 @@ In the above example, listing a client's or user's flows becomes as simple as:
 
     flows = flows_client.list_flows()["flows"]
 
-If cached tokens are missing, expired, or otherwise insufficient (e.g. the first time
+If cached tokens are missing, expired, or otherwise insufficient (e.g., the first time
 you run the script), the app will automatically initiate an auth flow to acquire new
-tokens. For a UserApp, this will print out a URL to terminal with a prompt instructing a
-the user to follow the link and enter the code they're given back into the terminal. For
-a ClientApp, the app will retrieve tokens programmatically through an Auth API.
+tokens. With a UserApp, the app will print a URL to the terminal with a prompt
+instructing a the user to follow the link and enter the code they're given back into the
+terminal. With a ClientApp, the app will retrieve tokens programmatically through a
+Globus Auth API.
 
 Once this auth flow has finished, the app will cache tokens for future use and
 invocation of your requested method will proceed as expected.
@@ -137,14 +138,14 @@ Manually Running Login Flows
 ----------------------------
 
 While your app will automatically initiate and oversee auth flows as detected, sometimes
-the programmer knows timing better. To manually trigger a login flow, call
-``GlobusApp.run_login_flow(...)``. This will initiate a flow requesting new tokens
-based on the app's currently defined scope requirements, caching the resulting tokens
-for future use.
+an author may want to explicitly control when an authorization occurs. To manually
+trigger a login flow, call ``GlobusApp.run_login_flow(...)``. This will initiate an auth
+flow requesting new tokens based on the app's currently defined scope requirements, and
+caching the resulting tokens for future use.
 
 This method accepts a single optional parameter, ``auth_params``, where a caller
 may specify additional session-based auth parameters such as requiring the use of an
-mfa token or rendering with a specific message:
+MFA token or rendering with a specific message:
 
 
 ..  code-block:: python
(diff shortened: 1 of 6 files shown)


docs/experimental/examples/oauth2/globus_app.rst
```diff
@@ -1,5 +1,202 @@
...
+
+In the above example, listing a client's or user's flows becomes as simple as:
+
+..  code-block:: python
+
+    flows = flows_client.list_flows()["flows"]
+
+If cached tokens are missing, expired, or otherwise insufficient (e.g. the first time
+you run the script), the app will automatically initiate an auth flow to acquire new
+tokens. For a UserApp, this will print out a URL to terminal with a prompt instructing a
```
kurtmckee, line 128:
```suggestion
tokens. For a UserApp, this will print a URL to the terminal with a prompt instructing
```

docs/experimental/examples/oauth2/globus_app.rst
```diff
@@ -1,5 +1,202 @@
...
+Once this auth flow has finished, the app will cache tokens for future use and
+invocation of your requested method will proceed as expected.
+
+
+Manually Running Login Flows
+----------------------------
+
+While your app will automatically initiate and oversee auth flows as detected, sometimes
+the programmer knows timing better. To manually trigger a login flow, call
+``GlobusApp.run_login_flow(...)``. This will initiate a flow requesting new tokens
```
kurtmckee, line 141: Above, the phrase "auth flow" is used; I recommend using that phrasing consistently.

```suggestion
``GlobusApp.run_login_flow(...)``. This will initiate an auth flow requesting new tokens
```

docs/experimental/examples/oauth2/globus_app.rst
```diff
@@ -1,5 +1,202 @@
...
+based on the app's currently defined scope requirements, caching the resulting tokens
```
kurtmckee, line 142:
```suggestion
based on the app's currently defined scope requirements, and caching the resulting tokens
```

docs/experimental/examples/oauth2/globus_app.rst
```diff
@@ -1,5 +1,202 @@
...
+for future use.
+
+This method accepts a single optional parameter, ``auth_params``, where a caller
+may specify additional session-based auth parameters such as requiring the use of an
+mfa token or rendering with a specific message:
```
kurtmckee, line 147:
```suggestion
MFA token or rendering with a specific message:
```

docs/experimental/examples/oauth2/globus_app.rst
```diff
@@ -1,5 +1,202 @@
...
+            session_message="Please authenticate with MFA",
+            session_required_mfa=True,
+        )
+    )
+
+
+Manually Defining Scope Requirements
+------------------------------------
+
+Globus service clients all maintain as a part of their class definition a list of
```
kurtmckee, line 167:
```suggestion
Globus service clients all maintain an internal list of
```

docs/experimental/examples/oauth2/globus_app.rst
```diff
@@ -1,5 +1,202 @@
...
+default scope requirements to be attached to any bound app. These scopes represent our
+best approximation of a "standard set" for this service; unfortunately however,
+this list will note be sufficient for all use cases.
```
kurtmckee, line 170:
```suggestion
default scope requirements to be attached to any bound app. These scopes represent an
approximation of a "standard set" for each service; however,
this list will not be sufficient for all use cases.
```

docs/experimental/examples/oauth2/globus_app.rst
```diff
@@ -1,5 +1,202 @@
...
+
+    ..  code-block:: python
+
+        from globus_sdk import Scope, FlowsClient
+
+        FlowsClient(app=my_app, app_scopes=[Scope(FlowsClient.scopes.manage_flows)])
+
+    This approach results in an app which only requires the ``flows:manage_flows``
+    scope. Neither default scope (``flows:view_flows`` and ``flows:run_status``) are
+    registered.
```
kurtmckee, line 189:
```suggestion
    scope. The default scopes (``flows:view_flows`` and ``flows:run_status``) are
    not registered.
```

docs/experimental/examples/oauth2/globus_app.rst
```diff
@@ -1,5 +1,202 @@
...
+
+#.  Through a service client's ``add_app_scope`` method.
+
+    ..  code-block:: python
+
+        from globus_sdk import Scope, FlowsClient
+
+        flows_client = FlowsClient(app=my_app)
+        flows_client.add_app_scope(Scope(FlowsClient.scopes.manage_flows))
```
kurtmckee, line 198: It feels very unnecessary that the "clearly already a scope" thing, which is being passed to something that "clearly wants a scope", has to be wrapped in a `Scope` object.

Is this strictly necessary today? Could it instead be:

```suggestion
        flows_client.add_app_scope(FlowsClient.scopes.manage_flows)
```

If so, that's what I want to see here.

src/globus_sdk/experimental/globus_app/globus_app.py
```diff
@@ -441,11 +440,11 @@ class ClientApp(GlobusApp):
     A ``GlobusApp`` using client credentials - useful for service accounts and
     automation use cases.
 
-    ``ClientApp``s are always used with confidential clients either by passing
+    ``ClientApp`` s are always used with confidential clients either by passing
```
kurtmckee, line 443: The 's' should still be attached; does it not render correctly when placed next to it? If not, perhaps a backslash is the solution here, though that may require this docstring to convert to an r-string (`r"""`).

```suggestion
    ``ClientApp``\s are always used with confidential clients either by passing
```

kurtmckee (changes requested):
I'm _really_ pleased with this new documentation! Thank you!

docs/experimental/examples/oauth2/globus_app.rst
```diff
@@ -1,5 +1,202 @@
...
+=================
+
+Programmatic communication with Globus services relies on the authorization of requests.
+Management and resolution of this authorization can become an arduous task, especially
+when a script needs to interact with different services each carrying an individual set
+of complex auth requirements. To assist with this task, this library provides a utility
+construct called a GlobusApp.
+
+A :py:class:`~GlobusApp` is a distinct object which will manage auth requirements
+(i.e. **scopes**) identified by their associated service (i.e. **resource server**).
```
ada-globus, line 16:
```suggestion
(i.e., **scopes**) identified by their associated service (i.e., **resource server**).
```

Does GlobusApp provide any special handling vis-a-vis session-related issues today?  I don't believe so, but want to double check, as scopes/consents are only one dimension of auth requirements.

docs/experimental/examples/oauth2/globus_app.rst
```diff
@@ -1,5 +1,202 @@
...
+In addition to storing these requirements, a GlobusApp provides a mechanism to resolve
+unmet ones through browser and api-based auth flows, supplying the resulting tokens to
+bound clients as requested.
+
+There are two flavors of GlobusApp:
+
+*   :py:class:`~UserApp`, a GlobusApp for interactions between an end user and Globus
+    services. Facilitates interactions *on behalf of a user*.
```
ada-globus, line 24: "On behalf of a user" feels a bit fuzzy, as an application could use a service account to do something on my behalf. One thought:

```suggestion
    services. Operations will be performed as the *user's identity.*.
```

docs/experimental/examples/oauth2/globus_app.rst
```diff
@@ -1,5 +1,202 @@
...
+
+*   :py:class:`~ClientApp`, a GlobusApp for interactions between a client
+    (i.e. service account) and Globus services. Facilitates interactions
+    *as the script itself*.
```
ada-globus, line 28:
```suggestion
    (i.e., service account) and Globus services. Operations will be performed
    as the *client's own identity*.
```

docs/experimental/examples/oauth2/globus_app.rst
```diff
@@ -1,5 +1,202 @@
...
+
+
+..  Note::
+
+    Even in the context of a UserApp, client info is required to interact with a
+    service. Those interactions will be performed on behalf of a user instead of the
+    client itself, but the client is necessary to facilitate the interaction.
+
+
+Once instantiated, an app can be passed to any service client using the init
```
ada-globus, line 63:
```suggestion
Once instantiated, a GlobusApp can be passed to any service client using the init
```

docs/experimental/examples/oauth2/globus_app.rst
```diff
@@ -1,5 +1,202 @@
...
+``app`` kwarg (e.g. ``TransferClient(app=my_app)``). Doing this will bind the app to the
+client, registering a default set of scopes requirements for the service client's
+resource server and configuring the app as the service client's auth provider.
+
+
+..  tab-set::
+
+    ..  tab-item:: UserApp
+
+        Construct a UserApp then bind it to a transfer client and a flows client.
```
ada-globus, line 73:
```suggestion
        Construct a UserApp, then bind it to a Transfer client and a Flows client.
```

docs/experimental/examples/oauth2/globus_app.rst
```diff
@@ -1,5 +1,202 @@
...
+From this point, the app manages scope validation, token caching, and authorizer
+supply for any the clients it is bound to.
+
+In the above example, listing a client's or user's flows becomes as simple as:
+
+..  code-block:: python
+
+    flows = flows_client.list_flows()["flows"]
+
+If cached tokens are missing, expired, or otherwise insufficient (e.g. the first time
```
ada-globus, line 126:
```suggestion
If cached tokens are missing, expired, or otherwise insufficient (e.g., the first time
```

docs/experimental/examples/oauth2/globus_app.rst
```diff
@@ -1,5 +1,202 @@
...
+
+Once this auth flow has finished, the app will cache tokens for future use and
+invocation of your requested method will proceed as expected.
+
+
+Manually Running Login Flows
+----------------------------
+
+While your app will automatically initiate and oversee auth flows as detected, sometimes
+the programmer knows timing better. To manually trigger a login flow, call
```
ada-globus, line 140:
```suggestion
an author may want to explicitly control when an authorization occurs. To manually trigger a login flow, call
```

src/globus_sdk/experimental/globus_app/globus_app.py
```diff
@@ -126,17 +126,17 @@ class GlobusAppConfig:
...
 
     A ``GlobusApp`` manages scope requirements across multiple resource servers,
     runs login flows through its ``LoginFlowManager``, handles token storage and
     validation through its ``ValidatingTokenStorage``, and provides up-to-date
     authorizers from its ``AuthorizerFactory`` through ``get_authorizer``.
 
-    In the near future, a ``GlobusApp`` will be accepted as an initialization parameter
-    to any Globus service client, automatically handling the client's default
-    scope requirements and providing the client an authorizer.
+    A ``GlobusApp`` is accepted as an initialization parameter to any SDK-hosted Globus
```
ada-globus, line 137: Is "hosted" a bit of an odd fit here? Maybe:

```suggestion
    A ``GlobusApp`` is accepted as an initialization parameter to any SDK-provided Globus
```

ada-globus:
Overall looks good—my comments are entirely proof-read and copy-edit oriented!

docs/experimental/examples/oauth2/globus_app.rst
```diff
@@ -1,5 +1,202 @@
...
+    (i.e. service account) and Globus services. Facilitates interactions
+    *as the script itself*.
+
+Setup
+-----
+
+A GlobusApp is heavily configurable object. For common scripting usage however,
+instantiation only requires two parameters:
+
+#.  **App Name** - A human readable slug to identify your app in http requests and token
```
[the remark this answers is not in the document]
derek-globus: The regex you described is not required but it is recommended.

We'll be using this to generate a file name so I'd trying to push users in the direction of "my-cool-app" instead of "My App".

docs/experimental/examples/oauth2/globus_app.rst
```diff
@@ -1,5 +1,202 @@
...
+        *   Use a *Native Client* when your client will be facilitating interactions
+            between a user and a service, particularly if it is bundled within a
+            script or cli tool to be distributed to end-user's machines.
+
+
+..  Note::
+
+    Even in the context of a UserApp, client info is required to interact with a
+    service. Those interactions will be performed on behalf of a user instead of the
+    client itself, but the client is necessary to facilitate the interaction.
```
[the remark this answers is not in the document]
derek-globus: The note is meant to highlight that despite these things being called a `UserApp` and a `ClientApp` both require a client, not just the one called `ClientApp`.

Reworded & indented to try to clarify this.

src/globus_sdk/experimental/globus_app/globus_app.py
derek-globus: Yep this was a rendering issue; ``ClientApp``\\s solved it

docs/experimental/examples/oauth2/globus_app.rst
derek-globus: It allows you to provide them on explicit login requests but not in the same way that scopes are managed, I'll sstick with `i.e.` for now.

Also, I know you're right so I'll fix it but I'd like to register my disdain for the period followed by the comma in `i.e.,`

ada-globus (approved):
🙌

docs/experimental/examples/oauth2/globus_app.rst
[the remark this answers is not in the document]
kurtmckee: If the user selects an in-memory token store, is that still a consideration?

My assumption here is that it's more important how this appears in the Web App when consenting to things, so I'd prefer to slugify the app name when creating a file-backed token store, but not use the word "slug" here.

docs/experimental/examples/oauth2/globus_app.rst
```diff
@@ -1,5 +1,203 @@
...
+
+
+    .. tab-item:: ClientApp
+
+        Construct a ClientApp, then bind it to a Transfer client and a Flows client.
+
+        ..  Note::
+
+            ``ClientApp.__init__(...)`` requires the `client_secret` keyword argument.
+            Native clients, which lack secrets, are not allowed.
```
kurtmckee, line 100: Should this read "Native client IDs, which lack..."?

```suggestion
            Native client IDs, which lack secrets, are not allowed.
```

docs/experimental/examples/oauth2/globus_app.rst
```diff
@@ -1,5 +1,203 @@
...
+            session_message="Please authenticate with MFA",
+            session_required_mfa=True,
+        )
+    )
+
+
+Manually Defining Scope Requirements
+------------------------------------
+
+Globus service client classes all maintain internal list of default scope requirements
```
kurtmckee, line 169:
```suggestion
Globus service client classes all maintain an internal list of default scope requirements
```

docs/experimental/examples/oauth2/globus_app.rst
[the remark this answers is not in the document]
derek-globus: If a user provides their own instantiated token storage, in-memory or not, this value is unused.

At this point we're debating behavior though not documentation.
Can we remove this from the documentation PR & discuss out-of-band whether to make the change you're suggesting?

docs/experimental/examples/oauth2/globus_app.rst
```diff
@@ -1,5 +1,203 @@
...
+
+This can be done in one of two ways:
+
+#.  Through a service client initialization, using the ``app_scopes`` kwarg.
+
+    ..  code-block:: python
+
+        from globus_sdk import Scope, FlowsClient
+
+        FlowsClient(app=my_app, app_scopes=[Scope(FlowsClient.scopes.manage_flows)])
```
kurtmckee, line 186: I'm assuming that this can drop the `Scope` instantiation, too.

```suggestion
        from globus_sdk import FlowsClient

        FlowsClient(app=my_app, app_scopes=[FlowsClient.scopes.manage_flows])
```

docs/experimental/examples/oauth2/globus_app.rst
```diff
@@ -1,5 +1,203 @@
...
+
+    This approach results in an app which only requires the ``flows:manage_flows``
+    scope. The default scopes (``flows:view_flows`` and ``flows:run_status``) are not
+    registered.
+
+#.  Through a service client's ``add_app_scope`` method.
+
+    ..  code-block:: python
+
+        from globus_sdk import Scope, FlowsClient
```
kurtmckee, line 196:
```suggestion
        from globus_sdk import FlowsClient
```

kurtmckee (changes requested):

docs/experimental/examples/oauth2/globus_app.rst
derek-globus: I don't think so, no.

You're not allowed to use native clients in a ClientApp. The client ID is the identifier for that native client that you aren't allowed to use.

docs/experimental/examples/oauth2/globus_app.rst
derek-globus: This one actually can't.

Maybe it should be able to, but for now it can't.

derek-globus: > Maybe it should be able to, but for now it can't.

This is mean btw in a similar vein as the slug comment above. I'd like to discuss that behavior outside of a documentation PR. We can discuss the merits of adding that and update documentation after adding that behavior if we decide to; I don't want to block this documentation from going out on that decision though.

kurtmckee (approved):

ada-globus (approved):

derek-globus:
Those last two commits update behavior and wording to reject confidential clients in UserApps.

I was testing UserApp locally and ran into issues with redirection in the case of Confidential Client (we don't have a mechanism for users to pass through their redirection url).

This makes sure our documentation is accurate and raises an error on instantiation rather than later down the line.

PR #1016, opened 2024-07-26
Added flows all scope

derek-globus:
## What?
* Added the `flows:all` to the `FlowsScopes` object.

## Testing

Used the following script to print my flows in sandbox using `flows:all`
```python
from globus_sdk import Scope, FlowsClient
from globus_sdk.experimental.globus_app import GlobusAppConfig, UserApp

SANDBOX_NATIVE_CLIENT = 'bdf9595f-2d0d-412b-a721-2767d7959404'
app = UserApp("test-flows-all", client_id=SANDBOX_NATIVE_CLIENT, config=GlobusAppConfig(environment="sandbox"))

flows = FlowsClient(app=app, app_scopes=[Scope(FlowsClient.scopes.all)])
flow = flows.list_flows()["flows"][0]

print("First Flow")
print(flow)
```

Output
```
Please authenticate with Globus here:
-------------------------------------
https://auth.sandbox.globuscs.info/v2/oauth2/authorize?client_id=bdf9595f-2d0d-412b-...trunc...
-------------------------------------
Enter the resulting Authorization Code here: ...trunc...

First Flow
{'id': '3792a8ac-a48f-42e7-9725-32c67a7b9a63', ...trunc...}
```

----
📚 Documentation preview 📚: https://globus-sdk-python--1016.org.readthedocs.build/en/1016/

### DIFF
--- a/changelog.d/20240726_111812_derek_flows_all_scope.rst
+++ b/changelog.d/20240726_111812_derek_flows_all_scope.rst
@@ -1,0 +1,6 @@
+
+Added
+~~~~~
+
+-   Added a reference to the new Flows all scope under
+    ``globus_sdk.scopes.FlowsScopes.all``. (:pr:`NUMBER`)
--- a/src/globus_sdk/scopes/data.py
+++ b/src/globus_sdk/scopes/data.py
@@ -103,6 +103,7 @@ FlowsScopes = _FlowsScopeBuilder(
     "flows.globus.org",
     "eec9b274-0c81-4334-bdc2-54e90e689b9a",
     known_url_scopes=[
+        "all",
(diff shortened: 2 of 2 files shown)


sirosen (approved):
At first, I wanted to hold this change back because I thought we should wait until after the release which adds support, but since the scope exists, it's at least non-harmful to release in the SDK. And probably these will release near the same time.

derek-globus:
I had a similar thought; worst case scenario (realistically) is that this gets pushed out a day or two before service support if we have an urgent SDK patch to make.

More likely scenario is that I can cut a release of the SDK next wednesday along with the flows release.

PR #1017, opened 2024-07-29
Client based token caching

derek-globus:
Made some changes to the way token storage is handled to improve cross-script usage & expanded the config interface to make specifying the correct tokenstorage easier.

## Changes

-   ``app_name`` is no longer required (defaults to "DEFAULT")

-   Token storage now defaults to including the client id in the path.

    -   Old (unix) : ``~/.globus/app/{app_name}/tokens.json``

    -   New (unix): ``~/.globus/app/{client_id}/{app_name}/tokens.json``

    -   Old (win): ``~\AppData\Local\globus\app\{app_name}\tokens.json``

    -   New (win): ``~\AppData\Local\globus\app\{client_id}\{app_name}\tokens.json``

-   ``GlobusAppConfig.token_storage`` now accepts shorthand string references:
     ``"json"`` to use a ``JSONTokenStorage``, ``"sqlite"`` to use a
     ``SQLiteTokenStorage`` and ``"memory"`` to use a ``MemoryTokenStorage``.

-   ``GlobusAppConfig.token_storage`` also now accepts a ``TokenStorageProvidable``,
     a class with a ``for_globus_app(...) -> TokenStorage`` class method.

----
📚 Documentation preview 📚: https://globus-sdk-python--1017.org.readthedocs.build/en/1017/

### DIFF
--- a/changelog.d/20240728_220953_derek_client_based_token_caching.rst
+++ b/changelog.d/20240728_220953_derek_client_based_token_caching.rst
@@ -1,0 +1,24 @@
+
+Changed
+~~~~~~~
+
+-   Changed the experimental ``GlobusApp`` class in the following way (:pr:`NUMBER`):
+
+    -   ``app_name`` is no longer required (defaults to "DEFAULT")
+
+    -   Token storage now defaults to including the client id in the path.
+
+        -   Old (unix) : ``~/.globus/app/{app_name}/tokens.json``
+
+        -   New (unix): ``~/.globus/app/{client_id}/{app_name}/tokens.json``
+
+        -   Old (win): ``~\AppData\Local\globus\app\{app_name}\tokens.json``
+
+        -   New (win): ``~\AppData\Local\globus\app\{client_id}\{app_name}\tokens.json``
+
+    -   ``GlobusAppConfig.token_storage`` now accepts shorthand string references:
+        ``"json"`` to use a ``JSONTokenStorage``, ``"sqlite"`` to use a
+        ``SQLiteTokenStorage`` and ``"memory"`` to use a ``MemoryTokenStorage``.
+
+    -   ``GlobusAppConfig.token_storage`` also now accepts a ``TokenStorageProvidable``,
+        a class with a ``for_globus_app(...) -> TokenStorage`` class method.
--- a/changelog.d/20240728_220953_derek_client_based_token_caching.rst
+++ b/changelog.d/20240728_220953_derek_client_based_token_caching.rst
@@ -20,5 +20,8 @@ Changed
         ``"json"`` to use a ``JSONTokenStorage``, ``"sqlite"`` to use a
         ``SQLiteTokenStorage`` and ``"memory"`` to use a ``MemoryTokenStorage``.
 
-    -   ``GlobusAppConfig.token_storage`` also now accepts a ``TokenStorageProvidable``,
+    -   ``GlobusAppConfig.token_storage`` also now accepts a ``TokenStorageProvider``,
         a class with a ``for_globus_app(...) -> TokenStorage`` class method.
+
+    -   Renamed the experimental ``FileTokenStorage`` attribute ``.filename`` to
+        ``.filepath``.
--- a/src/globus_sdk/experimental/globus_app/globus_app.py
+++ b/src/globus_sdk/experimental/globus_app/globus_app.py
@@ -196,26 +196,13 @@ class GlobusApp(metaclass=abc.ABCMeta):
         self.app_name = app_name
         self.config = config
 
-        if login_client:
-            if client_id or client_secret:
-                raise GlobusSDKUsageError(
-                    "login_client is mutually exclusive with client_id and "
-                    "client_secret."
-                )
-            if login_client.environment != self.config.environment:
-                raise GlobusSDKUsageError(
-                    f"[Environment Mismatch] The login_client's environment "
-                    f"({login_client.environment}) does not match the GlobusApp's "
-                    f"configured environment ({self.config.environment})."
-                )
-
-        self.client_id: UUIDLike | None
-        if login_client:
-            self._login_client = login_client
-            self.client_id = login_client.client_id
-        else:
-            self.client_id = client_id
-            self._initialize_login_client(client_secret)
+        self.client_id, self._login_client = self._determine_client_info(
+            app_name=app_name,
+            config=config,
+            client_id=client_id,
+            client_secret=client_secret,
+            login_client=login_client,
+        )
 
         self._scope_requirements = self._setup_scope_requirements(scope_requirements)
 
@@ -255,10 +242,66 @@ class GlobusApp(metaclass=abc.ABCMeta):
             for resource_server, scopes in scope_requirements.items()
         }
 
-    @abc.abstractmethod
-    def _initialize_login_client(self, client_secret: str | None) -> None:
+    def _determine_client_info(
+        self,
+        app_name: str,
+        config: GlobusAppConfig,
+        login_client: AuthLoginClient | None,
+        client_id: UUIDLike | None,
+        client_secret: str | None,
+    ) -> tuple[UUIDLike, AuthLoginClient]:
         """
-        Initializes self._login_client to be used for making authorization requests.
+        Extracts a client_id and login_client from GlobusApp initialization parameters.
+
+        Depending on which parameters were provided, this method will either:
+            1.  Create a new login client from the supplied credentials.
+                *   The actual initialization is performed by the subclass using the
+                    abstract method: _initialize_login_client``.
+            2.  Extract the client_id from a supplied login_client.
+
+        :returns: tuple of client_id and login_client
+        """
+        if login_client and client_id:
+            msg = "Mutually exclusive parameters: client_id and login_client."
+            raise GlobusSDKUsageError(msg)
+
+        if login_client:
+            # User provided an explicit login client, extract the client_id.
+            if login_client.client_id is None:
+                msg = "An explicit login_client must have a discoverable client_id."
+                raise GlobusSDKUsageError(msg)
+            if login_client.environment != config.environment:
+                raise GlobusSDKUsageError(
+                    "[Environment Mismatch] The login_client's environment "
+                    f"({login_client.environment}) does not match the GlobusApp's "
+                    f"configured environment ({config.environment})."
+                )
+
+            return login_client.client_id, login_client
+
+        elif client_id:
+            # User provided an explicit client_id, construct a login client
+            login_client = self._initialize_login_client(
+                app_name, config, client_id, client_secret
+            )
+            return client_id, login_client
+
+        else:
+            raise GlobusSDKUsageError(
+                "Could not to set up a globus login client. One of client_id or "
+                "login_client is required."
+            )
+
+    @abc.abstractmethod
+    def _initialize_login_client(
+        self,
+        app_name: str,
+        config: GlobusAppConfig,
+        client_id: UUIDLike,
+        client_secret: str | None,
+    ) -> AuthLoginClient:
+        """
+        Initializes and returns an AuthLoginClient to be used in authorization requests.
         """
 
     @abc.abstractmethod
@@ -428,24 +471,25 @@ class UserApp(GlobusApp):
 
         return provider.for_globus_app(app_name, login_client, config)
 
-    def _initialize_login_client(self, client_secret: str | None) -> None:
-        if self.client_id is None:
-            raise GlobusSDKUsageError(
-                "One of either client_id or login_client is required."
-            )
-
+    def _initialize_login_client(
+        self,
+        app_name: str,
+        config: GlobusAppConfig,
+        client_id: UUIDLike,
+        client_secret: str | None,
+    ) -> AuthLoginClient:
         if client_secret:
-            self._login_client = ConfidentialAppAuthClient(
-                app_name=self.app_name,
-                client_id=self.client_id,
+            return ConfidentialAppAuthClient(
+                app_name=app_name,
+                client_id=client_id,
                 client_secret=client_secret,
-                environment=self.config.environment,
+                environment=config.environment,
             )
         else:
-            self._login_client = NativeAppAuthClient(
-                app_name=self.app_name,
-                client_id=self.client_id,
-                environment=self.config.environment,
+            return NativeAppAuthClient(
+                app_name=app_name,
+                client_id=client_id,
+                environment=config.environment,
             )
 
     def _initialize_authorizer_factory(self) -> None:
@@ -531,17 +575,24 @@ class ClientApp(GlobusApp):
             config=config,
         )
 
-    def _initialize_login_client(self, client_secret: str | None) -> None:
-        if not (self.client_id and client_secret):
+    def _initialize_login_client(
+        self,
+        app_name: str,
+        config: GlobusAppConfig,
+        client_id: UUIDLike,
+        client_secret: str | None,
+    ) -> AuthLoginClient:
+        if not client_secret:
             raise GlobusSDKUsageError(
-                "Either login_client or both client_id and client_secret are required"
+                "A ClientApp requires a client_secret to initialize its own login "
+                "client."
             )
 
-        self._login_client = ConfidentialAppAuthClient(
-            client_id=self.client_id,
+        return ConfidentialAppAuthClient(
+            client_id=client_id,
             client_secret=client_secret,
-            app_name=self.app_name,
-            environment=self.config.environment,
+            app_name=app_name,
+            environment=config.environment,
         )
 
     def _initialize_authorizer_factory(self) -> None:
--- a/src/globus_sdk/experimental/globus_app/globus_app.py
+++ b/src/globus_sdk/experimental/globus_app/globus_app.py
@@ -49,7 +49,7 @@ else:
 
 
 @runtime_checkable
-class TokenStorageProvidable(Protocol):
+class TokenStorageProvider(Protocol):
     @classmethod
     def for_globus_app(
         cls, client_id: UUIDLike, app_name: str, config: GlobusAppConfig, namespace: str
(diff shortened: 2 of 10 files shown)


src/globus_sdk/experimental/globus_app/globus_app.py
```diff
@@ -144,7 +140,7 @@ class GlobusApp(metaclass=abc.ABCMeta):
 
     def __init__(
         self,
-        app_name: str,
+        app_name: str = "DEFAULT",
```
derek-globus, line 143: Open Question: Is there a better default than this?

This is the name that will be used by default in (1) token-based file caching (after being 'slugified'), (2) http request headers, and ultimately (3) native login flow prefilled named grants.

My initial version of this was named `"Unnamed Globus App"` (becomes /unamed-globus-app/tokens.json) but I swapped it to `"DEFAULT"`. Curious what opinion reviewers have.

kurtmckee: I think "Unnamed Globus App" looks good to me. It's longer but includes the keywords necessary for developers who care to figure out where it's coming from.

ada-globus: I think I quite like this change, FWIW.

src/globus_sdk/experimental/globus_app/globus_app.py
```diff
@@ -98,9 +93,10 @@ class GlobusAppConfig:
         be initialized with the app's ``login_client`` and this config's
         ``request_refresh_tokens``.
         If not given the default behavior will depend on the type of ``GlobusApp``.
-    :param token_storage: a ``TokenStorage`` instance or class that will
-        be used for storing token data. If not passed a ``JSONTokenStorage``
-        will be used.
+    :param token_storage: The interface for ``GlobusApp`` to store and retrieve tokens.
```
ada-globus, line 96: Would this be very slightly better as:

```suggestion
    :param token_storage: The interface for a ``GlobusApp`` to store and retrieve tokens.
```

src/globus_sdk/experimental/globus_app/globus_app.py
```diff
@@ -220,11 +200,102 @@ def __init__(
...
+        :returns: tuple of client_id and login_client
+        :raises: GlobusSDKUsageError if a single client ID or login client could not be
+            definitively resolved.
+        """
+        if login_client and client_id:
+            msg = "Mutually exclusive parameters: client_id and login_client."
+            raise GlobusSDKUsageError(msg)
+
+        if login_client:
+            # User provided an explicit login client, extract the client_id.
```
ada-globus, line 230: Any reason why not:
```suggestion
        if login_client:
            if client_id:
                msg = "Mutually exclusive parameters: client_id and login_client."
                raise GlobusSDKUsageError(msg)

            # User provided an explicit login client, extract the client_id.
```

...my preference is not strong, but I was just slightly surprised to see `login_client` checked twice in this way.

derek-globus: Just a stylistic choice.

I laid out the code in this pattern in an attempt to establish that neither `login_client` nor `client_id` were the "standard" or "expected" resolution paths. Your suggestion makes it seem a bit more like `login_client` is the standard resolution (mutex checking is a subcase of login_client existence).

My preference is also not strong here. If the check in question had the capability to be expensive at all I'd definitely swap but I can't imagine `bool(login_client)` doing much more than a `None` identity check.

src/globus_sdk/experimental/globus_app/globus_app.py
derek-globus: > I think I quite like this change, FWIW.

Do you mean the change to name it `"DEFAULT"`, the change to name if `"Unnamed Globus App"` or the decision at large to not require an explicit app name?

src/globus_sdk/experimental/tokenstorage/base.py
```diff
@@ -148,3 +180,47 @@ def user_only_umask(self) -> t.Iterator[None]:
             yield
         finally:
             os.umask(old_umask)
+
+
+def _default_globus_app_filename(
```
ada-globus, line 185: This isn't really the _filename_, though—it's the full path, right?

src/globus_sdk/experimental/tokenstorage/base.py
```diff
@@ -148,3 +180,47 @@ def user_only_umask(self) -> t.Iterator[None]:
...
+    client_id: UUIDLike, app_name: str, environment: str
+) -> str:
+    r"""
+    Construct a default TokenStorage filename for a globus app.
```
ada-globus, line 189:
```suggestion
    Construct a default TokenStorage filename for a GlobusApp.
```

src/globus_sdk/experimental/globus_app/globus_app.py
ada-globus: I couldn't see Kurt's comment when I wrote that cause I was in code view. I like "default", but my opinion is not that strong, I can stand down on it if Kurt feels strongly.

src/globus_sdk/experimental/globus_app/globus_app.py
ada-globus: I hadn't read it that way, but I'm totally fine with that reasoning!

ada-globus:
I've got some reservation around the use of "filename" when we're really looking for the path... but I also acknowledge this change potentially has broader impact than just the site where I marked it up. I'd like to advocate for a change, but if others prefer filename, I'll let it go without a fuss!

src/globus_sdk/experimental/globus_app/globus_app.py
```diff
@@ -37,35 +42,17 @@
...
-        # try to get the app data dir, preferring the local appdata
-        datadir = os.getenv("LOCALAPPDATA", os.getenv("APPDATA"))
-        if not datadir:
-            home = os.path.expanduser("~")
-            datadir = os.path.join(home, "AppData", "Local")
-        return os.path.join(datadir, "globus", "app", app_name, filename)
-    else:
-        return os.path.expanduser(f"~/.globus/app/{app_name}/{filename}")
+@runtime_checkable
+class TokenStorageProvidable(Protocol):
```
sirosen, line 51: "Providable" is an unusual relationship to declare. 
It looks to me like this is a "provider", "factory", "generator", etc. Is that understanding correct?

A naming scheme I've seen used for protocols in places like typeshed is `SupportsX`. So maybe this could be `SupportsAppTokenStorage`?

I'm not exactly against the current name, but I don't understand it. This seems to me like a case we should clarify, and possibly rename of we decide that there's a better choice.

src/globus_sdk/experimental/tokenstorage/base.py
```diff
@@ -148,3 +180,47 @@ def user_only_umask(self) -> t.Iterator[None]:
...
+
+
+def _slugify_app_name(app_name: str) -> str:
+    """
+    Slugify a globus app name for use in a file path.
+
+    This "slugification" will replace spaces with hyphens and lowercase the string so:
+      "My App" -> "my-app"
+    """
+    return app_name.replace(" ", "-").lower()
```
kurtmckee, line 226: `"My  App"` will become `"my--app"`, which is maybe okay. Does Auth have restrictions on the characters used?

For example, is there anything preventing question marks or angle brackets before this potentially reaches the filesystem?

tests/unit/experimental/globus_app/test_globus_app.py
```diff
@@ -72,28 +73,27 @@ def test_user_app_native():
...
     assert user_app.app_name == "test-app"
     assert user_app._login_client == mock_client
+    assert user_app.client_id == "mock-client_id"
 
 
 def test_user_app_no_client_or_id():
-    with pytest.raises(GlobusSDKUsageError) as exc:
+    msg = (
+        "Could not to set up a globus login client. One of client_id or "
+        "login_client is required."
```
kurtmckee, line 87: I didn't immediately see where this was raised in the code, but "Could not to set up" needs to be addressed in the original location.

```suggestion
        "Could not set up a Globus login client. "
        "One of client_id or login_client is required."
```

tests/unit/experimental/globus_app/test_globus_app.py
```diff
@@ -113,14 +113,39 @@ def test_user_app_default_token_storage():
...
     token_storage = user_app._authorizer_factory.token_storage._token_storage
     assert isinstance(token_storage, JSONTokenStorage)
+
     if os.name == "nt":
         # on the windows-latest run this was
-        # C:\Users\runneradmin\AppData\Roaming\globus\app\test-app\tokens.json
-        assert "\\globus\\app\\test-app\\tokens.json" in token_storage.filename
+        # C:\Users\runneradmin\AppData\Roaming\globus\app\mock_client_id\test-app\tokens.json
+        expected = "\\globus\\app\\mock_client_id\\test-app\\tokens.json"
+        assert expected in token_storage.filename
```
kurtmckee, line 121:
```suggestion
        assert token_storage.filename.endswith(expected)
```

kurtmckee (changes requested):

src/globus_sdk/experimental/globus_app/globus_app.py
kurtmckee: I don't feel strongly, but I think `"DEFAULT"` versus `"Unnamed Globus App"` is comparable to raising `Exception` versus `ZeroDivisionError`, insofar as it gives developers a clear indication what they need to be looking for.

ada-globus: Hmm, that's a good point.

I'd read this slightly differently, though: since this change makes the app name fully optional, calling it the "default" app is not only accurate, it's also precise. _But_ your point is well taken, and I do see the value of the alternative. It just yields slightly clunkier paths to the credentials, which is a _very modest_ concession.

I'm happy enough with either option, I'll step back from this one and let y'all go with what you prefer!

changelog.d/20240728_220953_derek_client_based_token_caching.rst
```diff
@@ -0,0 +1,24 @@
+
+Changed
+~~~~~~~
+
+-   Changed the experimental ``GlobusApp`` class in the following way (:pr:`NUMBER`):
+
+    -   ``app_name`` is no longer required (defaults to "DEFAULT")
+
+    -   Token storage now defaults to including the client id in the path.
```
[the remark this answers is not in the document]
ada-globus: I had suggested that in conversation with Derek as well, but I think the issue is that if you don't have the app name, then every app using that client will clobber each other app's tokens.

src/globus_sdk/experimental/globus_app/globus_app.py
kurtmckee: I've left separate feedback that I'm concerned including the app name in the token cache path may not be ideal anyway. If that feedback is accepted, then the clunkiness concern might be fully addressed.

derek-globus: I think Kurt's convinced me; I'll rename back to `Unnamed Globus App`.

src/globus_sdk/experimental/tokenstorage/base.py
derek-globus: Sure, I can rename.

src/globus_sdk/experimental/globus_app/globus_app.py
derek-globus: Yeah, I'm not entirely sure why I went with "Providable" in the end. It is a Provider but in Protocol form so I froze.

I added a similar thing to LoginFlowManagers in pr-1018 & called that one a `LoginFlowManagerProvider`. Are you happy with that?

changelog.d/20240728_220953_derek_client_based_token_caching.rst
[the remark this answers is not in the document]
derek-globus: Yeah it's a balancing act in my head.

It's true that in those situations, we'd effectively invalidate a user's token cache but the problem of token clobbering is much more of a common case that I think we should design around.

Consider a user who has 2 scripts both using the same native client id, `list_flows.py` and `update_flow.py`. `list_flows.py` declares a flows requirements for `FlowsScopes.view_flows` whereas `update_flow.py` declares a requirement for `FlowsScopes.manage_flows`.

If we only delineate by client id then any time the user switches between running the two scripts they'll be overriding the shared token location (only one token is stored per resource server), requiring a new login flow for the next time the other script is run. That user represents a more common source of annoyance that would result from not delineating by app_name.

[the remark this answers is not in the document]
kurtmckee: I don't yet understand why that would be a problem. It's the same client ID, so I would expect the tokens to be the shared.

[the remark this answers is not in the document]
derek-globus: The token infrastructure is shared but each script defines different scope requirements for those tokens.

So when `list_flows.py` goes to retrieve cached tokens it pulls the access token & scope associated with that token (manage_flows), evaluates it against it's defined scope requirement (view_flows) and determines that the token is invalid, so a new one is required.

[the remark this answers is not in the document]
kurtmckee: Ah, I see. (And, sorry, your response wasn't visible to me when I said "I don't yet understand" -- that was directed toward Ada's response!)

Thank you both for explaining!

src/globus_sdk/experimental/tokenstorage/base.py
derek-globus: Auth does not have any obvious restrictions for us to replicate locally ([thread w/ Josh](https://globus.slack.com/archives/C03N0A8J7/p1722377481158659)).

I don't take issue with `"My  App"` -> `"my-app"`.
We're doing some very modest modification to the app name to present directories a little bit cleaner but we shouldn't be doing much at all here.

kurtmckee: I'd like to strengthen this. Could I recommend something like:

```python
import re
import unicodedata

app_name = "I <3 Globus!"

path_name = app_name
path_name = "".join(
    c if unicodedata.category(c).startswith(("L", "N")) else "-" for c in path_name
)
path_name = re.sub("-+", "-", path_name).strip("-") or f"app-name-{len(app_name)}"
path_name = path_name.lower()

# path_name == "i-3-globus"
```

(I've broken the above into separate conceptual transformation steps.)

I don't want to have users accidentally creating directories with colons or brackets, and restricting the Unicode character class to letters and numbers seems permissive enough. The other transforms (replacing runs of hyphens, and lowercasing the whole string) aren't strictly important to me, but disallowing forbidden filesystem characters matters a lot to me.

derek-globus: Updated the slugify method to meet the following rules:

* Reserved filesystem characters are replaced with a '+'. ('a?' -> 'a+')
* Periods and Spaces are replaced with a '-'. ('a. b' -> 'a--b')
* Control characters are removed. ('a\0b' -> 'ab')
* The string is lowercased. ('AB' -> 'ab')

Additionally an error is raised if the resulting slug is 
* The empty string
* A reserved windows filename ('CON', 'NUL', 'LPT1', etc.)

src/globus_sdk/experimental/tokenstorage/base.py
```diff
@@ -216,11 +225,39 @@ def _default_globus_app_filepath(
...
 def _slugify_app_name(app_name: str) -> str:
     """
     Slugify a globus app name for use in a file path.
 
-    This "slugification" will replace spaces with hyphens and lowercase the string so:
-      "My App" -> "my-app"
+    * Reserved filesystem characters are replaced with a '+'. ('a?' -> 'a+')
+    * Periods and Spaces are replaced with a '-'. ('a. b' -> 'a--b')
+    * Control characters are removed. ('a\0b' -> 'ab')
+    * The string is lowercased. ('AB' -> 'ab')
```
kurtmckee, line 241: :heart:

src/globus_sdk/experimental/tokenstorage/base.py
```diff
@@ -216,11 +225,39 @@ def _default_globus_app_filepath(
...
-    return app_name.replace(" ", "-").lower()
+    app_name = _RESERVED_FS_CHARS.sub("+", app_name)
+    app_name = app_name.replace(".", "-").replace(" ", "-")
+    app_name = "".join(c for c in app_name if c.isprintable())
+    app_name = app_name.lower()
+
+    if _RESERVED_FS_NAMES.fullmatch(app_name):
+        msg = (
+            f'App name results in a reserved filename ("{app_name}"). Please choose a'
+            "different name."
```
kurtmckee, line 255: I recommend breaking at sentences; my experience has been that it's easier to find things when copied-and-pasted error messages aren't broken midway through the line.

```suggestion
            f'App name results in a reserved filename ("{app_name}"). '
            "Please choose a different name."
```

tests/unit/experimental/test_token_storage.py
```diff
@@ -0,0 +1,51 @@
...
+
+def test_app_name_slugification_removes_control_characters():
+    app_name = "glob\0\n\t\rus"
+    expected = "globus"
+    actual_slug = _slugify_app_name(app_name)
+    assert actual_slug == expected
+
+
+def test_app_name_slugification_rejects_empty_name():
+    with pytest.raises(GlobusSDKUsageError):
```
kurtmckee, line 35:
```suggestion
    with pytest.raises(GlobusSDKUsageError, match="name is empty"):
```

tests/unit/experimental/test_token_storage.py
```diff
@@ -0,0 +1,51 @@
...
+        _slugify_app_name("\n\t\r")
+
+
+@pytest.mark.parametrize("app_name", ("CON", "PRN\n", "nul", "LPT9", "COM1"))
+def test_app_name_slugification_rejects_reserved_names(app_name):
+    # # https://stackoverflow.com/a/31976060
+    with pytest.raises(GlobusSDKUsageError):
```
kurtmckee, line 42:
```suggestion
    with pytest.raises(GlobusSDKUsageError, match="reserved filename"):
```

kurtmckee (approved):
I've suggested changes that I think are best practices, but approve the PR overall!

src/globus_sdk/experimental/tokenstorage/base.py
derek-globus: ¯\_(ツ)_/¯

I don't agree so much as a rule.
In this instance it can't hurt, so I'm happy to change. But while I agree it's an improvement to searching to have a sentence not split over two lines I won't always do that as I think other things, like line management & structure is more important.

kurtmckee (approved):

PR #1027, opened 2024-08-02
Split globus_app into multiple modules

derek-globus:
## What?
Split the constructs in `globus_app.py` into multiple smaller modules.
The existing module had gotten too large to manage efficiently (600 - 700 loc) 

New module structure
```
src/experimental/globus_app
  |- app.py - Core GlobusApp class
  |- user_app.py - UserApp class
  |- client_app.py - ClientApp class
  |- config.py - UserAppConfig dataclass + configuration defaults
  |- _types.py - Common basic types
```

----
📚 Documentation preview 📚: https://globus-sdk-python--1027.org.readthedocs.build/en/1027/


sirosen:
Get out of my head! Quit thinking what I'm thinking!

I'll start reading right away. Big :+1: for this direction.

derek-globus:
> Get out of my head! Quit thinking what I'm thinking!

I had those 3 big PRs outstanding and didn't want any merge conflicts. But nothing is open against these paths now, so I felt I could sneak them in.

Very open to feedback if you don't agree with this structure btw. My only strong take is an anti-monolith-module one which it sounds like we share.

sirosen:
> Very open to feedback if you don't agree with this structure btw. My only strong take is an anti-monolith-module one which it sounds like we share.

Anti-monolith is, IMO, the important part.

The one thing I'm not sure about is using a module name which matches the containing package name, so `globus_app/globus_app.py`. The issue is that the name `globus_app` can be slightly ambiguous to a reader. i.e. `from .globus_app import Foo` requires slightly more context to read -- you need to be cautious as a reader that you are reading that import in the right relative dir frame of reference.
It's a _very_ minor issue, but it is my preference to do something like `base_globus_app.py` just to make that slightly easier.

sirosen (approved):

sirosen (approved):

PR #1029, opened 2024-08-09
Use `all` in default_scope_requirements for Flows

sirosen:
Update the `default_scope_requirements` field for `FlowsClient` to
list (only) the `all` scope. This increases the capabilities of a
default-built FlowsClient to include creation of a Flow.

This appears to have been a minor oversight in the initial
`default_scope_requirements` build-out, not an intentional decision to
limit capabilities. The `all` scope is newer than
`default_scope_requirements` and is the most appropriate choice now.

----
📚 Documentation preview 📚: https://globus-sdk-python--1029.org.readthedocs.build/en/1029/

### DIFF
--- a/changelog.d/20240809_155147_sirosen_use_flows_all_scope.rst
+++ b/changelog.d/20240809_155147_sirosen_use_flows_all_scope.rst
@@ -1,0 +1,7 @@
+Changed
+~~~~~~~
+
+.. rubric:: Experimental
+
+- The ``default_scope_requirements`` for ``globus_sdk.FlowsClient`` has been
+  updated to list the Flows ``all`` scope. (:pr:`NUMBER`)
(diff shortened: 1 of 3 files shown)


kurtmckee (approved):

derek-globus (changes requested):
I’m not 100% on whether we should be defaulting to write access. It’s a balancing act as with anything between security posture and ease of use but it has some pretty big security implications if by default any script distributed using a GlobusApp requests access to not only see your flows but also create, edit, and delete them in your name.

sirosen:
We discussed and agreed not to change this -- at least not for now.

The PR was opened under the misconception that this was an oversight, but providing read-only access to services was an intentional choice (and holds across other services where applicable, e.g. Search).

sirosen:
Given that we have revisited the topic, and now believe that `all` is best for Flows, I'm going to merge based on the prior approval. (I take Derek reopening it as implicit approval.)

PR #1030, opened 2024-08-23
Create a separate and user-facing SpecificFlowScopeBuilder

sirosen:
The SpecificFlowClient includes internal logic for building the appropriate scope(s) for a given flow by ID.
This logic can only be accessed by users today by creating a SpecificFlowClient and then accessing its `scopes` attribute.

That usage pattern will remain supported, but it will now be possible to directly instantiate the appropriate `scopes` data via a dedicated class, `SpecificFlowScopeBuilder`.
This matches the pattern provided by GCS for building endpoint and collection scopes.

In support of this, the `globus_sdk.scopes` subpackage is refactored to split the `data` module into a subpackage with per-service modules.
Most of these are small, but the Flows and GCS components are able to include their relevant logic in an organized way.
This also moves the class-level stub for SpecificFlowClient into this package, as a private attribute of `SpecificFlowScopeBuilder`.

----
📚 Documentation preview 📚: https://globus-sdk-python--1030.org.readthedocs.build/en/1030/


src/globus_sdk/scopes/data/flows.py
```diff
@@ -0,0 +1,117 @@
...
+import typing as t
+
+from globus_sdk._types import UUIDLike
+
+from ..builder import ScopeBuilder, ScopeBuilderScopes
+
+
+class _FlowsScopeBuilder(ScopeBuilder):
+    """
+    The Flows Service breaks the scopes/resource server convention:
```
kurtmckee, line 12:
```suggestion
    The Flows service breaks the scopes/resource server convention:
```

src/globus_sdk/scopes/data/flows.py
```diff
@@ -0,0 +1,117 @@
...
+      its resource server is a domain name but its scopes are built around the client id
```
kurtmckee, line 13:
```suggestion
      its resource server is a domain name but its scopes are built around the client ID.
```

src/globus_sdk/scopes/data/flows.py
```diff
@@ -0,0 +1,117 @@
...
+    Given that there isn't a simple way to support this more generally
+      (and we shouldn't encourage supporting this more generally), this class serves to
+      build out the scopes accurately specifically for Flows
```
kurtmckee, line 16: Below, another private class uses full sentences so I'm recommending the same here.

```suggestion
      build out the scopes accurately specifically for Flows.
```

src/globus_sdk/scopes/data/flows.py
```diff
@@ -0,0 +1,117 @@
...
+    def __getattribute__(self, name: str) -> t.Any:
+        if name == "resource_server":
+            _raise_attr_error("resource_server")
+        return object.__getattribute__(self, name)
+
+
+def _raise_attr_error(name: str) -> t.NoReturn:
+    raise AttributeError(
+        f"It is not valid to attempt to access the {name} of the "
+        "SpecificFlowClient class. "
```
kurtmckee, line 88: As descriptive text, should this say "the {name} attribute"?

```suggestion
        f"It is not valid to attempt to access the {name} attribute "
        "of the SpecificFlowClient class. "
```

Similar question for the next line as well.

src/globus_sdk/scopes/data/timer.py
kurtmckee, line 1: I think this filename should be `timers.py`.

src/globus_sdk/scopes/data/timer.py
```diff
@@ -0,0 +1,12 @@
+from ..builder import ScopeBuilder
+
+TimerScopes = ScopeBuilder(
+    "524230d7-ea86-4a52-8312-86065a9e0417",
+    known_url_scopes=[
+        "timer",
+    ],
+)
+"""Globus Timer scopes.
```
kurtmckee, line 9: Although this should likely be named "TimersScopes", the name of the service should be updated.

```suggestion
"""Globus Timers scopes.
```

kurtmckee (changes requested):

src/globus_sdk/scopes/data/flows.py
sirosen: :+1: This was a lift-and-shift, but I don't object to tweaking this while we're here.

src/globus_sdk/scopes/data/flows.py
sirosen: Yeah, I can get onboard with that. The goal is to be clear to the user, and I can imagine this helping.
I'm also going to single-quote it, as `'{name}'`, so that the boundary between parts is a little clearer.

src/globus_sdk/scopes/data/timer.py
sirosen: I suppose so! :dizzy:  We'll need to deal with renaming more broadly, but I can happily rename the module for now (and update the docstring, below).

src/globus_sdk/scopes/data/__init__.py
```diff
@@ -0,0 +1,20 @@
+from .auth import AuthScopes
+from .flows import FlowsScopes, SpecificFlowScopeBuilder
+from .gcs import GCSCollectionScopeBuilder, GCSEndpointScopeBuilder
+from .groups import GroupsScopes, NexusScopes
+from .search import SearchScopes
+from .timer import TimerScopes
```
kurtmckee, line 6: I'm betting this is upset now.

```suggestion
from .timers import TimerScopes
```

kurtmckee (changes requested):

src/globus_sdk/scopes/data/__init__.py
sirosen: Oh, it is. I'm sure it is; good catch.

kurtmckee (approved):

PR #1033, opened 2024-08-26
Update experimental root docs

derek-globus:
Updated the experimental root documentation to establish a policy of leaving behind aliases in the experimental module whenever constructs are brought over into the main module.

----
📚 Documentation preview 📚: https://globus-sdk-python--1033.org.readthedocs.build/en/1033/

### DIFF
--- a/docs/experimental/index.rst
+++ b/docs/experimental/index.rst
@@ -1,16 +1,14 @@
-.. warning::
-
-    The ``experimental`` subpackage contains new interfaces that should be considered
-    unstable. While we will make an effort to provide deprecation warnings for code
-    that is graduating to an official module, the contents of this subpackage *may*
-    change at any time or be removed entirely.
-
-    **Use at your own risk.**
-
 .. experimental_root:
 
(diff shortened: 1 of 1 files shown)


ada-globus (approved):

sirosen (approved):
I'm good with this.

I'll note that we do declare that things can change and be renamed here, so I'm not sure what value exactly this offers to a downstream consumer (presumably consumers need to pin, right?). But it does us basically 0 harm.

derek-globus:
>I'll note that we do declare that things can change and be renamed here, so I'm not sure what value exactly this offers to a downstream consumer (presumably consumers need to pin, right?). But it does us basically 0 harm.

Yeah, the idea would be less about "ensuring we maintain exact compatibility", we very explicitly give ourselves credence to not do that here. Rather the goal would be "when we have a construct we are moving over wholesale, we'll leave an alias behind."

kurtmckee (approved):

m1yag1 (approved):

PR #1034, opened 2024-08-29
'add_app_data_access_scope' accepts iterables

sirosen:
`collection_id: UUIDLike` has been converted to
`collection_id: UUIDLike | Iterable[UUIDLike]`

All other changes here serve to instrument, document, and test this
change. The helper now iterates over collection IDs as appropriate.
We still check that the input is valid UUID data in order to helpfully
error earlier on malformed data.

---

This change was made in response to feedback about the typical usage around Transfer submission.
It's typical to have an iterable of standard mapped collections, which can contain 0, 1, or 2 elements.
Allowing for an iterable (and potentially an empty one at that) makes such usages simpler.

----
📚 Documentation preview 📚: https://globus-sdk-python--1034.org.readthedocs.build/en/1034/

### DIFF
--- a/src/globus_sdk/services/transfer/client.py
+++ b/src/globus_sdk/services/transfer/client.py
@@ -114,22 +114,27 @@ class TransferClient(client.BaseClient):
     default_scope_requirements = [Scope(TransferScopes.all)]
 
     def add_app_data_access_scope(
-        self, collection_id: UUIDLike | t.Iterable[UUIDLike]
+        self, collection_ids: UUIDLike | t.Iterable[UUIDLike]
     ) -> TransferClient:
         """
-        Add a dependent ``data_access`` scope for one or more given ``collection_id``
+        Add a dependent ``data_access`` scope for one or more given ``collection_ids``
         to this client's ``GlobusApp``. Useful for resolving ``ConsentRequired`` errors
         when using standard Globus Connect Server mapped collections.
 
-        ``collection_id`` must be for standard Globus Connect Server mapped
-        collections or app driven authentication flows will hit unknown scope errors.
+        .. warning::
+
+            This method must only be used on ``collection_ids`` for non-High-Assurance
+            GCS Mapped Collections.
+
+            Use on other collection types, e.g., on GCP Mapped Collections or any form
+            of Guest Collection, will result in "Unknown Scope" errors during the login
+            flow.
 
         Returns ``self`` for chaining.
 
         Raises ``GlobusSDKUsageError`` if this client was not initialized with an app.
 
-        :param collection_id: an ID for a standard Globus Connect Server mapped
-            collection, or an iterable of IDs
+        :param collection_ids: a collection ID or an iterable of IDs.
 
         .. tab-set::
 
@@ -165,18 +170,18 @@ class TransferClient(client.BaseClient):
                     ...
                     res = client.submit_transfer({})
         """  # noqa: E501
-        if isinstance(collection_id, (str, uuid.UUID)):
-            utils.check_uuid(collection_id, name="collection_id")
+        if isinstance(collection_ids, (str, uuid.UUID)):
+            utils.check_uuid(collection_ids, name="collection_ids")
             # wrap the collection_ids input in a list for consistent iteration below
-            collection_ids = [collection_id]
+            collection_ids_ = [collection_ids]
         else:
             # copy to a list so that ephemeral iterables can be iterated multiple times
-            collection_ids = list(collection_id)
-            for i, c in enumerate(collection_ids):
-                utils.check_uuid(c, name=f"collection_id[{i}]")
+            collection_ids_ = list(collection_ids)
+            for i, c in enumerate(collection_ids_):
+                utils.check_uuid(c, name=f"collection_ids[{i}]")
 
         base_scope = Scope(TransferScopes.all)
-        for coll_id in collection_ids:
+        for coll_id in collection_ids_:
             data_access_scope = Scope(
                 GCSCollectionScopeBuilder(str(coll_id)).data_access,
                 optional=True,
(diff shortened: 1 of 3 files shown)


src/globus_sdk/services/transfer/client.py
```diff
@@ -112,44 +113,76 @@ class TransferClient(client.BaseClient):
...
-        Add a dependent ``data_access`` scope for a given ``collection_id`` to this
-        client's ``GlobusApp``. Useful for resolving ``ConsentRequired`` errors
+        Add a dependent ``data_access`` scope for one or more given ``collection_id``
+        to this client's ``GlobusApp``. Useful for resolving ``ConsentRequired`` errors
         when using standard Globus Connect Server mapped collections.
-        ``collection_id`` must be for a standard Globus Connect Server mapped
-        collection or app driven authentication flows will hit unknown scope errors.
+
+        ``collection_id`` must be for standard Globus Connect Server mapped
+        collections or app driven authentication flows will hit unknown scope errors.
```
kurtmckee, line 125:
```suggestion
        collections or app-driven authentication flows will hit unknown scope errors.
```

src/globus_sdk/services/transfer/client.py
```diff
@@ -112,44 +113,76 @@ class TransferClient(client.BaseClient):
...
         """
-        Add a dependent ``data_access`` scope for a given ``collection_id`` to this
-        client's ``GlobusApp``. Useful for resolving ``ConsentRequired`` errors
+        Add a dependent ``data_access`` scope for one or more given ``collection_id``
+        to this client's ``GlobusApp``. Useful for resolving ``ConsentRequired`` errors
         when using standard Globus Connect Server mapped collections.
-        ``collection_id`` must be for a standard Globus Connect Server mapped
-        collection or app driven authentication flows will hit unknown scope errors.
+
+        ``collection_id`` must be for standard Globus Connect Server mapped
```
kurtmckee, line 124: "must _be for_"...perhaps "must _refer to_ GCS mapped collections"?

Also, what is meant by "_standard_ GCS mapped collections"? Are there mapped collections that are not standard?

src/globus_sdk/services/transfer/client.py
```diff
@@ -112,44 +113,76 @@ class TransferClient(client.BaseClient):
     scopes = TransferScopes
     default_scope_requirements = [Scope(TransferScopes.all)]
 
-    def add_app_data_access_scope(self, collection_id: UUIDLike) -> TransferClient:
+    def add_app_data_access_scope(
+        self, collection_id: UUIDLike | t.Iterable[UUIDLike]
```
kurtmckee, line 117: `collection_id` can now be an iterable? I'm don't love that this is now a singular ID that can be an iterable.

src/globus_sdk/services/transfer/client.py
```diff
@@ -112,44 +113,76 @@ class TransferClient(client.BaseClient):
...
+        ``collection_id`` must be for standard Globus Connect Server mapped
+        collections or app driven authentication flows will hit unknown scope errors.
 
         Returns ``self`` for chaining.
 
         Raises ``GlobusSDKUsageError`` if this client was not initialized with an app.
 
-        :param collection_id: UUID for a standard Globus Connect Server mapped
-            collection.
+        :param collection_id: an ID for a standard Globus Connect Server mapped
```
kurtmckee, line 131: Same as above...is the word "standard" needed here?

src/globus_sdk/services/transfer/client.py
```diff
@@ -112,44 +113,76 @@ class TransferClient(client.BaseClient):
...
+            collection, or an iterable of IDs
```
kurtmckee, line 132:
```suggestion
            collection, or an iterable of IDs.
```

kurtmckee (changes requested):

src/globus_sdk/services/transfer/client.py
sirosen: I considered alternatives -- including renaming to `collection_ids`, as I don't think that's a significant breaking change.

I don't know that I like any of the options very well.
e.g.
```python
    def add_app_data_access_scope(
        self, *collection_ids: UUIDLike
```

Looks right, but requires that callers unpack their iterables.

Is there a particular option you'd argue for? I don't think any choice here is perfect.

kurtmckee: I would lean toward calling it `collection_ids` and then upgrade a singular item to an iterable, but nothing immediately seems perfect. :disappointed:

sirosen: I'm planning to come back to this in an hour or two. I'll go with `collection_ids` -- it's a case where nothing will be absolutely perfect and after a few days away I'm okay with that.

src/globus_sdk/services/transfer/client.py
sirosen: "Standard" was used, I believe, in contrast with "High-Assurance". I think that _some_ docs use "standard" as the antonym for HA, but it's hardly consistent enough that I'd argue in favor of it.

I would tend to put this note into a warning, since it's a big caveat, which gives us more visual room and latitude to explain. So... does the following look right?
```
.. warning::

    This method must only be used on ``collection_ids`` for non-High-Assurance GCS Mapped Collections.
    Use on other collection types, e.g., on GCP Mapped Collections or any form of Guest Collection,
    will result in "Unknown Scope" errors during the login flow.
```

src/globus_sdk/services/transfer/client.py
sirosen: I think I will pare this text down to `ID of a collection or an iterable of IDs.` and let the surrounding context -- primarily my suggested warning -- handle the nuance around correct usage.

sirosen:
I've put together a change which synthesizes all of the feedback -- it wasn't worth applying piecemeal because there was some redenting and so forth.

kurtmckee (approved):

PR #1035, opened 2024-08-29
Refine and fix the container types in GlobusApp

sirosen:
For user-input interfaces, we should generally declare types which
express our constraints vis-a-vis mutability of container types.
That is, if we can accept a `Mapping[K, V]`, we should declare that as
our input constraint, rather than `dict[K, V]` or
`MutableMapping[K, V]`.

The reason for this is that the variance of `K` and `V` over an
immutable mapping is more permissive. Immutable container types tend
to be covariant, whereas mutable containers are invariant.
Invariance can incorrectly penalize users by being overly strict,
rejecting subtypes of `K` and `V`.

Therefore, fix all of the input points for GlobusApp which
unnecessarily use `dict` to use `MutableMapping` instead.

Additionally, make `__init__` methods consistent in declaring a return
type of `None`. `mypy` is now lenient, allowing omission of the return
type on `__init__`, but it considers a no-arg unannotated init to be
untyped. This can lead to surprising behaviors -- for consistency,
`__init__` should always be annotated with a return type.

A typing regression test is included for the usage which first raised
this issue.

----
📚 Documentation preview 📚: https://globus-sdk-python--1035.org.readthedocs.build/en/1035/

### DIFF
--- a/src/globus_sdk/experimental/globus_app/errors.py
+++ b/src/globus_sdk/experimental/globus_app/errors.py
@@ -24,7 +24,7 @@ class MissingIdentityError(IdentityValidationError):
 class IdentityMismatchError(IdentityValidationError):
     """The identity in a token response did not match the expected identity."""
 
-    def __init__(self, message: str, stored_id: UUIDLike, new_id: UUIDLike):
+    def __init__(self, message: str, stored_id: UUIDLike, new_id: UUIDLike) -> None:
         super().__init__(message)
         self.stored_id = stored_id
         self.new_id = new_id
@@ -33,7 +33,7 @@ class IdentityMismatchError(IdentityValidationError):
 class MissingTokenError(TokenValidationError):
     """No token stored for a given resource server."""
 
-    def __init__(self, message: str, resource_server: str):
+    def __init__(self, message: str, resource_server: str) -> None:
         super().__init__(message)
         self.resource_server = resource_server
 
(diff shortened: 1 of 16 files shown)


src/globus_sdk/experimental/globus_app/errors.py
```diff
@@ -50,7 +50,9 @@ def __init__(self, expires_at_seconds: int):
 class UnmetScopeRequirementsError(TokenValidationError):
     """The token stored for a given resource server is missing required scopes."""
 
-    def __init__(self, message: str, scope_requirements: dict[str, list[Scope]]):
+    def __init__(
+        self, message: str, scope_requirements: dict[str, list[Scope]]
```
kurtmckee, line 54: Did this need to be a mapping as well?

sirosen: I intentionally omitted that based on three facts (not so much "reasons" as considerations or factors).

1. It's an error constructor.
    Error construction isn't the typical user-facing interface (usually it's consumption of built instances). So there isn't as much to be gained in this case.

2. Changing this would mean that we need to make a runtime change.
    It would be easy enough to accept a `Mapping[str, Sequence[Scope]]` and then convert it with `{k: list(v) for k, v ...`, but I'm not convinced in this case.

3. The declared type does not include a union in the invariant types.
    These sorts of changes are much more significant where a union is used (since `list[int | str]` will reject `list[int]`, etc). Since there's no union here, there aren't many cases in which a type which looks like a subtype would be rejected. If you subclass `Scope` you can get there, but that's about it.

kurtmckee (approved):

PR #1039, opened 2024-09-04
Allow better customization of CLI LoginFlowManager

sirosen:
The interface is changed here away from init parameters and towards subclassing as the method of customization. This allows for deeper customization for users.

The changelog not only explains the change, but includes an example usage with `click`.


src/globus_sdk/experimental/login_flow_manager/command_line_login_flow_manager.py
```diff
@@ -111,22 +103,34 @@ def run_login_flow(
...
+        self.print_authorize_url(authorize_url)
+        auth_code = self.prompt_for_code()
 
         return self.login_client.oauth2_exchange_code_for_tokens(auth_code)
 
-    def _print_and_prompt(self, authorize_url: str) -> str:
-        # Prompt the user to authenticate
+    def print_authorize_url(self, authorize_url: str) -> None:
+        """
+        Prompt the user to authenticate, using the computed ``authorize_url``.
```
ada-globus, line 113:
```suggestion
        Prompt the user to authenticate using the provided ``authorize_url``.
```

src/globus_sdk/experimental/login_flow_manager/command_line_login_flow_manager.py
```diff
@@ -111,22 +103,34 @@ def run_login_flow(
...
                 """
             )
         )
 
-        # Request they to enter the auth code
-        return input(f"{self.code_prompt}\n").strip()
+    def prompt_for_code(self) -> str:
+        """
+        After the authorization URL is printed, ask the user to enter the
+        authorization code.
```
ada-globus, line 133: If we want to express the order in which this method will be executed relative to other operations, I wouldn't do it inside of the method (i.e., nothing internal to the method interacts with the printing of the authorization URL or controls ordering). I would advocate that we address either in `run_login_flow` or at the class level, but I can't place suggestions at either location via GH.

```suggestion
        Prompt the user to enter an authorization code.
```

ada-globus (changes requested):
Overall, I'm a 'yes', but I'd like to revise the comments so that they're specific to what's in their associated scope (rather than suggesting orchestration happening outside of the a oparticular method).

derek-globus (approved):
Ignoring the docstrings as Ada is already on top of that shaping, I like this a lot! Especially the click example is nice.

src/globus_sdk/experimental/login_flow_manager/command_line_login_flow_manager.py
sirosen: I'm fine with this rephrase. In concert with the `click` example (which I think we'll move or copy from the changelog at some point), the usage mode will be clear.

ada-globus (approved):

PR #1040, opened 2024-09-04
Make `GlobusApp.token_store` public and documented

sirosen:
In service of this, the docstrings for GlobusApp needed to be fixed.
The `__init__` docstring is not used in our sphinx configuration, and
is therefore moved into the class docstring, where it will be properly
loaded and used for parameter doc. This allows the renamed attr to be
documented as an instance var.

----
📚 Documentation preview 📚: https://globus-sdk-python--1040.org.readthedocs.build/en/1040/

### DIFF
--- a/src/globus_sdk/experimental/globus_app/app.py
+++ b/src/globus_sdk/experimental/globus_app/app.py
@@ -32,6 +32,27 @@ class GlobusApp(metaclass=abc.ABCMeta):
     A ``GlobusApp`` is accepted as an initialization parameter to any SDK-provided
     service client, automatically handling the client's default scope requirements and
     providing the client with an authorizer.
+
+    :param app_name: A string to identify this app. Used for default tokenstorage
+        location and in the future will be used to set user-agent when this app is
+        attached to a service client
+    :param login_client: An ``AuthLoginCLient`` that will be used for running
+        authentication flows. Different classes of ``GlobusApp`` may require
+        specific classes of ``AuthLoginClient``. ``Mutually exclusive with
+        ``client_id`` and ``client_secret``.
+    :param client_id: A client UUID used to construct an ``AuthLoginCLient`` for running
+        authentication flows. The type of ``AuthLoginCLient`` will depend on the
+        type of ``GlobusApp``. Mutually exclusive with ``login_client``.
+    :param client_secret: The value of the client secret for ``client_id`` if it uses
+        secrets. Mutually exclusive with ``login_client``.
+    :param scope_requirements: A dictionary of scope requirements indexed by
+        resource server. The dict value may be a scope, scope string, or list of
+        scopes or scope strings.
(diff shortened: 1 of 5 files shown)


ada-globus (approved):
It looks like there's still an assertion referencing the old private name in `test_user_app_creates_consent_client` causing a test failure, but otherwise this looks good!

src/globus_sdk/experimental/globus_app/app.py
```diff
@@ -32,6 +32,27 @@ class GlobusApp(metaclass=abc.ABCMeta):
...
+    :param client_secret: The value of the client secret for ``client_id`` if it uses
+        secrets. Mutually exclusive with ``login_client``.
+    :param scope_requirements: A dictionary of scope requirements indexed by
+        resource server. The dict value may be a scope, scope string, or list of
+        scopes or scope strings.
+    :param config: A ``GlobusAppConfig`` used to control various behaviors of this app.
+
+    :ivar token_store: The ``ValidatingTokenStorage`` containing tokens for the app.
+        Authorization mediated by the app will use this object, so modifying this will
+        impact clients which are defined to use the app whenever they fetch tokens.
```
derek-globus, line 55: I spy with my little eye a naming inconsistency between the [parameter passed](https://github.com/globus/globus-sdk-python/blob/main/src/globus_sdk/experimental/globus_app/config.py#L88) and the instance variable name. 

```
user_app = UserApp("my-app", client_id=..., config=GlobusAppConfig(token_storage="memory"))
user_app.token_store
```

There's a separate question of the TokenStorage class vs token_store namespace but the config difference is more concerning to me.

derek-globus (changes requested):

src/globus_sdk/experimental/globus_app/app.py
sirosen: We proposed `token_store` very casually as the name for this, without a ton of discussion.
Do we maybe prefer `token_storage`? That would lead to fewer interface changes and less of a chance of inconsistency.

Depending on one's perspective, either it's a good thing or a bad thing that it matches the subpackage naming.

derek-globus: The config parameter is the one that I care more about than the token interface.

I'd say, my vote order would be:

1. Use `token_storage` instead of `token_store` (there's just one less word for users to question "is this a different thing than a tokenstorage or tokenstorageadapater?")
2. Update config to use `token_store` as well.

ada-globus: Good catch! My order of preference matches Derek's.

sirosen: I just amended with this change. On review, I like `app.token_storage` better in multiple ways.

derek-globus (approved):

PR #1042, opened 2024-09-05
Replace app requirement exploration method

derek-globus:
## What?
Replace `app.get_scope_requirements(resource_server: str) -> tuple[Scope]` with the @property method `app.scope_requirements` which will create and return a deep copy of the internal scope requirements.

## Why?
`app.get_scope_requirements(...)` was added to give developers a lens into the aggregate requirements registered on a GlobusApp. By making resource_server a required parameter however, we mandated that a developer have foreknowledge of the scopes they are exploring which, at least in part, defeats the purpose of supporting exploration.

This new approach allows for iteration over all resource servers and leverages deepcopy to ensure that users don't retrieve and modify their scope requirements outside of the App defined interfaces (`app.add_scope_requirements(...)`)

----
📚 Documentation preview 📚: https://globus-sdk-python--1042.org.readthedocs.build/en/1042/

### DIFF
--- a/changelog.d/20240905_134232_derek_better_app_scope_inspection.rst
+++ b/changelog.d/20240905_134232_derek_better_app_scope_inspection.rst
@@ -1,0 +1,9 @@
+
+Changed
+~~~~~~~
+
+-   The experimental ``GlobusApp`` construct's scope exploration interface has changed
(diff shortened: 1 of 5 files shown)


sirosen (approved):
As we discussed, this seems like the right way to take the feedback we got.

I wish we didn't need deepcopy, but it feels like the right choice right now.

PR #1044, opened 2024-09-06
Move 'serializable' helpers and expose ValidationError

sirosen:
The initial goal of this changeset was to move the `serializable` helpers and validators out of `experimental` and into internal modules.
A commit-by-commit review (below) explains why it has grown in scope to include publicizing `ValidationError` and new lazy-import behavior on `exc`.

Commit series:

- 1st commit: make the change via a very minimal route. It is strictly-speaking correct, but it exposes an issue with `ValidationError`.
  Having moved from `experimental`, the symbol has actually changed from public (via `auth_requirements_error`) to private (under `_guards`).
  Callers of some class methods (especially `from_dict` on classes like `TokenData`) no longer have an export of this name to use in their except blocks.

- 2nd commit: make `ValidationError` public, moving it into `exc` and re-exporting it in `__init__.py`.
  It also moves `check_uuid` to `_guards.validators.uuidlike` as an internal nicety for maintainers.

- 3rd commit: add more unit tests (hooray!).

- 4th commit: testing a full `tox` matrix reveals that `_guards` now imports `requests`, breaking the lazy import tests which protect our import-time behavior.
  To resolve, `globus_sdk.exc` lazily imports the specific component which handles `requests` most directly.
  (Alternative solution: that module could lazy-import requests, but deferral in this location seemed cleaner to explain to users.)

- 5th commit: Add a changelog and one more test case (to hold to the promise/declaration of the changelog).

## Rationale for `globus_sdk.exc` changes

If we're uncomfortable with the level of change here for an internal component, we can take alternative paths, but many of the options are, IMO, strictly *worse* than this one.

For example, we could "just" define `ValidationError` in `_guards`.
But to do so, we need to make other compromises -- we don't define the error inthe same subpackage as our other errors, we can't inherit from `GlobusError`, etc.

The *motivation* for most of the workarounds we could use are to avoid changing `globus_sdk.exc`.
But I think that ignores the fact that `globus_sdk.exc` importing `requests` is, itself, a defect -- and not a necessary one at that -- and I therefore rejected that rationale for taking another "simpler" course.

### Added

- `globus_sdk.exc.ValidationError` is a new error type used by the SDK for certain cases of `ValueError`\s which are caused by invalid content. It can also be accessed as `globus_sdk.ValidationError`.

### Changed

- Imports of `globus_sdk.exc` now defer importing `requests` so as to reduce import-time performance impact the library is not needed.

  The following error classes are now lazily loaded even when `globus_sdk.exc` is imported:

  - `GlobusConnectionError`
  - `GlobusConnectionTimeoutError`
  - `GlobusTimeoutError`
  - `NetworkError`

----
📚 Documentation preview 📚: https://globus-sdk-python--1044.org.readthedocs.build/en/1044/


ada-globus (approved):

ada-globus:
I'm a bit biased since at least some of the decisions flowed out of our discussion around this, so rather unsurprisingly, I'm happy with this change :)

PR #1046, opened 2024-09-06
Refactor Authorization docs (leaving stub pages behind)

sirosen:
- Move authorizer documentation into authorization/
- Move scope docs into `authorization/`

---

This is a preparatory move serving three purposes.
(1) It's just good cleanup.
(2) The new structure will let us document the `consents` module in a logical place when it moves out of experimental.
(3) It makes room for GlobusApp, new tokenstorage docs, and other improvements.

----
📚 Documentation preview 📚: https://globus-sdk-python--1046.org.readthedocs.build/en/1046/


docs/authorization/globus_authorizers.rst
```diff
@@ -0,0 +1,91 @@
...
+    :member-order: bysource
+    :show-inheritance:
+
+``GlobusAuthorizer`` objects which have a static authorization header are all
+implemented using the static authorizer class:
+
+.. autoclass:: StaticGlobusAuthorizer
+    :members:
+    :member-order: bysource
+    :show-inheritance:
```
derek-globus, line 58: I'd consolidate the preamble of these two interfaces and invert the order (static comes before renewing because it's simpler and helps establish the concept to which renewing adds more complexity).

Something like

```suggestion
Authorizers within this SDK fall into two categories: 
 * "Static Authorizers" already contain all authorization data and simply format it into the proper authorization header. These all inherit from the ``StaticGlobusAuthorizer`` class.
 * "Renewing Authorizer" take some initial parameters but internally define a functional behavior to acquire new authorization data as necessary. These all inherit from the ``RenewingGlobusAuthorizer`` class.
 
.. autoclass:: StaticGlobusAuthorizer
    :member-order: bysource
    :show-inheritance:

.. autoclass:: RenewingAuthorizer
    :member-order: bysource
    :show-inheritance:
```

---

Separately, I'm not convinced we should be showing the members for static & renewing authorizer. Seems like the duplicate the method definitions defined above in `GlobusAuthorizer` and just serve to take up screen space.

docs/authorization/scopes_and_consents/index.rst
```diff
@@ -0,0 +1,18 @@
...
+
+For more information, see
+`this docs.globus.org explanation of scopes and consents
+<https://docs.globus.org/guides/overviews/clients-scopes-and-consents/>`_.
+
+.. toctree::
+    :maxdepth: 1
+
+    scopes
+    mutable_scopes
```
derek-globus, line 18: Move this to the top of the page (or signal in a more apparent way that this is a nav link structure).

Otherwise it looks at a glance to me like "these docs have been moved to docs.globus.org/guides/overview/clients_scopes_and_consents."

derek-globus (changes requested):
I don't know how much of this copy is new vs copy/pasted into a new location.

I like the new structure of it a lot but think that the page text itself needs some massaging to get to a consumable state. In the globus authorizers page in particular, we should structure to be as approachable as possible; there's a good chance, this is the first time that users are interacting with globus authorization.

docs/authorization/globus_authorizers.rst
```diff
@@ -0,0 +1,91 @@
...
+
+.. autoclass:: RefreshTokenAuthorizer
+    :members:
+    :member-order: bysource
+    :show-inheritance:
+
+.. autoclass:: ClientCredentialsAuthorizer
```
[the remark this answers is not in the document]
sirosen: This might have been order of implementation. I don't recall. :joy: 

Let's reorder as you suggested; I think that's much better.

docs/authorization/scopes_and_consents/index.rst
sirosen: Yeah, I'll rearrange. I don't know exactly how yet; let me get in there and see what looks right.

docs/authorization/scopes_and_consents/scopes.rst
```diff
@@ -0,0 +1,218 @@
...
+    import globus_sdk
+    from globus_sdk.scopes import TransferScopes
+
+    CLIENT_ID = "<YOUR_ID_HERE>"
+
+    client = globus_sdk.NativeAppAuthClient(CLIENT_ID)
+    client.oauth2_start_flow(requested_scopes=[TransferScopes.all])
+    ...
+
+As Client Class Attributes
```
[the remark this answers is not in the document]
sirosen: Yeah, this is relatively old doc, so I need to review it a bit.
I think GCS & Flows might need a special callout here.

docs/authorization/scopes_and_consents/scopes.rst
```diff
@@ -0,0 +1,218 @@
...
+    >>> print(str(foo))
+    foo[bar[baz]]
+    >>> print(bar.serialize())
+    bar[baz]
+    >>> alpha = Scope("alpha")
+    >>> alpha.add_dependency("beta", optional=True)
+    >>> print(str(alpha))
+    alpha[*beta]
+    >>> print(repr(alpha))
+    Scope("alpha", dependencies=[Scope("beta", optional=True)])
```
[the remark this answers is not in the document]
sirosen: pycon is Python Console. It's more appropriate for this case because it contains output.

docs/authorization/scopes_and_consents/scopes.rst
```diff
@@ -0,0 +1,218 @@
...
+
+.. note::
+
+    ``TimersScopes`` is also available under the legacy name ``TimerScopes``.
+
+.. autodata:: globus_sdk.scopes.data.TimersScopes
+    :annotation:
+
+.. autodata:: globus_sdk.scopes.data.TransferScopes
+    :annotation:
```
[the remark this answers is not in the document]
sirosen: Can we save debate about this for future changes? This is a lift and shift of existing docs which are built via the custom sphinx plugin. Changing this extensively is something I'm open to doing, but want to keep out of this scope of work.

docs/authorization/scopes_and_consents/index.rst
sirosen: I broke things apart into sections, so that the overview/background on scopes with the docs.globus.org link could go in a separate section. Take a look and LMK if we can mark this resolved!

docs/authorization/scopes_and_consents/scopes.rst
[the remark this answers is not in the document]
sirosen: I've changed the header to `As Client Attributes` and covered class attributes, instance attributes, and classmethod access -- the three patterns we have in current use. You should take a look, but I'm marking this specific thread resolved.

docs/authorization/scopes_and_consents/scopes.rst
```diff
@@ -0,0 +1,218 @@
...
+
+    # use the `resource_server` of a ScopeBuilder to grab the associated token
+    # data from the response
+    tokendata = token_response.by_resource_server[TransferScopes.resource_server]
+
+Scope objects
+-------------
+
+In order to support optional and dependent scopes, an additional type is
+provided by ``globus_sdk.scopes``: the ``Scope`` class.
```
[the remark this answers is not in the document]
sirosen: I've reworded, not only as you suggested but also to draw the connection that having such a model is useful for representing and manipulating dependent scopes. Take a look and see if it's satisfactory -- I think you'll agree that it's an improvement but maybe we need to refine it further.

docs/authorization/scopes_and_consents/scopes.rst
```diff
@@ -0,0 +1,246 @@
...
+
+    from globus_sdk.scopes import GroupsScopes, Scope
+
+    group_scope = Scope.deserialize(GroupsScopes.all)
+
+``Scope`` objects primarily provide three main pieces of functionality:
+
+    * parsing (deserializing)
+    * stringifying (serializing)
+    * dynamic scope tree building
```
ada-globus, line 133: I think this helps to build _any_ scope, right?

```suggestion
    * scope tree construction
```

docs/authorization/scopes_and_consents/scopes.rst
```diff
@@ -0,0 +1,246 @@
...
+
+Dynamic Scope Construction
+~~~~~~~~~~~~~~~~~~~~~~~~~~
```
ada-globus, line 136:
```suggestion
Scope Construction
~~~~~~~~~~~~~~~~~~
```

docs/authorization/scopes_and_consents/scopes.rst
```diff
@@ -0,0 +1,246 @@
...
+stringified representation.
+
+Serializing Scopes
+~~~~~~~~~~~~~~~~~~
+
+Whenever scopes are being sent to Globus services, they need to be encoded as
+strings. All scope objects support this by means of their defined
+``serialize`` method. Note that ``__str__`` for a ``Scope`` is just an
+alias for ``serialize``. For example, the following is valid usage to demonstrate
+``str()``, ``repr()``, and ``serialize()``:
```
ada-globus, line 167:
```suggestion
alias for ``serialize``. For example, the following is an example of
``str()``, ``repr()``, and ``serialize()`` usage:
```

ada-globus (approved):
I'm marking as "approved" to indicate my overall endorsement of the restructuring. I have a few modest change requests, but if you're okay with those suggestions, this puts you in a position to just apply and merge so we can keep things moving.

docs/authorization/scopes_and_consents/scopes.rst
```diff
@@ -0,0 +1,246 @@
...
+.. currentmodule:: globus_sdk.scopes
+
+Scopes and ScopeBuilders
+========================
+
+OAuth2 Scopes for various Globus services are represented by ``ScopeBuilder``
+objects.
+
+A number of pre-set scope builders are provided and populated with useful data,
+and they are also accessible via the relevant client classes.
```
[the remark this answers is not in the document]
sirosen: It's intended to mean that my level of proficiency in English is much lower than it may appear. 😉

docs/authorization/scopes_and_consents/scopes.rst
sirosen: Yes, after a manner. The tree of scopes is primarily meaningful and useful vis-a-vis dynamic dependencies. You can construct a scope tree which mirrors static dependencies, but I'm not sure that doing so serves much purpose.

sirosen: It's possible, but I don't think that the `Scope` object we provide here matches what's provided via the Auth scope management API.

Scopes have underlying IDs, descriptions, and other attributes which we aren't modeling here.
That said, I think this is a nicer API for definition, and it would be interesting to look at it as a way of defining / declaring our static dependency data.

PR #1048, opened 2024-09-16
Promote `gare` from `experimental`

sirosen:
The `auth_requirements_error` module is moved to `globus_sdk.gare` and docs and importts are updated.

Included in this change, `GlobusAuthRequirementsError` has been renamed to `GARE`, as the `Error` suffix to a name typically indicates that a class is an exception class.

Aliasing has been put in place from `experimental` to the new code location.

As a minor, related fix `_serializable.Serializable` now uses a `Self` type annotation, rather than a type var. This improves the doc rendering of the `from_dict` and `to_dict` methods, as they show the self-type rather than an unbound type variable T.

----
📚 Documentation preview 📚: https://globus-sdk-python--1048.org.readthedocs.build/en/1048/

### DIFF
--- a/docs/authorization/gare.rst
+++ b/docs/authorization/gare.rst
@@ -1,29 +1,29 @@
 Auth Requirements Errors
 ========================
 
-Globus Auth Requirements Error is a response format that conveys to a client any
+'Globus Auth Requirements Error' is a response format that conveys to a client any
 modifications to a session (i.e., "boosting") that will be required
 to complete a particular request.
 
-The ``globus_sdk.experimental.auth_requirements_error`` module provides a
-number of tools to make it easier to identify and handle these errors when they occur.
+The ``globus_sdk.gare`` module provides a number of tools to make it easier to
+identify and handle these errors when they occur.
 
-GlobusAuthRequirementsError
----------------------------
+GARE
+----
 
-The ``GlobusAuthRequirementsError`` class provides a model for working with Globus
-Auth Requirements Error responses.
+The ``GARE`` class provides a model for working with Globus Auth Requirements Error
+responses.
 
 Services in the Globus ecosystem may need to communicate authorization requirements
 to their consumers. For example, a service may need to instruct clients to have the user
-consent to an additional scope, ``"foo"``. In such a case, ``GlobusAuthRequirementsError``
-can provide serialization into the well-known Globus Auth Requirements Error format:
+consent to an additional scope, ``"foo"``. In such a case, ``GARE`` can provide
+serialization into the well-known Globus Auth Requirements Error format:
 
 .. code-block:: python
 
-    from globus_sdk.experimental.auth_requirements_error import GlobusAuthRequirementsError
+    from globus_sdk.gare import GARE
 
-    error = GlobusAuthRequirementsError(
+    error_doc = GARE(
         code="ConsentRequired",
         authorization_parameters=GlobusAuthorizationParameters(
             required_scopes=["foo"],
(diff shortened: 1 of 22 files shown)


docs/authorization/gare.rst
```diff
@@ -118,22 +115,38 @@ Requirements Error, you can use, e.g.,:
...
 .. code-block:: python
 
-    auth_requirements_error.utils.to_auth_requirements_error(
-        other_error
-    )  # GlobusAuthRequirementsError
-    auth_requirements_error.utils.to_auth_requirements_errors(
-        [other_error]
-    )  # [GlobusAuthRequirementsError, ...]
+    gare.to_auth_requirements_error(other_error)  # GARE
+    gare.to_auth_requirements_errors([other_error])  # [GARE, ...]
```
ada-globus, line 119: Given that we've gotten ride of the "Auth Requirements Error" name, I have a fairly strong preference that these become `to_gare` and `to_gares`, otherwise I think we're potentially introducing some inconsistency/confusion here.

I think maybe we alias the old method names, perhaps with a v4 deprecation warning?

sirosen: Yeah, I can include that rename and agree with that rationale. I'll come back to this later today.

docs/authorization/gare.rst
```diff
@@ -1,29 +1,29 @@
 Auth Requirements Errors
 ========================
 
-Globus Auth Requirements Error is a response format that conveys to a client any
+'Globus Auth Requirements Error' is a response format that conveys to a client any
```
ada-globus, line 4: I can't comment on the heading, which is outside of this change, but perhaps worth revising to "Globus Auth Requirements Errors (GAREs)"? I don't feel super strongly, but it means 'GARE' will show up in the navigation, which I think is a usability win.

src/globus_sdk/gare/_auth_requirements_error.py
```diff
@@ -11,9 +11,9 @@ class GlobusAuthorizationParameters(Serializable):
     Data class containing authorization parameters that can be passed during
     an authentication flow to control how the user will authenticate.
 
-    When used with a GlobusAuthRequirementsError this represents the additional
-    authorization parameters needed in order to complete a request that had
-    insufficient authorization state.
+    When used with a GARE  this represents the additional authorization
```
ada-globus, line 14:
```suggestion
    When used with a GARE this represents the additional authorization
```

src/globus_sdk/gare/_functional_api.py
```diff
@@ -77,12 +72,11 @@ def to_auth_requirements_error(
...
 def to_auth_requirements_errors(
-    errors: list[GlobusAPIError | ErrorSubdocument | dict[str, t.Any]]
-) -> list[GlobusAuthRequirementsError]:
+    errors: list[exc.GlobusAPIError | exc.ErrorSubdocument | dict[str, t.Any]]
+) -> list[GARE]:
     """
     Converts a list of GlobusAPIErrors, ErrorSubdocuments, or dicts into a list of
-    GlobusAuthRequirementsErrors by attempting to match each error to
-    GlobusAuthRequirementsError (preferred) or legacy variants.
+    GAREs  by attempting to match each error to GARE (preferred) or legacy variants.
```
ada-globus, line 79:
```suggestion
    GAREs by attempting to match each error to GARE (preferred) or legacy variants.
```

ada-globus (commented):
Apart from the suggestions above, I'm happy with how this came out 🙌

docs/authorization/gare.rst
sirosen: I'm happy to include this -- I needed to touch this file anyway, due to the function renames. I'll include with the other feedback updates.

ada-globus (approved):
LOL @ the plaintive wail in that last commit. Yay for the line-length win everywhere, including our other products! 😁

PR #1049, opened 2024-09-16
Replace MANIFEST.in with pyproject settings

sirosen:
The manifest file specified inclusion of HTML files in a specific package dir.

Add a sanity test which ensures that the file is in the built distribution, then replace this with pyproject.toml configuration.

----
📚 Documentation preview 📚: https://globus-sdk-python--1049.org.readthedocs.build/en/1049/

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,0 @@
-include src/globus_sdk/experimental/login_flow_manager/local_server_login_flow_manager/html_files/*
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,6 +53,7 @@ namespaces = false
 
 [tool.setuptools.package-data]
 globus_sdk = ["py.typed"]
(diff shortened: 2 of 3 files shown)


sirosen:
Interesting -- this built correctly for me locally with `rm -r build/; tox ...`, but not in CI.
I may be missing a need for an additional `pyproject.toml` setting, but I don't know why it worked for me in one context but not another. Marking as a draft while I investigate.

sirosen:
I've fixed the issue. I was getting some cached data under tox builds. Doing full reinstalls with `tox r -r -e ...` made the tests accurate for me, and allowed me to zero in on what was needed.

The commit message and PR comment have been rephrased to be more accurate: this does not simply remove a redundant config, but moves it from `MANIFEST.in` to `pyproject.toml`.

kurtmckee (approved):

PR #1051, opened 2024-09-17
Add response subtypes for different oauth2 grants

sirosen:
The initial goal of this work is to allow a function handling a token
response, e.g., in a tokenstorage interface, to determine the context
in which the original call was made. It is now possible to check

    if isinstance(response, OAuthRefreshTokenResponse): ...

in a generic token handler.

Additionally, this opens the way for distinct token response objects
to implement methods which vary based on the known properties of their
respective grant types. For example, OAuthRefreshTokenResponse could
raise a more informative error when `decode_id_token` is called and
no `id_token` is present.

----
📚 Documentation preview 📚: https://globus-sdk-python--1051.org.readthedocs.build/en/1051/

### DIFF
--- a/docs/services/auth.rst
+++ b/docs/services/auth.rst
@@ -61,10 +61,6 @@ Auth Responses
    :members:
    :show-inheritance:
 
-.. autoclass:: OAuthDependentTokenResponse
-   :members:
-   :show-inheritance:
-
 .. autoclass:: AuthorizationCodeTokenResponse
    :members:
    :show-inheritance:
@@ -74,6 +70,10 @@ Auth Responses
    :show-inheritance:
 
 .. autoclass:: ClientCredentialsTokenResponse
+   :members:
+   :show-inheritance:
+
+.. autoclass:: OAuthDependentTokenResponse
    :members:
    :show-inheritance:
 
--- a/docs/services/auth.rst
+++ b/docs/services/auth.rst
@@ -61,15 +61,15 @@ Auth Responses
    :members:
    :show-inheritance:
 
-.. autoclass:: AuthorizationCodeTokenResponse
+.. autoclass:: OAuthAuthorizationCodeResponse
    :members:
    :show-inheritance:
 
-.. autoclass:: RefreshTokenResponse
+.. autoclass:: OAuthRefreshTokenResponse
    :members:
    :show-inheritance:
 
-.. autoclass:: ClientCredentialsTokenResponse
+.. autoclass:: OAuthClientCredentialsResponse
    :members:
    :show-inheritance:
 
--- a/docs/services/auth.rst
+++ b/docs/services/auth.rst
@@ -65,6 +65,18 @@ Auth Responses
    :members:
    :show-inheritance:
 
+.. autoclass:: AuthorizationCodeTokenResponse
+   :members:
+   :show-inheritance:
+
+.. autoclass:: RefreshTokenResponse
+   :members:
+   :show-inheritance:
+
+.. autoclass:: ClientCredentialsTokenResponse
+   :members:
+   :show-inheritance:
+
 .. autoclass:: GetConsentsResponse
    :members:
    :show-inheritance:
--- a/src/globus_sdk/authorizers/client_credentials.py
+++ b/src/globus_sdk/authorizers/client_credentials.py
@@ -13,7 +13,7 @@ log = logging.getLogger(__name__)
 
 
 class ClientCredentialsAuthorizer(
-    RenewingAuthorizer["globus_sdk.ClientCredentialsTokenResponse"]
+    RenewingAuthorizer["globus_sdk.OAuthClientCredentialsResponse"]
 ):
     r"""
     Implementation of a RenewingAuthorizer that renews confidential app client
@@ -47,7 +47,7 @@ class ClientCredentialsAuthorizer(
         POSIX timestamp (i.e. seconds since the epoch)
     :param on_refresh: A callback which is triggered any time this authorizer fetches a
         new access_token. The ``on_refresh`` callable is invoked on the
-        :class:`globus_sdk.ClientCredentialsTokenResponse` object resulting from the
+        :class:`globus_sdk.OAuthClientCredentialsResponse` object resulting from the
         token being refreshed. It should take only one positional argument, the token
         response object.
         This is useful for implementing storage for Access Tokens, as the
@@ -63,7 +63,7 @@ class ClientCredentialsAuthorizer(
         access_token: str | None = None,
         expires_at: int | None = None,
         on_refresh: (
-            None | t.Callable[[globus_sdk.ClientCredentialsTokenResponse], t.Any]
+            None | t.Callable[[globus_sdk.OAuthClientCredentialsResponse], t.Any]
         ) = None,
     ):
         # values for _get_token_data
@@ -77,7 +77,7 @@ class ClientCredentialsAuthorizer(
 
         super().__init__(access_token, expires_at, on_refresh)
 
-    def _get_token_response(self) -> globus_sdk.ClientCredentialsTokenResponse:
+    def _get_token_response(self) -> globus_sdk.OAuthClientCredentialsResponse:
         """
         Make a request for new tokens, using a 'client_credentials' grant.
         """
@@ -86,7 +86,7 @@ class ClientCredentialsAuthorizer(
         )
 
     def _extract_token_data(
-        self, res: globus_sdk.ClientCredentialsTokenResponse
+        self, res: globus_sdk.OAuthClientCredentialsResponse
     ) -> dict[str, t.Any]:
         """
         Get the tokens .by_resource_server,
--- a/src/globus_sdk/authorizers/client_credentials.py
+++ b/src/globus_sdk/authorizers/client_credentials.py
@@ -48,7 +48,7 @@ class ClientCredentialsAuthorizer(
     :param on_refresh: A callback which is triggered any time this authorizer fetches a
         new access_token. The ``on_refresh`` callable is invoked on the
         :class:`globus_sdk.ClientCredentialsTokenResponse` object resulting from the
-        token being refreshed. It should take only one argument, the token response
+        token being refreshed. It should take only one positional argument, the token response
         object.
         This is useful for implementing storage for Access Tokens, as the
         ``on_refresh`` callback can be used to update the Access Tokens and
--- a/src/globus_sdk/authorizers/client_credentials.py
+++ b/src/globus_sdk/authorizers/client_credentials.py
@@ -79,7 +79,7 @@ class ClientCredentialsAuthorizer(
 
     def _get_token_response(self) -> globus_sdk.ClientCredentialsTokenResponse:
         """
-        Make a client credentials request for new tokens.
+        Make a request for new client credentials tokens.
         """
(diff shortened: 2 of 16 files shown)


docs/services/auth.rst
```diff
@@ -65,6 +65,18 @@ Auth Responses
...
 
+.. autoclass:: AuthorizationCodeTokenResponse
+   :members:
+   :show-inheritance:
+
+.. autoclass:: RefreshTokenResponse
+   :members:
+   :show-inheritance:
+
+.. autoclass:: ClientCredentialsTokenResponse
```
kurtmckee, line 76: Should these be alphabetical?

src/globus_sdk/authorizers/client_credentials.py
```diff
@@ -75,15 +77,17 @@ def __init__(
 
         super().__init__(access_token, expires_at, on_refresh)
 
-    def _get_token_response(self) -> OAuthTokenResponse:
+    def _get_token_response(self) -> globus_sdk.ClientCredentialsTokenResponse:
         """
-        Make a client credentials grant
+        Make a client credentials request for new tokens.
```
kurtmckee, line 82:
```suggestion
        Make a request for new client credentials tokens.
```

src/globus_sdk/authorizers/client_credentials.py
```diff
@@ -47,22 +47,24 @@ class ClientCredentialsAuthorizer(RenewingAuthorizer):
         POSIX timestamp (i.e. seconds since the epoch)
     :param on_refresh: A callback which is triggered any time this authorizer fetches a
         new access_token. The ``on_refresh`` callable is invoked on the
-        :class:`OAuthTokenResponse <globus_sdk.OAuthTokenResponse>`
-        object resulting from the token being refreshed. It should take only one
-        argument, the token response object.
+        :class:`globus_sdk.ClientCredentialsTokenResponse` object resulting from the
+        token being refreshed. It should take only one argument, the token response
```
kurtmckee, line 51: ?

```suggestion
        token being refreshed. It should take only one positional argument, the token response
```

src/globus_sdk/authorizers/client_credentials.py
```diff
@@ -3,18 +3,18 @@
...
 
-if t.TYPE_CHECKING:
-    from globus_sdk.services.auth import ConfidentialAppAuthClient, OAuthTokenResponse
-
 log = logging.getLogger(__name__)
 
 
-class ClientCredentialsAuthorizer(RenewingAuthorizer):
+class ClientCredentialsAuthorizer(
+    RenewingAuthorizer["globus_sdk.ClientCredentialsTokenResponse"]
```
kurtmckee, line 16: Why quoted?

src/globus_sdk/authorizers/refresh_token.py
```diff
@@ -10,7 +10,7 @@
 log = logging.getLogger(__name__)
 
 
-class RefreshTokenAuthorizer(RenewingAuthorizer):
+class RefreshTokenAuthorizer(RenewingAuthorizer["globus_sdk.RefreshTokenResponse"]):
```
kurtmckee, line 13: Why quoted?

src/globus_sdk/authorizers/refresh_token.py
```diff
@@ -78,14 +77,14 @@ def __init__(
 
         super().__init__(access_token, expires_at, on_refresh)
 
-    def _get_token_response(self) -> globus_sdk.OAuthTokenResponse:
+    def _get_token_response(self) -> globus_sdk.RfreshTokenResponse:
```
kurtmckee, line 80: Why was this not getting caught by our linters?

```suggestion
    def _get_token_response(self) -> globus_sdk.RefreshTokenResponse:
```

kurtmckee (changes requested):

docs/services/auth.rst
sirosen: Arguably? Either alphabetical or in some other reasonable sorted order.
I was trying to put them in order from "most used to least used", but I actually think that necessitates putting the OAuthDependentTokenResponse below these. Would you be in favor of such a change?

src/globus_sdk/authorizers/client_credentials.py
sirosen: I actually don't like either of these phrasings.

> a client credentials request

:point_up: what does that mean?

> client credentials tokens

:point_up_2: who dis?

---

Neither of these refer to things in a precise way, and there's no well-known thing called "a client credentials token".

I'm planning to update to
> Make a request for new tokens, using a 'client_credentials' grant.

It might be pedantic, but it's accurate.

src/globus_sdk/authorizers/client_credentials.py
sirosen: Sure!

src/globus_sdk/authorizers/client_credentials.py
sirosen: As a quoted forward reference, we get deferred evaluation.

We don't necessarily _have to_ defer, but given that the goal of the lazy importer is to defer the high-impact imports (namely, `requests` and `urllib3`) where possible, I think we're generally safer using deferred evaluation of annotations whenever it's an option. I've sometimes found myself disentangling the import-time behavior of modules in order to ensure that a given module doesn't eagerly import `requests` (specifically for the CLI use-case), and I'm just trying to head off this problem where possible.

src/globus_sdk/authorizers/client_credentials.py
sirosen: I accidentally committed this (meant to hit the suggestion below), but I'll just update again "over the top". I'm un-resolving for now, but please re-resolve if you like the result!

docs/services/auth.rst
kurtmckee: Sorting by assumed usage stats is fine, just wanted to understand if there was an ordering decision made here.

src/globus_sdk/authorizers/renewing.py
```diff
@@ -17,8 +17,12 @@
 # possible delays or clock skew.
 EXPIRES_ADJUST_SECONDS = 60
 
+# the type of the response which is produced by the authorizer, received by it, and
+# passed to the `on_refresh` callback
+ResponseT = t.TypeVar("ResponseT", bound="OAuthTokenResponse")
 
-class RenewingAuthorizer(GlobusAuthorizer, metaclass=abc.ABCMeta):
+
+class RenewingAuthorizer(GlobusAuthorizer, t.Generic[ResponseT], metaclass=abc.ABCMeta):
```
derek-globus, line 25: What would happen to clients who have implemented their own custom RenewingAuthorizer if we add a new generic to the class signature?

src/globus_sdk/authorizers/refresh_token.py
sirosen: Found it. There's an issue with the lazy importer code that makes `mypy` ignore unknown attributes. I'll do a separate PR with a relevant test.

src/globus_sdk/authorizers/client_credentials.py
sirosen: Yeah, if it were trivial to add linting, I'd do it so that we enforce "use deferred annotation syntax where possible", but it's hard to think through what such a lint would look like.

kurtmckee: Does this apply to return values? For example, this PR introduces several unquoted references in these same files, so is anything gained by quoting this one annotation where others are unquoted?

src/globus_sdk/services/auth/response/oauth.py
```diff
@@ -250,12 +250,49 @@ def __str__(self) -> str:
         )
 
 
+class AuthorizationCodeTokenResponse(OAuthTokenResponse):
```
derek-globus, line 253: \<NamingWar>

I think it's a step too far to remove the OAuth prefix from these names. The existing 2 response classes + the many AuthLoginClient methods have established the precedent of prefixing this category of things with OAuth (or oauth2 in the case of methods) and even if I like the shorter names you set up here a bit better; I'll advocate that we should be consistent for relationship-hinting's sake.

My formal change submission would be:
 1. Add the prefix "OAuth" to label-wise consistent
 2. Replace the suffix "TokenResponse" with "Response".

These two work with  the existing custom grant-type response class (DependentToken) to set a clear naming pattern of `f"OAuth{GrantType}Response[OAuthTokenResponse]"`

---

If you buy that, we'd end up with:
```python
class OAuthAuthorizationCodeResponse[OAuthTokenResponse]: ...
class OAuthClientCredentialsResponse[OAuthTokenResponse]: ...
class OAuthDependentTokenResponse[OAuthTokenResponse]: ...
class OAuthRefreshTokenResponse[OAuthTokenResponse]: ...
```

derek-globus (changes requested):

src/globus_sdk/authorizers/renewing.py
sirosen: Implicitly, their subclasses become
```python
class MyAuthorizer(RenewingAuthorizer[Any]): ...
```

At runtime, they don't encounter any changes, but at typing-time, their adjustments will depend on what type checkers and type checker settings they are enforcing on their projects.
For example, in `mypy --strict`, a rule is in place which disallows the implicit-any for generics. So the following two equivalent formulations are treated differently:
```python
class MyAuthorizer(RenewingAuthorizer[Any]):  # ok

class MyAuthorizer(RenewingAuthorizer):  # not ok
```

I wouldn't consider it a breaking change, but it's possible that someone will need to update their types. (Typing time compatibility is a complex area; more to unpack here if you like.)

src/globus_sdk/services/auth/response/oauth.py
sirosen: > If you buy that

I can buy my way out of naming wars? Hush and take my money! :joy: 

---

I'm perfectly okay with your suggestion and will apply it. There's a mild discrepancy with the base class, but I don't mind it much.

Even if we had space for a clean-room design for how to name the responses for different variants on this API, I think we might struggle to determine what naming is best.

derek-globus: Yeah, I think if it were a clean namespace, I'd probably suggest that the parent class (and suffix?) be GrantResponse or OAuthGrantResponse but that's not really feasible right now.

src/globus_sdk/authorizers/client_credentials.py
sirosen: Because the annotations in question are in return type annotations, `from __future__ import annotations` is enough to get deferred forward reference behavior. The special thing about the use of a generic in a class hierarchy is that it's now in a runtime context, and therefore needs to be explicitly stringified.

As another example, there's
```
if TYPE_CHECKING:
    from some_module import SomeClass

T = TypeVar("T", bound="SomeClass")  # stringified, forward reference

# or

from some_module import SomeClass

T = TypeVar("T", bound=SomeClass)  # not stringified, evaluates at runtime
```

kurtmckee (approved):

src/globus_sdk/authorizers/renewing.py
derek-globus: Discussed offline; resolving this thread.

derek-globus (approved):

PR #1054, opened 2024-09-17
Rename 'TokenData' to 'TokenStorageData'

sirosen:
This is an experimental component which has not yet been documented
and does not appear in the changelog history; therefore, no
documentation changes or changelog are included with this change and
it is considered purely internal.

----
📚 Documentation preview 📚: https://globus-sdk-python--1054.org.readthedocs.build/en/1054/


ada-globus (approved):
I don't object to doing this as no-news, but maybe give a minute before smashing 'merge' in case anyone else does?

derek-globus (approved):

PR #1060, opened 2024-09-23
Only call token_validation_error_handler once

derek-globus:
## Changes

1. Fixed a bug where upgrading from access token to refresh token mode in a `GlobusApp` could result in multiple login prompts.
2. Removed the `skip_error_handling` optional kwarg from the method interface `GlobusApp.get_authorizer(...)`.

## Testing

Ran a script that requires dependent scopes twice, once with request_refresh_tokens set to false, once with it set to true. Observed a single login both invocations.

First script run:
  * Before fix: one login prompt
  * After fix: one login prompt

```py
from globus_sdk import TransferClient
from globus_sdk.experimental.globus_app import UserApp, GlobusAppConfig

app = UserApp(
    "refresh-tokens-with-consents",
    client_id="<client-id>",
    config=GlobusAppConfig(request_refresh_tokens=False),
)
transfer = TransferClient(app=app)

TUTORIAL_COLLECTION = "6c54cade-bde5-45c1-bdea-f4bd71dba2cc"
transfer.add_app_data_access_scope(TUTORIAL_COLLECTION)

ls_result = transfer.operation_ls(TUTORIAL_COLLECTION, "/")
print(ls_result)
# Results were printed
```

Second script run:
 * Before fix: two login prompts
 * After fix: one login prompt

```py
from globus_sdk import TransferClient
from globus_sdk.experimental.globus_app import UserApp, GlobusAppConfig

app = UserApp(
    "refresh-tokens-with-consents",
    client_id="<client-id>",
    config=GlobusAppConfig(request_refresh_tokens=True),
)
transfer = TransferClient(app=app)

TUTORIAL_COLLECTION = "6c54cade-bde5-45c1-bdea-f4bd71dba2cc"
transfer.add_app_data_access_scope(TUTORIAL_COLLECTION)

ls_result = transfer.operation_ls(TUTORIAL_COLLECTION, "/")
print(ls_result)
# Results were printed
```

----
📚 Documentation preview 📚: https://globus-sdk-python--1060.org.readthedocs.build/en/1060/

### DIFF
--- a/src/globus_sdk/experimental/globus_app/app.py
+++ b/src/globus_sdk/experimental/globus_app/app.py
@@ -264,7 +264,8 @@ class GlobusApp(metaclass=abc.ABCMeta):
         """
         for resource_server in self._scope_requirements.keys():
             try:
-                self.get_authorizer(resource_server, skip_error_handling=True)
+                with self._disabled_token_validation_error_handler():
+                    self.get_authorizer(resource_server)
             except TokenValidationError:
                 return True
         return False
@@ -325,23 +326,15 @@ class GlobusApp(metaclass=abc.ABCMeta):
 
         return auth_params
 
-    def get_authorizer(
-        self,
-        resource_server: str,
-        *,
-        skip_error_handling: bool | None = None,
-    ) -> GlobusAuthorizer:
+    def get_authorizer(self, resource_server: str) -> GlobusAuthorizer:
         """
         Get a ``GlobusAuthorizer`` from the app's authorizer factory for a specified
         resource server. The type of authorizer is dependent on the app.
 
         :param resource_server: The resource server for which the requested Authorizer
             should provide authorization headers.
-        :param skip_error_handling: If True, skip the configured token validation error
-            handler when a ``TokenValidationError`` is raised. Default: False.
         """
-        if skip_error_handling is None:
-            skip_error_handling = not self._token_validation_error_handling_enabled
+        error_handling_enabled = self._token_validation_error_handling_enabled
 
         try:
             # Disable token validation error handling while try to get an authorizer.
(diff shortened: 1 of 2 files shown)


sirosen (approved):
With the interface change we discussed (removal of that param), I'm happy for this to merge.

We should think about how to test, but it's not urgent given that we have manually checked on the behavior.

src/globus_sdk/experimental/globus_app/app.py
```diff
@@ -323,30 +326,44 @@ def _auth_params_with_required_scopes(
...
         :param resource_server: The resource server for which the requested Authorizer
             should provide authorization headers.
-        :param skip_error_handling: If True, skip the configured token validation error
-            handler when a ``TokenValidationError`` is raised. Default: False.
         """
+        error_handling_enabled = self._token_validation_error_handling_enabled
+
         try:
-            return self._authorizer_factory.get_authorizer(resource_server)
+            # Disable token validation error handling while try to get an authorizer.
```
kurtmckee, line 340:
```suggestion
            # Disable token validation error handling while trying to get an authorizer.
```

kurtmckee (approved):
Is there a test case that could be added?

I'm concerned that adding a test case will require such specific mocking and forced conditions that changes to the code end up forcing changes to the test case, which isn't ideal, so I'm not sure it's even practical...

derek-globus:
I'll take a stab at a unit-test replication of the script defined in the details section, but to Kurt's point, if I end up mocking out the universe to the point that it's not actually adding value, I'll report back here and move forward on the merit of the manual testing.

derek-globus:
I went down the mock-the-universe route and it's interesting but it involves a pretty big scope creep to this PR. I'd like to work on it as a follow-on so we don't block this work by a refactor of the globusapp test suite.

Gonna merge this as-is.

PR #1061, opened 2024-09-23
Rewrite ValidatingTokenStorage to be more modular

sirosen:
Separate validation criteria for TokenStorageData frm the core ValidatingTokenStorage definition.
This is now a class which operates over inner validators, and is largely generic.

The validators themselves contain much of the logic around identity_id and scope/consent validation. Refresh token and access token validation are also moved into dedicated validators.

----
📚 Documentation preview 📚: https://globus-sdk-python--1061.org.readthedocs.build/en/1061/


src/globus_sdk/experimental/globus_app/app.py
```diff
@@ -165,6 +157,39 @@ def _resolve_client_info(
                 "login_client is required."
             )
 
+    def _finalize_token_storage(self) -> None:
```
derek-globus, line 160: Not immediately clear to me that this full listing should live in GlobusApp itself. I'd love to push it into the "validing token storage" space somehow.

src/globus_sdk/experimental/globus_app/authorizer_factory.py
```diff
@@ -110,6 +111,11 @@ class AccessTokenAuthorizerFactory(AuthorizerFactory[AccessTokenAuthorizer]):
     def __init__(self, token_storage: ValidatingTokenStorage) -> None:
         super().__init__(token_storage)
         self._cached_authorizer_expiration: dict[str, int] = {}
+        # inject a validator which ensures that tokens have an expiration time
+        # in the future
+        self.token_storage.after_retrieve_validators.insert(
+            0, self._check_expires_at_seconds
+        )
```
derek-globus, line 118: The fact that authorizerfactory-specific attachments live in this module doesn't actually seem  necessary now that validators exist on their own.

I'd advocate that whatever context is building out the list of validators should own this too. Should make for simpler composition heuristics if validators only get added at one point.

src/globus_sdk/experimental/globus_app/_validating_token_storage.py
```diff
@@ -2,288 +2,124 @@
 
 import typing as t
```
derek-globus, line 3: `validating_token_storage` should be a dedicated module if we add two support modules.

Something like
```
  globus_app/
    |- app.py
    |- validating_token_storage/
      |- __init__.py
      |- token_validators.py
      |- identity_id_reader.py
      |- token_storage.py
```

src/globus_sdk/experimental/globus_app/_identity_id_reader.py
```diff
@@ -0,0 +1,75 @@
+from __future__ import annotations
+
+import typing as t
+
+from globus_sdk.experimental.tokenstorage import TokenStorage, TokenStorageData
+
+
+class IdentityIDReader:
```
derek-globus, line 8: Still thinking about this class.

At least one thing that strikes me as odd off the bat: this isn't responsible for reading identity id from a response object, that functionality has stayed in the tokenstorage object.

I do still have a broader question in my head about whether this is the right type of thing to isolate in context, but if it is isolated it should probably own all of the reading

src/globus_sdk/experimental/globus_app/authorizer_factory.py
sirosen: I think the changes I applied satisfy this pretty well. The validators exist independently and are associated in UserApp with the relevant authorizer factories.

src/globus_sdk/experimental/globus_app/_identity_id_reader.py
sirosen: This disappeared back into `ValidatingTokenStorage` in the last rewrites. I think we're in a pretty good place now.
I considered trying to define a reader which owns all of this functionality, but the `decode_id_token` case has a lot of potential subtlety to handle, so I'm not worrying about it for now.

src/globus_sdk/experimental/globus_app/app.py
sirosen: I didn't move it very far, but it is now in a dedicated helper method.

I'm resolving, but open to further discussion.

sirosen:
I consider this work now ready for review. I've only added limited additional testing -- after looking at coverage reporting, most of the codepaths are already tested by existing tests. I could do a bit more if there's some particular part of the change we want to make sure is explored by tests.

src/globus_sdk/experimental/globus_app/validating_token_storage/validators.py
```diff
@@ -0,0 +1,289 @@
...
+        self,
+        *,
+        resource_server: str,
+        token_data: TokenStorageData,
+        identity_id: str | None,
+        eval_dependent: bool = True,
+    ) -> None:
+        """
+        Given a particular resource server/token data, evaluate whether the token +
+            user's consent forest meet the attached scope requirements.
```
kurtmckee, line 185: This dangling "+" catches my eye as a shortcut for writing but not necessarily comprehension.

src/globus_sdk/experimental/globus_app/validating_token_storage/validators.py
```diff
@@ -0,0 +1,289 @@
...
+
+        Note: If consent_client was omitted, only root scope requirements are validated.
```
kurtmckee, line 187:
```suggestion
        ..  note::

            If ``consent_client`` was omitted, only root scope requirements are validated.
```

src/globus_sdk/experimental/globus_app/validating_token_storage/validators.py
```diff
@@ -0,0 +1,289 @@
...
+
+        :param resource_server: The resource server string to validate against.
+        :param token_data: The token data to validate against.
```
kurtmckee, line 190: ?

```suggestion
        :param token_data: The token data to validate.
```

src/globus_sdk/experimental/globus_app/validating_token_storage/validators.py
```diff
@@ -0,0 +1,289 @@
...
+        :param identity_id: The identity ID of the user, from the surrounding context.
+        :param eval_dependent: Whether to evaluate dependent scope requirements.
+        :raises UnmetScopeRequirements: if token/consent data does not meet the
```
kurtmckee, line 193:
```suggestion
        :raises UnmetScopeRequirements: If token/consent data does not meet the
```

src/globus_sdk/experimental/globus_app/validating_token_storage/validators.py
```diff
@@ -0,0 +1,289 @@
...
+        :returns: None if all scope requirements are met (or indeterminable).
+        """
+        required_scopes = self.scope_requirements.get(resource_server)
+
+        # Short circuit - No scope requirements are, by definition, met.
+        if required_scopes is None:
+            return
+
+        # 1. Does the token meet root scope requirements?
+        root_scopes = token_data.scope.split(" ")
```
kurtmckee, line 204: Under ideal circumstances, scopes should be whitespace normalized. Is there any way that the scope in `token_data` might be from an untrusted source that has two or more spaces next to each other?

```pycon
>>> "a  b".split(" ")
['a', '', 'b']

>>> "a  b".split()
['a', 'b']
```

```suggestion
        root_scopes = token_data.scope.split()
```

kurtmckee (changes requested):

src/globus_sdk/experimental/globus_app/app.py
```diff
@@ -84,22 +94,23 @@ def __init__(
...
         # initialize our authorizer factory
         self._initialize_authorizer_factory()
-
-        # create a consent client for token validation
-        # reducing the scope requirements to barebones openid (user identification)
-        # additionally, this will ensure that openid scope requirement is always
-        # registered (it's required for token identity validation).
-        consent_client = AuthClient(app=self, app_scopes=[Scope(AuthScopes.openid)])
-        self.token_storage.set_consent_client(consent_client)
+        self._authorizer_factory_initialized = True
```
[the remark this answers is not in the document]
sirosen: I don't think there's any perfect solution for this -- it's an attempt to capture the "partially initialized state" -- and I don't mind adjusting it.

We could definitely set it to False at the start of init, rather than as a class attr. I'd be fine with that.

If we don't set it to True at the end of init, each subclass will be required to set it in their implementations of the `_initialize_authorizer_factory` hook. I think that's not great, and risks having some subclass miss it. So I'd rather keep that behavior in the base class if possible.

Another solution for this whole problem is to implement a class `_InitStubAuthorizerFactory` which gets assigned at the start of init, and overwritten later. It can raise errors (except for the place where this bool is checked, which would be a no-op. We can then eliminate the separate bool. I only just thought of this idea, but let me know if it sounds worth pursuing.

src/globus_sdk/experimental/globus_app/validating_token_storage/context.py
```diff
@@ -0,0 +1,16 @@
...
+class TokenValidationContext:
+    """
+    This is the context object which will be passed to token validators.
+
+    It allows the ValidatingTokenStorage to capture state about its execution and pass
+    that information, nicely bundled, to validators.
+    """
+
+    prior_identity_id: str | None
+    token_data_identity_id: str | None
```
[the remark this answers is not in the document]
sirosen: It is not, but given that there's a minor perf penalty for using frozen dataclasses, I thought it would be fine to define it as such. (Since most objects are mutable) 

The perf hit is very minor though. I can mark it frozen if you like?

src/globus_sdk/experimental/globus_app/validating_token_storage/validators.py
sirosen: Yeah, this is lift and shift of an existing piece of code, of course. I'll adjust this one, but I do try to keep such changes separate from the functional rework. A dedicated commit for the various doc fixes will do the trick.

src/globus_sdk/experimental/globus_app/validating_token_storage/validators.py
sirosen: It looks like a bug to me at first, but if these data are from Auth then there will never be a double space unless a user manipulates the data directly.

I think a bare split call is more robust and better, but it's also an unreachable condition under most realistic usage. So I intend to apply the change but not to try to define a break-fix test case. Please unresolve if you disagree and want a test for it.

src/globus_sdk/experimental/globus_app/validating_token_storage/context.py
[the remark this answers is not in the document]
kurtmckee: I consider all dataclasses to be a performance penalty, so my preference would be to make this a named tuple.

```python
class TokenValidationContext(typing.NamedTuple):
    """
    This is the context object which will be passed to token validators.
    It allows the ValidatingTokenStorage to capture state about its execution and pass
    that information, nicely bundled, to validators.
    """

    prior_identity_id: str | None
    token_data_identity_id: str | None
```

src/globus_sdk/experimental/globus_app/validating_token_storage/validators.py
kurtmckee: No need for a test for this, I just prefer bare `.split()` calls when the intent is to split on whitespace. When an argument is given, I assume it's a critical element.

Thanks for updating this!

src/globus_sdk/experimental/globus_app/app.py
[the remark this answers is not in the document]
kurtmckee: I lean toward setting it to `False` at the start and setting it to `True` at the end. :+1:

src/globus_sdk/experimental/globus_app/validating_token_storage/context.py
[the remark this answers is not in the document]
sirosen: I need to think about this... To put my thoughts into text:

We currently have several datatype construction paradigms in the SDK (by my count, 3: dataclass, PayloadWrapper, and Serializable).
One is "our old customized UserDict", one was "built for GAREs but turned out to have other valid use-cases", and one is "'the' stdlib construct we use". Even if it's superior in this case, introducing namedtuples adds a fourth.
Each was well-justified in its own context, and if you need 4 things to express what you mean, then you need 4 things. Arguably there's no issue.

I think the only thing I dislike is that we would have namedtuples _and_ frozen dataclasses. If at all possible, we should choose between those two. (And I'm onboard with choosing namedtuple.)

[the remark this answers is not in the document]
sirosen: The only frozen dataclass currently in use is the `GlobusAppConfig`.

I'll raise this with the team after my some meetings; I'm not sure if there would be any harm in making that also a typing.NamedTuple.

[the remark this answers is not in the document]
derek-globus: GlobusAppConfig uses a `defaultfactory` to set a default environment ([ref](https://github.com/globus/globus-sdk-python/blob/main/src/globus_sdk/experimental/globus_app/config.py#L93)); I don't believe there's any equivalent functionality in NamedTuple.

src/globus_sdk/experimental/globus_app/app.py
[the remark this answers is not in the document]
derek-globus: I wonder if it'd read cleaner if the flag was just `_initialized` rather than being specific to the authorizer.

---

I hadn't considered the fake authorizer factory route. That would clean this up pretty well...

src/globus_sdk/experimental/globus_app/validating_token_storage/context.py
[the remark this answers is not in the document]
sirosen: There is not. The only way to get that with a namedtuple, as far as I know, is to have a builder of some kind which makes the whole object.

I'll bring this up in the team chat -- the question I have is whether or not "minor perf hit but fewer different types of objects" is a good reason to make this object a dataclass (frozen or otherwise) rather than a named tuple. Obviously either one "works"; the open question is which one is _best_.

src/globus_sdk/experimental/globus_app/app.py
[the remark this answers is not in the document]
sirosen: I tried it both ways and settled on the bool. I've pushed the stub-object version followed by a revert commit so that we can all see the two side-by-side. I think the simplicity of the bool wins out by a lot, when you write the whole thing out. Having to provide the stub with all of the methods of the ABC is one of the issues with that approach, IMO.

src/globus_sdk/experimental/globus_app/validating_token_storage/validators.py
```diff
@@ -0,0 +1,293 @@
...
+        self,
+        *,
+        resource_server: str,
+        token_data: TokenStorageData,
+        identity_id: str | None,
+        eval_dependent: bool = True,
+    ) -> None:
+        """
+        Given a particular resource server and token data, evaluate whether or not the
+        user's consent forest meet the attached scope requirements.
```
kurtmckee, line 185:
```suggestion
        user's consent forest meets the attached scope requirements.
```

kurtmckee (changes requested):

src/globus_sdk/experimental/globus_app/validating_token_storage/validators.py
sirosen: D'oh!

kurtmckee (approved):

PR #1063, opened 2024-09-25
Enhance the 'ensure-exports-are-documented' lint

sirosen:
1.  Make the linter able to understand the relevant sphinx directives
    (via regex matching), looking for autoclass, py:data, etc

2.  Make the linter able to check additional public subpackage
    contents, starting with `globus_sdk.scopes`

2a. Document `scopes_to_str` and `scopes_to_scope_list` via
    `autofunction` directives. These are now considered public parts
    of `globus_sdk.scopes`.

    This was ambiguous before. Although we might not want to treat
    them as public in the future, there is an existing ambiguity.
    Because they are small and simple, simply committing to these
    public interfaces makes our policy clearer with minimal cost.

3.  Add support for listing deprecated names in the linter. This
    allows it to ignore deprecated `Timer*` names. If a deprecated
    name disappears from the __all__ lists, it will be treated as
    an error.

The lint is still imperfect -- it doesn't understand module context
(e.g. `.. currentmodule: ...`) and match names appropriately via
qualname or similar -- but it's sufficiently improved that it won't be
trivially fooled anymore.

----
📚 Documentation preview 📚: https://globus-sdk-python--1063.org.readthedocs.build/en/1063/

### DIFF
--- a/docs/authorization/scopes_and_consents/scopes.rst
+++ b/docs/authorization/scopes_and_consents/scopes.rst
@@ -195,6 +195,10 @@ Scope Reference
 
 .. autoclass:: ScopeCycleError
 
+.. autofunction:: scopes_to_str
+
+.. autofunction:: scopes_to_scope_list
+
 ScopeBuilders
 -------------
 
--- a/docs/authorization/scopes_and_consents/scopes.rst
+++ b/docs/authorization/scopes_and_consents/scopes.rst
@@ -195,6 +195,11 @@ Scope Reference
 
 .. autoclass:: ScopeCycleError
 
+.. rubric:: Utility Functions
+
+``globus_sdk.scopes`` also provides helper functions which are used to
+manipulate scope objects.
+
 .. autofunction:: scopes_to_str
 
 .. autofunction:: scopes_to_scope_list
--- a/scripts/ensure_exports_are_documented.py
+++ b/scripts/ensure_exports_are_documented.py
@@ -13,6 +13,11 @@ REPO_ROOT = pathlib.Path(__file__).parent.parent
 
 _ALL_NAME_PATTERN = re.compile(r'\s+"(\w+)",?')
 
+PACKAGE_DIRS_TO_SCAN = (
+    "globus_sdk",
+    "globus_sdk/scopes",
+)
+
 DEPRECATED_NAMES = {
     "TimerAPIError",
     "TimerClient",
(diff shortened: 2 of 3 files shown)


docs/authorization/scopes_and_consents/scopes.rst
```diff
@@ -195,6 +195,10 @@ Scope Reference
 
 .. autoclass:: ScopeCycleError
 
+.. autofunction:: scopes_to_str
+
+.. autofunction:: scopes_to_scope_list
```
derek-globus, line 200: These utilities needs a bit more precursor explanation if they're  going in the primary scopes page. 

Maybe a dedicated subsection? Hard to say if that's overkill, but at least something disambiguating.

scripts/ensure_exports_are_documented.py
```diff
@@ -52,10 +84,18 @@ def get_names_from_all_list() -> list[str]:
 
 def ensure_exports_are_documented() -> bool:
     success = True
-    for name in get_names_from_all_list():
-        if not is_documented(name):
-            print(f"{name} is not documented")
-            success = False
+    used_deprecations = set()
+    for package_dir in ("globus_sdk", "globus_sdk/scopes"):
```
derek-globus, line 88: nit:

We'll probably want to include more than just sdk & scopes, maybe pull this tuple into a static-defined constant to make it more clear where we would add those?

scripts/ensure_exports_are_documented.py
```diff
@@ -52,10 +84,18 @@ def get_names_from_all_list() -> list[str]:
...
+        for name in get_names_from_all_list(package_dir):
+            if name in DEPRECATED_NAMES:
+                used_deprecations.add(name)
+                continue
+            if not is_documented(name):
+                print(f"{name} is not documented")
```
derek-globus, line 94: I know this isn't new, but this error message could use a bit of work.

I'd ask that we convey:
  1. What is required in terms of documentation (an `.. auto<x>` method)
  2. Ideally some sort of trace to this file (not sure what the best format for that would be? maybe just a file name?). In the case of an exception it can be difficult to trace down and understand exactly what's being evaluated and why.

docs/authorization/scopes_and_consents/scopes.rst
sirosen: I didn't want to worry about a full subsection, but I could do a rubric + a short paragraph.

scripts/ensure_exports_are_documented.py
sirosen: Yeah, I can put it at top of script. I definitely want to grow this list.

docs/authorization/scopes_and_consents/scopes.rst
derek-globus: Yeah, even just a sentence informing that these are convenience utility functions would be good in my book.

derek-globus (approved):
Docs build is failing but I'm 👍 to merge this once that's fixed.

kurtmckee (approved):

PR #1066, opened 2024-09-30
Move ValidatingTokenStorage out of experimental

derek-globus:
## Changes
1. Move `ValidatingTokenStorage` from `globus_sdk.experimental.globus_app` to `globus_sdk.tokenstorage`.
   * With Stephen's recent work, this token storage no longer requires includes any `GlobusApp`-specific references. It can be used as an extensible class to validate token storages in any context.
1. Move `globus_app` errors into the `validating_token_storage` module space.
   * The errors defined here are all ValidatingTokenStorage errors (subclasses of `TokenValidationError`). They are referenced in `GlobusApp` to catch and retry behavior but are properly associated with the token storage itself.
1. Add docs for key `ValidatingTokenStorage` constructs to `docs/authorization/token_storages.rst`

## Note

This PR does not intend to introduce any behavioral changes, only cosmetic/import location.

If you observe any, please raise them as bugs.

----
📚 Documentation preview 📚: https://globus-sdk-python--1066.org.readthedocs.build/en/1066/


src/globus_sdk/experimental/globus_app/types.py
kurtmckee, line 1: It is common to avoid naming modules after packages in Python's standard library, though it's not a technical necessity like it was in Python 2.

Python 2's import machinery used implicit relative imports. This meant that writing `import types` would first look in the same directory as the module that was trying to import `types`, which was such a hassle.

Anyway, this catches my eye. I still have a preference for this to be named `_types` (if it's supposed to be private) or `types_` (if it's public).

As I said, now that absolute imports are the standard, this isn't a technical issue anymore, but I wanted to express my preference here.

src/globus_sdk/tokenstorage/v2/validating_token_storage/context.py
```diff
@@ -0,0 +1,19 @@
...
+import dataclasses
+
+
+@dataclasses.dataclass
+class TokenValidationContext:
+    """
+    A data object of information available to
+    :class:`TokenDataValidators <TokenDataValidator>` during validation.
+
+    :ivar str | None prior_identity_id: The identity ID associated with the
```
kurtmckee, line 12: I want to thank you for introducing this `ivar` syntax. I think you've used it in another PR or two but I don't think I've seen it elsewhere, and it's something that I'll be using moving forward. Thanks! :heart:

src/globus_sdk/tokenstorage/v2/validating_token_storage/validators.py
```diff
@@ -61,7 +62,7 @@ def after_retrieve(
         """
 
 
-class _OnlyBeforeValidator(TokenDataValidator):
+class _OnlyBeforeValidator(TokenDataValidator, ABC):
```
kurtmckee, line 65: I thought this wasn't needed.

`_OnlyBeforeValidator` already inherits from `abc.ABC`, and I thought it couldn't be instantiated unless a `before_retrieve()` is defined in a further subclass.


```suggestion
class _OnlyBeforeValidator(TokenDataValidator):
```

src/globus_sdk/tokenstorage/v2/validating_token_storage/validators.py
```diff
@@ -3,27 +3,28 @@
 import abc
 import time
 import typing as t
+from abc import ABC
```
kurtmckee, line 6: `abc` is already imported.

```suggestion
```

kurtmckee (changes requested):

src/globus_sdk/experimental/globus_app/types.py
derek-globus: I'll actually swap it to `protocols`.

Within the current structure, it only holds protocols (mostly defining how non-globus-app-specific constructs are integrated with globus-app).

src/globus_sdk/tokenstorage/v2/validating_token_storage/validators.py
derek-globus: I added these initially because PyCharm was complaining.

image

---

You're correct that it's not needed. At runtime instantiating an `_OnlyBeforeValidator` fails because no `before_store` is defined whether `abc.ABC` is explicitly defined or not.

Conceptually though, this class flags that the subclasses are intended to be abstract as well, not concrete implementations. Given that it doesn't actually affect the class's mro (I just tested that locally) which is the only place that I could envision a negative runtime impact coming from, I'd argue this is just a clearer way to write it.

I don't feel passionately about this though; if folks disagree, ultimately I can change.

src/globus_sdk/experimental/globus_app/types.py
sirosen: You can still be bitten by this in some cases even under Python 3.
For example, `cd src/globus_sdk/; cp _types.py types.py; python _generate_init.py`. You'll get an error when this finally hits an import from `types` because it picks up on the `types.py` module in the working directory.

I also prefer that we avoid well-known stdlib module names where possible, to avoid such cases, but don't hold to it as a strict rule. `protocols.py` sounds great! :+1:

src/globus_sdk/tokenstorage/v2/validating_token_storage/validators.py
kurtmckee: Ah, then yes, please add that in. I, too, don't want to hear about it from PyCharm!

kurtmckee (approved):

src/globus_sdk/experimental/globus_app/types.py
kurtmckee: > it picks up on the `types.py` module in the working directory.

:musical_note: Hello Darkness my old friend :musical_note:

src/globus_sdk/tokenstorage/v2/validating_token_storage/validators.py
sirosen: It know it's probably irrational, but I don't like seeing `ABC` listed as an explicit base when it's already been inherited and the class is an ABC. It reads to me like a programming mistake, "Oh, you didn't realize but this was already an abstract class."

Consistent with that, I think that PyCharm lint is wrong. Cases like this one are not uncommon, in which an intermediate class can fill in some but not all of the details of a base class.

_That said_,
- I understand what the lint is catching (accidentally missing an abstract method)
- This has no runtime impact: `__mro__` is the same either way (oh, hey, you tested it too! :grin: )
- It's fine and this class is an internal implementation detail anyway

So it's okay, and I don't see a reason to make you fight with PyCharm, but I'll register the fact that visually I don't like it. :shrug:

sirosen (approved):

src/globus_sdk/tokenstorage/v2/validating_token_storage/validators.py
kurtmckee: > I think that PyCharm lint is wrong. Cases like this one are not uncommon

It's also not uncommon to overlook an abstract method implementation, so I don't share the interpretation that the lint is wrong.

PR #1072, opened 2024-10-04
Enforce more lazy imports of `requests`

sirosen:
The benefits of this change are twofold:

- it ensures that the lazy import semantics provided by the SDK's
  top-level `__init__` are preserved when a greater variety of symbols
  are imported from it

- it pushes for a more uniform import style, in which the SDK itself
  demonstrates the best practice of importing the package and then
  accessing its members -- this avoids eagerly doing an import when a
  module is loaded and is especially valuable when modules only need
  an SDK symbol for type annotations

The change was made first by expanding and reorganizing the list of
modules which are tested for lazy imports, and then making any
necessary src/ changes to satisfy the test.

----
📚 Documentation preview 📚: https://globus-sdk-python--1072.org.readthedocs.build/en/1072/

### DIFF
--- a/src/globus_sdk/login_flows/login_flow_manager.py
+++ b/src/globus_sdk/login_flows/login_flow_manager.py
@@ -29,9 +29,7 @@ class LoginFlowManager(metaclass=abc.ABCMeta):
         request_refresh_tokens: bool = False,
         native_prefill_named_grant: str | None = None,
     ) -> None:
-        if not isinstance(
-            login_client, globus_sdk.NativeAppAuthClient
-        ) and not isinstance(login_client, globus_sdk.ConfidentialAppAuthClient):
+        if not isinstance(login_client, (globus_sdk.NativeAppAuthClient, globus_sdk.ConfidentialAppAuthClient)):
             raise globus_sdk.GlobusSDKUsageError(
                 f"{type(self).__name__} requires a NativeAppAuthClient or "
                 f"ConfidentialAppAuthClient, but got a {type(login_client).__name__}."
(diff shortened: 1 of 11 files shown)


src/globus_sdk/login_flows/login_flow_manager.py
```diff
@@ -30,15 +24,15 @@ class LoginFlowManager(metaclass=abc.ABCMeta):
...
         request_refresh_tokens: bool = False,
         native_prefill_named_grant: str | None = None,
     ) -> None:
-        if not isinstance(login_client, NativeAppAuthClient) and not isinstance(
-            login_client, ConfidentialAppAuthClient
-        ):
-            raise GlobusSDKUsageError(
+        if not isinstance(
+            login_client, globus_sdk.NativeAppAuthClient
+        ) and not isinstance(login_client, globus_sdk.ConfidentialAppAuthClient):
```
kurtmckee, line 34: `isinstance()` accepts a tuple of possible values, like `str.startswith()`:

```suggestion
        if not isinstance(login_client, (globus_sdk.NativeAppAuthClient, globus_sdk.ConfidentialAppAuthClient)):
```

kurtmckee (changes requested):

src/globus_sdk/login_flows/login_flow_manager.py
sirosen: Yep; good call to tidy this up!

sirosen: I am 0% surprised that black reflowed this. (Since I plan to squash, I don't mind.)

kurtmckee (approved):

PR #1073, opened 2024-10-04
Implement 'set_subscription_id'

### LINKED ISSUE #316: Add support for new API for setting endpoints to managed
Transfer 7.0 has added an API which lets users who do not have Administrator on an endpoint to set endpoints to Managed via `PUT /endpoint/<uuid>/subscription`

I'm not sure if this docs for this are live yet.
We need to get clarity on the API a little and add a `TransferClient` method for this.

sirosen:
closes #316

----
📚 Documentation preview 📚: https://globus-sdk-python--1073.org.readthedocs.build/en/1073/

### DIFF
--- a/src/globus_sdk/services/transfer/client.py
+++ b/src/globus_sdk/services/transfer/client.py
@@ -275,7 +275,9 @@ class TransferClient(client.BaseClient):
         return self.put(f"endpoint/{endpoint_id}", data=data, query_params=query_params)
 
     def set_subscription_id(
-        self, collection_id: UUIDLike, subscription_id: UUIDLike
+        self,
+        collection_id: UUIDLike,
+        subscription_id: UUIDLike | Literal["DEFAULT"] | None,
     ) -> response.GlobusHTTPResponse:
         """
         Set the ``subscription_id`` on a mapped collection.
@@ -289,7 +291,18 @@ class TransferClient(client.BaseClient):
         cannot have a ``subscription_id`` directly set in this way.
 
         :param collection_id: The collection ID which is having its subscription set.
-        :param subscription_id: The ID of the subscription to assign.
+        :param subscription_id: The ID of the subscription to assign, the special
+            string ``"DEFAULT"``, or ``None``.
+
+        .. note::
+
+            Setting ``subscription_id="DEFAULT"`` results in the service choosing your
+            subscription ID, but requires that you only have one subscription.
+            If you have multiple subscriptions, using ``"DEFAULT"`` will result in an
+            error.
+
+            Setting ``subscription_id=None`` clears any existing subscription from the
+            collection.
 
         .. tab-set::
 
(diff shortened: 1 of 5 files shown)


src/globus_sdk/services/transfer/client.py
```diff
@@ -274,6 +274,52 @@ def update_endpoint(
...
+
+        This operation requires membership in a Globus subscription group and
+        has authorization requirements which depend upon the caller's roles on
+        the subscription group and the endpoint or collection.
+
+        Guest Collections inherit the subscriptions of their Mapped Collections and
+        cannot have a ``subscription_id`` directly set in this way.
+
+        :param collection_id: The collection ID which is having its subscription set.
+        :param subscription_id: The ID of the subscription which is being assigned.
```
ada-globus, line 292:
```suggestion
        :param subscription_id: The ID of the subscription to assign.
```

src/globus_sdk/services/transfer/client.py
```diff
@@ -274,6 +274,52 @@ def update_endpoint(
         log.info(f"TransferClient.update_endpoint({endpoint_id}, ...)")
         return self.put(f"endpoint/{endpoint_id}", data=data, query_params=query_params)
 
+    def set_subscription_id(
+        self, collection_id: UUIDLike, subscription_id: UUIDLike
```
ada-globus, line 278: `"DEFAULT"` is an allowable value. I believe since `UUIDLike` just unions `UUID` and `str` types, it doesn't pose a problem, though I do wonder if it'd be helpful to add an explicit literal here (and/or a note to the `:param` below) indicating that it's a valid, "special" value?

Also, though the docs don't mention it, is `null` valid for `subscription_id` (for instance, to clear a subscription)?

sirosen: Good question about `null`. I should test and we can decide based on what it does.

I considered taking extra steps with `subscription_id="DEFAULT"`. I think I'm -0 on changing the annotation -- it doesn't really harm anything, but I don't know that it's helpful -- but doc is where I'm a bit less confident of my judgement to omit it. Since you've asked, I'll add it.

sirosen:
```
$ globus api transfer PUT /endpoint/$GCP_ID/subscription --body '{"subscription_id": null}'
```

worked.

So yes, we should allow `None`. :slightly_smiling_face:

sirosen: I tried a variant of this which used `Literal["DEFAULT"]` even though that's a subtype of `UUIDLike`. LMK if you think it's nicer or not -- I think of this as a little experiment in "how does this look?"

ada-globus: Yeah, I think you're a -0 and I'm a +0. I think it's a little nicer (I am specifically thinking about this as used by a type-savvy IDE) but I will defer to you if you still dislike it. I'll approve now for expediency, and will be fine with either outcome.

ada-globus (approved):

src/globus_sdk/services/transfer/client.py
sirosen: The law of "that's the way it is in in the PR" wins. Merging :wink:

PR #1075, opened 2024-10-04
Fix prefill_named_grant when set via GlobusApp

sirosen:
This uses a "nice" format of `<app> on <host>`. e.g.,

    sample-script on porgy-and-bess

Where `app_name="sample-script"` and the hostname is
`"porgy-and-bess"`.

----
📚 Documentation preview 📚: https://globus-sdk-python--1075.org.readthedocs.build/en/1075/

### DIFF
--- a/changelog.d/20241004_164201_sirosen_fix_prefill_named_grant.rst
+++ b/changelog.d/20241004_164201_sirosen_fix_prefill_named_grant.rst
@@ -1,0 +1,6 @@
+Fixed
+~~~~~
+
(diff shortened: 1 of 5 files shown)


ada-globus (approved):
Approving under the presumption that this was the agreed outcome of the discussion around how to handle this (since IIRC there was some disagreement). I'm a strong 👍 for use of the hostname.

sirosen:
Yeah, this is where the discussion landed. Also, I'm pretty sure this won't conflict with other work, so I'm going ahead with the merge.

PR #1076, opened 2024-10-07
Move globus app

derek-globus:
* Move GlobusApp constructs out of experimental, documenting the GlobusApp & Config-based constructs.

----
📚 Documentation preview 📚: https://globus-sdk-python--1076.org.readthedocs.build/en/1076/


derek-globus:
Rebased against main

docs/authorization/globus_app/apps.rst
```diff
@@ -0,0 +1,45 @@
...
+
+GlobusApps
+==========
+
+:class:`GlobusApp` is a high level construct designed to simplify the process of
+of authenticating for Globus service clients.
+
+A ``GlobusApp`` requires a programmatic Globus Auth entity called a "client" to be
+supplied as the requester driving authentication flows. A client can be created and
+managed through the `Globus Developer Console <https://developers.globus.org/>`_.
```
sirosen, line 13: I'm open to debating this, but I think we should send users to

```suggestion
managed through the `Globus Developer Console <https://app.globus.org/settings/developers>`_.
```

now that it's integrated into the webapp.

docs/authorization/globus_app/apps.rst
```diff
@@ -0,0 +1,45 @@
...
+There are two types of GlobusApps; :class:`UserApp`, for contexts in which a client
+should submit requests on behalf of a user, and :class:`ClientApp`, for contexts which
+a client should submit requests on behalf of themselves.
```
sirosen, line 16: Because this is intro text, I'm giving it a bit of extra scrutiny. I think the notion of "clients acting on behalf of themselves" is very abstract and hard for a new user to grasp.

I would phrase these in terms of "login to the app" and "the client acts as an automated user", leading to the following rephrase:
> ```rst
> There are two types of GlobusApp: :class:`UserApp` and :class:`ClientApp`.
> A :class:`UserApp` requires users to login to the application, allowing the application to
> submit requests on behalf of the user.
> A :class:`ClientApp` uses its own credentials to act as a user, and is suitable
> for a variety of automation use cases.
> ```

docs/authorization/globus_app/apps.rst
```diff
@@ -0,0 +1,45 @@
...
+
+The interfaces of these classes, defined below, intentionally include many
+"sane defaults" (i.e., storing oauth2 access tokens in a json file). These defaults
+however can be overridden if the behavior is not desired. For more information on what
+you can customize and how, see :ref:`globus_app_config`.
```
sirosen, line 21:
```suggestion
can be overridden to customize the app's behavior. For more information on what
you can customize and how, see :ref:`globus_app_config`.
```

docs/authorization/globus_app/index.rst
```diff
@@ -0,0 +1,13 @@
+.. _globus_app:
+
+Globus App
+==========
+
+For a narrative-style introduction to the concepts contained within GlobusApp, see
+:ref:`using_globus_app`.
```
sirosen, line 7: :rocket: 

Great choice to split up the doc into reference vs narrative at this point in the tree!

docs/authorization/globus_app/config.rst
```diff
@@ -0,0 +1,33 @@
...
+
+Reference
+---------
+
+Data Model
+^^^^^^^^^^
+
+.. autoclass:: globus_sdk.GlobusAppConfig()
+    :members:
+    :exclude-members: token_validation_error_handler
```
sirosen, line 16: I'm surprised by the decision to exclude this field from the documentation about the config. Can you share the reasoning behind this?

If we wanted it to be internal, we could have named it to be private. So I'm not clear on why it would be a public but undocumented attribute.

(Aside: I just looked at the default implementation and was once again gladdened by the renaming you did to get us to `app.login(force=True)` -- it makes the handler very easy to grasp.)

scripts/ensure_exports_are_documented.py
```diff
@@ -41,19 +42,22 @@ def iter_all_documented_names() -> t.Iterable[str]:
...
     #   .. autoclass:: <name>
+    #   .. autoclass:: <name>(<args>)
     #   .. autofunction:: <name>
     #   .. autoexception:: <name>
     #   .. autodata:: <name>
+    auto_directive = r"auto(?:class|function|exception|data)"
+    name_capturer = r"(?:\w+\.)*(\w+)(?:\(.*\))?"
     autodoc_pattern = re.compile(
-        r"^\.\.\s+auto(?:class|function|exception|data)\:\:\s+(?:\w+\.)*(\w+)$",
+        rf"^\.\.\s+{auto_directive}::\s+{name_capturer}$",
```
sirosen, line 52: May I suggest instead of creating a `rf` string, that we switch to a verbose regex and embed any necessary comments?

I think this is harder to read for breaking the expression up into multiple assembled strings, but I understand the impulse to break it into multiple physical lines.

sirosen: As a separate matter, I don't understand why it's desirable to allow for the empty `()` after a classname.
I would rather see
```rst
.. autoclass:: Noodle
```
than
```rst
.. autoclass:: Noodle()
```

simply on the grounds that the bare name is the norm in sphinx documentation. I didn't even know until a few weeks ago that the empty args were _valid_, that's how unusual it is in my experience!

Could you share why you are including the empty args to classnames in autoclass directives?

scripts/ensure_exports_are_documented.py
```diff
@@ -41,19 +42,22 @@ def iter_all_documented_names() -> t.Iterable[str]:
...
         flags=re.MULTILINE,
     )
     # names under these directives
     #
     #   .. class:: <name>
     #   .. py:data:: <name>
     pydoc_pattern = re.compile(
-        r"^\.\.\s+(?:py\:data|class)\:\:\s+(?:\w+\.)*(\w+)$",
+        r"^\.\.\s+(?:py:data|class)::\s+(?:\w+\.)*(\w+)$",
```
sirosen, line 60: :+1: Thanks for catching that!

src/globus_sdk/login_flows/__init__.py
```diff
@@ -6,10 +6,10 @@
...
 
-__all__ = [
+__all__ = (
     "CommandLineLoginFlowManager",
     "LocalServerLoginError",
     "LocalServerEnvironmentalLoginError",
     "LocalServerLoginFlowManager",
     "LoginFlowManager",
-]
+)
```
sirosen, line 15: Remind me to add a lint for this. IMO `__all__` should almost always be a tuple, not a list.

sirosen (changes requested):
I have some comments from my first read, but I want to see the rendered docs to see if that makes anything jump out. Overall, obviously it's very close, so mostly these are nits.

sirosen:
I think that given that the new docs link to `Using a GlobusApp`, now would be a good time to move _that doc in particular_ out of the `Updated Examples` section. We can punt this to a follow-up PR if you prefer.
Here's what the sidebar looks like right now:
image

I think `Using a GlobusApp` could go under `Getting Started`, alongside the tutorial. Is that okay for an initial step? Docs can always be moved more later.

docs/authorization/globus_app/config.rst
derek-globus: It is still documented; it's just documented as an ivar instead of a method. Sphinx exposes `@properties` as though they are callables even though that's not the actual invocation pattern. This my way of saying "don't do that, I documented it elsewhere".

image

scripts/ensure_exports_are_documented.py
derek-globus: > and embed any necessary comments

Not sure what you mean by this. Like embed the comments within the regex?

docs/authorization/globus_app/config.rst
sirosen: Oh, that makes sense... sort of silly. :shrug: 

I don't think any action is needed here, but I think we _could_ add a comment to the doc so that anyone reading it will understand that this is why this "unexpected directive is needed".

scripts/ensure_exports_are_documented.py
derek-globus: > As a separate matter, I don't understand why it's desirable to allow for the empty () after a classname.

Omitting the `()` instructs sphinx to auto-insert init args into the generated docs page.  Adding a `()` tells spinx that "I will explicitly list out the parameters to this class's init".

In these cases I used the empty `()` to tell sphinx not to list out any init-args. That shouldn't be the default there are cases where it's valuable to have that ability; specifically:
 * Abstract base classes that a user shouldn't be instantiating (`GlobusApp`).
 * Data classes where the parameters are listed more visibly as ivars (`GlobusAppConfig`).
 * Protocols. (Providers and handlers in this case)

Here's some side-by-sides of what gets generated in those different cases that hopefully demonstrate why I opted to pull redundant or unnecessary info that in some cases was taking up a good amount of screen space.

### GlobusApp

In this case there's 5 added lines of terse arg listing that will be skipped over basically every time someone reads this page; so better to omit it.

**With `()`**

image

**Without `()`**

image


### GlobusAppConfig

**With `()`**

image


** Without `()`**

image


### Protocols

**With `()`**

image


** Without `()`**

image


---

Long message, but it's important to look at what documentation we're actually generating and choose what'll be most helpful there not just what looks cleanest in the rst format; and I hold that each of these instances have a good justification for omitting those init params.

docs/authorization/globus_app/config.rst
derek-globus: I can give it a shot but [rst comments are weird](https://stackoverflow.com/a/4783854)... I'd be afraid that as a language restriction inserting a comment might make it less readable than leaving it undocumented.

scripts/ensure_exports_are_documented.py
derek-globus: Thank PyCharm!

derek-globus:
> I think that given that the new docs link to Using a GlobusApp, now would be a good time to move that doc in particular out of the Updated Examples section.

I agree but want to punt. This PR is already pretty massive as-is; so unless something is actively broken I think we separate that work.

docs/authorization/globus_app/apps.rst
derek-globus: Yeah, I didn't know which was more desirable; I copied from other locations in the SDK (and have used this link in past PRs) but for the files I've touched as a part of this change I can update to using the app-based one.

scripts/ensure_exports_are_documented.py
sirosen: Ooh, okay, I didn't know that `()` actually had an impact on Sphinx behavior.

I agree with the protocol case, where the `*args, **kwargs` inclusion is basically meaningless.

I disagree with the others, but I'm not sure I want to block this PR on resolving our difference of opinion on this topic.
The initializer for a class -- including for an abstract class -- is one of its interfaces.

That said, I hate the generated content for `GlobusApp`'s signature, since it expands the config. I would like best
```
class globus_sdk.GlobusApp(
    app_name='Unnamed Globus App',
    *,
    login_client=None,
    client_id=None,
    client_secret=None,
    scope_requirements=None,
    config=<DEFAULT_CONFIG>
)
```

(Aside: I looked into getting `furo` to generate this kind of spacing, and learned that it's basically infeasible unless something changes in the generated sphinx html.)

---

I'm happy to accept this for now, and if I feel that we need to document `__init__` for a class separately, I can submit my follow-up for us to discuss.

derek-globus: My strength of opinion is:
 * Strongest: protocols
 * Stronger: GlobusApp
 * Medium Strength: Data Class

But we can resolve this discussion for now if you're happy enough as-is & schedule some time in the arena to duke it out;

docs/authorization/globus_app/config.rst
sirosen: Yeah, acknowledged that they're awkward. In practice, I've never had an issue with the inline comment syntax which that poster labels as "avoid". If you're writing normal comments, they're very unlikely to be valid rst directives.

But for this case, my suggestion would be something like
```rst
..
    In the class below, omit token_validation_error_handler as a member.
    We document it explicitly as an ivar to avoid having sphinx think it's a method.
```

docs/authorization/globus_app/apps.rst
derek-globus: I just opened a thread in #auth to confirm what they'd like us to do.

derek-globus: Confirmed with Brigitte that `https://app.globus.org/settings/developers` should replace all instances of `https://developers.globus.org/`.

Updated all of the usages in this PR.

docs/authorization/globus_app/config.rst
derek-globus: Just looked back & I realize I've been talking about a different attribute (GlobusApp.scope_requirements) than your initial comments identifies. I'll update both locations to have a comment explaining why.

The one you called out is a similar but even more obvious bug in my eyes:

``GlobusAppConfig`` is a data class which holds some number of values. One of those values is a callable which may be provided at init time called "token_validation_error_handler. Because it's a callable attribute, sphinx renders it like a method you'd call on the class, rather than a value you'd retrieve from the class.  Formally the latter does work, but I would not recommend that users almost ever do that. (It is similarly documented in ivars).

image

scripts/ensure_exports_are_documented.py
derek-globus: Updated regex back to inline w/ `re.X`

docs/authorization/globus_app/apps.rst
derek-globus: Rephrased it again...

sirosen: FWIW, your rewrite is fantastic!

Marking this thread resolved.

derek-globus: Glad to hear you like it! Thanks for pushing me on it.

I'm going to unresolve this thread actually; I pinged Ada & Brigitte about this section to get their sign off as well. Want to leave it open in case they give comments here.

sirosen (approved):
The future has arrived.

docs/authorization/globus_app/apps.rst
```diff
@@ -0,0 +1,89 @@
...
+
+.. currentmodule:: globus_sdk
+
+GlobusApps
+==========
+
+
+.. warning::
+
+    Currently ``GlobusApp`` only supports scripting-based use cases.
```
ada-globus, line 11: Workshopping...

```suggestion
    Currently, ``GlobusApp`` can be used in scripts (e.g., notebooks or automation) and local applications launched directly by a user.
```

docs/authorization/globus_app/apps.rst
```diff
@@ -0,0 +1,89 @@
...
+
+    Hosted applications (i.e. web servers) should see :ref:`globus_authorizers` instead.
+
+
+:class:`GlobusApp` is a high level construct designed to simplify the process of
+of authenticating for Globus service clients.
```
ada-globus, line 17:
```suggestion
:class:`GlobusApp` is a high level construct designed to simplify authentication
for Globus service clients.
```

docs/authorization/globus_app/apps.rst
sirosen: If we're making notes on this warning, I would like to see it converted from a `warning` to a `note` to be less of an attention-grab at the start of the doc.

I might phrase it as
> ``GlobusApp`` only supports client applications, where a user or client logs in or authenticates.
> Resource server implementations should not use ``GlobusApp``.

This phrasing reveals my belief that any abstraction aimed at Resource Server implementations will be fully distinct from ``GlobusApp``.

docs/authorization/globus_app/apps.rst
```diff
@@ -0,0 +1,89 @@
...
+
+A ``GlobusApp`` will always require a programmatic entity called a "client". Depending
+on the type of ``GlobusApp`` in use, the client will play different roles in the
+request flow.
```
ada-globus, line 21:
```suggestion
A ``GlobusApp`` requires a programmatic entity called a "client". Depending
on the type of ``GlobusApp`` in use, the client will play different roles in the
request flow.
```

docs/authorization/globus_app/apps.rst
```diff
@@ -0,0 +1,89 @@
...
+
+GlobusApps
+==========
+
+
+.. warning::
+
+    Currently ``GlobusApp`` only supports scripting-based use cases.
+
+    Hosted applications (i.e. web servers) should see :ref:`globus_authorizers` instead.
```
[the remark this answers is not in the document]
sirosen: I believe that a hosted webapp which uses 3-legged oauth to authenticate users can be developed on top of `GlobusApp` but requires that you implement your own login flow manager.

This is a case we haven't discussed much, but I haven't seen it as necessary to build explicit support, since it will always require custom login flow code.

But I think there's some blurring of lines happening (already in the current doc) here: a hosted client application and a resource server are different cases, and should be described as distinct.

docs/authorization/globus_app/apps.rst
sirosen: I'm not sure why, but this rephrase is highlighting for me something else:
> Globus service clients.

"service clients" might mislead some folks into thinking that a "service client" is something meaningful vis-a-vis Globus Auth.

Perhaps...

> authentication for applications using globus-sdk clients.

?

docs/authorization/globus_app/apps.rst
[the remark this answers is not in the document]
derek-globus: Web services that require dependent tokens effectively don't work with GlobusApp today.

Somebody probably could encapsulate that behavior in a login flow manager, but I would be very surprised if they did (and did it well). Given that we do have very plug-and-play interfaces for scripting cases I'm of the opinion that our messaging is "this works for scripting, not web services".

docs/authorization/globus_app/apps.rst
```diff
@@ -0,0 +1,89 @@
...
+
+A ``GlobusApp`` will always require a programmatic entity called a "client". Depending
+on the type of ``GlobusApp`` in use, the client will play different roles in the
+request flow.
+
+*   A :class:`UserApp` drives interaction starting with a browser based login
+    flow. With this type of ``GlobusApp``, newly created resources (collections,
+    flows, etc.) are owned by the human user who has logged in (and thus granted the
+    app authorization tokens). Similarly resource access is evaluated based on that
+    user's permissions.
```
ada-globus, line 27:
```suggestion
*   A :class:`UserApp` drives interaction using a web-browser-based login
    flow. With this type of ``GlobusApp``, all service interactions are performed as the
    Globus-identity of the user who logged in. For instance, newly created resources
    (collections, flows, etc.) may be owned by that user. Similarly, access to resources
    will be evaluated based on that user's permissions.
```

docs/authorization/globus_app/apps.rst
sirosen: > Wondering if the first use of client here needs to link out to the relevant Auth docs: https://docs.globus.org/api/auth/developer-guide/#developing-apps

I vote (strongly) against this. We should have SDK-specific documentation on this subject, and having two "sources of truth" for this information will confuse users.

> ...also wondering if we should indicate which of the options from the provided list "Service Account, Portal / Science Gateway, Thick Client, API, or Advanced Application" are relevant?

:+1: for this.
I have some minor gripes about those categories being what they are[^1], but I think matching the web experience in our docs will help users.

[^1]: This part of the webapp is only for developers, but intentionally choosing use-case names rather than names meant to match the SDKs and Auth docs _assumes_ that the user in question is non-technical. It's dissonant and I don't see that it was necessary to introduce this discrepancy. I strongly believe that this is a poor UX decision, but I only minorly care about it.

docs/authorization/globus_app/apps.rst
```diff
@@ -0,0 +1,89 @@
...
+
+    A :class:`UserApp` should be used in cases where a script simplifies a process
+    for a user, but the user is ultimately responsible for the actions taken by the
+    script. The `Globus CLI <https://docs.globus.org/cli/>`_ is a good example of
+    one such use case.
```
ada-globus, line 32:
```suggestion
    A :class:`UserApp` is appropriate when a script or local application simplifies a process
    for a user, but the user is still ultimately responsible for the actions performed.
    The `Globus CLI <https://docs.globus.org/cli/>`_ is one such example.
```

docs/authorization/globus_app/apps.rst
ada-globus: I'm totally okay with not doing the first suggestion if we do the second 😄 on reflection, I think that's actually the one I care more about.

docs/authorization/globus_app/apps.rst
```diff
@@ -0,0 +1,89 @@
...
+
+    A :class:`UserApp` **should use a "native"-type client**. "Confidential"-type
+    clients are allowed, but prove more difficult to configure.
```
ada-globus, line 35:
```suggestion
    A :class:`UserApp` **should use a "native" client**. "Confidential"
    clients are allowed, but prove more difficult to configure.
```

ada-globus: Follow-up question: Is there _ever_ a compelling reason to use a confidential client in this case? (i.e., if there is a situation where that's actually the correct solution, it _might_ be useful to say something like "Confidential clients are allowed but should only be used if you need to ...")

docs/authorization/globus_app/apps.rst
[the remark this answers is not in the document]
sirosen: > Web services that require dependent tokens effectively don't work with GlobusApp today.

That's not what I'm describing as what should work -- that's the resource server case.

I'm talking about a web application which is a pure client, and does the 3-legged OAuth login flow (authorization code grant). Think something like a heavyweight django app which is a client to Globus services.
That's a hosted web app but it's not a resource server.

docs/authorization/globus_app/apps.rst
```diff
@@ -0,0 +1,89 @@
...
+
+*   A :class:`ClientApp` by contrast does not require any manual login flow. Instead
+    newly created resources (collections, flows, etc.) are owned by the client
+    itself, not a human. Similarly, resource access will be evaluated based on the
+    client's permissions.
```
ada-globus, line 40:
```suggestion
*   A :class:`ClientApp` does not use a login flow.
    All service interactions are performed as the application itself, rather
    than as a Globus user's identity. Resources created (collections, flows, etc.)
    are may be owned by the application's client. Similarly, resource access will be
    evaluated based on the application's client's permissions.
```

docs/authorization/globus_app/apps.rst
```diff
@@ -0,0 +1,89 @@
...
+
+    A :class:`ClientApp` should be used in cases where a service account, potentially
+    managed by multiple humans, is responsible for the actions taken by the script or
+    process. These cases mostly involve some kind of automation, i.e., frequent data
+    movement or flow invocation.
```
ada-globus, line 45:
```suggestion
    A :class:`ClientApp` is appropriate when a service account, potentially
    managed by multiple humans, is responsible for actions taken by the script or
    process. These cases mostly involve some kind of automation, e.g., frequent data
    movement or flow invocation.
```

docs/authorization/globus_app/apps.rst
```diff
@@ -0,0 +1,89 @@
...
+
+    A :class:`ClientApp` **must use a "confidential"-type client**. "Native"-type
+    clients don't have the necessary security properties (a client secret) to own their
+    own resources and will be rejected.
```
ada-globus, line 49:
```suggestion
    A :class:`ClientApp` **must use a "confidential" client**. "Native"
    clients don't have the necessary security properties (a client secret) to manage
    resources and will be rejected.
```

docs/authorization/globus_app/apps.rst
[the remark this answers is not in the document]
ada-globus:
```suggestion
    Web services (and other hosted applications that act as resource servers) should see :ref:`globus_authorizers` instead.
```

docs/authorization/globus_app/apps.rst
```diff
@@ -0,0 +1,89 @@
...
+
+    .. note::
+
+        Not all Globus services or contexts support ``ClientApp``-based interactions.
+
+        GCS collections with sensitive data, for instance, may require that a human user
+        has authenticated with their university email recently before allowing data
+        interaction. In these cases, a ``UserApp`` must be used instead.
```
ada-globus, line 57:
```suggestion
        GCS collections with sensitive data, for instance, may require that a Globus user
        has authenticated with their university email recently before allowing data
        interaction. In these cases, a ``UserApp`` must be used instead.
```

docs/authorization/globus_app/apps.rst
```diff
@@ -0,0 +1,89 @@
...
+
+
+Both "native" and "confidential" type clients, may be created and managed as a part
+of a `Globus Project <https://app.globus.org/settings/developers>`_.
```
ada-globus, line 61:
```suggestion
Both "native" and "confidential" clients may be created and managed as part
of a `Globus Project <https://app.globus.org/settings/developers>`_.
```

docs/authorization/globus_app/apps.rst
sirosen: My short answer here is "yes".

There are only two specific cases I know of: 
- a hosted client (e.g., a Django app)
- a templated client (e.g., actual cli implementation) 

I still wonder a little if we should split UserApp into multiple classes for the different cases, but I figured that we can add subclasses in the future if we want.

docs/authorization/globus_app/apps.rst
derek-globus: I'm hapyp to replace "Globus" with "globus-sdk" but I want to keep "service clients".

The term "client" is overloaded , particularly in this section of the docs so I want to be as clear as possible about what type of client we mean. I'll make the term "service client" be a link to the "[Service Clients](https://globus-sdk-python.readthedocs.io/en/stable/services/index.html)" section of our docs which hopefully will alleviate the concern that it's a Globus Auth thing.

sirosen: > The term "client" is overloaded

No contest!

:+1: for <code>globus-sdk :ref:&#96;service clients <service_clients>&#96;</code>

I'm not sure if we ever intentionally made "service client" a term of art -- other than it being the heading on that page, I don't think the term is used anywhere. And I would tend to pedantically call them "globus-sdk client classes" or "globus-sdk client objects" in most contexts. But I don't really object to the terminology and -- hey! -- it's the name used in the `services/index.rst` doc.

src/globus_sdk/globus_app/app.py
```diff
@@ -283,7 +271,7 @@ def login_required(self) -> bool:
 
     def logout(self) -> None:
         """
-        Logout an auth entity from the app.
+        Logout an auth entity out from the app.
```
sirosen, line 274: I'm not sure why I'm only seeing this now, but I think either this needs to revert or it needs to change to
`Log ... out from the app`.

Otherwise, this is saying "logout out of the app".

Trying to find my "best liked" phrasing without spending more than 30s on this, I come up with...
```
Log the current user or client out of the app.
```

sirosen (changes requested):

docs/authorization/globus_app/apps.rst
```diff
@@ -0,0 +1,99 @@
...
+
+There are two types of ``GlobusApps`` available in the SDK: :class:`UserApp` and
+:class:`ClientApp`, each modeling a different style of service interaction:
+
+.. list-table::
+    :widths: 50 50
+    :header-rows: 1
+
+    *   - **UserApp**
+        - **ClientApp**
```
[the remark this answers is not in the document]
sirosen: I looked over at the docs prior to this PR, and I'm wondering why we aren't relying on the definitions which are currently available in the `Using a GlobusApp` doc. They read extremely well to me:

> There are two flavors of GlobusApp:
>
> - [UserApp](https://globus-sdk-python.readthedocs.io/en/stable/experimental/globus_app.html#globus_sdk.experimental.globus_app.UserApp), a GlobusApp for interactions between an end user and Globus services. Operations are performed as a user identity.
>
> - [ClientApp](https://globus-sdk-python.readthedocs.io/en/stable/experimental/globus_app.html#globus_sdk.experimental.globus_app.ClientApp), a GlobusApp for interactions between a client (i.e. service account) and Globus services. Operations are performed as a client identity.

These definitions are both accurate and short.
I think we should start with these at the top-level, and that any details on things like mapping use-cases or behaviors to the two types should go somewhere later.

docs/authorization/globus_app/apps.rst
```diff
@@ -0,0 +1,99 @@
...
+
+    *   - Appropriate for simplifying a process otherwise performed by a user (e.g.,
+          the `Globus CLI <https://docs.globus.org/cli/>`_)
```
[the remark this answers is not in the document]
sirosen: My view has refined and softened as I've become more of a proponent of making the definitions separate.

docs/authorization/globus_app/apps.rst
derek-globus: Reclassified as a note & updated the wording here; I pulled from both of your suggestions for the final copy.

docs/authorization/globus_app/apps.rst
derek-globus: Ended up going with `"globus-sdk services" where "services" is a link to our "service clients" page.

I also made an effort to qualify other usages of "client" across the page as "OAuth2 client"

docs/authorization/globus_app/apps.rst
derek-globus: Refactored this section a good amount; the linkage to specific webapp based terminology (point two) is included now in the comparison table.

docs/authorization/globus_app/apps.rst
derek-globus: This is now in the table with updated wording.

docs/authorization/globus_app/apps.rst
derek-globus: The content for this piece now reads (in a table):

> Commonly used to perform actions as a specific end user (e.g., the `Globus CLI <https://docs.globus.org/cli/>`_)

docs/authorization/globus_app/apps.rst
derek-globus: This concept is reflected in the table now.

docs/authorization/globus_app/apps.rst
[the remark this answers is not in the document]
derek-globus: Added these definitions to the start.

docs/authorization/globus_app/apps.rst
```diff
@@ -0,0 +1,99 @@
...
+        - Created resources (e.g., collections or flows) by default are owned by the
+          OAuth2 client
+
+    *   - Existing resource access is evaluated based on a user's permissions
+        - Existing resource access is evaluated based on the OAuth2 client's permissions
+
+    *   - Should use a "native" OAuth2 client
+
+          (`Register a thick client
+          <https://app.globus.org/settings/developers/registration/public_installed_client>`_)
```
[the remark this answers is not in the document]
derek-globus: Expanded this to `Should typically use a "native"... May use a "confidential"...`

docs/authorization/globus_app/apps.rst
```diff
@@ -0,0 +1,99 @@
...
+application at the `Globus Developer's Console <https://app.globus.org/developers>`_.
+
+The following section provides a comparison of of the specific types of ``GlobusApps``
+to aid in selecting the proper one for your use case.
+
+Types of GlobusApps
+-------------------
+
+There are two types of ``GlobusApps`` available in the SDK: :class:`UserApp` and
+:class:`ClientApp`, each modeling a different style of service interaction:
```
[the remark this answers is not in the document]
derek-globus: Pulled this into its own sentence.

src/globus_sdk/globus_app/app.py
derek-globus: Updated to use "a user or client" in this method and login

docs/authorization/globus_app/apps.rst
derek-globus: I switched back to "appropriate for"; with the other wording shift, I do think that "appropriate" is more accurate and reads better than "commonly used to"

sirosen (approved):

PR #1078, opened 2024-10-09
Add ComputeClient methods to manage functions

rjmello:
Add `.register_function()` and `.delete_function()` methods to the `ComputeClient` class.

----
📚 Documentation preview 📚: https://globus-sdk-python--1078.org.readthedocs.build/en/1078/

### DIFF
--- a/changelog.d/20241008_214456_30907815+rjmello_compute_client_function_methods.rst
+++ b/changelog.d/20241008_214456_30907815+rjmello_compute_client_function_methods.rst
@@ -1,0 +1,8 @@
+Added
+~~~~~
+
+- Added ``ComputeClient.register_function()`` and ``ComputeClient.delete_function()``
+  methods. (:pr:`NUMBER`)
+
+- Added ``ComputeFunctionDocument`` and ``ComputeFunctionMetadata`` data classes for
+  use with the ``ComputeClient.register_function()`` method. (:pr:`NUMBER`)
--- a/src/globus_sdk/_generate_init.py
+++ b/src/globus_sdk/_generate_init.py
@@ -136,6 +136,8 @@ _LAZY_IMPORT_TABLE: list[tuple[str, tuple[str, ...]]] = [
         (
             "ComputeClient",
             "ComputeAPIError",
+            "ComputeFunctionDocument",
+            "ComputeFunctionMetadata",
         ),
     ),
     (
--- a/src/globus_sdk/_testing/data/compute/delete_function.py
+++ b/src/globus_sdk/_testing/data/compute/delete_function.py
@@ -1,0 +1,13 @@
+from globus_sdk._testing.models import RegisteredResponse, ResponseSet
+
+from ._common import FUNCTION_ID
+
+RESPONSES = ResponseSet(
+    metadata={"function_id": FUNCTION_ID},
+    default=RegisteredResponse(
+        service="compute",
+        path=f"/v2/functions/{FUNCTION_ID}",
+        method="DELETE",
+        json={"result": 302},
+    ),
+)
--- a/src/globus_sdk/services/compute/data.py
+++ b/src/globus_sdk/services/compute/data.py
@@ -1,0 +1,53 @@
+from __future__ import annotations
+
+from globus_sdk import utils
+from globus_sdk._types import UUIDLike
+from globus_sdk.utils import MISSING, MissingType
+
+
+class ComputeFunctionMetadata(utils.PayloadWrapper):
+    """A wrapper for function metadata.
+
+    :param python_version: The Python version used to serialize the function.
+    :param sdk_version: The Globus Compute SDK version used to serialize the function.
+    """
+
+    def __init__(
+        self,
+        *,
+        python_version: str | MissingType = MISSING,
+        sdk_version: str | MissingType = MISSING,
+    ):
+        super().__init__()
+        self["python_version"] = python_version
+        self["sdk_version"] = sdk_version
+
+
+class ComputeFunctionDocument(utils.PayloadWrapper):
+    """A function registration document.
+
+    :param function_name: The name of the function.
+    :param function_code: The serialized function source code.
+    :param description: The description of the function.
+    :param metadata: The metadata of the function.
+    :param group: Restrict function access to members of this Globus group.
+    :param public: Indicates whether the function is public.
+    """
+
+    def __init__(
+        self,
+        *,
+        function_name: str,
+        function_code: str,
+        description: str | MissingType = MISSING,
+        metadata: ComputeFunctionMetadata | MissingType = MISSING,
+        group: UUIDLike | MissingType = MISSING,
+        public: bool = False,
+    ):
+        super().__init__()
+        self["function_name"] = function_name
+        self["function_code"] = function_code
+        self["description"] = description
+        self["metadata"] = metadata
+        self["group"] = group
+        self["public"] = public
--- a/docs/services/compute.rst
+++ b/docs/services/compute.rst
@@ -14,6 +14,17 @@ Globus Compute
         ..  listknownscopes:: globus_sdk.scopes.ComputeScopes
             :base_name: ComputeClient.scopes
 
+Helper Objects
+--------------
+
+.. autoclass:: ComputeFunctionDocument
+   :members:
+   :show-inheritance:
+
+.. autoclass:: ComputeFunctionMetadata
+   :members:
+   :show-inheritance:
+
 Client Errors
 -------------
 
--- a/src/globus_sdk/__init__.py
+++ b/src/globus_sdk/__init__.py
@@ -71,6 +71,8 @@ _LAZY_IMPORT_TABLE = {
     "services.compute": {
         "ComputeClient",
         "ComputeAPIError",
+        "ComputeFunctionDocument",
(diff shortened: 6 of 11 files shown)


src/globus_sdk/_generate_init.py
```diff
@@ -136,6 +136,8 @@ def __getattr__(name: str) -> t.Any:
         (
             "ComputeClient",
             "ComputeAPIError",
+            "FunctionDocument",
+            "FunctionMetadata",
```
sirosen, line 140: I _think_ we should name these `ComputeFunctionDocument` and `ComputeFunctionMetadata`, although I could be convinced otherwise.

Right now, we have a very flat namespace for the service-related SDK exports -- `globus_sdk.X` for all `X`.
In most cases they're unambiguously associated with a given service, e.g., `StorageGatewayPolicies` is associated with GCS and couldn't have meaning with respect to anything else.

In cases where user confusion seems possible, we opt for longer names, e.g., `SearchScrollQuery` rather than `ScrollQuery`.

There's no hard-and-fast rule here, so there's nothing definitively _wrong_ with `FunctionDocument` and `FunctionMetadata` as names. The underlying problem we're looking at is that services haven't been given their own namespaces to populate freely -- and it's something I've been mulling for a few years without having any clear plan of action yet.

When I look at these names not only in light of the Globus ecosystem, but in comparison with the broader Python ecosystem, I think `FunctionMetadata` looks too much like the type of object one would expect from `inspect` or `types` in the stdlib -- and the SDK does contain some special values in the Python context, like `MISSING`, which make that kind of confusion possible.

src/globus_sdk/services/compute/data.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def __init__(
+        self,
+        *,
+        function_name: str,
+        function_code: str,
+        description: t.Optional[str] = None,
+        metadata: t.Optional[FunctionMetadata] = None,
+        group: t.Optional[UUIDLike] = None,
+        public: bool = False,
```
sirosen, line 39: Should the defaults for these be that they are present with `null` values, or that they be omitted?

In newer document types and APIs we're trying to use `MISSING` instead of `None`. Very frequently, `None` has been used, historically, as a default where `MISSING` would be better. It doesn't apply to all APIs, however. `MISSING` isn't _inherently better_. It's just different in a way which we should consider.

For an example of usage, look at the GCS EndpointDocument type:
https://github.com/globus/globus-sdk-python/blob/6613e15a8928db39149b630441b630a73a1bcc23/src/globus_sdk/services/gcs/data/endpoint.py#L92

When the SDK is encoding a document as JSON to send, it _strips out_ `MISSING` values. So the effect is the difference between the following two encodings:
```python
# as written today
encode(
  FunctionDocument('foo', 'bar')
)
# {"function_name": "foo", "function_code": bar", "description": null, "metadata": null, "group": null, "public": False}

# if it were written using MISSING for defaults
encode(
  FunctionDocument('foo', 'bar')
)
# {"function_name": "foo", "function_code": bar"}
```

Note that I also assumed in this example that we make `public: bool | MissingType = MISSING` the default.

The difference comes into play most prominently with APIs which use a null field to indicate that a value should change / be removed in an update call. But it also matters here: if `public = False` is the default, then that's a client-side default. If `public = MISSING` is the default, then that's a server-side default.

---

If you're curious about how the behavior is implemented, here's MISSING:
https://github.com/globus/globus-sdk-python/blob/6613e15a8928db39149b630441b630a73a1bcc23/src/globus_sdk/utils.py#L25

And here's where MISSING gets filtered from payloads:
https://github.com/globus/globus-sdk-python/blob/6613e15a8928db39149b630441b630a73a1bcc23/src/globus_sdk/transport/encoders.py#L96-L99

(It also gets filtered from headers and query params.)

---

The exact behavior you will want depends on the behavior of the service, so I'm not advocating for a change so much as I'm making sure we discuss which is appropriate for the Compute API.

src/globus_sdk/services/compute/data.py
```diff
@@ -0,0 +1,47 @@
...
+    """A wrapper for function metadata.
+
+    :param python_version: The Python version used to serialize the function.
+    :param sdk_version: The Globus Compute SDK version used to serialize the function.
+    """
+
+    def __init__(self, *, python_version: str, sdk_version: str):
+        super().__init__()
+        self["python_version"] = python_version
+        self["sdk_version"] = sdk_version
```
sirosen, line 17: Can we define defaults for these fields from the current execution environment?

The former looks like it's formatted `sys.version_info` data. I just tried out this little ditty and it seems to do the right thing:
```python
import sys


def fmt_version(v) -> str:
    if v.releaselevel == "final":
        return f"{v.major}.{v.minor}.{v.micro}"
    else:
        designator = {
            "candidate": "rc",
            "beta": "b",
            "alpha": "a",
        }[v.releaselevel]
        return f"{v.major}.{v.minor}.{v.micro}{designator}{v.serial}"


print(fmt_version(sys.version_info))
```

We could do something similar to pull the `globus-compute-sdk` version using importlib metadata.

---

This is as much a question about understanding of the service as it is about what we could do in this class.
When are callers expected to construct this object? Is it typically referring to the current runtime environment or are there significant cases in which it is meant to refer to some _other_ runtime when data was constructed?

src/globus_sdk/_testing/data/compute/delete_function.py
```diff
@@ -0,0 +1,13 @@
...
+
+from ._common import FUNCTION_ID
+
+RESPONSES = ResponseSet(
+    metadata={"function_id": FUNCTION_ID},
+    default=RegisteredResponse(
+        service="compute",
+        path=f"/v2/functions/{FUNCTION_ID}",
+        method="DELETE",
+        json={"result": 302},
```
sirosen, line 11: I'm curious, what does the 302 in this context mean?

changelog.d/20241008_214456_30907815+rjmello_compute_client_function_methods.rst
```diff
@@ -0,0 +1,6 @@
+Added
+~~~~~
+
+- Added a ``ComputeClient.register_function()`` method. (:pr:`NUMBER`)
+
+- Added a ``ComputeClient.delete_function()`` method. (:pr:`NUMBER`)
```
sirosen, line 6: I don't feel strongly about phrasing, but I think these should be combined:

```suggestion
- Add ``ComputeClient.register_function()`` and ``ComputeClient.delete_function()``. (:pr:`NUMBER`)
```

Also, I think another bullet point is needed for the addition of helper classes.

src/globus_sdk/_generate_init.py
rjmello: No objections from me. This will be more relevant for the endpoint methods because we need to differentiate between GCS and Compute endpoints.

src/globus_sdk/services/compute/data.py
rjmello: The register function route currently handles null and missing values the same, so I'm fine with the proposed change.

src/globus_sdk/_testing/data/compute/delete_function.py
rjmello: It's an HTTP status code integer meant to indicate a successful deletion. This is a relic that we will change in the v3 API route.

src/globus_sdk/services/compute/data.py
sirosen: If they're equivalent (which is common for many APIs), I have a slight preference for MISSING. That way, if the service _adds_ a default behavior -- e.g., "default description is `<name> created by <user> on <date>`" -- the SDK picks it up seamlessly.

src/globus_sdk/services/compute/data.py
rjmello: It refers to the runtime environment when serializing the function. We currently set these values in the Compute SDK, where the serialization takes place. I prefer to leave it as-is for now.

sirosen: Okay. It's easier to add defaults than it is to take them away (the former is backwards compatible but the latter is not), so I see no reason to block on this. Mostly was just curious. :slightly_smiling_face:

sirosen (approved):
I assume we want to wait for a Compute maintainer to read and approve as well, so I'll refrain from merging for now.

khk-globus (approved):
I have not tried this out, but a static analysis looks good to me.

changelog.d/20241008_214456_30907815+rjmello_compute_client_function_methods.rst
```diff
@@ -0,0 +1,8 @@
+Added
+~~~~~
+
+- Added ``ComputeClient.register_function()`` and ``ComputeClient.delete_function()``
+  methods. (:pr:`NUMBER`)
+
+- Added ``ComputeFunctionDocument`` and ``ComputeFunctionMetadata`` data classes for
+  use with the ``ComputeClient.register_function()`` method. (:pr:`NUMBER`)
```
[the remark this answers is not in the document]
sirosen: I've wanted to enhance our doc build pipeline to give a rendered context for these. Because we're using `scriv` to avoid changelog conflicts, it's possible to... :bulb: :exclamation: 
We'll probably change style to use sphinx roles if I can make things work the way I think we can.

PR #1079, opened 2024-10-11
Support `version` field in `SearchClient.post_search()`

m1yag1:
https://app.shortcut.com/globus/story/35200/python-sdk-support-version-field-in-searchclient-post-search

## Summary of changes

* Added ignore for `SearchQuery` in pyproject.toml filterwarnings for pytest
* Updated SphinxDocs for `SearchQueryV1`
* Added `SearchQueryV1` to the _generate_init.py script
* Added `SearchQueryV1` class for new v1 search request version
* Added deprecation warning to the `SearchQuery` class during init
* Added unit and functional tests for `SearchQueryV1` and `SearchQuery` deprecation warnings.

----
📚 Documentation preview 📚: https://globus-sdk-python--1079.org.readthedocs.build/en/1079/

### DIFF
--- a/changelog.d/20241011_113822_8730430+m1yag1_sc_35200_version_field_post_search.rst
+++ b/changelog.d/20241011_113822_8730430+m1yag1_sc_35200_version_field_post_search.rst
@@ -1,0 +1,6 @@
+Development
+~~~~~~~~~~~
+
+- ``SearchQueryV1`` is a new class for submitting complex queries replacing
+the legacy ``SearchQuery`` class. A deprecation warning has been added to the
+``SearchQuery`` class. (:pr:`1079`)
--- a/changelog.d/20241011_113822_8730430+m1yag1_sc_35200_version_field_post_search.rst
+++ b/changelog.d/20241011_113822_8730430+m1yag1_sc_35200_version_field_post_search.rst
@@ -1,5 +1,5 @@
-Development
-~~~~~~~~~~~
+Added
+~~~~~
 
 - ``SearchQueryV1`` is a new class for submitting complex queries replacing
 the legacy ``SearchQuery`` class. A deprecation warning has been added to the
--- a/changelog.d/20241011_113822_8730430+m1yag1_sc_35200_version_field_post_search.rst
+++ b/changelog.d/20241011_113822_8730430+m1yag1_sc_35200_version_field_post_search.rst
@@ -2,5 +2,5 @@ Added
 ~~~~~
 
 - ``SearchQueryV1`` is a new class for submitting complex queries replacing
-the legacy ``SearchQuery`` class. A deprecation warning has been added to the
-``SearchQuery`` class. (:pr:`1079`)
+  the legacy ``SearchQuery`` class. A deprecation warning has been added to the
+  ``SearchQuery`` class. (:pr:`1079`)
--- a/docs/services/search.rst
+++ b/docs/services/search.rst
@@ -25,6 +25,10 @@ only to document the methods it provides to its subclasses.
    :members:
    :show-inheritance:
 
+.. autoclass:: SearchQueryV1
+    :members:
+    :show-inheritance:
+
 .. autoclass:: SearchScrollQuery
    :members:
    :show-inheritance:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,6 +67,7 @@ testpaths = ["tests"]
 norecursedirs = ["tests/non-pytest"]
 filterwarnings = [
     "error",
+    "ignore:'SearchQuery'*:DeprecationWarning",
 ]
 
 [tool.scriv]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,7 +67,6 @@ testpaths = ["tests"]
 norecursedirs = ["tests/non-pytest"]
 filterwarnings = [
     "error",
-    "ignore:'SearchQuery'*:DeprecationWarning",
 ]
 
 [tool.scriv]
--- a/src/globus_sdk/services/search/data.py
+++ b/src/globus_sdk/services/search/data.py
@@ -115,9 +115,7 @@ class SearchQuery(SearchQueryBase):
         additional_fields: dict[str, t.Any] | None = None,
     ):
         super().__init__()
-        exc.warn_deprecated(
-            "'SearchQuery' is a deprecated name. Use 'SearchQueryV1' instead."
-        )
+        exc.warn_deprecated("'SearchQuery' is deprecated. Use 'SearchQueryV1' instead.")
         if q is not None:
             self["q"] = q
         if limit is not None:
@@ -240,12 +238,17 @@ class SearchQueryV1(utils.PayloadWrapper):
 
     def __init__(
         self,
-        q: str | None = None,
         *,
+        q: str | None = None,
         limit: int | None = None,
         offset: int | None = None,
         advanced: bool | None = None,
         additional_fields: dict[str, t.Any] | None = None,
+        filters: list[dict[str, t.Any]] | utils.MissingType = utils.MISSING,
+        facets: list[dict[str, t.Any]] | utils.MissingType = utils.MISSING,
+        post_facet_filters: list[dict[str, t.Any]] | utils.MissingType = utils.MISSING,
+        boosts: list[dict[str, t.Any]] | utils.MissingType = utils.MISSING,
+        sorts: list[dict[str, t.Any]] | utils.MissingType = utils.MISSING,
     ):
         super().__init__()
         self["@version"] = "query#1.0.0"
@@ -260,11 +263,11 @@ class SearchQueryV1(utils.PayloadWrapper):
         if additional_fields is not None:
             self.update(additional_fields)
 
-        self["filters"] = []
-        self["facets"] = []
-        self["post_facet_filters"] = []
-        self["boosts"] = []
-        self["sorts"] = []
+        self["filters"] = filters
+        self["facets"] = facets
+        self["post_facet_filters"] = post_facet_filters
+        self["boosts"] = boosts
+        self["sorts"] = sorts
 
 
 class SearchScrollQuery(SearchQueryBase):
--- a/src/globus_sdk/services/search/data.py
+++ b/src/globus_sdk/services/search/data.py
@@ -238,6 +238,7 @@ class SearchQueryV1(utils.PayloadWrapper):
     :param post_facet_filters: a list of filters to apply after facet
         results are returned
     :param boosts: a list of boosts to apply to the query
+    :param sort: a list of fields to sort results
     :param additional_fields: additional data to include in the query document
     """
 
--- a/src/globus_sdk/services/search/data.py
+++ b/src/globus_sdk/services/search/data.py
@@ -233,6 +233,11 @@ class SearchQueryV1(utils.PayloadWrapper):
     :param advanced: Whether to enable (``True``) or not to enable (``False``) advanced
         parsing of query strings. The default of ``False`` is robust and guarantees that
         the query will not error with "bad query string" errors
+    :param filters: a list of filters to apply to the query
+    :param facets: a list of facets to apply to the query
+    :param post_facet_filters: a list of filters to apply after facet
+        results are returned
+    :param boosts: a list of boosts to apply to the query
     :param additional_fields: additional data to include in the query document
     """
 
@@ -243,12 +248,12 @@ class SearchQueryV1(utils.PayloadWrapper):
         limit: int | utils.MissingType = utils.MISSING,
         offset: int | utils.MissingType = utils.MISSING,
         advanced: bool | utils.MissingType = utils.MISSING,
-        additional_fields: dict[str, t.Any] | utils.MissingType = utils.MISSING,
         filters: list[dict[str, t.Any]] | utils.MissingType = utils.MISSING,
         facets: list[dict[str, t.Any]] | utils.MissingType = utils.MISSING,
         post_facet_filters: list[dict[str, t.Any]] | utils.MissingType = utils.MISSING,
         boosts: list[dict[str, t.Any]] | utils.MissingType = utils.MISSING,
         sorts: list[dict[str, t.Any]] | utils.MissingType = utils.MISSING,
+        additional_fields: dict[str, t.Any] | utils.MissingType = utils.MISSING,
     ):
         super().__init__()
         self["@version"] = "query#1.0.0"
--- a/src/globus_sdk/services/search/data.py
+++ b/src/globus_sdk/services/search/data.py
@@ -239,11 +239,11 @@ class SearchQueryV1(utils.PayloadWrapper):
     def __init__(
         self,
         *,
-        q: str | None = None,
-        limit: int | None = None,
-        offset: int | None = None,
-        advanced: bool | None = None,
-        additional_fields: dict[str, t.Any] | None = None,
+        q: str | utils.MissingType = utils.MISSING,
+        limit: int | utils.MissingType = utils.MISSING,
+        offset: int | utils.MissingType = utils.MISSING,
+        advanced: bool | utils.MissingType = utils.MISSING,
+        additional_fields: dict[str, t.Any] | utils.MissingType = utils.MISSING,
         filters: list[dict[str, t.Any]] | utils.MissingType = utils.MISSING,
         facets: list[dict[str, t.Any]] | utils.MissingType = utils.MISSING,
         post_facet_filters: list[dict[str, t.Any]] | utils.MissingType = utils.MISSING,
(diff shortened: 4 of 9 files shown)


changelog.d/20241011_113822_8730430+m1yag1_sc_35200_version_field_post_search.rst
```diff
@@ -0,0 +1,6 @@
+Development
+~~~~~~~~~~~
```
sirosen, line 2:
```suggestion
Added
~~~~~
```

changelog.d/20241011_113822_8730430+m1yag1_sc_35200_version_field_post_search.rst
```diff
@@ -0,0 +1,6 @@
...
+
+- ``SearchQueryV1`` is a new class for submitting complex queries replacing
+the legacy ``SearchQuery`` class. A deprecation warning has been added to the
+``SearchQuery`` class. (:pr:`1079`)
```
sirosen, line 6:
```suggestion
  the legacy ``SearchQuery`` class. A deprecation warning has been added to the
  ``SearchQuery`` class. (:pr:`1079`)
```

pyproject.toml
```diff
@@ -67,6 +67,7 @@ testpaths = ["tests"]
 norecursedirs = ["tests/non-pytest"]
 filterwarnings = [
     "error",
+    "ignore:'SearchQuery'*:DeprecationWarning",
```
sirosen, line 70: I am against us ignoring our own warnings in our testsuite. We should resolve whatever warnings we have, or appropriate tests should be updated to apply a warning filter or (even better) to test for and confirm the warning emissions with `pytest.warns`.

src/globus_sdk/services/search/data.py
```diff
@@ -115,6 +115,9 @@ def __init__(
         additional_fields: dict[str, t.Any] | None = None,
     ):
         super().__init__()
+        exc.warn_deprecated(
+            "'SearchQuery' is a deprecated name. Use 'SearchQueryV1' instead."
```
sirosen, line 119:
```suggestion
            "'SearchQuery' is deprecated. Use 'SearchQueryV1' instead."
```

This is a minor edit to the warning string, but semantically significant. `SearchQuery` and `SearchQueryV1` are not different names for the same thing. They are different things.

changelog.d/20241011_113822_8730430+m1yag1_sc_35200_version_field_post_search.rst
```diff
@@ -0,0 +1,6 @@
+Development
```
m1yag1, line 1: I updated the changelog only for Development because the new `SearchQueryV1` is not fully implemented to replace the legacy search query.

pyproject.toml
m1yag1, line 70: Max and I attempted to use the `pytest.mark.filterwarnings` decorator but it doesn't work with classes that are being instantiated within pytest.mark.parametrize. I felt it was best to add the ignore to the `pyproject.toml` for now until the legacy SearchQuery is fully deprecated and then it could be removed.

src/globus_sdk/services/search/data.py
```diff
@@ -221,6 +224,49 @@ def add_sort(
...
+    :param additional_fields: additional data to include in the query document
+    """
+
+    def __init__(
+        self,
+        q: str | None = None,
+        *,
+        limit: int | None = None,
+        offset: int | None = None,
+        advanced: bool | None = None,
```
sirosen, line 247: I believe that all of these should have defaults of `MISSING`, which will allow us to set them in the body without `if`-checks on their values as well.

We recently discussed this over on the Compute-related PR, so I think the explanation I wrote there is really solid and explains the `None` vs `MISSING` divide:
https://github.com/globus/globus-sdk-python/pull/1078#discussion_r1794192893

But we can discuss here if there's anything unclear in that.

docs/services/search.rst
```diff
@@ -25,6 +25,10 @@ only to document the methods it provides to its subclasses.
    :members:
    :show-inheritance:
 
+.. autoclass:: SearchQueryV1
```
m1yag1, line 28: Within the docs I noticed were was an array for [deprecated items](https://github.com/globus/globus-sdk-python/blob/main/scripts/ensure_exports_are_documented.py#L25-L29). I wasn't sure if I needed to add `SearchQuery` to that array. Let me know if it does need to be added.

src/globus_sdk/services/search/data.py
```diff
@@ -221,6 +224,49 @@ def add_sort(
...
+    :param advanced: Whether to enable (``True``) or not to enable (``False``) advanced
+        parsing of query strings. The default of ``False`` is robust and guarantees that
+        the query will not error with "bad query string" errors
+    :param additional_fields: additional data to include in the query document
+    """
+
+    def __init__(
+        self,
+        q: str | None = None,
+        *,
```
sirosen, line 244: I'm not sure about this, but I think we want to move primarily to keyword-only arguments.

With a named positional, you're actually declaring two supported usages: `SearchQueryV1("*")` and `SearchQueryV1(q="*")`. Typically it's fine (we lived with it until Python 3, after all), until the following two things happen:
- you add a second named-positional argument
- you deprecate or remove the first named-positional argument

After such changes to an API, positional usage becomes problematic. So then you're in the position of only supporting keyword-arg usage anyway.

In Python APIs, I've increasingly tended towards keyword-only arguments:

```suggestion
        *,
        q: str | None = None,
```

One thing about this is that it's a reversible decision, whereas declaring support for a positional arg is not.
Once `SearchQueryV1("*")` is supported, support must be maintained.
But if `SearchQueryV1(q="*")` is the only supported usage, it can easily be updated to also allow `SearchQueryV1("*")` in the future.

src/globus_sdk/services/search/data.py
```diff
@@ -221,6 +224,49 @@ def add_sort(
...
+        if advanced is not None:
+            self["advanced"] = advanced
+        if additional_fields is not None:
+            self.update(additional_fields)
+
+        self["filters"] = []
+        self["facets"] = []
+        self["post_facet_filters"] = []
+        self["boosts"] = []
+        self["sorts"] = []
```
sirosen, line 267: I believe that all of these should be accepted as arguments of type `list[dict[str, t.Any]] | MissingType` with defaults of `MISSING`.

The difference here is between sending several empty arrays in a payload vs not sending them at all.

tests/unit/helpers/test_search.py
```diff
@@ -26,6 +26,37 @@ def test_init():
         assert param_query[par] == add_params[par]
 
 
+def test_init_legacy_deprecation_warning():
+    with pytest.warns(
+        DeprecationWarning,
```
sirosen, line 31: Let's check for `RemovedInV4Warning`, rather than the superclass here. It's a minor difference, but since we know what the type should be, let's verify that our contract is met. We don't want to send our users bare `DeprecationWarning`s, given that we document that we emit `RemovedInV4Warning`s.

tests/unit/helpers/test_search.py
```diff
@@ -26,6 +26,37 @@ def test_init():
...
+    params = {"q": "foo", "limit": 10, "offset": 0, "advanced": False}
+    param_query = SearchQueryV1(**params)
+    for par in params:
+        assert param_query[par] == params[par]
+
+    # init with additional_fields
+    add_params = {"param1": "value1", "param2": "value2"}
+    param_query = SearchQueryV1(additional_fields=add_params)
+    for par in add_params:
+        assert param_query[par] == add_params[par]
```
sirosen, line 57: A test for the initializer itself is good, but I would like us to mirror this with a functional test as well -- one which builds a `SearchQueryV1` document, sends it, and then fetches the last request to verify the payload contents.
Such a test is a good way to confirm that `MISSING` values are set in all of the appropriate slots.

i.e. I believe that the ideal is that we get this:
```python
test_client.post_search(SearchQueryV1(q="*"))
last_req = get_last_request()
parsed_body = json.loads(last_req.body)
assert parsed_body == {"q": "*"}
```

changelog.d/20241011_113822_8730430+m1yag1_sc_35200_version_field_post_search.rst
sirosen: To clarify, since our comments went in simultaneously:
Any additive change which we release is available to our users as soon as we release it.

The `Development` header is for infrastructural changes which primarily impact us as maintainers, but which also could impact unusual consumers like repackagers. For example, if we switched from `tox` to `nox`, that would have a significant impact on repackaging of the SDK, but would not impact the typical end-user who consumes it as an installed package.

docs/services/search.rst
sirosen: That list of elements exists to explicitly exempt deprecated symbols from the documentation requirement.
It would not be appropriate to remove documentation for the `SearchQuery` class, since users may continue to use it up until we remove it in the 4.0.0 release.

pyproject.toml
m1yag1: @sirosen I'll go through the tests and update them as necessary. The SearchQuery class is used in lot of places. If I turn deprecations to warnings it gives an idea of the number of errors.

```
src/globus_sdk/services/search/data.py:118: 2 warnings
tests/unit/helpers/test_search.py: 11 warnings
  /Users/blackbird/Projects/00_globus/00_automate/globus-sdk-python/src/globus_sdk/services/search/data.py:118: RemovedInV4Warning: 'SearchQuery' is deprecated. Use 'SearchQueryV1' instead.
    exc.warn_deprecated("'SearchQuery' is deprecated. Use 'SearchQueryV1' instead.")

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

sirosen: My opinion is that that's simply the work required here, in that case.

If we take a shortcut here, we're simply making someone do that work later. Right now, the problem is obvious and apparent. We should solve it correctly today.

m1yag1:
@sirosen I believe I've taken care of all the comments except the test updates for `pytest.mark.filterwarnings`. I'll work on each to figure out the necessary test changes and let you know here.

pyproject.toml
m1yag1: btw, I'm not advocating doing the wrong thing today. Happy to work through the tests and do it right 👍. I was interested in knowing the blast radius so now that I have numbers to count down as I work I know I've hit them all 😄 .

sirosen: Sure, it's not an accusation, after all!

It's more or less what follows from my stance that we shouldn't be ignoring our own warnings.
If we don't agree that it's wrong to have the ignore rule, then it's fine to leave it in place. But if we _do_ agree that the ignore rule is wrong, then we should fix it now.

src/globus_sdk/services/search/data.py
sirosen: This was marked resolved but I still see `... | None = None` in the diff, so I'm unresolving.

pyproject.toml
kurtmckee: Perhaps this could be addressed via an auto-used fixture?

sirosen: You mean a module-scoped ignore fixture for this? Possibly, if that's what's appropriate.
But if some of these are constructed in the module context (e.g., parametrize arguments), that won't work cleanly.

m1yag1: @sirosen I added the `filterwarning` marker where needed and split off the SearchQuery legacy helper into a separate test where required. At this point, I should have covered all the comments.

src/globus_sdk/services/search/data.py
```diff
@@ -221,6 +222,51 @@ def add_sort(
...
+    """
+
+    def __init__(
+        self,
+        *,
+        q: str | utils.MissingType = utils.MISSING,
+        limit: int | utils.MissingType = utils.MISSING,
+        offset: int | utils.MissingType = utils.MISSING,
+        advanced: bool | utils.MissingType = utils.MISSING,
+        additional_fields: dict[str, t.Any] | utils.MissingType = utils.MISSING,
```
sirosen, line 246: Although it doesn't matter in terms of semantics, we conventionally put `additional_fields` and `query_params` (where each of these apply) at the end of the list of parameters. The reason being, they are the "escape valves" if none of the options above apply.
I'd like us to shift this one to match that style.

src/globus_sdk/services/search/data.py
```diff
@@ -221,6 +222,51 @@ def add_sort(
...
+    A specialized dict which has helpers for creating and modifying a Search
+    Query document. Replaces the usage of ``SearchQuery``.
+
+    :param q: The query string. Required unless filters are used.
+    :param limit: A limit on the number of results returned in a single page
+    :param offset: An offset into the set of all results for the query
+    :param advanced: Whether to enable (``True``) or not to enable (``False``) advanced
+        parsing of query strings. The default of ``False`` is robust and guarantees that
+        the query will not error with "bad query string" errors
+    :param additional_fields: additional data to include in the query document
```
sirosen, line 236: I'm a little surprised that pylint is passing, given that we have some undocumented parameters. Maybe I need to look at its settings for `__init__`... Regardless, we need param doc for `filters`, etc. It should be pretty short and easy to add.

sirosen (changes requested):
Some minor final edits requested, and then we should be good to go!

src/globus_sdk/services/search/data.py
kurtmckee: pylint catches that? PyCharm notifies me about parameter mismatches in docstrings but I haven't spotted that in other linters.

sirosen: Well, it's _supposed to_, under our config:
```toml
[tool.pylint]
load-plugins = ["pylint.extensions.docparams"]
accept-no-param-doc = "false"

[tool.pylint."messages control"]
disable = [  # simplified -- we have other disabled rules
    "missing-raises-doc",
]
```

But I'm guessing that it doesn't understand that the class docstring applies to the init method. Maybe it's configurable and we can fix it.

src/globus_sdk/services/search/data.py
```diff
@@ -221,6 +222,56 @@ def add_sort(
...
+        *,
+        q: str | utils.MissingType = utils.MISSING,
+        limit: int | utils.MissingType = utils.MISSING,
+        offset: int | utils.MissingType = utils.MISSING,
+        advanced: bool | utils.MissingType = utils.MISSING,
+        filters: list[dict[str, t.Any]] | utils.MissingType = utils.MISSING,
+        facets: list[dict[str, t.Any]] | utils.MissingType = utils.MISSING,
+        post_facet_filters: list[dict[str, t.Any]] | utils.MissingType = utils.MISSING,
+        boosts: list[dict[str, t.Any]] | utils.MissingType = utils.MISSING,
+        sorts: list[dict[str, t.Any]] | utils.MissingType = utils.MISSING,
```
sirosen, line 255: Sorry for not catching this earlier, but it's `sort`, not `sorts`.

```suggestion
        sort: list[dict[str, t.Any]] | utils.MissingType = utils.MISSING,
```

src/globus_sdk/services/search/data.py
```diff
@@ -221,6 +222,56 @@ def add_sort(
...
+
+        self["q"] = q
+        self["limit"] = limit
+        self["offset"] = offset
+        self["advanced"] = advanced
+        self["filters"] = filters
+        self["facets"] = facets
+        self["post_facet_filters"] = post_facet_filters
+        self["boosts"] = boosts
+        self["sorts"] = sorts
```
sirosen, line 269:
```suggestion
        self["sort"] = sort
```

sirosen (changes requested):
I just caught a minor issue doing my final pre-approval pass. Sorry for not noticing it sooner.

Also, `sort` needs to be added to the docstring -- I didn't see it there.

m1yag1:
@sirosen no worries! I've pushed updates for the last review comment on sort.

src/globus_sdk/services/search/data.py
```diff
@@ -221,6 +222,56 @@ def add_sort(
...
+    :param limit: A limit on the number of results returned in a single page
+    :param offset: An offset into the set of all results for the query
+    :param advanced: Whether to enable (``True``) or not to enable (``False``) advanced
+        parsing of query strings. The default of ``False`` is robust and guarantees that
+        the query will not error with "bad query string" errors
+    :param filters: a list of filters to apply to the query
+    :param facets: a list of facets to apply to the query
+    :param post_facet_filters: a list of filters to apply after facet
+        results are returned
+    :param boosts: a list of boosts to apply to the query
```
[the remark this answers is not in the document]
m1yag1: OMG what is wrong with me. I've pushed an update that should have the `sort` param.

sirosen (approved):

PR #1080, opened 2024-10-14
Add a pre-build scriv step to RTD builds

sirosen:
When running a readthedocs PR build, automatically run `scriv collect` if there are changelog fragments, with a `-dev` suffixed version number.
In support of this, install `scriv` into docs environments, so that it can be run by readthedocs. (Requires a run of `tox r -m freezedeps`.)

The end result is that PR builds will be able to see a preview of the changelog. It won't necessarily be "fully cleaned up" in cases where we want to make some edits at the time of the release, but it will be close and in many cases perfect.

---

This change is a bit of an experiment. I'm curious if it will work in the way I would like. The RTD build on *this* PR will be a good test.

----
📚 Documentation preview 📚: https://globus-sdk-python--1080.org.readthedocs.build/en/1080/

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -8,6 +8,9 @@ build:
   os: "ubuntu-22.04"
   tools:
     python: "3.11"
+  jobs:
+    pre_build:
+      - bash ./scripts/rtd-pre-sphinx-build.sh
 
 python:
   install:
--- a/requirements/docs.in
+++ b/requirements/docs.in
@@ -6,3 +6,6 @@ sphinx-design
 
 # required for testing modules to load
 responses
+
+# required for the PR pre-build step in RTD
+scriv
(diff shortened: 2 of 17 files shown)


sirosen:
It works! Sample of the output:
image

(current in-process release is huge, which makes this both good and bad as a test case)

kurtmckee (approved):
Can CI be updated to follow this practice, so we're seeing build failures in CI (say, due to an incorrect reference, like `` :class:`SearchClein` ``)?

Or will a failing Read the Docs build trigger a CI failure? I'm not actually clear on this.

sirosen:
It should count as a CI failure, which would catch those bad references. (I think!)
But we could move/rename the script and run it in GitHub Actions if we find that RTD isn't doing what we want.

I'm looking a small additional change, but if it doesn't work I'll revert to the starting commit here.

PR #1088, opened 2024-10-16
Add Mike Arbelaez to codeowners

ada-globus:
----
📚 Documentation preview 📚: https://globus-sdk-python--1088.org.readthedocs.build/en/1088/

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,12 +6,12 @@
 #
 
 # default rule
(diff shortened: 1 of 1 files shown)


kurtmckee (approved):
Unbelievable that I overlooked Mike when adding Max in a previous PR.

Mike even went to the effort of 1337-izing his username so it would sort ahead of Max's and I still missed it! /smh

PR #1089, opened 2024-10-17
Switch from `.in` files to `[dependency-groups]`

sirosen:
`pip-tools` does not yet support Dependency Groups, but it is
in their tracker. Therefore, a small amount of shimming is needed to
apply it here.

1. Move the dependency groups from `requirements/*.in` files to
   `[dependency-groups]` in pyproject.toml
2. Add a tox environment (`freezedeps-print`) which installs
   `dependency-groups` (the package) and uses it to print dependency
   groups to new `.<name>.in` files, hidden and gitignored
3. Feed those files into `pip-compile` as before

Finally, run `tox r -m freezedeps` to confirm that the result works.

----
📚 Documentation preview 📚: https://globus-sdk-python--1089.org.readthedocs.build/en/1089/


pyproject.toml
```diff
@@ -1,3 +1,41 @@
...
+    "typing-extensions>=4.0",
+    # although 'responses' is provided by the test requirements, it also
+    # must be installed for proper type-checking on globus_sdk._testing
+    "responses",
+]
+typing-mindeps = [
+    {include-group = "typing"},
+    "typing-extensions==4.0",
+]
+
```
sirosen, line 38: I just looked at this in the GitHub render. Should this table come after the `[project]` and `[tools.setuptools]` tables? Maybe push to the bottom of the file?

kurtmckee: I'm inclined to have this come after -- at least -- `[build-system]`. It's not tool-specific, so it seems fine to have it above the tool-specific sections. :+1:

requirements/.gitignore
kurtmckee, line 1: I'd prefer to have this in the top level `.gitignore`.

kurtmckee (approved):

requirements/.gitignore
sirosen: Sure! But I still want it restricted to this dir.

kurtmckee (approved):

pyproject.toml
sirosen: I put it after the `[project]` table. I'm still fussing about "is it right?! is it right?!" internally, but IMO we can live with this and move it around later if we dislike it.

PR #1090, opened 2024-10-18
Minor test cleanup

sirosen:
- Do not use parametrize with single value
- Prefer scoped warning handlers (context manager + specific warning type)

----
📚 Documentation preview 📚: https://globus-sdk-python--1090.org.readthedocs.build/en/1090/

### DIFF
--- a/tests/unit/helpers/test_search.py
+++ b/tests/unit/helpers/test_search.py
@@ -4,29 +4,28 @@ Unit tests for globus_sdk.SearchQuery
 
 import pytest
 
-import globus_sdk
-from globus_sdk import SearchQuery, SearchQueryV1, utils
-from globus_sdk.exc.warnings import RemovedInV4Warning
+from globus_sdk import RemovedInV4Warning, SearchQuery, SearchQueryV1, utils
 
 
 def test_init_legacy():
-    """Creates SearchQuery and verifies results"""
-    with pytest.warns(
-        globus_sdk.exc.RemovedInV4Warning, match="'SearchQuery' is deprecated"
-    ):
+    params = {"q": "foo", "limit": 10, "offset": 0, "advanced": False}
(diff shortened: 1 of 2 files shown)


tests/unit/helpers/test_search.py
```diff
@@ -4,14 +4,17 @@
...
 
 
-@pytest.mark.filterwarnings("ignore:'SearchQuery'*:DeprecationWarning")
 def test_init_legacy():
     """Creates SearchQuery and verifies results"""
-    query = SearchQuery()
+    with pytest.warns(
+        globus_sdk.exc.RemovedInV4Warning, match="'SearchQuery' is deprecated"
+    ):
+        query = SearchQuery()
```
kurtmckee, line 17: These imports, and the subsequent usage here, aren't looking okay to me.

`globus_sdk` doesn't need to be imported, and the `RemovedInV4Warning` doesn't need to be qualified.

```python
    with pytest.warns(RemovedInV4Warning, match="'SearchQuery' is deprecated"):
        query = SearchQuery()
```

kurtmckee (changes requested):

tests/unit/helpers/test_search.py
sirosen: Oh, :facepalm: , I missed that it was already imported by name.

That said, we document it as `globus_sdk.RemovedInV4Warning`, so I think it's best if the testsuite matches that usage. (Which is also not what I wrote!)

I'll make a change right now which is fairly minimal. Although I generally prefer the verbose style of importing a package name (in our case, `globus_sdk`) and accessing all of the qualified names we need from that name, rewriting the whole test module to use that style isn't the small change I had in mind for this PR.

kurtmckee (approved):

PR #1093, opened 2024-10-24
Improve EOF error messaging during CLI login flows

chris-janidlo:
Motivated by observed issues in Compute - users can set up working Compute endpoints, change their setups slightly around access to STDIN, and then see cryptic EOF errors when they eventually need to re-auth.

NB: as of the time of this commit, Compute handles this by raising an error before running login flows if the current environment does not appear to be an interactive terminal. The goal of this commit is to move that behavior to a more centralized location.

----
📚 Documentation preview 📚: https://globus-sdk-python--1093.org.readthedocs.build/en/1093/

### DIFF
--- a/src/globus_sdk/login_flows/command_line_login_flow_manager.py
+++ b/src/globus_sdk/login_flows/command_line_login_flow_manager.py
@@ -2,8 +2,10 @@ from __future__ import annotations
 
 import textwrap
 import typing as t
+from contextlib import contextmanager
 
 import globus_sdk
+from globus_sdk.exc.base import GlobusError
 from globus_sdk.gare import GlobusAuthorizationParameters
 from globus_sdk.utils import get_nice_hostname
 
@@ -11,6 +13,13 @@ from .login_flow_manager import LoginFlowManager
 
 if t.TYPE_CHECKING:
     from globus_sdk.globus_app import GlobusAppConfig
+
+
+class CommandLineLoginFlowEOFError(GlobusError, EOFError):
+    """
+    An error raised when a CommandLineLoginFlowManager reads an EOF when prompting for
+    a code.
+    """
 
 
 class CommandLineLoginFlowManager(LoginFlowManager):
@@ -128,4 +137,16 @@ class CommandLineLoginFlowManager(LoginFlowManager):
         :returns: The authorization code entered by the user.
         """
         code_prompt = "Enter the resulting Authorization Code here: "
-        return input(code_prompt).strip()
+        with self._handle_input_errors():
+            return input(code_prompt).strip()
+
+    @contextmanager
+    def _handle_input_errors(self) -> t.Iterator[None]:
+        try:
+            yield
+        except EOFError as e:
+            msg = (
+                "An EOF was read when an authorization code was expected."
+                " (Are you running this in an interactive terminal?)"
+            )
+            raise CommandLineLoginFlowEOFError(msg) from e
--- a/changelog.d/20241024_113828_chris_command_line_login_flow_manager_catch_eof.rst
+++ b/changelog.d/20241024_113828_chris_command_line_login_flow_manager_catch_eof.rst
@@ -1,0 +1,5 @@
+Changed
+~~~~~~~
+
+- Improved error messaging around EOF errors when prompting for code during a command
+  line login flow (:pr:`NUMBER`)
--- a/docs/authorization/login_flows.rst
+++ b/docs/authorization/login_flows.rst
@@ -52,6 +52,9 @@ Example Code:
     :member-order: bysource
     :show-inheritance:
 
+
+.. autoexception:: CommandLineLoginFlowEOFError
+
 Local Server
 ------------
 
--- a/src/globus_sdk/login_flows/__init__.py
+++ b/src/globus_sdk/login_flows/__init__.py
@@ -1,4 +1,7 @@
-from .command_line_login_flow_manager import CommandLineLoginFlowManager
+from .command_line_login_flow_manager import (
+    CommandLineLoginFlowEOFError,
+    CommandLineLoginFlowManager,
+)
 from .local_server_login_flow_manager import (
     LocalServerEnvironmentalLoginError,
     LocalServerLoginError,
@@ -8,6 +11,7 @@ from .login_flow_manager import LoginFlowManager
 
 __all__ = (
     "CommandLineLoginFlowManager",
+    "CommandLineLoginFlowEOFError",
     "LocalServerLoginError",
     "LocalServerEnvironmentalLoginError",
     "LocalServerLoginFlowManager",
--- a/tests/functional/login_flows/test_login_flow_manager.py
+++ b/tests/functional/login_flows/test_login_flow_manager.py
@@ -1,11 +1,16 @@
 from unittest.mock import Mock, patch
 
+import pytest
+
 from globus_sdk import ConfidentialAppAuthClient, NativeAppAuthClient
 from globus_sdk._testing import load_response
 from globus_sdk.gare import GlobusAuthorizationParameters
 from globus_sdk.login_flows import (
     CommandLineLoginFlowManager,
     LocalServerLoginFlowManager,
+)
+from globus_sdk.login_flows.command_line_login_flow_manager import (
+    CommandLineLoginFlowEOFError,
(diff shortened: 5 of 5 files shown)


src/globus_sdk/login_flows/command_line_login_flow_manager.py
```diff
@@ -128,4 +137,16 @@ def prompt_for_code(self) -> str:
         :returns: The authorization code entered by the user.
         """
         code_prompt = "Enter the resulting Authorization Code here: "
-        return input(code_prompt).strip()
+        with self.handle_input_errors():
```
sirosen, line 140: Should this handler be up in `run_login_flow`, rather than here?

Here's the use-case that I'm thinking about:
```python
import contextlib

import click
from globus_sdk.experimental.login_flow_manager import CommandLineLoginFlowManager


class ClickLoginFlowManager(CommandLineLoginFlowManager):
    def print_authorize_url(self, authorize_url: str) -> None:
        click.echo(click.style("Login here for a code:", fg="yellow"))
        click.echo(authorize_url)

    def prompt_for_code(self) -> str:
        return click.prompt("Enter the code here:")

    @contextlib.contextmanager
    def handle_input_errors(self) -> t.Iterator[None]:
        try:
            yield
        except click.Abort as e
            raise CommandLineLoginFlowEOFError("User aborted when prompted for input.") from e
```

I don't have a clear-cut answer on this, but I think it bears consideration because it fundamentally asks the question:
What is the contract declared by `handle_input_errors`?

src/globus_sdk/login_flows/command_line_login_flow_manager.py
```diff
@@ -2,8 +2,10 @@
 
 import textwrap
 import typing as t
+from contextlib import contextmanager
```
sirosen, line 5: Ultra-minor style note, but I'll include it since I'm leaving comments already:

I prefer to import stdlib modules by name, not using `from ... import ...` syntax. This results in the fully qualified name being seen at the usage sites, which I think is beneficial, given how often we're reading code without the benefits of fancy IDE tools (e.g. if you're a crusty old man like me, and you don't use a fancy IDE).

So my style is generally to `import contextlib` and then use `@contextlib.contextmanager` where it's needed.

There are various counterarguments to this style, but I've personally never been convinced. I think `contextlib.contextmanager`, `textwrap.dedent`, `datetime.datetime`, and `socket.error`[^1] are less ambiguous than `contextmanager`, `dedent`, `datetime`, and `error`.

[^1]: `socket.error` is 100% me cheating to make a point. That's an alias for `OSError`. :grin: 

---

You'll see both styles used in the SDK, so you haven't violated a style rule. This is a _personal_ preference, and one that I haven't felt any need to force on my teammates or to battle-it-out until we reach consensus.

src/globus_sdk/login_flows/command_line_login_flow_manager.py
```diff
@@ -128,4 +137,16 @@ def prompt_for_code(self) -> str:
         :returns: The authorization code entered by the user.
         """
         code_prompt = "Enter the resulting Authorization Code here: "
-        return input(code_prompt).strip()
+        with self.handle_input_errors():
+            return input(code_prompt).strip()
+
+    @contextmanager
+    def handle_input_errors(self) -> t.Iterator[None]:
```
sirosen, line 144: Two notes:

- Because this is a public interface, it should have a docstring which tells users what it does
- Should it be named `handle_input_errors` or something specific to the prompt step, i.e. `handle_prompt_errors`?

Under that (renaming) interpretation, combined with the potential rescoping I mentioned above, we'd get something like...

```python
@contextlib.contextmanager
def handle_prompt_errors(self) -> t.Iterator[None]:
    """
    A contextmanager which wraps the ``prompt_for_input`` step when a login flow is run, and may
    catch and reshape errors emitted by that method.

    The default implementation converts ``EOFError``\s to ``CommandLineLoginFlowEOFError``\s.
    """
```

sirosen (commented):
I actually don't recall the original use-case which we discussed that well.

I'm perfectly happy for us to wrap EOFErrors like this if it makes external consumption easier -- it makes the error specific to the SDK, so it has at least some utility.

However, I want to make sure we're clear about what the interface we're defining is, so I've left some questions with a soft suggestion for some rearrangement.

src/globus_sdk/login_flows/command_line_login_flow_manager.py
```diff
@@ -13,6 +15,13 @@
     from globus_sdk.globus_app import GlobusAppConfig
 
 
+class CommandLineLoginFlowEOFError(GlobusError):
```
derek-globus, line 18:
```suggestion
class CommandLineLoginFlowEOFError(GlobusError, EOFError):
```

Best to keep this as catchable with an `except EOFError` even if we want custom typing/messaging.

sirosen: Just checked to confirm, and the inheritance is
```
BaseException
 └── Exception
      └── EOFError
```
which is key to my agreement here. I don't think we're contaminating our error class here with an unrelated inheritance structure if we do this, so IMO it's a good idea.

src/globus_sdk/login_flows/command_line_login_flow_manager.py
derek-globus: I'd request we don't put this on someone external to the team.

As you mention it's not an agreed upon style guide within the SDK package itself, better to hold off on pushing our personal preferences until we can come to a consensus on them as maintainer preferences.

sirosen: Agreed; this wasn't even meant as a request for change. I wanted to share my perspective, note that there's no enforced style in the SDK, and leave it at that.

src/globus_sdk/login_flows/command_line_login_flow_manager.py
chris-janidlo: In the original discussion, we landed on keeping `handle_input_errors` narrowly scoped to exactly that - handling errors that can happen with `input` - and calling it inside `prompt_for_code`. That was so that we could avoid trying to predict what kinds of errors might happen in `prompt_for_code` overrides, and instead let consumers handle errors that they need to handle as they come up. (And of course we would still provide `handle_input_errors` as a convenience method if anyone wants to use it.)

I also believe @derek-globus was specifically thinking about an SMTP-based implementation that would be impacted by a design like this (but I might be mixing that up with another topic; correct me if I'm wrong).

sirosen: I think the SMTP example was intentionally pushing the envelope on what might be supported/valid, not a known, real use-case.

If the method here is so narrowly scoped, I guess I'll ask: should it be public or private?

If it's only used in this one place, I would tend to rename it to `_handle_eof_errors` or even just inline it here. If it's meant for broader use, and to support subclassing+overriding, then I find it a little odd to have it only used inside of the default implementation of `prompt_for_code`.

chris-janidlo: I'm on board with an underscore prefix. Should I still add a doc string in that case?

sirosen: I just grabbed Derek for a quick sync -- and confirmed that we agree.

The simplest change is to make it private with the underscore. Docstring is optional in that case. This is such a small function, I don't think it's necessary.

---

If we kept it public, we'd probably move it to run outside of `prompt_for_code`. But with it private, the click example I gave becomes:

```python
import contextlib

import click
from globus_sdk.experimental.login_flow_manager import CommandLineLoginFlowManager


class ClickLoginFlowManager(CommandLineLoginFlowManager):
    def print_authorize_url(self, authorize_url: str) -> None:
        click.echo(click.style("Login here for a code:", fg="yellow"))
        click.echo(authorize_url)

    def prompt_for_code(self) -> str:
        try:
            return click.prompt("Enter the code here:")
        except click.Abort as e:
            raise CommandLineLoginFlowEOFError("User aborted when prompted for input.") from e
```

so I think we're on good footing.

derek-globus (approved):
This looks great, thanks for submitting a contribution Chris!

sirosen (approved):

PR #1094, opened 2024-10-25
Add ComputeClient methods to submit & get tasks

rjmello:
- Added the `.get_task()`, `.get_task_batch()`, `.get_task_group()`, and `.submit()` methods to the `ComputeClientV2` class.
- Added the `.submit()` method to the `ComputeClientV3` class.

----
📚 Documentation preview 📚: https://globus-sdk-python--1094.org.readthedocs.build/en/1094/

### DIFF
--- a/src/globus_sdk/services/compute/client.py
+++ b/src/globus_sdk/services/compute/client.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 import typing as t
 
-from globus_sdk import GlobusHTTPResponse, client
+from globus_sdk import GlobusHTTPResponse, client, utils
 from globus_sdk._types import UUIDLike
 from globus_sdk.scopes import ComputeScopes, Scope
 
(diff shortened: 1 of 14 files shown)


src/globus_sdk/services/compute/client.py
```diff
@@ -71,3 +71,66 @@ def delete_function(self, function_id: UUIDLike) -> GlobusHTTPResponse:
...
+
+            .. tab-item:: API Info
+
+                .. extdoclink:: Get Task
+                    :service: compute
+                    :ref: Tasks/operation/get_task_status_and_result_v2_tasks__task_uuid__get
+        """  # noqa: E501
+        return self.get(f"/v2/tasks/{task_id}")
+
+    def get_task_batch(self, task_ids: list[UUIDLike]) -> GlobusHTTPResponse:
```
sirosen, line 90: `list[T]` is problematic in public interfaces because it's invariant, meaning that this won't accept values of type `list[str]` or `list[UUID]`.
Typically, we accept `Sequence[T]` or `Iterable[T]` in such cases.

A good example would be here in the Auth client:
https://github.com/globus/globus-sdk-python/blob/b571dd28680eff8a964f4ee9e02e82296bba6658/src/globus_sdk/services/auth/client/service_client.py#L577

One thing you need to watch out for is that `str` is a `Sequence[str]`, so if we update the type to `t.Iterable[UUIDLike]`, you'll find that `client.get_task_batch(self, "foo")` is accepted by the type checker.

To deal with this, we have several cases which accept singular values in some APIs which require lists, and then explicitly wrap the content into a list. The whole situation represents a slight impedance mismatch between API contracts and Python typing constructs, so we need to flex things a little bit in one way or another to get clean types and runtime-safe code. For the example method from above, the handling code is here:
https://github.com/globus/globus-sdk-python/blob/b571dd28680eff8a964f4ee9e02e82296bba6658/src/globus_sdk/services/auth/client/service_client.py#L628

I recommend the same approach here, so that the signature becomes
```python
    def get_task_batch(self, task_ids: UUIDLike | t.Iterable[UUIDLike]) -> GlobusHTTPResponse:
```
and we use the "safe iter" helper from the `utils` module to "fix it in post", as it were.

sirosen: Oh, also, if you spot any existing SDK code using `list[T]` where `T` is a union type, in a public interface, that's probably a typing bug. If you can point those out to us, I'll work on fixes!

rjmello: Good call; I just pushed some updates.

sirosen (approved):
This looks good to me, but I want to confirm with the team that `get_task_batch` looks good to us. Since we don't yet have consistency about always passing a payload object vs passing unpacked fields, I just want to confirm that we're comfortable adding more APIs of both styles (I am, as I think consistency of this sort is only valuable if we can get it library-wide, which will require a major effort with some clear direction).

src/globus_sdk/_testing/data/compute/submit.py
```diff
@@ -0,0 +1,41 @@
...
+}
+
+RESPONSES = ResponseSet(
+    metadata={
+        "endpoint_id": ENDPOINT_ID,
+        "function_id": FUNCTION_ID,
+        "task_id": TASK_ID,
+        "task_id_2": TASK_ID_2,
+        "task_args": TASK_ARGS,
+        "task_args_2": TASK_ARGS_2,
```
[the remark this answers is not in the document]
rjmello: That's correct. I'll move them directly to the test to clarify things.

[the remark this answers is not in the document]
sirosen: I'll wait for this and then intend to re-approve and merge.

I think it's slightly easier to read if these are just baked into the relevant test, since they aren't used for request matching or anything.

tests/functional/services/compute/v2/test_get_task_batch.py
```diff
@@ -0,0 +1,33 @@
...
+
+import pytest
+
+import globus_sdk
+from globus_sdk._testing import load_response
+
+
+@pytest.mark.parametrize(
+    "task_id_style", ("string", "list", "set", "uuid", "uuid_list")
+)
```
kurtmckee, line 11: In place of strings and a conditional with an `else: raise` block, I recommend using transformation lambdas directly. Below, I've added IDs so that if errors ever occur, pytest will clearly show which transformer failed:

```suggestion
@pytest.mark.parametrize(
    "transform",
    (
        pytest.param(lambda x: x, id="string"),
        pytest.param(lambda x: [x], id="list"),
        pytest.param(lambda x: {x}, id="set"),
        pytest.param(lambda x: uuid.UUID(x), id="uuid"),
        pytest.param(lambda x: [uuid.UUID(x)], id="uuid_list"),
    ),
)
```

Then, in the test itself, the conditional just collapses to

```python
task_ids = transform(task_ids)
```

kurtmckee (approved):

tests/functional/services/compute/v2/test_get_task_batch.py
rjmello: I like it!

kurtmckee (approved):

src/globus_sdk/services/compute/client.py
```diff
@@ -71,3 +71,69 @@ def delete_function(self, function_id: UUIDLike) -> GlobusHTTPResponse:
...
+            .. tab-item:: API Info
+
+                .. extdoclink:: Get Task Batch
+                    :service: compute
+                    :ref: operation/get_batch_status_v2_batch_status_post
+        """
+        task_ids = list(utils.safe_strseq_iter(task_ids))
+        return self.post("/v2/tasks/batch", data={"task_ids": task_ids})
+
+    def get_task_group(self, task_group_id: UUIDLike) -> GlobusHTTPResponse:
```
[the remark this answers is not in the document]
rjmello: We may be interested to do this at some point, but I'll leave it as-is for now.

PR #1101, opened 2024-11-05
Update RTD config to use 'uv' for faster builds

sirosen:
This PR is an initial exploration of ways in which we can safely drop `uv` into our existing process without any requisite changes to the local dev experience.
RTD is chosen because
- it is simple
- it is well isolated from the rest of our work
- we should be able to measure build times (impact) trivially

---

Using `build.commands`, we completely override the build process for RTD. Per their documentation (linked in a comment), this is the way to convert a doc build to use `uv`.

The install is done using the `uv pip` interface, which has a few advantages:
- it can be treated as a plug-in replacement for `pip ...`
- it supports `-r` options (so we don't need to change how we're locking dependencies here)
- it ensures that we don't see any surprising new semantics from the uv lockfile (which could be hard to troubleshoot locally)

Because we're now using `build.commands`, there's no point in keeping `jobs.pre_build` as separate config. Simply inline the relevant command into our overridden process.

----
📚 Documentation preview 📚: https://globus-sdk-python--1101.org.readthedocs.build/en/1101/

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -11,9 +11,9 @@ build:
   # use 'uv pip install' to apply build customization in a pip-compatible way
   # see: https://docs.readthedocs.io/en/stable/build-customization.html#install-dependencies-with-uv
   commands:
-    - bash ./scripts/rtd-pre-sphinx-build.sh
     - asdf plugin add uv
     - asdf install uv latest
     - asdf global uv latest
     - uv pip install . -r "requirements/py3.11/docs.txt"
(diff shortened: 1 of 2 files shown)


sirosen:
I think this PR shows an interesting historical record but is not worth pursuing.

One RTD build time before this PR: 1m19s
One RTD build time at the HEAD of this PR: 1m4s

You can see how much I had to munge things to get it to work. I think `uv` is not a big win for us here.

PR #1109, opened 2024-11-12
add support for filter_entity_type on transfer client endpoint search

pjhinton-globus:
Added a new keyword argument to TransferClient.endpoint_search() that supports filtering by the collection entity type. Updated filter_non_functional argument to indicate its mutual exclusivity with filter_entity_type.

Once this has passed through review, a separate PR will be opened for the CLI that adds a `--filter-entity-type` option to `globus endpoint search`.

refs sc-37389

----
📚 Documentation preview 📚: https://globus-sdk-python--1109.org.readthedocs.build/en/1109/


src/globus_sdk/services/transfer/client.py
```diff
@@ -392,6 +392,7 @@ def endpoint_search(
         filter_owner_id: str | None = None,
         filter_host_endpoint: UUIDLike | None = None,
         filter_non_functional: bool | None = None,
+        filter_entity_type: str | None = None,
```
sirosen, line 395: I would like us to annotate this with the `Literal` enumeration of valid strings.
I'll explain why below; first, the suggestion:
```suggestion
        filter_entity_type: Literal[
            "GCP_mapped_collection",
            "GCP_guest_collection",
            "GCSv5_endpoint",
            "GCSv5_mapped_collection",
            "GCSv5_guest_collection",
            "GCSv4_host",
            "GCSv4_share",
        ] | None = None,
```

There are a few reasons I have for preferring the `Literal` here.

With a literal annotation, at runtime the value is just a `str`. That means that if the service adds a new valid value which isn't reflected in the client, you can just pass `filter_entity_type="GCSv6_NewHotness",  # type: ignore` and it will work. In summary: "annotating as a `Literal` doesn't change the runtime type at all".

But `Literal` also provides
- documentation as rendered in our sphinx docs
- documentation as loaded/used by code completion and other analyzers (e.g., autocomplete)
- value checking/linting via type checkers

src/globus_sdk/services/transfer/client.py
```diff
@@ -411,7 +412,10 @@ def endpoint_search(
             endpoint. May cause BadRequest or PermissionDenied errors if the endpoint ID
             given is not valid for this operation.
         :param filter_non_functional: Limit search to endpoints which have the
-            'non_functional' flag set to True or False.
+            'non_functional' flag set to True or False. Mutually exclusive with
+            ``filter_entity_type``.
+        :param filter_entity_type: Limit search to endpoints or collections by their
+            entity type. Mutually exclusive with ``filter_non_functional``.
```
sirosen, line 418: I have a question about this, which I realized when suggesting the Literal annotation above and don't see clarified here or in the Transfer docs. Are users limited to passing exactly one entity type here, or can they make any of these requests to combine them?
```
/endpoint_search?filter_entity_type=GCSv4_host/GCSv5_endpoint
/endpoint_search?filter_entity_type=GCSv4_host,GCSv5_endpoint
/endpoint_search?filter_entity_type=GCSv4_host&filter_entity_type=GCSv5_endpoint
```

If any of these are possible, I think it's best to handle an iterable of those `Literal` values, allowing for `filter_entity_type="GCSv5_endpoint"` as well as `filter_entity_type=["GCSv5_endpoint"]` and `filter_entity_type=("GCSv5_endpoint", "GCSv4_host")`.
It would make the annotations quite tricky but I'm happy to help if that reflects reality.

If these usages are not possible I count that as a reason in favor of the `Literal` annotation -- it clarifies that you can only pass one.

Either way, I think we should make a small tweak to clarify the Transfer API docs. I'll make a suggestion over there once I have clarity on what is/is not supported.

sirosen (changes requested):

src/globus_sdk/services/transfer/client.py
pjhinton-globus: @sirosen asked:

> I have a question about this, which I realized when suggesting the Literal annotation above and don't see clarified here or in the Transfer docs. Are users limited to passing exactly one entity type here, or can they make any of these requests to combine them?
> 
> ```
> /endpoint_search?filter_entity_type=GCSv4_host/GCSv5_endpoint
> /endpoint_search?filter_entity_type=GCSv4_host,GCSv5_endpoint
> /endpoint_search?filter_entity_type=GCSv4_host&filter_entity_type=GCSv5_endpoint
> ```
> 
> If any of these are possible, I think it's best to handle an iterable of those `Literal` values, allowing for `filter_entity_type="GCSv5_endpoint"` as well as `filter_entity_type=["GCSv5_endpoint"]` and `filter_entity_type=("GCSv5_endpoint", "GCSv4_host")`. It would make the annotations quite tricky but I'm happy to help if that reflects reality.
> 
> If these usages are not possible I count that as a reason in favor of the `Literal` annotation -- it clarifies that you can only pass one.
> 
> Either way, I think we should make a small tweak to clarify the Transfer API docs. I'll make a suggestion over there once I have clarity on what is/is not supported.

The value of `filter_entity_type`, if specified, must be one of the five recognized entity types. Combinations of entity types are not supported.

The Transfer API operation is a thin wrapper around a lower level function.

https://github.com/globusonline/koa/blob/ede225ab1e22b7661d242b9b04437f61f314becd/tapi/globus_org_rest/view/endpoint_search.py#L30-L41

That lower level function just checks for whether it is specified and in a set list of values.

https://github.com/globusonline/koa/blob/ede225ab1e22b7661d242b9b04437f61f314becd/pykoa/api/endpoint_search.py#L124-L132

changelog.d/20241112_220331_pjhinton_sc_37389_trasfer_client_collection_search_filter_entity_type.rst
```diff
@@ -0,0 +1,53 @@
...
+.. GitHub Actions will automatically replace it when the PR is merged.
+..
+.. Python Support
+.. ~~~~~~~~~~~~~~
+..
+.. - A bullet item for the Python Support category. (:pr:`NUMBER`)
+..
+.. Added
+.. ~~~~~
+..
```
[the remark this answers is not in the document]
pjhinton-globus: So noted! This and other scriv cleanup recommendations implemented in commit 55cad902d0ddd8b929cd05425be14ac33e57cd66./

src/globus_sdk/services/transfer/client.py
pjhinton-globus: Implemented in commit 55cad902d0ddd8b929cd05425be14ac33e57cd66.

changelog.d/20241112_220331_pjhinton_sc_37389_trasfer_client_collection_search_filter_entity_type.rst
```diff
@@ -0,0 +1,4 @@
+Added
+~~~~~
+
+- Add `filter_entity_type` keyword argument on `TransferClient.endpoint_search()`. (:pr:`NUMBER`)
```
sirosen, line 4: Sorry for not catching this on my first review pass. RST requires double-backticks for inline-code, and reserves single backticks for some of its special syntaxes (like sphinx roles).

```suggestion
- Add ``filter_entity_type`` keyword argument on ``TransferClient.endpoint_search()``. (:pr:`NUMBER`)
```

sirosen (changes requested):
Looks great; thanks for making those adjustments!

I spotted one final thing before we can approve+merge, and then we should be good to go.

sirosen (approved):

changelog.d/20241112_220331_pjhinton_sc_37389_trasfer_client_collection_search_filter_entity_type.rst
pjhinton-globus: @sirosen wrote:

> Sorry for not catching this on my first review pass. RST requires double-backticks for inline-code, and reserves single backticks for some of its special syntaxes (like sphinx roles).

Thanks for your patience as I cleanup up beginner errors! Addressed in commit f8052c9440caf22b2b2bfd3278809b76c10f2578.

PR #1111, opened 2024-11-21
Remove identity ID consistency validation from ClientApp

derek-globus:
## What?
* Remove identity id consistency validation from ClientApps (UserApps are unaffected).

## Why?
1. ClientApps don't have the same token refresh erganomics as UserApps.
   * If an access token is expired in a UserApp, it will raise an error which can be by the app and initiate a login flow.
   * If an access token is expired in a ClientApp, it will still return a ClientCredentialAuthorizer expecting that the access token will be refreshed by that authorizer down the line.
1. ClientCredentialAuthorizers exclusively request one token at a time.

Because of these two facts, a refresh initiated by a ClientCredentialAuthorizer will only ever contain identity info if it's refreshing specifically an globus auth token because that identity info only comes if you specify the special auth-managed "openid" scope.

Because this is a pretty widespread bug (any usage of a ClientApp), I advocate that we simply remove identity id consistency validation from ClientApps (thank you @rjmello for this suggestion) while planning to move off of `ClientCredntialAuthorizer` in the future.

## Testing
1. Added a unit test that recreates a scenario users reported. Verified it failed before the fix & succeeded after it.
2. Simulated a client app token expiry scenario locally & observed that the groups token was successfully refreshed.

### DIFF
--- a/changelog.d/20241121_133309_derek_client_app_expired_token_bug.rst
+++ b/changelog.d/20241121_133309_derek_client_app_expired_token_bug.rst
@@ -2,7 +2,7 @@
 Changed
 ~~~~~~~
 
--   Removed identity id validation from ``ClientApps``. (:pr:`NUMBER`)
+-   Removed identity ID consistency validation from ``ClientApp``. (:pr:`NUMBER`)
 
 Fixed
 ~~~~~
--- a/changelog.d/20241121_133309_derek_client_app_expired_token_bug.rst
+++ b/changelog.d/20241121_133309_derek_client_app_expired_token_bug.rst
@@ -1,0 +1,10 @@
+
+Changed
+~~~~~~~
+
+-   Removed identity id validation from ``ClientApps``. (:pr:`NUMBER`)
+
+Fixed
+~~~~~
+
+-   Fixed a bug that would cause ``ClientApp`` token refreshes to fail. (:pr:`NUMBER`)
--- a/src/globus_sdk/globus_app/app.py
+++ b/src/globus_sdk/globus_app/app.py
@@ -192,7 +192,7 @@ class GlobusApp(metaclass=abc.ABCMeta):
         # our initial scope requirements
         scope_validator = ScopeRequirementsValidator(scope_requirements, consent_client)
 
-        # use validators to enforce invariants about scopes and identity ID
+        # use validators to enforce invariants about scopes
         validating_token_storage.validators.append(scope_validator)
 
         return validating_token_storage
--- a/src/globus_sdk/globus_app/app.py
+++ b/src/globus_sdk/globus_app/app.py
@@ -14,7 +14,6 @@ from globus_sdk.tokenstorage import (
     ScopeRequirementsValidator,
     TokenStorage,
     TokenValidationError,
-    UnchangingIdentityIDValidator,
     ValidatingTokenStorage,
 )
 
@@ -92,7 +91,7 @@ class GlobusApp(metaclass=abc.ABCMeta):
 
         # create the requisite token storage for the app, with validation based on
         # the provided parameters
-        self.token_storage = _build_default_validating_token_storage(
+        self.token_storage = self._initialize_validating_token_storage(
             token_storage=self._token_storage,
             consent_client=consent_client,
             scope_requirements=self._scope_requirements,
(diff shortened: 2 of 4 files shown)


changelog.d/20241121_133309_derek_client_app_expired_token_bug.rst
```diff
@@ -0,0 +1,10 @@
+
+Changed
+~~~~~~~
+
+-   Removed identity id validation from ``ClientApps``. (:pr:`NUMBER`)
```
kurtmckee, line 5:
```suggestion
-   Removed identity ID validation from ``ClientApp``. (:pr:`NUMBER`)
```

src/globus_sdk/globus_app/app.py
```diff
@@ -178,6 +177,26 @@ def _initialize_login_client(
...
         """
 
+    def _initialize_validating_token_storage(
+        self,
+        token_storage: TokenStorage,
+        consent_client: AuthClient,
+        scope_requirements: t.Mapping[str, t.Sequence[Scope]],
+    ) -> ValidatingTokenStorage:
+        """
+        Initializes the validating token storage for the app.
```
kurtmckee, line 187:
```suggestion
        Initialize the validating token storage for the app.
```

src/globus_sdk/globus_app/app.py
```diff
@@ -178,6 +177,26 @@ def _initialize_login_client(
         Initializes and returns an AuthLoginClient to be used in authorization requests.
         """
 
+    def _initialize_validating_token_storage(
+        self,
```
kurtmckee, line 181: It looks like this can be a static method, but I don't know how that will affect subclasses. Is adding `@staticmethod` worth considering here?

kurtmckee (changes requested):

src/globus_sdk/globus_app/app.py
derek-globus: "Initializes" matches the other docstrings for the other `_initialize_*` methods ([here](https://github.com/globus/globus-sdk-python/blob/main/src/globus_sdk/globus_app/app.py#L178) and [here](https://github.com/globus/globus-sdk-python/blob/main/src/globus_sdk/globus_app/app.py#L226)). I'd like to stay consistent with those.

src/globus_sdk/globus_app/app.py
derek-globus: I don't think so. This structure is inline with the rest of the class & gives us the ability to reference the instance being instantiated if we need it in the future.

What benefit would we see from making this method static?

tests/unit/globus_app/test_globus_app.py
```diff
@@ -455,6 +462,38 @@ def test_user_app_expired_authorizer_triggers_login():
...
+    config = GlobusAppConfig(token_storage=memory_storage)
+    client_app = ClientApp("test-app", **client_creds, config=config)
+
+    transfer = TransferClient(app=client_app, app_scopes=[Scope(meta["scope"])])
+
+    load_response(transfer.task_list)
+    transfer.task_list()
+
+    access_token = memory_storage.get_token_data(meta["resource_server"]).access_token
+    assert access_token == meta["access_token"]
```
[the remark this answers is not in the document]
derek-globus: Sure, I can do that.

src/globus_sdk/globus_app/app.py
kurtmckee: Nothing technical that I'm aware of, I'm just attuned to it because PyCharm complains by default.

src/globus_sdk/globus_app/app.py
kurtmckee: Ah, I see. In general, other functions' docstrings appear to use active verbs, and I recommend using active verbs in all places. I advocate to break with the other initializer functions and introduce this new function with active verb phrasing.

It's not critical to me, but I recommend always using active verbs in all places.

derek-globus: I don't agree in this case (not necessarily about the overall guidance; rather about making this class's docstrings less internally consistent).

Since this is a pretty small decision (a stylistic debate over a single word in an internal docstring), you okay if I proceed as is & resolve this thread?

kurtmckee (approved):

ada-globus (approved):
🙌

PR #1113, opened 2024-11-22
Add ComputeClientV3 methods to manage endpoints 

rjmello:
Added `.register_endpoint()`, `.update_endpoint()`, `.lock_endpoint()`, and `.get_endpoint_allowlist()` methods to the `ComputeClientV3` class.

----
📚 Documentation preview 📚: https://globus-sdk-python--1113.org.readthedocs.build/en/1113/


tests/functional/services/compute/v3/test_register_endpoint_v3.py
```diff
@@ -0,0 +1,39 @@
...
+        "display_name": "My Endpoint",
+        "version": "2.31.0",
+        "multi_user": False,
+        "allowed_functions": [str(uuid.uuid1())],
+        "authentication_policy": str(uuid.uuid1()),
+        "subscription_uuid": str(uuid.uuid1()),
+        "metadata": {
+            "endpoint_config": ENDPOINT_CONFIG.strip(),
+            "user_config_template": "",
+            "user_config_schema": {},
```
khk-globus, line 26: Given that this is testing the happy path, perhaps we don't include the fields not relevant to this non-MEP registration request?  Certainly these two fields are not necessary (right?  I haven't double-checked just now, maybe I'm missing something).  `multi_user` is similarly not required here.

In terms of the need for this test's purpose, I'm not sure about some of the other fields either.  (Have to run at this time; let me know if I'm missing the thought-process.)

khk-globus (commented):
Also thinking at this time that you may want to add yourself (and anyone else appropriate from our team) to CODEOWNERS.  Maybe not for this PR&nbsp;&mdash;&nbsp;or maybe for the associated paths in this PR?

src/globus_sdk/services/compute/client.py
```diff
@@ -236,6 +236,69 @@ class ComputeClientV3(client.BaseClient):
...
+    def lock_endpoint(self, endpoint_id: UUIDLike) -> GlobusHTTPResponse:
+        """Temporarily block registration requests for the endpoint.
+
+        :param endpoint_id: The ID of the Globus Compute endpoint.
+
+        .. tab-set::
+
+            .. tab-item:: API Info
+
+                .. extdoclink:: Delete Endpoint
```
kurtmckee, line 281: This doesn't look right; it's in the `lock_endpoint()` method.

tests/functional/services/compute/v2/test_lock_endpoint_v2.py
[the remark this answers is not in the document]
kurtmckee: IIRC this is needed when `__init__.py` files aren't present.

I'm not suggesting adding `__init__.py` files! ...plz don't do that. :grinning:

kurtmckee (changes requested):

tests/functional/services/compute/v2/test_lock_endpoint_v2.py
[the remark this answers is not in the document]
kurtmckee: IIRC this is needed when `__init__.py` files aren't present.

I'm not suggesting adding `__init__.py` files! ...plz don't do that. :grinning:

src/globus_sdk/services/compute/client.py
rjmello: Ah, good catch.

tests/functional/services/compute/v3/test_register_endpoint_v3.py
rjmello: I'm fine trimming the example payload down, but I also see benefit to showing all possible fields.

tests/functional/services/compute/v2/test_lock_endpoint_v2.py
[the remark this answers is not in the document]
rjmello: Thoughts on these alternative structures?
- Move the test files directly under `compute/`, drop the `_v*` suffix, then create multiple versioned test functions within each:
  ```python
  # tests/functional/services/compute/test_lock_endpoint.py
  
  def test_lock_endpoint_v2():
    ...
  
  def test_lock_endpoint_v3():
    ...
  ```
- Add a `_v*` suffix to every test file then move them directly under `compute/`.
- Add a `_v*` suffix to every test file and leave them as-is.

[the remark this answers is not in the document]
kurtmckee: I tested locally, and discovered that it's not necessary to add `__init__.py` files to every directory up the chain.

```
touch tests/functional/services/compute/v2/__init__.py
touch tests/functional/services/compute/v3/__init__.py
```

Then, remove the `_v*` suffix for all of the test files. pytest will be satisfied knowing that the files are imported with names like `v2.test_lock_endpoint` and `v3.test_lock_endpoint`.

This eliminates the filename suffixes duplicating the directory name without requiring lots of empty `__init__.py` files, and maintains the `/v2/` and `/v3/` directory separation that I think looks nice.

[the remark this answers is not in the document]
rjmello: Ah, very nice. 🍻

tests/functional/services/compute/v3/test_register_endpoint_v3.py
khk-globus: As mentioned outside of this thread, I don't see the value-add.  Who's the audience for this test data and slightly-bloated interaction?  To my understanding, this is test data ... for tests.

It's just a test, so I'm not overly chuffed about it, but I do observe that it doesn't match what we expect the real-world interaction to be, even if it does work.  If it's to flesh out edge cases, then we have more tests to write to do just that.  But if it's just to exercise the `globus_sdk` code and new function, then I'd gently encourage a more expected data set.

Regardless, this observation is not a blocker.

khk-globus (approved):
Approving, so as to not block the general consensus.

tests/functional/services/compute/v3/test_register_endpoint_v3.py
rjmello: From previous conversations, I understood these to partially serve as examples for usage (e.g., [test_collections.py](https://github.com/globus/globus-sdk-python/blob/8df9c5cac8845180ca03085f03bd985da6087b65/tests/functional/services/gcs/test_collections.py#L88-L113)). If that's wrong, then I agree with limiting the request payload to what is necessary for the test.

@kurtmckee, can you provide some clarification?

kurtmckee: I'll bet that it's been said that these test documents are examples for usage, but I wouldn't target the tests for that purpose; I would focus on excellent documentation for that purpose.

PR #1126, opened 2025-01-10
Improve `RegisteredResponse` init doc & signature

sirosen:
`RegisteredResponse.__init__` allowed any keyword arguments via `**kwargs` usage, but would later pass these to `responses.Response` via the `add()` or `replace()` methods, at which point there could be a `TypeError`.

This can make it hard to understand what features are being exposed during development. e.g., "Was the parameter name `status` or `status_code`?"

To resolve and provide an improved experience for development with `RegisteredResponse`, the init parameters are better documented and the signature is improved to explicitly enumerate supported args.

1. Add `:param ...:` docstring items
2. Explicitly enumerate parameters, removing `**kwargs: Any`
3. Document which parameters come from `responses.Response`, which come from `responses.BaseResponse`, and which ones are omitted (and why)
4. Update `service` and `method` to use Literals to show the values

In terms of lost functionality, technically the following three parameters are removed in this change:

    - auto_calculate_content_length
    - passthrough
    - match_querystring

However, given that `_testing` has a weaker backwards compatibility contract than the rest of the SDK and there are no known usages for these, this seems like a good change rather than a bad one. We can always add `auto_calculate_content_length` in the future. The other two would require stronger justification, since

    - match_querystring is deprecated in `responses`
    - passthrough doesn't seem applicable to any normal uses of `globus_sdk._testing`
    - we don't forbid anyone from using `responses` directly

---

I was motivated to tackle this while playing with some editor support for code completion and docstring inspection, and I found that `RegisteredResponse` didn't have the information I wanted.
In particular, I wanted to make sure I was using correct parameter names, and the loaded class defintion and docstring didn't provide this information.

----
📚 Documentation preview 📚: https://globus-sdk-python--1126.org.readthedocs.build/en/1126/

### DIFF
--- a/src/globus_sdk/_testing/models.py
+++ b/src/globus_sdk/_testing/models.py
@@ -76,8 +76,19 @@ class RegisteredResponse:
             ]
             | None
         ) = None,
-        # method derives from `responses.BaseResponse`
-        method: Literal["GET", "PUT", "POST", "PATCH", "HEAD", "DELETE"] = "GET",
+        # method will be passed through to `responses.Response`, so we
+        # support all of the values which it supports
+        method: Literal[
+            "GET",
+            "PUT",
+            "POST",
+            "PATCH",
+            "HEAD",
+            "DELETE",
+            "OPTIONS",
+            "CONNECT",
+            "TRACE",
+        ] = "GET",
         # these parameters derive from `response.Response`
         body: str | None = None,
         content_type: str | None = None,
--- a/src/globus_sdk/_testing/models.py
+++ b/src/globus_sdk/_testing/models.py
@@ -89,15 +89,14 @@ class RegisteredResponse:
             "CONNECT",
             "TRACE",
         ] = "GET",
-        # these parameters derive from `response.Response`
+        # these parameters are passed through to `response.Response` (or omitted)
         body: str | None = None,
         content_type: str | None = None,
         headers: dict[str, str] | None = None,
         json: None | list[t.Any] | dict[str, t.Any] = None,
         status: int = 200,
         stream: bool | None = None,
-        # match derives from `responses.BaseResponse`; it has a default of `()` there
-        match: t.Sequence[t.Callable[..., tuple[bool, str]]] = (),
+        match: t.Sequence[t.Callable[..., tuple[bool, str]]] | None = None,
         # metadata is globus-sdk specific
         metadata: dict[str, t.Any] | None = None,
         # the following are known parameters to `responses.Response` which
@@ -162,7 +161,6 @@ class RegisteredResponse:
             "status": self.status,
             "stream": self.stream,
             "match_querystring": None,
-            "match": self.match,
         }
         if self.json is not None:
             kwargs["json"] = self.json
@@ -170,6 +168,8 @@ class RegisteredResponse:
             kwargs["body"] = self.body
         if self.content_type is not None:
             kwargs["content_type"] = self.content_type
+        if self.match is not None:
+            kwargs["match"] = self.match
 
         if requests_mock is None:
             use_requests_mock: responses.RequestsMock | types.ModuleType = responses
--- a/src/globus_sdk/_testing/models.py
+++ b/src/globus_sdk/_testing/models.py
@@ -19,6 +19,25 @@ class RegisteredResponse:
     A mock response along with descriptive metadata to let a fixture "pass data
     forward" to the consuming test cases. (e.g. a ``GET Task`` fixture which
     shares the ``task_id`` it uses with consumers via ``.metadata["task_id"]``)
+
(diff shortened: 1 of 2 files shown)


src/globus_sdk/_testing/models.py
```diff
@@ -40,14 +59,44 @@ class RegisteredResponse:
     def __init__(
         self,
         *,
+        # path and service are glbous-sdk specific
+        # in `responses`, these are just `url`
         path: str,
-        service: str | None = None,
-        method: str = responses.GET,
+        service: (
+            Literal[
```
kurtmckee, line 66: Looks like we have a pyupgrade configuration issue: above, `Literal` is imported conditionally for Python 3.7 and below, which we no longer support.

This should become `t.Literal` in a subsequent PR.

src/globus_sdk/_testing/models.py
```diff
@@ -40,14 +59,44 @@ class RegisteredResponse:
...
+                "gcs",
+                "groups",
+                "timer",
+                "flows",
+                "compute",
+            ]
+            | None
+        ) = None,
+        # method derives from `responses.BaseResponse`
+        method: Literal["GET", "PUT", "POST", "PATCH", "HEAD", "DELETE"] = "GET",
```
kurtmckee, line 80: There are other methods, like `OPTIONS`. Would it makes sense to construct the literals from the `http.HTTPMethod` enum?

Also, the "derives from" phrasing isn't clear to me. Does that mean these literal values are the only ones supported by `responses.BaseResponse`, and should perfectly match it? Does that mean that `method` will be passed to a constructor that is originally defined in `responses.BaseResponse`?

Ultimately, these are passed as grouped keyword arguments to a single location, so it may be clearer to group these based on that final target, rather than noting the subclassing hierarchy.

src/globus_sdk/_testing/models.py
```diff
@@ -40,14 +59,44 @@ class RegisteredResponse:
...
+        content_type: str | None = None,
         headers: dict[str, str] | None = None,
-        metadata: dict[str, t.Any] | None = None,
         json: None | list[t.Any] | dict[str, t.Any] = None,
-        body: str | None = None,
-        **kwargs: t.Any,
+        status: int = 200,
+        stream: bool | None = None,
+        # match derives from `responses.BaseResponse`; it has a default of `()` there
+        match: t.Sequence[t.Callable[..., tuple[bool, str]]] = (),
```
kurtmckee, line 89: Because the implementation's default is noted here, should this be updated if it ever changes in `responses`? If not, then we're choosing a default for our own purposes, and the implementation's default doesn't matter.

src/globus_sdk/_testing/models.py
```diff
@@ -40,14 +59,44 @@ class RegisteredResponse:
...
+        # metadata is globus-sdk specific
+        metadata: dict[str, t.Any] | None = None,
+        # the following are known parameters to `responses.Response` which
+        # `RegisteredResponse` does not support:
+        #   - url: provided via (path, service)
```
kurtmckee, line 94: Perhaps:

```suggestion
        #   - url: calculated from (path, service)
```

src/globus_sdk/_testing/models.py
sirosen: Let me pull the exact strings from `http.HTTPMethod` in case I missed others other than `OPTIONS`.

Dynamically building the Literal at runtime is a no-go. Type checkers and other analyzers won't be able to read the data because they won't execute the code.
We can add a test-case which asserts that the two are equal, which is what I will do.

src/globus_sdk/_testing/models.py
sirosen: I think I'm going to change this, since the default in `responses` could change and we wouldn't track it with this behavior.
Instead, let's use `t.Sequence[...] | None = None` and then at the usage site (below) we'll only pass it if it has a non-`None` value.

src/globus_sdk/_testing/models.py
sirosen: I'm not certain that I got exactly what you were trying to suggest here.
I've just implemented a change which I think resolves this, but feel free to un-resolve if you think there's more to discuss.

src/globus_sdk/_testing/models.py
sirosen: I also did my best to clarify the comment wording. `BaseResponse` vs `Response` only really matters if you are looking at the _current_ `responses` source code and trying to understand where the arguments for `Response` are defined.

Since that kind of examination is a rare activity, and `responses` could change, it now all says that we pass values to `responses.Response`, which is more robust (as a statement) to changes in `responses`.

kurtmckee (approved):

PR #1127, opened 2025-01-14
Document auto-login disabling

derek-globus:
* Documentation didn't explain what happened if `GlobusAppConfig.token_validation_error_handler` was set to None (or explain that it could be set to None). Added that explanation.

----
📚 Documentation preview 📚: https://globus-sdk-python--1127.org.readthedocs.build/en/1127/

### DIFF
--- a/changelog.d/20250114_150601_derek_fix_app_config_doc_ivar.rst
+++ b/changelog.d/20250114_150601_derek_fix_app_config_doc_ivar.rst
@@ -1,0 +1,6 @@
+
+Documentation
+~~~~~~~~~~~~~
+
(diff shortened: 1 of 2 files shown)


sirosen (approved):
I think those infix type declarations in `ivar` blocks for Sphinx might not be needed, since sphinx can read annotations. But let me look at fixing that after this merges, so that we can have the doc content itself updated.

sirosen:
FWIW: I was wrong. autodoc_typehints doesn't support loading the annotations off of dataclass fields.
I'd like to get that working, but it probably requires a custom plugin. :(

PR #1131, opened 2025-01-18
Extract and re-inject base client sphinx param doc into client classes, for better sphinx doc

sirosen:
Add three helpers to `utils` in order to improve the runtime
docstrings for client classes.

1. `read_sphinx_params` -- a simple `:param:` parser
2. `inject_sphinx_params` -- a `__doc__` rewriter which takes parsed
        params and inserts them in place of a special marker comment
3. `inject_sphinx_params_of` -- a decorator version of these which
        chains together reading the docstring of one object and
        modifying the docstring of another

Then apply these to write

```python
@utils.inject_sphinx_params_of(client.BaseClient)
class FlowsClient(client.BaseClient):
    ...
```

and so forth. This is applied to all client classes except for the
Auth login clients, which do not inherit all of the parameters of
`BaseClient` -- in particular, we do not allow construction of a
login client with an `app`.

`read_sphinx_params` gets a functools cache, but the rest of the work
is still done dynamically, so there is a little bit of repetition.

This results in improved docstrings for sphinx and for runtime usages
like `help(globus_sdk.TransferClient)`. Static analyzers (e.g., IDEs)
will pick up on the original source string, which will show the marker
comment used to drive this behavior:

```
.. globus-sdk-inject-doc-params
```

---

before | after
---|---
image | image

----
📚 Documentation preview 📚: https://globus-sdk-python--1131.org.readthedocs.build/en/1131/


src/globus_sdk/services/search/client.py
```diff
@@ -14,10 +14,13 @@
 log = logging.getLogger(__name__)
 
 
+@utils.inject_sphinx_params_of(client.BaseClient)
 class SearchClient(client.BaseClient):
```
derek-globus, line 18: Is there a way to do this without the annotation by looking at the client class's mro instead of annotating?

src/globus_sdk/utils.py
```diff
@@ -276,3 +278,87 @@ def classproperty(func: t.Callable[[T], R]) -> _classproperty[T, R]:
...
+
+    All produced param strings are dedented.
+    """
+    docstring = textwrap.dedent(docstring)
+
+    result: list[str] = []
+    current: list[str] = []
+    for line in docstring.splitlines():
+        if not current:
+            if line.startswith(":param"):
```
derek-globus, line 305: Surely sphinx loads these as a part of its docstring parsing these. Do we have any mechanism to tap into that in a plugin instead of doing it ourselves?

sirosen: I don't think there's an easy way to do this via sphinx. I'm sure the info is in there, but complex sphinx extensions are pretty hard to write. 

I also like that this impacts `help()` in the REPL and maybe some IDEs as well. But I'm not overly attached to that. 

We could rewrite from a decorator to a special sphinx directive in the docstring. That kind of extension is on the easy side to write. But I'd stick with parsing the data ourselves -- I think it's actually a lot easier than getting that data from sphinx, partly because it introduces a requirement that we get the ordering of events correct. (Sphinx parse has to be done before we can access it.)

src/globus_sdk/services/search/client.py
sirosen: I'm pretty strongly against this being _only_ implicit, but I would be okay with an implicit default. 

Imagining this as a sphinx directive, I'd be okay with
```
.. sdk-autoparams:: BaseClient
.. sdk-autoparams
```
as equivalent. 

The reason I want explicit support is that I can imagine usages where this isn't inheritance based, but sibling methods share params. Especially for mapped/guest collections, which sometimes get slightly different but closely related APIs.

sirosen:
In response not only to feedback but some direct discussions, I've done a refactor to make this a sphinx directive. Here's the commit message, which also explains why I made some of the decisions I made:

[Convert copy-params to custom sphinx directive](https://github.com/globus/globus-sdk-python/pull/1131/commits/32a173316bbeacbba4505bcd58403f9dfaa097d5)

- Remove utility functions for adding parsed `:param:` doc to a runtime docstring
- Add a new `CopyParams` sphinx directive registered under the name `sdk-sphinx-copy-params`, and use this to implement param copy/injection functionality as a documentation-time customization.

There are some particulars about how this looks in the Sphinx/docutils context which drive the implementation to look a little different from discussed ideas:

1. The source for doc has to be explicit. `:inherit:` or similar ideas about picking up the current class and class hierarchy aren't easily implemented in a sphinx plugin. The context of directive execution isn't populated with easy values for inspection, and finding the "current object being documented" is much more work than it is worth. So each usage specifies `BaseClient` as the source. (required param)

2. We support added parameters as directive contents. Having a custom directive adjacent to a `:param:` list causes docutils errors which kill the entire build. Having blank lines between `:param:` lists causes multiple `PARAMETERS` blocks to get rendered. So we need adjacency but we can't have _literal_ adjacency. If we put params into the content area, we can re-emit them from the directive.

2a. Supporting added parameters before and/or after the copied params with "<YIELD>". Although not strictly necessary, I've tested to ensure we can add params before or after the generated/copied params by having a special line in the content of "<YIELD>". The first time the directive sees this line, it will stop emitting content, emit the copied params, and then resume.

src/globus_sdk/_sphinxext.py
```diff
@@ -335,6 +335,36 @@ def gen_rst(self):
...
+        if self.content:
+            content = iter(self.content)
+        else:
+            content = ()
+
+        for line in content:
+            if line.strip() == "<YIELD>":
+                break
+            yield line
+
```
kurtmckee, line 361: I don't see that this is getting used in this PR. Is it needed now or in the near-term?

sirosen: It is not being used right now. I could remove it but only if you feel strongly.

I wanted to make sure that it was possible to control where the static content is placed relative to the generated content. I've demonstrated that, so technically we could remove it with the knowledge that we can always reimplement something like this if it's needed.

As an alternative, we could define this bit of extension code to _only_ insert content where it finds a `<YIELD>` line (when there is content), so that we'd put `<YIELD>` into the usages with content and it would be used (trivially). That might be nice because it's more explicit about the ordering, rather than an "implicit append" rule.

sirosen: (Oh, and I'm not overly attached to `<YIELD>` as the syntax, if you have other ideas about that.)

kurtmckee: I'm not strictly tied to "using or losing" this code, but since it's not used it caught my eye.

kurtmckee (approved):

derek-globus (changes requested):
One minor naming update request.

src/globus_sdk/_sphinxext.py
```diff
@@ -335,6 +335,36 @@ def gen_rst(self):
...
+    has_content = True
+    required_arguments = 1
+    optional_arguments = 0
+
+    def gen_rst(self):
+        import globus_sdk
+
+        source_object_name: str = self.arguments[0]
+        path = source_object_name.split(".")
+        cur = globus_sdk
```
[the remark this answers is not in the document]
sirosen: `cur` or `current` is "the current target object for docstring lookup".
Since this is traversal code, `cur` seems to me like an appropriate shorthand.

The only meaningful annotation it can be given is `t.Any`, so I don't think hints will help.

Would this comment satisfy?
```python
# traverse the SDK namespace, starting at the top level module
cur = globus_sdk
```

src/globus_sdk/utils.py
sirosen: Yeah, per our out-of-band discussion I want to refactor the sphinx extension into a subpackage. So I agree that this should have a better home -- and that could lead to us being able to write some unit tests for our sphinx extension, which I would very much like.

src/globus_sdk/_sphinxext.py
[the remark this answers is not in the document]
sirosen: I added a comment similar to this one, and renamed `cur` to `source-object`.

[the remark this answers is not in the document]
derek-globus: That's great 👍 
Thanks

PR #1136, opened 2025-02-19
Implement JWTDecoder and IDTokenDecoder, use them to reimplement decoding and improve the behavior and customizability of GlobusApp

sirosen:
Update: The implementation described here was updated in https://github.com/globus/globus-sdk-python/pull/1136/commits/b7f00b92b87c1d4243a6539dff4bc2a391c06bb6 to remove the abstract `JWTDecoder` class, as we are not convinced that (1) we want to invest in expressing our types precisely for this niche case (2) we would want to express "JWT decoding" as a supported SDK feature vs only supporting "id_token decoding".

## Why

This changeset addresses two notable issues with `id_token` handling in GlobusApp.

1.  GlobusApp usage calls out to Globus Auth after each login flow to get OIDC config and JWKs.

    These are currently not cached between usages, even though this would be possible in theory.

2.  There is no way to customize the parameters used for decoding `id_token` inside of GlobusApp.

    For example, if a user wanted to adjust the JWT leeway or disable audience verification, they would not be able to do so today.

## What

Introduce `globus_sdk.JWTDecoder` and `globus_sdk.IDTokenDecoder` classes, which implement `id_token` decoding.

These new classes define how OIDC `id_token`s should be decoded. When `OAuthTokenResponse.decode_id_token` is called, it now internally instantiates a new `IDTokenDecoder` and uses it to perform the decode.

The default JWT leeway of 5 minutes is now declared as an instance var and init param, `IDTokenDecoer.jwt_leeway`. This allows users to subclass and override this value trivially.

`IDTokenDecoder` objects cache OpenID configuration data and JWKs after looking them up.
If a decoder is used multiple times, it will reuse the cached data.

Token storage constructs can now contain a `JWTDecoder` in their `id_token_decoder` property. The decoder is used preferentially when trying to read the `sub` field from an `id_token` to store.

`GlobusAppConfig` can now contain a `JWTDecoder`, which will be attached to the underlying token storage during app initialization. Or it may have a `JWTDecoderProvider` -- any implementer of a `for_globus_app` constructor -- which defaults to the `IDTokenDecoder` class.

`GlobusApp` initialization will now attach the provided `JWTDecoder` to the token storage which is used, constructing via the `for_globus_app` call if necessary.

### Explanation of the token storage interface

Because abstract base classes can't add new requirements without breaking 3rd party implementations, token storage cannot require a decoder. It therefore must be optional.
The decoder is attached to the underlying token storage as a new attribute, and the attribute is runtime-checked to see if it should be used for decoding.

The attribute is wrapped in a property simply to provide better documentation facilities (property getter docstring).

### Naming

The original PR proposed names `IDTokenDecoder` and `DefaultIDTokenDecoder`, but the interfaces were different.
After discussion, I decided that what we really wanted was to have `IDTokenDecoder` as the (only) implementation provided by the SDK, but allow extension and customizability. At the same time, the contract needed for the rest of the codebase has to be declared -- some object with a `decode()` method taking a string. It was unclear how to rename the original base class which declares the contract, what does `IDTokenDecoder` become if `DefaultIDTokenDecoder` becomes `IDTokenDecoder`?

I strongly considered making it a protocol, but then the number of pure interfaces being exposed to SDK users rises to an uncomfortable degree for the size of this feature. (One for the thing itself, and one for the provider type.)
Eventually, I went with `JWTDecoder`, which is the more generic _type of thing_ which an `id_token` is. Just as "an `id_token` is a JWT", so too "an `IDTokenDecoder` is a `JWTDecoder`".

## Example Usage

To customize `jwt_leeway`, any of the following examples works. All three are fully supported.

### Override via Subclass Init

```python
import globus_sdk

class RelaxedIDTokenDecoder(globus_sdk.IDTokenDecoder):
    # override the init-time default of 300s to make it 1800s
    def __init__(self, login_client, *, jwt_leeway=1800, jwt_options=None):
        super().__init__(login_client, jwt_leeway=jwt_leeway, jwt_options=jwt_options)

config = globus_sdk.GlobusAppConfig(id_token_decoder=RelaxedIDTokenDecoder)
app = globus_sdk.UserApp("myapp", config=config, client_id=...)
```

### Create a Custom Provider

```python
import globus_sdk

class RelaxedDecoderProvider:
    @classmethod
    def for_globus_app(cls, *, app_name, config, login_client):
        return globus_sdk.IDTokenDecoder(login_client, jwt_leeway=1800)

config = globus_sdk.GlobusAppConfig(id_token_decoder=RelaxedDecoderProvider)
app = globus_sdk.UserApp("myapp", config=config, client_id=...)
```

### Instantiate a Customized Decoder

```python
import globus_sdk

login_client = globus_sdk.NativeAppAuthClient(...)
decoder = globus_sdk.IDTokenDecoder(login_client, jwt_leeway=1800)

config = globus_sdk.GlobusAppConfig(id_token_decoder=decoder)
app = globus_sdk.UserApp("myapp", config=config, login_client=login_client)
```

----
📚 Documentation preview 📚: https://globus-sdk-python--1136.org.readthedocs.build/en/1136/


src/globus_sdk/services/auth/id_token_decoder.py
```diff
@@ -0,0 +1,173 @@
...
+    of use.
+
+    The main method of a decoder is ``decode()``, which accepts an ``id_token`` string
+    and parses it into a dict.
+
+    :param auth_client: The client which should be used to callout to Globus Auth as
+        needed. Any AuthClient or AuthLoginClient will work for this purpose.
+    """
+
+    def __init__(self, auth_client: SupportsJWKMethods) -> None:
```
derek-globus, line 113: There's a couple spots in here where code paths need to branch to account for the fact that auth_client may not actually be an auth client. `SupportsJWKMethods` protocol over `AuthLoginClient` here is (to my reading) the root of that complexity so I want to make sure we've got a good justification for it.

Is there a history or use case informing why we we would say "to decode an id token you need an auth login client-ish" instead of "to decode an id token you need an auth login client" (my understanding of the difference between these)?

src/globus_sdk/services/auth/id_token_decoder.py
```diff
@@ -0,0 +1,173 @@
...
+        self._jwk = jwk
+
+    @property
+    def default_audience(self) -> str | None:
+        """
+        The audience for JWT verification defaults to the client's client ID.
+        """
+        if not hasattr(self._auth_client, "client_id"):
+            return None
+        return self._auth_client.client_id
```
derek-globus, line 148: The protocol mandates the client_id attribute exist ([ref](https://github.com/globus/globus-sdk-python/blob/main/src/globus_sdk/services/auth/_common.py#L90)); it just might be None (unless I don't understand protocols which is very possible).

```suggestion
        return self._auth_client.client_id
```

src/globus_sdk/services/auth/response/oauth.py
```diff
@@ -191,7 +191,8 @@ def decode_id_token(
             step. If ``"leeway"`` is included, it will be passed as the ``leeway``
             parameter, and all other values are passed as ``options``.
         """
-        logger.info('Decoding ID Token "%s"', self["id_token"])
+        id_token = self["id_token"]
+        logger.info('Decoding ID Token "%s"', id_token)
```
derek-globus, line 195: This is tangential do your actual change but flags me as one of those times where the sdk is louder than is should be.
If out-of-scope, you can discard this without comment.

```suggestion
        logger.debug('Decoding ID Token "%s"', id_token)
```

src/globus_sdk/tokenstorage/v2/base.py
```diff
@@ -29,6 +29,19 @@ class TokenStorage(metaclass=abc.ABCMeta):
...
+    def id_token_decoder(self) -> globus_sdk.IDTokenDecoder | None:
+        """
+        An ID Token decoder to use when decoding ``id_token`` JWTs from Globus Auth.
+        By default, a new decoder is used each time decoding is performed.
+        """
+        return self._id_token_decoder
+
+    @id_token_decoder.setter
+    def id_token_decoder(self, value: globus_sdk.IDTokenDecoder) -> None:
+        self._id_token_decoder = value
```
derek-globus, line 44: Would a plain-jane attribute be simpler than a property?

```suggestion
        # An ID Token decoder to use when decoding ``id_token`` JWTs from Globus Auth.
        # By default, a new decoder is used each time decoding is performed.
        self.id_token_decoder: globus_sdk.IDTokenDecoder | None = None
```

src/globus_sdk/tokenstorage/v2/base.py
```diff
@@ -111,8 +124,11 @@ def _extract_identity_id(
         if isinstance(token_response, globus_sdk.OAuthDependentTokenResponse):
             return None
 
-        if token_response.get("id_token"):
-            decoded_id_token = token_response.decode_id_token()
+        if id_token := token_response.get("id_token"):
+            if self.id_token_decoder:
+                decoded_id_token = self.id_token_decoder.decode(id_token)
+            else:
+                decoded_id_token = token_response.decode_id_token()
```
derek-globus, line 131: This path gets mentioned in another code path & I imagine you've thought through it, but would benefit from hearing your thinking.

The approach here is to use a new IDTokenDecoder for each decoding because that's what `token_response.decode_id_token` does anyways. Do we get a parallelism or simplicity benefit from pushing that down into the token response instead of just enforcing that there is an id token decoder?

derek-globus (commented):
I've got some questions about interfaces & implementation details.

ID token decoding has a good amount of history & is a pretty nuanced problem so I fully admit that I may just be missing some history/use case but I'd benefit from discussing those a bit.

src/globus_sdk/globus_app/app.py
```diff
@@ -239,6 +249,29 @@ def _resolve_token_storage(
...
+
+        This must run near the end of app initialization, when the `_token_storage`
+        (inner) and `token_storage` (validating storage, outer) storages have both
+        been initialized.
+        """
+        id_token_decoder = (
+            self.config.id_token_decoder
+            if self.config.id_token_decoder is not None
+            else DefaultIDTokenDecoder(self._login_client)
+        )
```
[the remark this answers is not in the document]
sirosen: I was going to push back on this, but I see that your formulation lets us remove the `None` possibility. I like that.

I'm going to go after this, but I'm including my original response (unmodified) below just to share some thoughts.

-----

I agree with everything you've said. But I have an additional three beliefs which make me favor the explicit option only here.

1. The provider interface also introduces complexity to our codebase.
2. The only use-cases in which a custom decoder would be used are advanced ones in which a user wants to take much more control over the internal behaviors of GlobusApp.
3. It is easy and safe to add `IDTokenDecoderProvider` if there is ever demand for such a feature.

There's no requirement that an `IDTokenDecoder` (the generic type) contains an Auth Client, and the login client usage inside of it isn't necessarily coordinated with usage in Globus App (since it only is needed to reach public APIs described by the OIDC spec). So I'm not sure that the additional burden of the simpler (for us) interface is significant.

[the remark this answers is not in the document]
derek-globus: FWIW, I don't entirely disagree with your theoretical pushback; especially number 1. I still hold my original stance & am glad you're going after it but you do raise a very valid con in the shape of complexity.

src/globus_sdk/services/auth/id_token_decoder.py
sirosen: `AuthClient` does not inherit from `AuthLoginClient` but implements `SupportsJWKMethods`.

The modeling of the auth clients is one of the things for which I have a v4 issue in our backlog. I think we can do better on it, but I'm not sure how and I'd rather tackle the problem when we have more freedom to move things around.

One of the possible outcomes of such a refactor could be that we decide that `AuthClient` is not a valid place to implement `SupportsJWKMethods` but we keep the protocol. Or we could decide to remove the protocol as well, and say that usage of `SupportsJWKMethods` should be replaced with `AuthLoginClient`.
I don't know what we'll do when we come back to this, but I'm just trying to express the relationship in terms of the protocol for now because `DefaultIDTokenDecoder` doesn't really care what it's given as long as it can be used to retrieve JWK and OIDC config data and a client ID.

---

As mostly an aside, I actually think `SupportsJWKMethods` is a good declaration of the interface here.

You could imagine
```python
class AuthCacheAdapter:
    def __init__(self, auth_client):
        self.auth_client = auth_client
        self.client_id = auth_client.client_id
    def get_jwk(...): ...
    def get_openid_configuration(...): ...
    def oauth2_introspect_token(...): ...
    def oauth2_dependent_tokens(...): ...
```

And such an object is "obviously" acceptable to `DefaultIDTokenDecoder`.

Protocols and interfaces are much closer to Python's historic allowance for duck typing than strict inheritance requirements. Especially given that type checks are only enforced by static analyzers and not by the language itself, I think it's better to think in terms of protocols.

src/globus_sdk/services/auth/response/oauth.py
sirosen: Oh, yeah, this can be made into debug. In general the SDK is quite chatty in its logging since the expectation is that library logging will be turned off and that if someone goes out of their way to turn it on, they want _a lot_ of detail on what's happening.

src/globus_sdk/tokenstorage/v2/base.py
sirosen: Yeah, I noticed this as well after finishing the implementation. Let me chew on it a little.
My reason for keeping it as a property was that the documentation is easier to chunk for sphinx in a property, but maybe that's a bad reason.

src/globus_sdk/services/auth/id_token_decoder.py
derek-globus: I don't object to protocols wholesale (in fact I'm advocating for one in a different comment) but I do hold that they have some big complexity cons with them and that this interface is better suited by inheritance.

I realize that this change is inheriting (no pun intended) art from other places in the SDK and we're going to reconsider it in v4 so I'll drop this thread.

src/globus_sdk/tokenstorage/v2/base.py
sirosen: Okay, why don't I do that! I was hesitant but why not, right?

sirosen: It is, IMO, a bit uglier in the rendered docs, but I can live with it in exchange for saving us some unnecessary method wrappers.

src/globus_sdk/tokenstorage/v2/base.py
derek-globus: My instinct is to push on the idea of `IDTokenDecoder` being an ABC.

There's a line being drawn between `IDTokenDecoder` and `DefaultIDTokenDecoder` right now that the invocation of `jwt.decode()` is a common concept (pulled up into a parent class) but the acceptance and assignment of a `SupportsJWKMethods` object is just an implementation detail of the default decoder. I don't see a reason to draw that line & it makes the `decode` method more complicated (it needs to reason about "what if I'm not decoding for a default id token decoder?")

If I were to make my own IDTokenDecoder, I'd extend `DefaultIDTokenDecoder` not `IDTokenDecoder` as it has builtin logic that I can just override. What do you think of pushing `decode` into `DefaultIDTokenValidator`, renamed it `IDTokenValidator`, and removing the parent abstract class (or keeping it as an interface without any actual implementations defined)?

src/globus_sdk/globus_app/app.py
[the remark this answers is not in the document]
derek-globus: > do we want to add any new specific tests for the provider mechanism

I'd probably do something along the lines of my snippet from the original message.
Doesn't need to be anything crazy; just encode in tests that we support passing in a subclass of IDTokenDecoder that provides itself via the interface so that we don't break that integration. (Technically it doesn't need to be an IDTokenDecoder subclass but 9/10 times would be).

src/globus_sdk/tokenstorage/v2/base.py
sirosen: Let me get an easy one out of the way: I am strongly against renaming the class to `Validator`. It decodes a JWT; it doesn't only validate it -- it's the proper interface for taking that (encoded) representation of the data and unpacking (decoding) it.
The JWT spec ([RFC 7519](https://www.rfc-editor.org/rfc/rfc7519)) doesn't define decoding and validation as separable activities. So decoding includes validation, but that's just part of the nature of JWTs. (And the `pyjwt` implementation.)

As for rearranging the class hierarchy... I am against that too, but less strongly. Here's a pair of valid decoders:
```python
class SimpleFilesysDecoder(IDTokenDecoder):
    def get_openid_configuration(self) -> dict[str, t.Any]:
        with open("/opt/globus-data/oidc_config.json") as fp:
            return json.load(fp)

    def get_jwk(self) -> RSAPublicKey:
        import jwt

        with open("/opt/globus-data/jwk.json") as fp:
            return jwt.algorithms.RSAAlgorithm.from_jwk(json.load(fp))


class DirectRequestsDecoder(IDTokenDecoder):
    def get_openid_configuration(self) -> dict[str, t.Any]:
        with open("/opt/globus-data/oidc_config.json") as fp:
            return requests.get("https://auth.globus.org/.well-known/openid-configuration").json()

    def _get_raw_jwk(self) -> dict[str, t.Any]:
        with open("/opt/globus-data/oidc_config.json") as fp:
            return requests.get("https://auth.globus.org/jwk.json").json()["keys"][0]

    def get_jwk(self) -> RSAPublicKey:
        import jwt

        return jwt.algorithms.RSAAlgorithm.from_jwk(self._get_raw_jwk())
```

That is, I don't think it is correct to require an AuthClient or AuthLoginClient object be directly available in order to define a decoder. What's necessary is that you have access to the OIDC configuration and the JWKs. That can be driven via the network, but you can have whatever storage or cache you want for these data.

So if we combine the two, we lose the proper distinction between decoding -- a data driven activity -- and the _default_ way we suggest that decoding is done -- in which those data are fetched dynamically via HTTPS on first use.

I don't agree that this makes the `decode` method more complicated. It properly abstracts away the source of the data which it needs, so it just calls some abstract methods which get it the relevant data. I would say it makes it simpler.

derek-globus: > Let me get an easy one out of the way: I am strongly against renaming the class to Validator.

Sorry, typo; didn't mean to suggest that. `message.replace("Validator", "Decoder")`

sirosen: Oh, and to make sure I'm clear, I'll call this out:

> If I were to make my own IDTokenDecoder, I'd extend `DefaultIDTokenDecoder` not `IDTokenDecoder` as it has builtin logic that I can just override.

Yes! I would expect that's the common case.

But it's not a strict requirement in order to define JWT decoding, and I don't want the fact that it's the common case to drive us to define an interface which combines multiple responsibilities which I think are better distinguished.

sirosen: This thread drifted from the original question -- about the choice to dispatch here -- into the class structure itself. I've just pushed an update which changes that significantly, so I'm marking resolved. There may still be issues worth discussing, but I've done my best to address any concerns with the refactor I applied.

sirosen:
I've just pushed a changeset which changes the abstract interface to
```python
class JWTDecoder(ABC):
    def decode(self, token: str, /) -> dict[str, Any]: ...
```
and therefore can give the `IDTokenDecoder` name to the primary implementation.

I think this tightens things up rather significantly.

I did want to try an approach using structural subtyping (protocols) rather than nominal subtyping (ABCs), since that would have allowed for the implementation to take additional optional arguments, but it seemed messy when defining the provider protocol in addition. So I stuck with nominal subtyping and made the interface _really_ narrow.

changelog.d/20250219_000456_sirosen_id_token_decoder.rst
```diff
@@ -0,0 +1,32 @@
+Added
+~~~~~
+
+- Introduce ``globus_sdk.IDTokenDecoder`` and
+  ``globus_sdk.JWTDecoder`` classes, which implement ``id_token``
+  decoding. (:pr:`NUMBER`)
+
+  - ``JWTDecoder`` is an abstract class declaring the interface, while
+    ``IDTokenDecoder`` is the primary implementation.
```
kurtmckee, line 9: Why is the abstract class a top-level class?

sirosen: IMO it's an unfortunate necessity until we plan out new namespacing to give service-specific components an appropriate exported name. If you look at `__init__.pyi`, you'll see that _all_ of the usable names in `globus_sdk.services.**.py` are top-level exports.

I don't yet have a handle on how we can organize things such that we consistently provide service-specific functionality under good names. There are _options_ we could pursue -- e.g., use `globus_sdk.services.auth` as the namespace -- but we haven't done that to-date and I don't want to introduce a change in style in any particular feature PR. I think that if we want to shift all of our naming to use that mode, we should have a more complete plan which encompasses things like all of the GCS document types.

I wouldn't be particularly in favor of `globus_sdk.services.auth.JWTDecoder` as the name we ask a user to access. It's two-layers deep, and I think it's incongruous with `globus_sdk.AuthClient` being available. I'm pretty sure we don't want
```python
from globus_sdk.services.auth import AuthClient
from globus_sdk.services.transfer import TransferClient
```

Perhaps what we need is to _not_ have per-service silos at all, but functional subpackages:
```python
from globus_sdk.clients import AuthClient, TransferClient
```

That works, but where do the document datatypes and helper objects like this, often specific to various services, go?

```python
from globus_sdk.documents import TransferData, StorageGatewayDocument
```
:+1: 

```python
from globus_sdk.helpers import IdentityMap, JWTDecoder
```
:-1: 

So what's the namespace for these?

---

As you can see, I have thoughts on this subject, but not clear conclusions as of yet. Until we set aside some time to figure this out, ideally in v4, I'm keeping things as top-level exports.

src/globus_sdk/services/auth/id_token_decoder.py
```diff
@@ -0,0 +1,170 @@
...
+    :param auth_client: The client which should be used to callout to Globus Auth as
+        needed. Any AuthClient or AuthLoginClient will work for this purpose.
+
+    :ivar float | datetime.timedelta jwt_leeway: The JWT leeway to use during decoding,
+        as a number of seconds or a timedelta. The default is 5 minutes.
+    :ivar dict jwt_options: The ``options`` passed to the underlying JWT decode
+        function. Defaults to an empty dict.
+    """
+
+    def __init__(self, auth_client: SupportsJWKMethods) -> None:
```
derek-globus, line 51:
```suggestion
    def __init__(
        self,
        auth_client: SupportsJWKMethods,
        *,
        jwt_leeway: float = 300.0,  # seconds
        jwt_options: dict[str, t.Any] | MissingType = Missing,
    ) -> None:
```

derek-globus: How would you feel about making these instance vars configurable as init params?

This makes jwt leeway default a part of the class signature, not just an internal comment in the code & allows for some more interesting extension patterns if users want to change it.

Extension 1:
```python
decoder = IDTokenDecoder(auth_client, jwt_leeway=10.0)
config = GlobusAppConfig(id_token_decoder=decoder)
```

Extension 2:
```python
config = GlobusAppConfig(
  id_token_decoder=functools.partial(IDTokenDecoder, jwt_leeway=10.0)
)
```

derek-globus (commented):
Small nit: PR title & details need to be updated to reflect the new names of stuff.

src/globus_sdk/services/auth/id_token_decoder.py
sirosen: Yeah, I like that. Let me refine to allow for it.

It might also simplify one specific odd mock for a test.

src/globus_sdk/globus_app/app.py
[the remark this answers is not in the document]
sirosen: I've got some basic tests in, so I'm marking resolved. Obviously, as with any software, more testing is always possible. :slightly_smiling_face:

src/globus_sdk/globus_app/config.py
```diff
@@ -115,6 +121,9 @@ class GlobusAppConfig:
     token_validation_error_handler: TokenValidationErrorHandler | None = (
         resolve_by_login_flow
     )
+    id_token_decoder: globus_sdk.JWTDecoder | JWTDecoderProvider = (
+        globus_sdk.IDTokenDecoder
+    )
```
derek-globus, line 126: Sorry, I missed this on the first pass. Should be my last note.

This key should become `jwt_decoder` now, right?
Our default provider is the `IDTokenDecoder` class but the range of things you can pass to this config values are `JWTDecoders` not `IDTokenDecoders`.

sirosen: I'm torn. It was intentional and not an oversight, but I could be talked into changing it. I'll use the concept names and not class names to explain my thought process:

The naming of `JWTDecoder` follows from
> id_token is a JWT
> ID Token Decoders are JWT Decoders

But in practical fact _here_ (where a JWT Decoder is used) is that the _type_ of JWT which you will be given is an `id_token`.

So the usage _in GlobusApp_ is that the type of JWT is an ID token. The contract which we need fulfilled is that of "any valid JWT Decoder", but specifically ID token decoding is how that decoder will be used.

Result:
- I left it named `id_token_decoder` :grey_question: 
- Which one is it really? :confused:

derek-globus: Whichever it is, it should be consistent.

This comment says "The thing needs to be an IDTokenDecoder" but the types say "it can be a JWTDecoder".
If those concepts don't line up properly, then we either named things improperly (type names don't match the concepts their strings refer to) or are mis-applying well-defined names.

As it stands, it's sort of a more complex version of:
```python
class Rectangle: ...
class Square: ...

class SquareFrobulator:
  def frobulate(square: Rectangle) -> None: 
    """
    :param square: A rectangle that will always be a square but I didn't type annotate as a Square because ???
    """
```

sirosen: I think that example is a mismatch for this situation to a sufficient degree that it's not useful to discuss. i.e. It reduces away the _necessary_ details which make this situation nuanced.

Here's how I documented this attribute of the `GlobusAppConfig`:
```
    :ivar ``JWTDecoder`` | ``JWTDecoderProvider`` id_token_decoder:
        A JWT decoder or a provider which produces a decoder. The decoder is used when
        decoding ``id_token`` JWTs from Globus Auth. Defaults to ``IDTokenDecoder``.
```

That to me explicitly explains why the name is `id_token_decoder`. It is an "`id_token` Decoder", not a short-hand for `IDTokenDecoder`.

I'm not saying that I won't be convinced to change the name, but IMO the above is not a good argument. The type and concept names _are_ aligned here. It's just that `id_token_decoder` _happens_ to also look like the snake-case spelling of `IDTokenDecoder`.

PR #1137, opened 2025-02-26
Improve the initialization ordering of GlobusApp

sirosen:
We have a cyclic dependency in GlobusApp init, in which we want a
self-registered client for `openid` scoped consent API usage, but we
need this as an input to various constructs before we are ready to
register any clients in the app. We have a "cycle breaker" bool flag
in app init which gets used to work around this behavior.

We cannot break the cycle fully in the current design: apps require
an auth client, and clients support use with apps.
But we can rephrase the cycle to make it easier to understand and
maintain.

1.  Externalize the app-attachment behavior of a client in a new
    `BaseClient.attach_globus_app()` method. This method is used to
    implement the client init behaviors (except for a slightly subtle
    one around environment setting), and replaces
    `BaseClient._finalize_app()`.

    This allows for a client to be instantiated and -- so long as no
    app, scopes, or authorizer are set on that client -- late-bound to a
    GlobusApp.

2.  Update `GlobusApp.__init__` to create the openid-scoped auth client
    without passing `app=self` or `app_scopes`; instead, late-bind
    these parameters to the client after the app is initialized to the
    point that it can handle the client's callbacks to the app.

    This is a more straightforward cycle-breaker to read and maintain
    than the prior internal bool tracking. Semantically, it does not
    change the behavior of the app (the now-unguarded cache-clearing
    call is a no-op on a new app).

3.  As an unanticipated benefit, `BaseClient.__init__` has
    significantly improved readability because app-attachment logic is
    all consolidated (moreso than it was under `_finalize_app()`).

Net-net, we have a new method to support,
`BaseClient.attach_globus_app()`, but it has a strict contract about
being called once per client and rejecting any internal data
conflicts. It also documents itself as being called during
initialization if an app is passed.

----
📚 Documentation preview 📚: https://globus-sdk-python--1137.org.readthedocs.build/en/1137/

### DIFF
--- a/src/globus_sdk/client.py
+++ b/src/globus_sdk/client.py
@@ -90,22 +90,18 @@ class BaseClient:
             raise exc.GlobusSDKUsageError(
                 f"A {type(self).__name__} cannot use both an 'app' and an 'authorizer'."
             )
-        if app and environment and environment != app.config.environment:
-            raise exc.GlobusSDKUsageError(
-                f"[Environment Mismatch] {type(self).__name__}'s environment "
-                f"({environment}) does not match the GlobusApp's configured "
-                f"environment ({app.config.environment})."
-            )
 
         # Determine the client's environment
-        # Either derived from the app used, or else from the provided kwarg
+        # Either the provided kwarg or derived from the app used
         #
         # If neither is specified, fallback to the GLOBUS_SDK_ENVIRONMENT environment
         # variable.
-        if app:
+        if environment:
+            self.environment = environment
+        elif app:
             self.environment = app.config.environment
         else:
-            self.environment = config.get_environment_name(environment)
+            self.environment = config.get_environment_name()
 
         # resolve the base_url for the client (see docstring for resolution precedence)
         self.base_url = self._resolve_base_url(base_url, self.environment)
@@ -229,14 +225,17 @@ class BaseClient:
                 "Unable to use an 'app' with a client with no "
                 "'resource_server' defined."
             )
-        # the client's environment must match the app's -- we can't tell which of these
-        # were set explicitly programmatically and which were implicit (e.g., from the
-        # env var), so erroring is the safest option
+        # the client's environment must match the app's
         #
-        # although we can track the value used in init (e.g., add
-        # `self._init_environment = environment` to init), it's harder to keep track of
-        # usages like
-        # `client = FooClient(); client.environment = "test"; client.base_url = ...`
+        # there are only two ways to get to a mismatch:
+        #
+        # 1. pass an explicit environment which doesn't match the app, e.g.,
+        #   `MyClient(environment="a", app=app)` where `app.config.environment="b"`
+        #
+        # 2. initialize a client without an app and later attach an app which doesn't
+        #    match, e.g., `MyClient(environment="a").attach_globus_app(app)`
+        #
+        # in these cases, the user has explicitly given us conflicting instructions
         if self.environment != app.config.environment:
             raise exc.GlobusSDKUsageError(
                 f"[Environment Mismatch] {type(self).__name__}'s environment "
--- a/src/globus_sdk/client.py
+++ b/src/globus_sdk/client.py
@@ -245,17 +245,12 @@ class BaseClient:
 
         # now, assign the app, app_name, and scopes
         self._app = app
-        self.app_scopes = app_scopes
+        self.app_scopes = app_scopes or self.default_scope_requirements
         if self.app_name is None:
             self.app_name = app.app_name
 
-        if self.app_scopes:
-            scope_requirements = self.app_scopes
-        else:
-            scope_requirements = self.default_scope_requirements
-
         # finally, register the scope requirements on the app side
-        self._app.add_scope_requirements({self.resource_server: scope_requirements})
+        self._app.add_scope_requirements({self.resource_server: self.app_scopes})
 
     def add_app_scope(self, scope_collection: ScopeCollectionType) -> BaseClient:
         """
--- a/src/globus_sdk/client.py
+++ b/src/globus_sdk/client.py
@@ -81,6 +81,16 @@ class BaseClient:
         app_name: str | None = None,
         transport_params: dict[str, t.Any] | None = None,
     ) -> None:
+        # check for input parameter conflicts
+        if app_scopes and not app:
+            raise exc.GlobusSDKUsageError(
+                f"A {type(self).__name__} must have an 'app' to use 'app_scopes'."
+            )
+        if app and authorizer:
+            raise exc.GlobusSDKUsageError(
+                f"A {type(self).__name__} cannot use both an 'app' and an 'authorizer'."
+            )
+
         # Determine the client's environment.
         if app is not None:
             # If we're using a GlobusApp, the client's environment must either match the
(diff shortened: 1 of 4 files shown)


src/globus_sdk/client.py
```diff
@@ -81,21 +81,30 @@ def __init__(
...
+        if app and authorizer:
+            raise exc.GlobusSDKUsageError(
+                f"A {type(self).__name__} cannot use both an 'app' and an 'authorizer'."
+            )
+        if app and environment and environment != app.config.environment:
+            raise exc.GlobusSDKUsageError(
+                f"[Environment Mismatch] {type(self).__name__}'s environment "
+                f"({environment}) does not match the GlobusApp's configured "
+                f"environment ({app.config.environment})."
+            )
```
derek-globus, line 98: I want this error to go away but am not sure whether there's a super clean way to do that.

`attach_globus_app` raises the exact same error message if the client's environment variable doesn't match the app's.  I think the ideal version of environment assignment has a clear precedence ordering; something like:

```python
class BaseClient:
   def __init__(...):
      ...
      self.environment = environment or app.config.environment or config.get_environment_name
```

then later during app attachment if environment was explicitly passed and that differs from the app's environment, we raise the error.  Not to mention that the above errors if `app` is none. 

---

What do you think? Am I onto something here or is this pushing in an unhelpful direction?

src/globus_sdk/client.py
```diff
@@ -176,20 +174,89 @@ def _resolve_base_url(cls, init_base_url: str | None, environment: str) -> str:
...
+        # now, assign the app, app_name, and scopes
+        self._app = app
+        self.app_scopes = app_scopes
+        if self.app_name is None:
+            self.app_name = app.app_name
+
+        if self.app_scopes:
+            scope_requirements = self.app_scopes
+        else:
+            scope_requirements = self.default_scope_requirements
```
derek-globus, line 256: Any benefit to assigning `app_scopes` as a client-level instance attribute comes from exposure of "this is what we passed to the app". In the case that app scopes aren't supplied explicitly, we pass the default scope requirements to the app & should reflect that in the attribute.

```suggestion
        self.app_scopes = app_scopes or self.default_scope_requirements
        if self.app_name is None:
            self.app_name = app.app_name
```

derek-globus (commented):
I'm basically a yes; just one idea/question about error deduplication and one comment about app_scopes.

src/globus_sdk/client.py
sirosen: It's something I would vaguely like as well, but as I recall I didn't find a way of making it happen which felt sufficiently clean. I'll take another look before I decide whether I want to try to do this now vs save for later -- IMO that's the decision we should make, whether or not we want to try to solve it right now.

src/globus_sdk/client.py
sirosen: I'm a bit reticent to change this, since the prior logic produced a `BaseClient.app_scopes` of `None` in the default case and this would change that. It's a very small change, but I don't want to change the behavior of that attribute without first acknowledging that we would be doing so.

Given that `default_scope_requirements` already exists as an attribute, I think it's better for `self.app_scopes` to be defined as "the scopes which were passed on init or when the app was attached", not "the scopes which were passed to the app itself".

src/globus_sdk/client.py
derek-globus: Fine with that approach; will consider this not a blocker for merging.

src/globus_sdk/client.py
derek-globus: Fair callout; this would be a slight behavioral change.

It's quite confusing to me that this would be the case:
```python
app = UserApp(client_id='my-cool-client')
flows = FlowsClient(app=app)

# Error: None != [...whatever the default flows scopes are...]
assert flows.app_scopes == app.scope_requirements[flows.resource_server]
```

So much so that I'm happy to have that slight behavioral shift.

sirosen: I disagree because I _don't_ consider that comparison mismatching to be confusing. Specifically because of these two cases:
```python
# setup
app = UserApp(client_id='my-cool-client')
flows_client = FlowsClient(app=app)

...

# case 1: another client comes along
other_flows_client = FlowsClient(app=app, app_scopes=FlowsClient.scopes.manage_runs)

...

# case 2: explicit scope manipulation
app.add_scope_requirements({flows_client.resource_server: [...]})

...

# boom!
assert flows_client.app_scopes == app.scope_requirements[flows_client.resource_server]
```

If you want to know what scopes an app will use, you need to ask the app itself -- asking the client is unsafe.

I'm finding the argument to change it unconvincing because I don't think there's a clear principle that we'd be following here. So it's not outweighing the my default towards backwards compatibility.

---

One thing that I'm unclear about as I look at this body of code is this:

Why is `BaseClient.app_scopes` an attribute at all? Why are we recording the value which (as of today) is passed on init?

In the case of explicit scope manipulation on the app, it may be recorded with a non-`None` value but inaccurate.

I thought we were just recording data from init-time for potential debug purposes. But you seem to be saying you have some other expectation about what that value means, and that leaves me confused.

src/globus_sdk/client.py
sirosen: Okay, I have an improvement for this, and it includes a comment which explains the (only) two paths to getting to an environment conflict.

I kept thinking I found cases which could slip through and then convincing myself that those were unreachable. Ultimately, writing down the two real cases helped me stop going in circles.

src/globus_sdk/client.py
derek-globus: I'd also be happy to remove the attribute.

From my initial comment:
> Any benefit to assigning app_scopes as a client-level instance attribute comes from exposure of "this is what we passed to the app"

That was very intentionally phrased cause I'm not convinced that there is actually any benefit to assigning app_scopes as a client-level instance attribute.

---

In your rebuttal example, I wouldn't expect `flows_client` to combine all scopes registered on the app into its `app_scopes` attribute. I would expect it to report the portion of the combined scopes that it brought to the party.

If flows_client_1 brings the default `chips & dip` to the party, flows_client_2 brings `soda`, and flows_client_3 brings `dino nuggets` then I expect that:
  * `flows_client_1.what_did_you_bring()` returns `chips & dip`
  * `flows_client_2.what_did_you_bring()` returns `soda`
  * `flows_client_3.what_did_you_bring()` return `dino nuggets`
and ultimately
  * `party_app.what_did_all_the_flows_clients_brings()` returns `[chips & dip, soda, dino nuggets]`

sirosen: Okay, let's save removal for future work. If we're both game to remove it, that's an option to consider.
(I tend slightly towards making it a private attribute which records what was passed, since it could help in a debugging session but isn't otherwise important to record in that place.)

I do find your party_app example pretty delightful. And I think the analogy is strong -- the rationale there is "`app_scopes` records what this client passed to the app".

The only real issue I have is that the result should have been `[chips, dip, soda, dino nuggests]`. Clearly chips and dip are separate items.

Given that `client.app_scopes` is already not super-useful, I'm okay with including the change now that I have a more complete rationale for why we should change it. It's such a marginal incompatibility, it's unlikely to hurt anyone. I will note it in the changelog though.

derek-globus: > The only real issue I have is that the result should have been [chips, dip, soda, dino nuggests]. Clearly chips and dip are separate items.

Your feedback is appreciated; will update in a followon PR.

sirosen: I gave this nice, neutral-sounding language in the changelog which doesn't call this a fix or a breaking change. Considering this resolved, please reopen if you feel the need to respond to my `chips, dip` provocation!

derek-globus (approved):

PR #1145, opened 2025-03-07
Initial tutorial rewrite for GlobusApp

sirosen:
This rewrite is the first step towards creating a better "Getting Started" experience for our users without having to redirect them *away* from the tutorial.

The new tutorial is intentionally very light on detail. It is not a GlobusApp user-guide and it is not concerned with explaining advanced usage scenarios or even such details as how tokenstorage is wired up by default. That is left for future getting-started docs.

**Update**: The new doc layout now also separates out "Getting Started" docs so that the user guide can grow to encompass more narrative documentation.

~The original tutorial is (at least temporarily) retained under the name `legacy_tutorial` so that we can continue to accommodate users who are building or maintaining applications built in that way.~
~Ideally it would be removed, but it would probably (?) need replacement with new narrative documentation which covers how you can build applications without GlobusApp. Otherwise, users who have existing applications will be left without the means to understand the code they have inherited.~

~The legacy tutorial is hardly changed. It now has a warning at the start and no longer links to the legacy examples.~

**Update**: The legacy tutorial has been removed in recent updates. There is now a new doc which replaces the legacy tutorial in documenting how to write a script without GlobusApp.

The new tutorial copies only the client creation documentation, but otherwise involves new steps and examples.

----
📚 Documentation preview 📚: https://globus-sdk-python--1145.org.readthedocs.build/en/1145/

### DIFF
--- a/docs/user_guide/getting_started/minimal_script.rst
+++ b/docs/user_guide/getting_started/minimal_script.rst
@@ -10,9 +10,9 @@ This is a basic tutorial in the use of the Globus SDK.
 You can jump right in by using the ``CLIENT_ID`` seen in the example code blocks below!
 That is the ID of the tutorial client, which lets you get started quickly and easily.
 
-When you are ready to create your own application, follow :ref:`the tutorial on
-registering a Native App <tutorial_register_app>` and use its ``CLIENT_ID`` in
-the rest of the tutorial to get your own app setup.
+When you are ready to create your own application, follow
+:ref:`How to Register an App in Globus Auth <tutorial_register_app>` and use
+its ``CLIENT_ID`` in the rest of the tutorial to get your own app setup.
 
 For readers who prefer to start with complete working examples, jump ahead to the
 :ref:`example scripts <minimal_script_complete_examples>` at the end before reviewing
@@ -49,7 +49,7 @@ Start by defining an application object using :class:`UserApp`:
 .. note::
 
     The default behavior for a :class:`UserApp` is to do a CLI-based login flow.
-    This behavior can be disabled or customized in numerous ways.
+    This behavior, and more, can be disabled or customized in numerous ways.
 
     For the full menu of options, look at the documentation about :ref:`Using a
     GlobusApp! <using_globus_app>`
@@ -60,8 +60,9 @@ Access the APIs via Clients
 Once you have an app defined, you can use it with client objects to access
 various Globus APIs.
 
-When you attempt to interact with a service, the app will automatically prompt
-you to login.
+When you attempt to interact with a service using an app-bound service client,
+the app will automatically prompt you to login if valid credentials are
+unavailable.
 
 Start by defining the client object:
 
--- a/docs/user_guide/getting_started/register_app.rst
+++ b/docs/user_guide/getting_started/register_app.rst
@@ -6,9 +6,10 @@ How to Register an App in Globus Auth
 An OAuth2 Client, also sometimes called an "app", represents an application or
 script.
 
-When a user logs in to an application, they grant consent for the client to
-perform certain actions, and the client is issued appropriate credentials for
-those actions.
+When a user logs into an application, they must grant consent for the client to
+perform any required actions.
+At the end of a login flow, Globus Auth issues the client appropriate
+credentials for those actions.
 
 Having your own app registered with the service also lets you set certain app-level
 settings for user logins, and keeps your application isolated from everyone
@@ -27,18 +28,19 @@ the `Clients, Scopes, and Consents
 <https://docs.globus.org/guides/overviews/clients-scopes-and-consents/>`_
 documentation.
 
-App Types & Native App
-----------------------
+Creating a Native App
+---------------------
 
 There are several different types of applications which you can register with
 Globus Auth.
 
-For simplicity, this tutorial will only cover how to create a
-"Native App" -- this is the suitable type of application for scripts and
-distributed applications which can't have a secret associated with them.
+For simplicity, this tutorial will only cover registration of the "Native App"
+type.
+This is the suitable type of application for scripts and distributed
+applications which aren't able to use secrets.
 
-Creating a Native App
----------------------
+Steps
+^^^^^
 
 1. Navigate to the `Developer Site <https://app.globus.org/settings/developers>`_
 


docs/tutorial/tutorial_recap.py
```diff
@@ -0,0 +1,34 @@
+import globus_sdk
```
derek-globus, line 1: Could we include a `__main__` check in these scripts?

```python
if __name__ == "__main__":
  main()
```

I expect customers will frequently download and modify this as-is (as they're advertised to be runnable starting points). This should push those of them less familiar with python very minorly in the direction of good and not-really-inhibiting python habits without affecting readability or comprehension negatively (imo).

docs/tutorial/index.rst
```diff
@@ -0,0 +1,195 @@
...
+When you attempt to interact with a service, the app will automatically prompt
+you to login.
+
+.. note::
+
+    The default behavior for a ``UserApp`` is to do a CLI-based login flow.
+    This behavior can be disabled or customized in numerous ways.
+
+    For the full menu of options, look at the documentation about :ref:`Using a
+    GlobusApp! <using_globus_app>`
```
derek-globus, line 119: Love this note. I wouldn't have thought to phrase it likes this but it's really good.

docs/tutorial/tutorial_recap.py
```diff
@@ -0,0 +1,34 @@
...
+            [
+                group["id"],
+                # note that 'name' could actually have commas in it, so quote it
+                f'"{group["name"]}"',
+                group["group_type"],
+                session_enforcement,
+                roles,
+            ]
+        )
+    )
```
derek-globus, line 34: My gut reaction is that this this processing bit seems like it takes up too much screen space in an otherwise very small script when it's not exactly what the script is trying to teach. 

I'm not sure what suggestions I have though for other use cases to more simply demonstrate "using a response" though. I'll think about it some more but if I can't come up with a better alternative can be okay with this.

docs/index.rst
```diff
@@ -25,7 +25,7 @@ Table of Contents
     :maxdepth: 1
 
     installation
-    tutorial
+    Tutorial <tutorial/index>
```
derek-globus, line 28: This structure is inconsistent with the existing nav-bar structure (a structure that I think I agree with).

In the navbar if an element is a directory of subelements, e.g, [Authorization](https://globus-sdk-python--1145.org.readthedocs.build/en/1145/authorization/index.html) [Service Clients](https://globus-sdk-python--1145.org.readthedocs.build/en/1145/services/index.html) that page is a annotated listing of its subdocuments.

This is a directory element that is also the main content itself.

---

What if this page was phrased using a different term than "tutorial"? Feels like it should stand out in a certain way from the typical "you have a use case, here's the tutorial to show you how to do that" page. Something similar to flask's "Quickstart -> [A Minimal Application](https://flask.palletsprojects.com/en/stable/quickstart/#a-minimal-application)".
image

derek-globus (changes requested):
Love this idea. Thanks for doing it!

Generally, looks very good in terms of the level of abstraction and what you're trying to convey (which is a style of writing that I find particularly difficult, so kudos).
I was a lot more nitpicky about prose and structure than I typically am because of how front and center this page will be. All threads are open to debate of course but there a couple pieces I wanted to poke at.

docs/tutorial/index.rst
```diff
@@ -0,0 +1,195 @@
...
+#. :ref:`Create an OAuth2 Client <tutorial_step1>`
+#. :ref:`Define your App Object <tutorial_step2>`
+#. :ref:`Access the APIs via Clients <tutorial_step3>`
+
+That should be enough to get you up and started.
+
+.. _tutorial_step1:
+
+Step 1: Create an OAuth2 Client
+-------------------------------
```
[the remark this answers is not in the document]
sirosen: Yes to both of those ideas. I prefer "Register".

I'd _like_ to separate it from the other two steps, but it's also part of the first steps you need to work through as a user. So I'm left still thinking about naming/structure. Let me look at it again and see if maybe it can go in a separate "Get ready!" / "Setup" section. I think it's important that users are guided proactively towards registering clients, but agree that it is a "different kind of step". (Numbering the steps may, itself, be the wrong approach...)

docs/tutorial/index.rst
```diff
@@ -0,0 +1,195 @@
...
+
+.. _tutorial_step2:
+
+Step 2: Define your App Object
+------------------------------
+
+Accessing Globus APIs as a user requires that you login to your new app and get
+it tokens, credentials providing access the service.
+
+Start by defining an application object using ``globus_sdk.UserApp``.
```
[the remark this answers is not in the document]
sirosen: I won't promise to take this exact wording, but I'll include some form of this change and mark the commit co-authored.

docs/tutorial/index.rst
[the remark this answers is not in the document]
sirosen: Oh, good callout. I thought of it here because this is the part of the workflow where the user will observe that behavior, but it does better belong with the earlier steps.

docs/tutorial/index.rst
```diff
@@ -0,0 +1,195 @@
...
+                    roles,
+                ]
+            )
+        )
+
+When ``groups_client.get_my_groups()`` runs in the example above, the SDK
+will prompt you to login.
+
+Summary: Complete Examples
+--------------------------
```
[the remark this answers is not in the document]
sirosen: Sounds like someone's been reading the `pydantic` docs! :stuck_out_tongue_winking_eye:

docs/tutorial/tutorial_recap.py
sirosen: First point of disagreement during this review! :boom: 
Because we disagree, I'll go into `--verbose` mode so you can see what I'm thinking.

---

I see a couple of harms in adding such a check. Both minor, but I think they outweigh the benefits.

Harm 1: If we put the examples fully under `main()` functions, then we have to indent all of the code, which I think hurts readability in a minor way -- but it matters for a tutorial.

As some additional nuance, if we split it up between some being global (e.g., the app declaration) but other parts being under `main()`, then we're introducing a pattern which demands explanation about which parts should be global and which should not. So I think it either all gets indented or none of it does (simplest model of use to teach a new user about).

Harm 2: new Python developers may not be familiar with `__name__` and the magical `"__main__"` value. We want to teach them about the SDK, not the language, and avoid introducing new concepts when they may be distracting.

The utility of a `__main__` block is that we are preserving the usability of a file as an importable module. But a novice SDK user may not be interested in that at all, since they may be in one of these two cases:
a) writing scripts, not modules
ii) only writing modules, never intending to use the tutorial code "as is"

We shouldn't assume that the user is interested in Python, and therefore in "learning best practices". And IMO this is a case where "best practices for application developers" != "best practices for sysadmins" -- it depends too much on what the user actually wants to do with the code.

docs/index.rst
sirosen: I think the way to do that is to restructure more... Which maybe is necessary.
Because I think the path here is that `Getting Started` becomes a section with `Installation, Tutorial, Legacy Tutorial` as docs under it. I'm worried about the changeset itself getting away from us a bit. I'll try some things and see if I can find a good path.

sirosen: I'm tackling this before the other changes (since things may move around as a result), and I think I have a clever solution to multiple problems.

The `tutorial/` space can contain a few things
- `minimal_app.rst` (most of the current tutorial)
- `registering_an_app.rst` (the app bit, cross-linked from `minimal_app.rst`)
- `legacy_tutorial.rst` (as now)

And we can name it `Tutorials` (plural), implying that we may/will have multiple tutorials.

And it may have some meta-content at the index. I'm trying this out right now.

docs/tutorial/index.rst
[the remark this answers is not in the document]
sirosen: I went a little bit longer because I felt like the tutorial had room for it once app registration is moved off to another doc. Overall, I'm happy with how this came out.

docs/tutorial/tutorial_recap.py
sirosen: I managed to squish it down to 4 lines total of printing/looping, which I think is good enough that the tutorial doc reads much better. I'm calling that good, but we can reopen this topic as needed.

docs/index.rst
sirosen: I'm marking this resolved because I've done the above rework (some minor changes) and copied out the minimal app example into a quickstart at the root.

I admit that I did this because the resulting doc page
1. felt stumpy
2. made the legacy tutorial too visible and I wanted to "push it down"

But I think this still obeys the spirit of the suggestion.

Marking resolved, but please reopen if we should make more or different changes right now. :smile:

docs/tutorial/tutorial_recap.py
derek-globus: Discussed offline; I'd like to have a larger conversation about this w/ the team & our style-guide across the sdk but will resolve for this particular thread and am okay to leave it as is.

sirosen:
Although it may extend review a bit, after the last round of feedback there really wasn't any good option I could see short of replacing the legacy tutorial. Once I had done 70% of the requisite work, _not_ doing the remaining 30% (which naturally took 90% of the time...) seemed incorrect as docs became redundant.

The refined docs now make mild updates to layout and replace the legacy tutorial with a document which intentionally parallels the new "create a minimal script" doc.

docs/user_guide/getting_started/register_app.rst
```diff
@@ -0,0 +1,72 @@
...
+
+There are several different types of applications which you can register with
+Globus Auth.
+
+For simplicity, this tutorial will only cover how to create a
+"Native App" -- this is the suitable type of application for scripts and
+distributed applications which can't have a secret associated with them.
+
+Creating a Native App
+---------------------
```
derek-globus, line 39: This first section is small enough that I'd consolidate it into a pre-amble to the next one.


```suggestion
Creating a Native App
---------------------

There are 2 different types of clients available for registration in Globus Auth.

For simplicity, this tutorial will only cover registration of the "Native App" type.
This is suitable for scripts and distributed applications which aren't able to
use secrets.

Steps
^^^^^
```

docs/user_guide/getting_started/minimal_script.rst
```diff
@@ -0,0 +1,115 @@
...
+Creating A Minimal Script
+=========================
+
+This is a basic tutorial in the use of the Globus SDK.
+
+You can jump right in by using the ``CLIENT_ID`` seen in the example code blocks below!
+That is the ID of the tutorial client, which lets you get started quickly and easily.
+
+When you are ready to create your own application, follow :ref:`the tutorial on
+registering a Native App <tutorial_register_app>` and use its ``CLIENT_ID`` in
```
derek-globus, line 14: I would inherit the name of the document as the link text rather than writing a a custom one.
It flows nicely enough in this sentence and helps codify in the users mind the name as a thing they can go back to.

docs/user_guide/getting_started/minimal_script.rst
```diff
@@ -0,0 +1,115 @@
...
+    CLIENT_ID = "61338d24-54d5-408f-a10d-66c06b59f6d2"
+
+    # create your app
+    my_app = globus_sdk.UserApp("my-user-app", client_id=CLIENT_ID)
+
+
+.. note::
+
+    The default behavior for a :class:`UserApp` is to do a CLI-based login flow.
+    This behavior can be disabled or customized in numerous ways.
```
derek-globus, line 52:
```suggestion
    This behavior, and more, can be disabled or customized in numerous ways.
```

docs/user_guide/getting_started/minimal_script.rst
```diff
@@ -0,0 +1,115 @@
...
+    GlobusApp! <using_globus_app>`
+
+Access the APIs via Clients
+---------------------------
+
+Once you have an app defined, you can use it with client objects to access
+various Globus APIs.
+
+When you attempt to interact with a service, the app will automatically prompt
+you to login.
```
derek-globus, line 64:
```suggestion
When you attempt to interact with a service using an app-bound service client, the app will
automatically prompt your user to login if valid credentials are unavailable.
```

docs/user_guide/getting_started/minimal_script.rst
```diff
@@ -0,0 +1,115 @@
...
+
+    # print in CSV format
+    # ('name' could actually have commas in it, so it is quoted)
+    print("ID,Name,Roles")
+    for group in groups_client.get_my_groups():
+        roles = "|".join({m["role"] for m in group["my_memberships"]})
+        print(",".join([group["id"], f'"{group["name"]}"', roles]))
+
+When ``groups_client.get_my_groups()`` runs in the example above, the SDK
+will prompt you to login.
```
derek-globus, line 85:
```suggestion
When ``groups_client.get_my_groups()`` runs in the example above, the SDK
prompts the user to login.
```

docs/user_guide/getting_started/minimal_script_noapp.rst
```diff
@@ -0,0 +1,273 @@
...
+
+3.  Customizing and extending :class:`GlobusApp` to suit your use case may
+    require understanding the underlying components.
+
+.. note::
+
+    Prior to the introduction of :class:`GlobusApp`, these tools were the only
+    way that an application could be written with the ``globus_sdk``. If you
+    are maintaining an existing application, you may need to be strategic about
+    when and how to upgrade such usages.
```
derek-globus, line 48: Tangential thought: this note makes me wonder if we should have a migration guide for adopting GlobusApp that we can reference for case #2.

derek-globus (changes requested):
I like the structure of everything and think you covered the concepts really well.

I've got a couple minor nitpicky things where I suggested minor reorganizations or rephrasing of things since these docs are so front and center.

docs/user_guide/getting_started/index.rst
```diff
@@ -0,0 +1,25 @@
...
+scenarios which are not supported by it. Reading the non-``GlobusApp`` example
+may also enhance your understanding of how ``GlobusApp`` works.
+
+.. toctree::
+    :caption: How To Use the SDK
+    :maxdepth: 1
+
+    Register an App in Globus Auth <register_app>
+    Create a Minimal Script <minimal_script>
+    Create a Script Without GlobusApp <minimal_script_noapp>
```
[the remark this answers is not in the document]
sirosen: I didn't like the discrepancy in name format, but I couldn't figure out how to square it with my preferred naming.

I think I prefer imperative for a menu item ("Do X, Do Y"), but it strikes me as strange for a page title.
I'll update to
```
- Register an App ... -> How to Register an App ...
- Create a Minimal Script -> How to Create a Minimal Script
- Create a Minimal Script Without GlobusApp -> How to ...
```

That way, the menu-ing rule is clear and unsurprising -- we've dropped "how to" from the titles to make for more compact menu entries. And made a clever switch between imperative and indicative.

docs/user_guide/getting_started/register_app.rst
```diff
@@ -0,0 +1,72 @@
...
+
+Registering a Globus Auth App
+=============================
+
+An OAuth2 Client, also sometimes called an "app", represents an application or
+script.
+
+When a user logs in to an application, they grant consent for the client to
+perform certain actions, and the client is issued appropriate credentials for
+those actions.
```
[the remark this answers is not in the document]
sirosen: I don't agree with the latter half of this, but I can take `into` and `must grant consent`.

There is a direct connection between the login flow and the first tokens issued to the client. I think it's important that a login flow terminates with the client having been issued the tokens associated with that login. You don't login sans-tokens and then later have the client callout to get tokens.

docs/user_guide/getting_started/register_app.rst
sirosen: I like moving this under the header, but
> There are 2 different types of clients available for registration in Globus Auth.

This is not true. There are actually _many_ different options for registration (including GCSv5 as a special case, for example) -- we just limit ourselves to three or four different types which are supported via Globus App, represented by our two different GlobusApp subtypes.

At a minimum we have: native app (UserApp with client ID only), client credentials grant only (ClientApp), thick client / portal (UserApp with client ID and secret), and templated client app (UserApp with client ID and secret, same as a thick client?).

In Auth, clients are not strictly categorized in this way (or at least, didn't used to be, maybe it has changed), but have a listing of permitted grant types to use. So really the suite of possible app types is something like the full set of combinations over
- client credentials
- authorization code grant
- pkce authorization code grant
- implicit grant

And separately there's a question of parent/child client relationships, which I'm not sure how it impacts these.

docs/user_guide/getting_started/minimal_script.rst
sirosen: Mostly :+1: , but I prefer "prompt you to login" over "prompt your user", since the expected usage in this guide is that you're writing a script for your own use.

docs/user_guide/getting_started/minimal_script.rst
sirosen: Per above, I disagree with this change. The target user is the reader of the doc. Simplifying the model to "you write the code, you login" helps keep the scope tighter.

"You write the code, they login" can be probably be served by identical code, but it's a different target use case.

docs/user_guide/getting_started/minimal_script_noapp.rst
sirosen: I think we need to write such a doc, but it doesn't exist yet. Part of why I'm eager to land this, to clear the way.

docs/user_guide/getting_started/register_app.rst
sirosen: I moved it under the header, took the second paragraph of text from the suggestion, but kept the original first sentence.

docs/user_guide/getting_started/index.rst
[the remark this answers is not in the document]
derek-globus: That's fine. It doesn't bother me to use imperative for the page title but I'm also happy with what you proposed.

derek-globus (approved):
Looks good!

PR #1146, opened 2025-03-13
Convert all info-level logging to debug, and lint to enforce this

sirosen:
Two commits:
- Add a custom flake8 plugin to lint for `log.info`
- Apply new no-info-logging rule

The changelog intentionally does not state this as a new policy, only as a new *fact* about the SDK.

Also, some log lines have been joined together (serial log invocations looked like they could use cleanup).

----
📚 Documentation preview 📚: https://globus-sdk-python--1146.org.readthedocs.build/en/1146/

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -5,3 +5,8 @@ extend-ignore = W503,W504,E203
 
 # in pyi stubs, spacing rules are different (black handles this)
 per-file-ignores = *.pyi:E302,E305
+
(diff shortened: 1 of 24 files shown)


sirosen:
Rebased onto #1147 to see if that clears CI. If it does, we can merge in order, with #1147 first.

kurtmckee (approved):

PR #1150, opened 2025-03-14
Create new 'Data Transfer' section of the User Guide docs

sirosen:
_Update: After initial review, the changes in point (3) below were made._ 

This documentation is mostly moved or recycled; very little of it is wholly new.

1. Move Initiating a Transfer doc into the User Guide

This was one of the "updated examples" under GlobusApp that is now rehomed into this section.
It is almost entirely untouched.

2. Add 'Scheduling a Repeating Transfer' doc

This is mostly a clone of the "managing timers" example doc, but with more narrative and explanation.
The example is left as-is, but the new doc is introduced to flesh out the Data Transfer section of the user guide and make its scope clearer.

3. Add `Usage Patterns` to the `User Guide` as a section containing `Data Transfer`.

The menu layout is now `User Guide > Usage Patterns > Data Transfer > Initiating a Transfer`.
The additional layer of nesting means that we can expand this section of the docs to potentially replace the entire example gallery over time without negatively impacting the sidebar with many specific categories of docs.
It's not yet clear what else might go at the end of the user guide after `Usage Patterns` -- maybe nothing, or maybe docs like the `Upgrading` guide.

> [!NOTE]
> The name "Usage Patterns" is still open to debate. It was chosen to avoid conflicting with and causing confusion with "Examples."
> Other names such as "Scenarios", "Recipes", etc. are all possible, subject to someone making a case for one of them.

----
📚 Documentation preview 📚: https://globus-sdk-python--1150.org.readthedocs.build/en/1150/


docs/user_guide/usage_patterns/data_transfer/scheduled_transfers/create_timer.py
```diff
@@ -0,0 +1,62 @@
...
+import globus_sdk
+from globus_sdk.experimental.globus_app import UserApp
+
+# Tutorial Client ID - <replace this with your own client>
+NATIVE_CLIENT_ID = "61338d24-54d5-408f-a10d-66c06b59f6d2"
+USER_APP = UserApp("manage-timers-example", client_id=NATIVE_CLIENT_ID)
+
+# Globus Tutorial Collection 1
+# https://app.globus.org/file-manager/collections/6c54cade-bde5-45c1-bdea-f4bd71dba2cc
+SRC_COLLECTION = "6c54cade-bde5-45c1-bdea-f4bd71dba2cc"
```
kurtmckee, line 12: Is this UUID now guaranteed? I was under the impression that collection might be recreated from time-to-time.

docs/user_guide/usage_patterns/data_transfer/scheduled_transfers/create_timer.py
```diff
@@ -0,0 +1,62 @@
...
+# scopes for the collections
+timers_client = globus_sdk.TimersClient(app=USER_APP)
+# Omit this step if the collections are either
+#   (1) A guest collection or (2) high assurance.
+timers_client.add_app_transfer_data_access_scope((SRC_COLLECTION, DST_COLLECTION))
+
+# submit the creation request to the service, printing out the ID of your new timer
+# after it's created -- you can find it in https://app.globus.org/activity/timers
+timer = timers_client.create_timer(
+    timer=globus_sdk.TransferTimer(
```
kurtmckee, line 52: Is the `timer` keyword argument required when calling `.create_timer()`? If not, I'd prefer to leave it out.

kurtmckee (approved):

docs/user_guide/usage_patterns/data_transfer/scheduled_transfers/create_timer.py
sirosen: I think there was a brief period when we transitioned from the original tutorial endpoints (go#ep1 and go#ep2) to the new tutorial collections during which we didn't have that guarantee. But a lot of us were not happy with that situation because it would mean that we can't rely on these IDs -- it's a problem not only here but also for docs.globus.org  -- so I'm pretty sure we "fixed" that instability.

docs/user_guide/usage_patterns/data_transfer/scheduled_transfers/create_timer.py
sirosen: Looks like it's a positional-or-keyword parameter, so yeah, I agree. Let's remove `timer=` from these examples.

PR #1154, opened 2025-03-21
Add more data transfer docs to the user guide

sirosen:
New docs under the user guide:

- Move Creating a Guest Collection to user guide
- Replace 'transfer relative deadline' doc example

The "creating a guest collection" doc is almost untouched (lift-and-shift)
whereas the transfer-relative-deadline one is mostly rewritten from the
original example content.

----
📚 Documentation preview 📚: https://globus-sdk-python--1154.org.readthedocs.build/en/1154/


docs/user_guide/usage_patterns/data_transfer/transfer_relative_deadline/index.rst
```diff
@@ -0,0 +1,103 @@
...
+as an ``offset``, an amount of time into the future.
+This gives us a generic phrasing of getting a future date:
+
+.. code-block:: python
+
+    import datetime
+
+
+    def make_relative_deadline(offset: datetime.timedelta) -> str:
+        now = datetime.datetime.now(tz=datetime.UTC)
```
kurtmckee, line 39: `datetime.UTC` is an alias that is only available on Python 3.11 and higher. This needs to be `datetime.timezone.utc` for broad compatibility.

```suggestion
        now = datetime.datetime.now(tz=datetime.timezone.utc)
```

docs/user_guide/usage_patterns/data_transfer/transfer_relative_deadline/index.rst
```diff
@@ -0,0 +1,103 @@
...
+    )
+    transfer_request.add_item(SRC_PATH, DST_PATH)
+
+This is then valid to submit with a :class:`TransferClient`:
+
+.. code-block:: python
+
+    tc = globus_sdk.TransferClient(...)
+
+    tc.submit_transfer(transfer_request)
```
kurtmckee, line 84: ?

```python
    tc = globus_sdk.TransferClient(...)

    tc.submit_transfer(transfer_request)
    
    tc.close()
```

docs/user_guide/usage_patterns/data_transfer/transfer_relative_deadline/submit_transfer_relative_deadline.py
```diff
@@ -0,0 +1,42 @@
...
+SRC_PATH = "/home/share/godata/file1.txt"
+
+# Globus Tutorial Collection 2
+# https://app.globus.org/file-manager/collections/31ce9ba0-176d-45a5-add3-f37d233ba47d
+DST_COLLECTION = "31ce9ba0-176d-45a5-add3-f37d233ba47d"
+DST_PATH = "/~/example-transfer-script-destination.txt"
+
+
+def make_relative_deadline(offset: datetime.timedelta) -> str:
+    now = datetime.datetime.now(tz=datetime.UTC)
```
kurtmckee, line 22: As noted in another comment:

```suggestion
    now = datetime.datetime.now(tz=datetime.timezone.utc)
```

kurtmckee (changes requested):

docs/user_guide/usage_patterns/data_transfer/transfer_relative_deadline/index.rst
sirosen: Ach! Thanks for this catch. I was overly focused on my local dev environment.

docs/user_guide/usage_patterns/data_transfer/transfer_relative_deadline/index.rst
```diff
@@ -0,0 +1,103 @@
...
+Setting a Relative Deadline for a Transfer
+==========================================
+
+When transferring or deleting data via Globus Transfer, users are able to set a
+``deadline`` for their tasks.
+This allows you to declare a time by which the task must be completed -- if the
+deadline is reached and the task is still in progress, it will be cancelled.
+
+The ``deadline`` field in a Transfer task takes a date and time in ISO 8601
+format. Because of the use of a standard format, it is easy to use the Python
```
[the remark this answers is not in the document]
sirosen: Transfer accepts some things which are broader than RFC 3339. The Transfer docs themselves are not super clear on this and I agree, we should point folks towards the narrower safe subset. But I don't want to go off-book too much vs the service docs.

Thinking about wording... What about... this?
> The ``deadline`` field in a Transfer task takes a date and time as a string, with
> support for common ISO 8601 formats.
> Because of the use of a standard format ...
>
> .. note::
>
> > Not all ISO 8601 syntaxes are supported.
> > We recommend sticking to the restrictive subset defined by RFC 3339, in which date-time data is
> > typically formatted as "YYYY-MM-DD'T'HH:mm:SSZ".

docs/user_guide/usage_patterns/data_transfer/transfer_relative_deadline/index.rst
sirosen: There's no `close()` method on a TransferClient or BaseClient. Not sure what this is suggesting?

docs/user_guide/usage_patterns/data_transfer/transfer_relative_deadline/index.rst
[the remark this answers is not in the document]
kurtmckee: :100: Yep, sounds good!

docs/user_guide/usage_patterns/data_transfer/transfer_relative_deadline/index.rst
kurtmckee: If it doesn't exist then no worries, but my suggestion is along the lines of "think about what best practices look like in our examples?" ? ? ? I think?

It popped out at me so I wrote it down.

sirosen: I think you're getting at the fact that there's resource leakage in a lot of current SDK usages -- clients contain transports, transports contain sessions (`requests`), sessions contain connection pools (`urllib3`), and nobody does explicit cleanup of these today.

This is what we were talking about the other day WRT where we should put the responsibility for closing those resources.

There's nothing clearly correct we can tell users to do right now, but I'd like to establish a pattern and then tell users to do that.

docs/user_guide/usage_patterns/data_transfer/transfer_relative_deadline/index.rst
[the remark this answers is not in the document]
sirosen: I added a little intersphinx link out to `datetime.datetime.isoformat()` too, to note that it's a good fit for the format. I hope you like it! Marking resolved.

kurtmckee (approved):

PR #1155, opened 2025-03-21
Add support for Search `index_list`

sirosen:
Recently noticed this was missing. The addition covers:

- a response type for iterability
- a client method
- a test fixture
- basic usage tests

----
📚 Documentation preview 📚: https://globus-sdk-python--1155.org.readthedocs.build/en/1155/

### DIFF
--- a/src/globus_sdk/_testing/data/search/index_list.py
+++ b/src/globus_sdk/_testing/data/search/index_list.py
@@ -1,50 +1,58 @@
+from __future__ import annotations
+
+import typing as t
 import uuid
 
 from globus_sdk._testing.models import RegisteredResponse, ResponseSet
 
 INDEX_IDS = [str(uuid.uuid1()), str(uuid.uuid1())]
-SUBSCRIPTIONS = [None, str(uuid.uuid1())]
+INDEX_ATTRIBUTES: dict[str, dict[str, t.Any]] = {
+    INDEX_IDS[0]: {
+        "subscription_id": None,
+        "creation_date": "2038-07-17 16:48:24",
+        "display_name": "Index of Indexed Awesomeness",
+        "description": "Turbo Awesome",
+        "max_size_in_mb": 1,
+        "size_in_mb": 0,
+        "num_subjects": 0,
+        "num_entries": 0,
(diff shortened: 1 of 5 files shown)


src/globus_sdk/_testing/data/search/index_list.py
```diff
@@ -0,0 +1,51 @@
+import uuid
+
+from globus_sdk._testing.models import RegisteredResponse, ResponseSet
+
+INDEX_IDS = [str(uuid.uuid1()), str(uuid.uuid1())]
+SUBSCRIPTIONS = [None, str(uuid.uuid1())]
+
+
+RESPONSES = ResponseSet(
+    metadata={"index_ids": INDEX_IDS, "subscription_ids": SUBSCRIPTIONS},
```
kurtmckee, line 10: The metadata structure here seems disjointed; it seems like it should be a dict of index IDs to subscription IDs, or a tuple, or something that relates these.

```python
INDEX_MAPPING = {
    str(uuid.uuid1()): None,  # First index has no subscription
    str(uuid.uuid1()): str(uuid.uuid1()),  # Second has a subscription
}
```

or

```python
INDICES = (
    (str(uuid.uuid1()), None),
    (str(uuid.uuid1()), str(uuid.uuid1()),
)
```

As it stands, the index ID and subscription ID mappings are independent of each other except by force of will or convention.

kurtmckee (approved):

src/globus_sdk/_testing/data/search/index_list.py
sirosen: I hesitated to make these into a dict because it doesn't communicate the ordering as strongly, but I think that might be best.
It's awkward because there's a question of "what data can we provide which doesn't simply recapitulate the same data as is in the response?"

A dict of the following form seems good to me:
```python
index_id0 = str(uuid.uuid1())
index_id1 = str(uuid.uuid1())
INDEX_ATTRIBUTES = {
    index_id0: {"subscription_id": None},
    index_id1: {"subscription_id": str(uuid.uuid1())},
}
```

I'll refactor that shortly.

PR #1157, opened 2025-03-25
Add initial user guide doc on handling GAREs

sirosen:
_This documentation is pretty rudimentary in some senses, but it's meant to give us a foothold to start writing more about GARE parsing and handling._

---

In order to handle a GARE from Transfer error, you need to
- try the operation
- parse the error to a GARE
- call `GlobusApp.login()` with the parsed params
- retry the operation

A new documentation section covers "Sessions & Consents" to cover session and consent manipulations for users. In that section, a single new doc covers how to handle `authorization_parameters` from Transfer.

----
📚 Documentation preview 📚: https://globus-sdk-python--1157.org.readthedocs.build/en/1157/

### DIFF
--- a/docs/user_guide/usage_patterns/index.rst
+++ b/docs/user_guide/usage_patterns/index.rst
@@ -28,3 +28,4 @@ transfers.
     :maxdepth: 1
 
     data_transfer/index
+    sessions_and_consents/index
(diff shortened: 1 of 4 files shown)


ada-globus (approved):
I'm happy enough with this as-is, though it's making me a little itchy to write a convenience method on `UserApp` to redrive when provided a GARE (as this seems like it could be generalized without a huge amount of friction).

sirosen:
It's making me itchy to write some different helpers -- like `TransferAPIError.to_gare()` -- let's try to satisfy those desires later, yah?

m1yag1 (approved):

PR #1162, opened 2025-04-07
Add user guide docs for detecting data_access

sirosen:
This is a new doc in the user guide for handling transfers, covering inspection of collections to detect the `data_access` scope requirement.
It is based on the similar function used in the timer submission doc, but explains a little bit further and focuses only on this one piece of functionality.

----
📚 Documentation preview 📚: https://globus-sdk-python--1162.org.readthedocs.build/en/1162/

### DIFF
--- a/docs/user_guide/usage_patterns/data_transfer/detecting_data_access/index.rst
+++ b/docs/user_guide/usage_patterns/data_transfer/detecting_data_access/index.rst
@@ -47,17 +47,13 @@ the service:
 
     collection_doc = transfer_client.get_endpoint(COLLECTION_ID)
 
-.. caution::
+.. note::
 
     Careful readers may note that we use the :meth:`TransferClient.get_endpoint`
     method to lookup a collection.
 
-    The Transfer service contains both Endpoints and Collections, and for
-    historical reasons both document types are available from the Get Endpoint
-    API.
-    This is the correct API to use not only when looking for information about
-    a Collection, but also for detecting whether an ID belongs to a Collection
-    or Endpoint, in cases where it is ambiguous.
+    The Transfer service contains both Endpoints and Collections,
+    and both document types are available from the Get Endpoint API.
 
 
 Reading Collection Type
@@ -70,21 +66,21 @@ First, whether or not the collection is a GCSv5 Mapped Collection:
 
 .. code-block:: python
 
-    entity_type: str = collection_doc["entity_type"]
-    is_v5_mapped_collection: bool = entity_type == "GCSv5_mapped_collection"
+    entity_type = collection_doc["entity_type"]
+    is_v5_mapped_collection = entity_type == "GCSv5_mapped_collection"
 
 Second, whether or not the collection is a High Assurance Collection:
 
 .. code-block:: python
 
-    high_assurance: bool = collection_doc["high_assurance"]
+    is_high_assurance = collection_doc["high_assurance"]
 
 Once we have this information, we can deduce whether or not ``data_access`` is
 needed with the following boolean assignment:
 
 .. code-block:: python
 
-    collection_uses_data_access: bool = is_v5_mapped_collection and not high_assurance
+    collection_uses_data_access = is_v5_mapped_collection and not is_high_assurance
 
 Converting Logic to a Helper Function
 -------------------------------------
--- a/docs/user_guide/usage_patterns/data_transfer/detecting_data_access/index.rst
+++ b/docs/user_guide/usage_patterns/data_transfer/detecting_data_access/index.rst
@@ -1,0 +1,161 @@
+.. currentmodule:: globus_sdk
+
+.. _userguide_detecting_data_access:
+
+Detecting data_access
+=====================
+
+Globus Collections come in several varieties, but only some of them have a
+``data_access`` scope.
+
+``data_access`` scopes control application access to collections, allowing
+users to revoke access for an application independent from other application
+permissions.
+Revoking consent stops data transfers and other operations.
+
+Because only some collection types have ``data_access`` scopes, application
+authors interacting with these collections may need to detect the type of
+collection and determine whether or not the scope will be needed.
+
+For readers who prefer to start with complete working examples, jump ahead to the
+:ref:`example script <userguide_detecting_data_access_example>`.
+
+Accessing Collections in Globus Transfer
+----------------------------------------
+
+The Globus Transfer service acts as a central registration hub for collections.
+Therefore, in order to get information about an unknown collection, we will
+need a :class:`TransferClient` with credentials.
+
+The following snippet creates a client and uses it to fetch a collection from
+the service:
+
+.. code-block:: python
+
+    import globus_sdk
+
+    # Tutorial Client ID - <replace this with your own client>
+    NATIVE_CLIENT_ID = "61338d24-54d5-408f-a10d-66c06b59f6d2"
+    USER_APP = globus_sdk.UserApp("detect-data-access-example", client_id=NATIVE_CLIENT_ID)
+
+    # Globus Tutorial Collection 1
+    # https://app.globus.org/file-manager/collections/6c54cade-bde5-45c1-bdea-f4bd71dba2cc
+    # replace with your own COLLECTION_ID
+    COLLECTION_ID = "6c54cade-bde5-45c1-bdea-f4bd71dba2cc"
+
+    transfer_client = globus_sdk.TransferClient(app=USER_APP)
+
+    collection_doc = transfer_client.get_endpoint(COLLECTION_ID)
+
+.. caution::
+
+    Careful readers may note that we the :meth:`TransferClient.get_endpoint`
+    method to lookup a collection.
+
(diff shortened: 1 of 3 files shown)


docs/user_guide/usage_patterns/data_transfer/detecting_data_access/index.rst
```diff
@@ -0,0 +1,161 @@
...
+    # replace with your own COLLECTION_ID
+    COLLECTION_ID = "6c54cade-bde5-45c1-bdea-f4bd71dba2cc"
+
+    transfer_client = globus_sdk.TransferClient(app=USER_APP)
+
+    collection_doc = transfer_client.get_endpoint(COLLECTION_ID)
+
+.. caution::
+
+    Careful readers may note that we the :meth:`TransferClient.get_endpoint`
```
kurtmckee, line 52:
```suggestion
    Careful readers may note that we use the :meth:`TransferClient.get_endpoint`
```

docs/user_guide/usage_patterns/data_transfer/detecting_data_access/index.rst
```diff
@@ -0,0 +1,161 @@
...
+    method to lookup a collection.
+
+    The Transfer service contains both Endpoints and Collections, and for
+    historical reasons both document types are available from the Get Endpoint
+    API.
+    This is the correct API to use not only when looking for information about
+    a Collection, but also for detecting whether an ID belongs to a Collection
+    or Endpoint, in cases where it is ambiguous.
```
kurtmckee, line 60: Are these intended to be separate paragraphs?

Separately, it might not be worthwhile to mention "for historical reasons", but rather to say simply that "this API supports both collections and endpoints" (but I'm not convinced that there's a strong need to even have this caution block).

docs/user_guide/usage_patterns/data_transfer/detecting_data_access/index.rst
```diff
@@ -0,0 +1,161 @@
...
+
+There are two attributes we need from the collection document to determine
+whether or not a ``data_access`` scope is used.
+
+First, whether or not the collection is a GCSv5 Mapped Collection:
+
+.. code-block:: python
+
+    entity_type: str = collection_doc["entity_type"]
+    is_v5_mapped_collection: bool = entity_type == "GCSv5_mapped_collection"
```
kurtmckee, line 74: Let's rip out the type annotations, which don't help with readability IMHO.

```suggestion
    entity_type = collection_doc["entity_type"]
    is_v5_mapped_collection = entity_type == "GCSv5_mapped_collection"
```

docs/user_guide/usage_patterns/data_transfer/detecting_data_access/index.rst
```diff
@@ -0,0 +1,161 @@
...
+
+Second, whether or not the collection is a High Assurance Collection:
+
+.. code-block:: python
+
+    high_assurance: bool = collection_doc["high_assurance"]
```
kurtmckee, line 80:
```suggestion
    high_assurance = collection_doc["high_assurance"]
```

docs/user_guide/usage_patterns/data_transfer/detecting_data_access/index.rst
```diff
@@ -0,0 +1,161 @@
...
+
+Once we have this information, we can deduce whether or not ``data_access`` is
+needed with the following boolean assignment:
+
+.. code-block:: python
+
+    collection_uses_data_access: bool = is_v5_mapped_collection and not high_assurance
```
kurtmckee, line 87: Et cetera. In general, the code snippets don't need type annotations IMO. We can perhaps have them in the example script, but I don't think we need anything other than the bare code logic throughout here.

```suggestion
    collection_uses_data_access = is_v5_mapped_collection and not high_assurance
```

docs/user_guide/usage_patterns/data_transfer/detecting_data_access/index.rst
```diff
@@ -0,0 +1,161 @@
...
+.. code-block:: python
+
+    def uses_data_access(
+        transfer_client: globus_sdk.TransferClient, collection_id: str
+    ) -> bool:
+        """
+        Use the provided `transfer_client` to lookup a collection by ID.
+
+        Based on the record, return `True` if it uses a `data_access` scope and `False`
+        otherwise.
```
kurtmckee, line 107:
```suggestion
        Return `True` if the collection uses a `data_access` scope
        and `False` otherwise.
```

kurtmckee (changes requested):

docs/user_guide/usage_patterns/data_transfer/detecting_data_access/index.rst
sirosen: Not separate paragraphs, but a linebreak after the period to make them more readable as raw text (maybe poor style? we don't have a strong convention around this as of yet).

I'd rather not nix the whole block, although I understand and agree that removing it wouldn't be the _worst_. I think some users might notice that we've told them to use `get_endpoint` to lookup a **collection** and think we've made a mistake in our docs. Maybe I should tone it down to a `note` block?

I'll look at rephrasing it, and maybe reflow it when I do if you have any thoughts on the linebreak style.

docs/user_guide/usage_patterns/data_transfer/detecting_data_access/index.rst
sirosen: Although in this case I think it's clear what the types are from context (and will take the suggestion to remove), I think annotating our assignments may help users understand payload shapes in some cases. That _assumes_ (and, IMO, this is a big assumption, but one that I have been making) that type-annotated Python has proliferated to the point that our users are likely to be familiar with the syntax.

docs/user_guide/usage_patterns/data_transfer/detecting_data_access/index.rst
sirosen: Per the above comment, and contrasting with that case, in this case I do not believe that the code sample is better without telling the user that this field is a bool.

Options, as I see it:
- add a comment, stating it's a bool
- try to use a suggestive name, like `is_high_assurance`
- use a type annotation

docs/user_guide/usage_patterns/data_transfer/detecting_data_access/index.rst
sirosen: I think we can trust users to do basic deductions that "the outcome of boolean ops on bools is a bool", or "if I compared a string against the field, it's probably a string".

So in this case I'm more open to removal because the type is _knowable_ without prior knowledge of the payload data. However, in the `"high_assurance"` field case, I think indicating the type is important. If we use annotations there, it might be best to be consistent and use them on all vars. If we use a non-annotation way of indicating the type, I would strip all of the annotations (again, for internal consistency).

docs/user_guide/usage_patterns/data_transfer/detecting_data_access/index.rst
kurtmckee: I generally err toward breaks at conjunctions and prepositions, like:

```suggestion
    The Transfer service contains both Endpoints and Collections,
    and for historical reasons both document types are available
    from the Get Endpoint API.
    This is the correct API to use not only when looking for information about
    a Collection, but also for detecting whether an ID belongs to a Collection
    or Endpoint, in cases where it is ambiguous.
```

Downgrading from `caution` to `note` will alleviate some of my concerns here. :+1:

docs/user_guide/usage_patterns/data_transfer/detecting_data_access/index.rst
kurtmckee: For a complete script, that may be the case but for code snippet that converts from an English sentence to a Python line I think it may get in the way.

I don't know for a fact that it's adding to -- nor whether it's harming -- readability, but I think it _might_ be obscuring the simple English->Python value that these inline code snippets offer.

docs/user_guide/usage_patterns/data_transfer/detecting_data_access/index.rst
kurtmckee: I like the `is_high_assurance` suggestion, which mirrors the `is_v5_mapped_collection` above.

sirosen: Okay, let's go with that style and I'll remove the annotations from these examples!

This back and forth has let me share why I added those, and I know we're in sync about trying to provide the best user/reader experience here, so we'll iterate based on how people (including our other colleagues) respond to this. :smile:

docs/user_guide/usage_patterns/data_transfer/detecting_data_access/index.rst
sirosen: I made it a `note`, folded at the `"and"`, and also cut the last sentence as unnecessary / too long. :tada:

kurtmckee (approved):

PR #1164, opened 2025-04-10
Added run manager and monitor support to FlowsClient

m1yag1:
[sc-40144](https://app.shortcut.com/globus/story/40144/cli-add-support-for-flow-run-roles-to-flow-commands)

----
📚 Documentation preview 📚: https://globus-sdk-python--1164.org.readthedocs.build/en/1164/

### DIFF
--- a/changelog.d/20250410_102306_8730430+m1yag1_sc_40144_add_run_manager_monitor_support.rst
+++ b/changelog.d/20250410_102306_8730430+m1yag1_sc_40144_add_run_manager_monitor_support.rst
@@ -1,5 +1,5 @@
 Added
 ~~~~~
 
-- Added support to the create and update operations of the ``FlowsClient`` for
-  including ``run_managers`` and/or ``run_monitors`` principals. (:pr:`NUMBER`)
+- ``FlowsClient.create_flow`` and ``FlowsClient.update_flow`` now support ``run_managers``
+  and ``run_monitors``. (:pr:`NUMBER`)
--- a/changelog.d/20250410_102306_8730430+m1yag1_sc_40144_add_run_manager_monitor_support.rst
+++ b/changelog.d/20250410_102306_8730430+m1yag1_sc_40144_add_run_manager_monitor_support.rst
@@ -1,0 +1,5 @@
+Added
+~~~~~
+
+- Added support to the create and update operations of the ``FlowsClient`` for
+  including ``run_managers`` and/or ``run_monitors`` principals. (:pr:`NUMBER`)
--- a/src/globus_sdk/_testing/data/flows/_common.py
+++ b/src/globus_sdk/_testing/data/flows/_common.py
@@ -123,6 +123,12 @@ TWO_HOP_TRANSFER_FLOW_DOC = {
     "flow_administrators": [
         "urn:globus:auth:identity:05d29dab-bd26-4510-9290-468972e8ac01"
     ],
+    "run_managers": [
+        "urn:globus:auth:identity:7d6064ef-5368-473a-b15b-e99c3561aa9b",
+    ],
+    "run_monitors": [
+        "urn:globus:auth:identity:57088a17-d5cb-4cfa-871a-c5cce48f2aec",
+    ],
 }
 
 TWO_HOP_TRANSFER_RUN_ID = "36ad9f9a-ad29-488f-beb4-c22ab729643a"
--- a/src/globus_sdk/_testing/data/flows/_common.py
+++ b/src/globus_sdk/_testing/data/flows/_common.py
@@ -11,6 +11,8 @@ RUN_ID = str(uuid.uuid1())
 USER1 = f"urn:globus:auth:identity:{uuid.uuid1()}"
 USER2 = f"urn:globus:auth:identity:{uuid.uuid1()}"
 GROUP = f"urn:globus:groups:id:{uuid.uuid1()}"
+RUN_MANAGER_PRINCIPAL = f"urn:globus:auth:identity:{str(uuid.uuid1())}"
+RUN_MONITOR_PRINCIPAL = f"urn:globus:auth:identity:{str(uuid.uuid1())}"
 
 TWO_HOP_TRANSFER_FLOW_USER_SCOPE = (
     "https://auth.globus.org/scopes/"
@@ -123,12 +125,8 @@ TWO_HOP_TRANSFER_FLOW_DOC = {
     "flow_administrators": [
         "urn:globus:auth:identity:05d29dab-bd26-4510-9290-468972e8ac01"
     ],
-    "run_managers": [
-        "urn:globus:auth:identity:7d6064ef-5368-473a-b15b-e99c3561aa9b",
-    ],
-    "run_monitors": [
-        "urn:globus:auth:identity:57088a17-d5cb-4cfa-871a-c5cce48f2aec",
-    ],
+    "run_managers": [RUN_MANAGER_PRINCIPAL],
+    "run_monitors": [RUN_MONITOR_PRINCIPAL],
 }
 
 TWO_HOP_TRANSFER_RUN_ID = "36ad9f9a-ad29-488f-beb4-c22ab729643a"
--- a/tests/functional/services/flows/test_flow_crud.py
+++ b/tests/functional/services/flows/test_flow_crud.py
@@ -1,9 +1,11 @@
 import json
 
 import pytest
+from responses import matchers
 
 from globus_sdk import FlowsAPIError
 from globus_sdk._testing import get_last_request, load_response
+from globus_sdk._testing.models import RegisteredResponse
 
 
 @pytest.mark.parametrize("subscription_id", [None, "dummy_subscription_id"])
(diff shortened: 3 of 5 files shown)


changelog.d/20250410_102306_8730430+m1yag1_sc_40144_add_run_manager_monitor_support.rst
```diff
@@ -0,0 +1,5 @@
+Added
+~~~~~
+
+- Added support to the create and update operations of the ``FlowsClient`` for
+  including ``run_managers`` and/or ``run_monitors`` principals. (:pr:`NUMBER`)
```
sirosen, line 5: We should refer to the methods by name. A client could have multiple operations which are 'create' or 'update' (for different resource types), and even when that's not the case, which of the following two is easier to resolve to some specific bit of code:
- "create operation of FlowsClient"
- "FlowsClient.create_flow"

I would probably go with
```suggestion
- ``FlowsClient.create_flow`` and ``FlowsClient.update_flow`` now support ``run_managers``
  and ``run_monitors``. (:pr:`NUMBER`)
```

src/globus_sdk/_testing/data/flows/_common.py
```diff
@@ -123,6 +123,12 @@
     "flow_administrators": [
         "urn:globus:auth:identity:05d29dab-bd26-4510-9290-468972e8ac01"
     ],
+    "run_managers": [
+        "urn:globus:auth:identity:7d6064ef-5368-473a-b15b-e99c3561aa9b",
+    ],
+    "run_monitors": [
+        "urn:globus:auth:identity:57088a17-d5cb-4cfa-871a-c5cce48f2aec",
+    ],
```
sirosen, line 131: I know this is adding to existing art, not creating something new, but the fact that there are a _bunch_ of hardcoded IDs here and that they're matched across the other fixture data (`TWO_HOP_TRANSFER_RUN`) seems like something we should ease out of a bit.

Can we at least, for now, define these near the top of the file and then re-use the same constant, rather than carefully matching string literals? i.e.
```python
RUN_MANAGER_PRINCIPAL = f"urn:globus:auth:identity:{str(uuid.uuid1())}"
RUN_MONITOR_PRINCIPAL = ...
```
and then use those here and in the run doc below?

sirosen (changes requested):
We should have at least one test for each method which sends these values, pulls the last request, and validates that the data was actually in the payload. There should be examples if you look at `get_last_request()` usages in the testsuite. An inverse test that validates their absence when `None` would also be nice.
When we come back to these and update to use `globus_sdk.MISSING` in a major release, those tests will ensure that we do so correctly.[^1]

[^1]: For the sake of keeping methods internally consistent, I don't think we should use `MISSING` today. But we will very likely switch to it in 4.0 or 5.0.

Also, this is extremely minor, but I would like it if you updated the PR comment to _not_ say that we're just doing this for the sake of CLI support. The CLI is an important consumer of the SDK, but we don't put features into the SDK just to support the CLI. This is about exposing the new `run_managers` and `run_monitors` fields to all SDK users, not just our own team.

src/globus_sdk/_testing/data/flows/_common.py
m1yag1: @sirosen I believe I've addressed all the comments for this PR.

tests/functional/services/flows/test_flow_crud.py
```diff
@@ -23,12 +21,34 @@ def test_create_flow(flows_client, subscription_id):
...
+@pytest.mark.parametrize("value", [None, [], "dummy_value"])
+@pytest.mark.parametrize("key", ["run_managers", "run_monitors"])
+def test_create_flow_request_data(flows_client, key, value):
+    meta = load_response(flows_client.create_flow).metadata
+
+    # Remove the key or set as empty list
+    if value is None:
+        meta["params"].pop(key)
+    else:
+        meta["params"][key] = value
```
sirosen, line 33: `load_response` is pulling global data from a registry.
Therefore, this is modifying a global dict.

We should not modify a loaded `RegisteredResponse.metadata` -- or any other attributes of that global object -- in the course of a test. We can create these constructs in local scope if we need modifiable versions, but the globals should be treated as immutable.

tests/functional/services/flows/test_flow_crud.py
```diff
@@ -23,12 +21,34 @@ def test_create_flow(flows_client, subscription_id):
...
+
+
 def test_create_flow_error_parsing(flows_client):
-    metadata = load_response(
+    meta = load_response(
         flows_client.create_flow, case="bad_admin_principal_error"
     ).metadata
     with pytest.raises(FlowsAPIError) as excinfo:
-        flows_client.create_flow(**metadata["params"])
+        flows_client.create_flow(**meta["params"])
```
sirosen, line 51: As above, since we aren't changing this test, let's not modify variable names in it, to improve the usability of our git history.

tests/functional/services/flows/test_flow_crud.py
```diff
@@ -59,9 +79,26 @@ def test_update_flow(flows_client):
...
+    else:
+        assert req_body[key] == value
+
+
 def test_delete_flow(flows_client):
-    metadata = load_response(flows_client.delete_flow).metadata
+    meta = load_response(flows_client.delete_flow).metadata
 
-    resp = flows_client.delete_flow(metadata["flow_id"])
+    resp = flows_client.delete_flow(meta["flow_id"])
```
sirosen, line 102: Again, as above, let's not rename this unnecessarily.

tests/functional/services/flows/test_flow_crud.py
```diff
@@ -59,9 +79,26 @@ def test_update_flow(flows_client):
         assert resp[k] == v
 
 
+@pytest.mark.parametrize("value", [None, [], "dummy_value"])
+@pytest.mark.parametrize("key", ["run_managers", "run_monitors"])
+def test_update_flow_request_data(flows_client, key, value):
```
sirosen, line 84: I would prefer a more descriptive test name, e.g. `test_update_flow_run_role_serialization`.

That holds for the create flow case as well.

The test name is the primary thing you'll see in the output if it fails. So it can either tell you
- "you broke update flow data"
or
- "you broke serialization of run roles"

tests/functional/services/flows/test_flow_crud.py
```diff
@@ -23,12 +21,34 @@ def test_create_flow(flows_client, subscription_id):
         assert "subscription_id" not in req_body
 
 
+@pytest.mark.parametrize("value", [None, [], "dummy_value"])
```
sirosen, line 24: Both tests pass `"dummy_value"`, but the field is typed as `list[str]`.
I assume this is accidental, and `["dummy_value"]` was intended?

We aren't checking the types of parameters at runtime (and I don't think we necessarily should, at least at present), so the test works fine, but it would break if we ever decided to check in the future, so it's fragile if we're trying to test the normal operating modes for the function under test.

sirosen (changes requested):

tests/functional/services/flows/test_flow_crud.py
```diff
@@ -8,11 +8,9 @@
 
 @pytest.mark.parametrize("subscription_id", [None, "dummy_subscription_id"])
 def test_create_flow(flows_client, subscription_id):
-    metadata = load_response(flows_client.create_flow).metadata
+    meta = load_response(flows_client.create_flow).metadata
 
-    resp = flows_client.create_flow(
-        **metadata["params"], subscription_id=subscription_id
-    )
+    resp = flows_client.create_flow(**meta["params"], subscription_id=subscription_id)
```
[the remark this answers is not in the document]
m1yag1: It was "metadata" in some tests and "meta" in others. The gain was meant to make the file more consistent.  I'll revert the change as requested.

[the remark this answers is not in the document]
m1yag1: @sirosen I reverted the changes, updated the test names, and corrected the dummy values to be a list. I had to force push because I reverted the last commit. Apologies if that causes any detached comments for review.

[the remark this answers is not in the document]
sirosen: No worries about review workflow. :+1: 
I admit it's arguable whether or not using the same variable name across tests is worthwhile consistency. If it were 0 cost to do so, I'd probably do it. But it's not zero cost, and the gain is _terribly_ marginal.

tests/functional/services/flows/test_flow_crud.py
```diff
@@ -23,6 +23,29 @@ def test_create_flow(flows_client, subscription_id):
...
+@pytest.mark.parametrize("key", ["run_managers", "run_monitors"])
+def test_create_flow_run_role_serialization(flows_client, key, value):
+    metadata = load_response(flows_client.create_flow).metadata
+    params = metadata["params"]
+
+    # Remove the key or set as empty list
+    if value is None:
+        params.pop(key)
+    else:
+        params[key] = value
```
sirosen, line 36: This still appears to be modifying the global dict. Based on comments, maybe a change was not pushed?

tests/functional/services/flows/test_flow_crud.py
m1yag1: @sirosen Ok, I'm hoping I finally got it this time. I registered a new response within the scope of that test. 🤞

sirosen (approved):
Excellent; thanks for working through this!

I think this one should probably squash-merge, on account of the number of commits relative to the size of the change -- and in particular because part of the review touched on not being disruptive to git-blame.

PR #1167, opened 2025-04-16
Add 'activity_notification_policy' to run_flow()

sirosen:
- Add a new data-constructor helper, `RunActivityNotificationPolicy`
- Add `activity_notification_policy` as an argument of type
  `RunActivityNotificationPolicy | dict | None`

One intentional oddity here is that for consistency with other
`run_flow` arguments, the default is `None`, which we omit, but in the
payload helper itself we use `MISSING`.

----
📚 Documentation preview 📚: https://globus-sdk-python--1167.org.readthedocs.build/en/1167/

### DIFF
--- a/src/globus_sdk/_testing/data/flows/run_flow.py
+++ b/src/globus_sdk/_testing/data/flows/run_flow.py
@@ -20,6 +20,15 @@ RESPONSES = ResponseSet(
         json=TWO_HOP_TRANSFER_RUN,
         match=[matchers.json_params_matcher(params=_request_params)],
     ),
+    # 'lenient' is the same as the 'default' above, but without requiring
+    # payload matching, which makes it more usable for a wide variety of tests
+    # TODO: revisit which of the fixtures should have which param matchers
+    lenient=RegisteredResponse(
+        service="flows",
+        method="POST",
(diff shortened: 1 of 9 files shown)


src/globus_sdk/_testing/data/flows/run_flow.py
```diff
@@ -20,6 +20,15 @@
         json=TWO_HOP_TRANSFER_RUN,
         match=[matchers.json_params_matcher(params=_request_params)],
     ),
+    # 'lenient' is the same as the 'default' above, but without requiring
+    # payload matching, which makes it more usable for a wide variety of tests
+    # TODO: revisit which of the fixtures should have which param matchers
+    lenient=RegisteredResponse(
```
ada-globus, line 26: 😲

ada-globus (approved):

src/globus_sdk/_testing/data/flows/run_flow.py
sirosen: Yeah, I have my own opinion here, which is a bit stronger than what I've left in the comments. But for now, I just needed an escape-hatch.

PR #1169, opened 2025-04-18
Remove an unnecessarily strict matcher

sirosen:
Similar to #1168

Trying to use a mock response with payload param matching makes
development more difficult because a downstream testsuite *MUST*
provide all of the exact parameters declared or it gets no data. This
contradicts the goal of `_testing`, which is to provide easy,
low-friction mocking for downstream consumers.

If stricter matchers are a desirable feature, they can be opt-in by
making non-`default` case mocks which have these requirements.
(Essentially, the inverse of the `lenient` variant mock which is being
removed here.)

----
📚 Documentation preview 📚: https://globus-sdk-python--1169.org.readthedocs.build/en/1169/

### DIFF
--- a/src/globus_sdk/_testing/data/flows/run_flow.py
+++ b/src/globus_sdk/_testing/data/flows/run_flow.py
@@ -1,5 +1,3 @@
-from responses import matchers
-
 from globus_sdk._testing.models import RegisteredResponse, ResponseSet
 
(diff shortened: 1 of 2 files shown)


ada-globus (approved):
This has been a huge source of friction for me in the past, so happy to sign off here. I would like for us to use strict matchers when we can, but strongly concur based on my experience w/ how high friction this can be, that it should be opt-in rather than opt-out.

PR #1172, opened 2025-04-23
add `activity_notification_policy` to GCS `CollectionDocument` class.

pjhinton-globus:
This change is needed to add support for setting activity notification policies on GCS guest collections in the Globus CLI.

Add support for the `activity_notification_policy` field on the GCS `CollectionDocument` and associated it with schema version 1.14.0.

----
📚 Documentation preview 📚: https://globus-sdk-python--1172.org.readthedocs.build/en/1172/


sirosen (approved):
I think this part is easy to approve+merge. I also want to go back over the various fields we may be missing, since we haven't been keeping up with GCS.

PR #1173, opened 2025-04-23
Better support parsing a GARE from the error root

sirosen:
When parsing a GARE from an API error object, we previously would
consider the sub-errors in `err.errors` (where `err` is a
`GlobusAPIError`) but would not inspect the root of the error object.
We now parse a GARE from the root error in these situations.

Because the default parse for non-JSON:API errors populates
`err.errors` with `data["errors"]` if present, *and falls back* to
`[data]` otherwise, there's some ambiguity we need to resolve. It's
important that `to_gares`, the plural/list variant, does not create
two GAREs when there is only one, found at the root.

To make the revised code simple, the conversion of
`dict | None -> GARE | None` is separated from the main `to_gare`
interface in a minor refactor.

----
📚 Documentation preview 📚: https://globus-sdk-python--1173.org.readthedocs.build/en/1173/

### DIFF
--- a/src/globus_sdk/gare/_functional_api.py
+++ b/src/globus_sdk/gare/_functional_api.py
@@ -13,12 +13,14 @@ from ._variants import (
     LegacyConsentRequiredTransferError,
 )
 
+AnyErrorDocumentType: t.TypeAlias = (
+    "exc.GlobusAPIError | exc.ErrorSubdocument | dict[str, t.Any]"
+)
+
 log = logging.getLogger(__name__)
 
 
-def to_gare(
-    error: exc.GlobusAPIError | exc.ErrorSubdocument | dict[str, t.Any],
-) -> GARE | None:
+def to_gare(error: AnyErrorDocumentType) -> GARE | None:
     """
     Converts a GlobusAPIError, ErrorSubdocument, or dict into a
     GARE by attempting to match to GARE (preferred) or legacy variants.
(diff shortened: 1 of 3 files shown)


src/globus_sdk/gare/_functional_api.py
```diff
@@ -89,24 +84,63 @@ def to_gares(
...
         if isinstance(error, GlobusAPIError):
-            # Use the ErrorSubdocuments
-            candidate_errors.extend(error.errors)
+            # Use the ErrorSubdocuments when handling API error types
+            maybe_gares.extend(to_gare(e) for e in error.errors)
+            # Also use the root document, but only if there is an `"errors"`
+            # key inside of the error document
+            # Why? Because the *default* for `.errors` when there is no inner
+            # `"errors"` array is an array containing the root document as a
+            # subdocument
```
ada-globus, line 98: 💡

ada-globus: Mentioned this in conversation yesterday—I think we may want to revisit the use of subdocuments in parsing, as we're now blurring the GARE handling with SDK internal logic, which I think makes it harder to reason about the characteristics of the implementation here (Exhibit A: _This bug_).

I think as a follow-up, we should introduce a list of valid locations that clients must support to the GARE specification, and then implement it explicitly in the parsing here. (And find a place to publicly document the GARE spec, since this is still only officially documented internally.)

That said, this change looks correct to me and allows us to handle cases we're currently missing, so I am happy to hit approve in the interim to unblock users. 👍

ada-globus (approved):

src/globus_sdk/gare/_functional_api.py
sirosen: I was very surprised by how complex this particular scenario turned out to be.
Intuitively, it seems simple -- look at the root doc, then "look at sub-errors" -- but it turns out to be complicated by the sub-error behavior.

I think supporting GAREs in JSON:API errors will be more important for us in the future, so we should incorporate that into our planning.

PR #1174, opened 2025-04-29
Add support for `filter_roles` option when listing flows and runs

m1yag1:
https://app.shortcut.com/globus/story/40419
### Added
 - `FlowsClient.list_flows` and `FlowsClient.list_runs` now support the
   `filter_roles` parameter to filter results by one or more roles.

----
📚 Documentation preview 📚: https://globus-sdk-python--1174.org.readthedocs.build/en/1174/

### DIFF
--- a/src/globus_sdk/services/flows/client.py
+++ b/src/globus_sdk/services/flows/client.py
@@ -3,7 +3,14 @@ from __future__ import annotations
 import logging
 import typing as t
 
-from globus_sdk import GlobusHTTPResponse, GlobusSDKUsageError, client, paging, utils
+from globus_sdk import (
+    GlobusHTTPResponse,
+    GlobusSDKUsageError,
+    client,
+    exc,
+    paging,
+    utils,
+)
 from globus_sdk._types import UUIDLike
 from globus_sdk.authorizers import GlobusAuthorizer
 from globus_sdk.globus_app import GlobusApp
@@ -341,6 +348,9 @@ class FlowsClient(client.BaseClient):
         if query_params is None:
             query_params = {}
         if filter_role is not None:
+            exc.warn_deprecated(
+                "The `filter_role` parameter is deprecated. Use `filter_roles` instead."
+            )
             query_params["filter_role"] = filter_role
         if filter_roles is not None:
             query_params["filter_roles"] = utils.commajoin(filter_roles)
--- a/src/globus_sdk/services/flows/client.py
+++ b/src/globus_sdk/services/flows/client.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 import typing as t
 
-from globus_sdk import GlobusHTTPResponse, client, paging, utils
+from globus_sdk import GlobusHTTPResponse, GlobusSDKUsageError, client, paging, utils
 from globus_sdk._types import UUIDLike
 from globus_sdk.authorizers import GlobusAuthorizer
 from globus_sdk.globus_app import GlobusApp
@@ -231,6 +231,7 @@ class FlowsClient(client.BaseClient):
         self,
         *,
         filter_role: str | None = None,
+        filter_roles: str | list[str] | None = None,
         filter_fulltext: str | None = None,
         orderby: str | t.Iterable[str] | None = None,
         marker: str | None = None,
@@ -240,14 +241,18 @@ class FlowsClient(client.BaseClient):
         List deployed flows
 
         :param filter_role: A role name specifying the minimum permissions required for
-            a flow to be included in the response.
+            a flow to be included in the response. Mutually exclusive with
+            **filter_roles**.
+        :param filter_roles: A list of role names specifying the roles the user must
+            have for a flow to be included in the response. Mutually exclusive with
+            **filter_role**.
         :param filter_fulltext: A string to use in a full-text search to filter results
         :param orderby: A criterion for ordering flows in the listing
         :param marker: A marker for pagination
         :param query_params: Any additional parameters to be passed through
             as query params.
 
-        **Role Values**
+        **Filter Role Values**
 
         The valid values for ``role`` are, in order of precedence for ``filter_role``:
 
@@ -259,6 +264,16 @@ class FlowsClient(client.BaseClient):
         For example, if ``flow_starter`` is specified then flows for which the user has
         the ``flow_starter``, ``flow_administrator`` or ``flow_owner`` roles will be
         returned.
+
+        The role values ``run_manager`` and ``run_monitor`` are also supported but do
+        not follow an order of precedence.
+
+        **Filter Roles Values**
+
+        The valid values for ``filter_roles`` are the same as those for ``filter_role``.
+        However, unlike ``filter_role``, no order of precedence is evaluated. Only flows
+        where the user has one or more of the specified roles will be included in the
+        results.
 
         **OrderBy Values**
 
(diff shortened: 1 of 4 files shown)


src/globus_sdk/services/flows/client.py
```diff
@@ -260,6 +265,16 @@ def list_flows(
         the ``flow_starter``, ``flow_administrator`` or ``flow_owner`` roles will be
         returned.
 
+        The role values ``run_manager`` and ``run_monitor`` are also supported but do
+        not follow an order of precedence.
+
+        **Filter Roles Values**
+
+        The valid values for ``filter_roles`` are the same as those for ``filter_role``.
```
m1yag1, line 273: I wasn't sure how to describe the differences between the `filter_role` and `filter_roles`. I tried to avoid using the term' legacy" and wanted to treat them as separate things. The addition of the `run_managers` and `run_monitors` not being part of the precedence order also made this a bit awkward. I'm open to any suggestions on how to make this documentation better.

sirosen (changes requested):
We primarily need one minor fix: comma-joining the lists.

I've also left some notes on doc improvements and handling of empty strings and empty lists, which I think are worth our while, but the comma-joining is the only blocker I see here.

src/globus_sdk/services/flows/client.py
```diff
@@ -327,8 +342,15 @@ def list_flows(
             query_params = {}
         if filter_role is not None:
             query_params["filter_role"] = filter_role
+        if filter_roles is not None:
+            query_params["filter_roles"] = filter_roles
         if filter_fulltext is not None:
             query_params["filter_fulltext"] = filter_fulltext
+
+        if filter_role and filter_roles:
```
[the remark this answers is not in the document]
m1yag1: I updated this and added those values to the mutually exclusive test case.

tests/functional/services/flows/test_list_flows.py
```diff
@@ -109,3 +110,25 @@ def test_list_flows_orderby_multi(flows_client, orderby_style, orderby_value):
...
+
+    res = flows_client.list_flows(filter_roles=filter_roles)
+    assert res.http_status == 200
+
+    req = get_last_request()
+    assert req.body is None
+    parsed_qs = urllib.parse.parse_qs(urllib.parse.urlparse(req.url).query)
+
+    if filter_roles:
+        assert parsed_qs["filter_roles"] == filter_roles
```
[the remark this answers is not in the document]
m1yag1: I added updated changes to this test. I hope I understood correctly and attempted to add parametrized values that tested the comma-joining logic.

src/globus_sdk/services/flows/client.py
```diff
@@ -327,8 +342,15 @@ def list_flows(
             query_params = {}
         if filter_role is not None:
             query_params["filter_role"] = filter_role
+        if filter_roles is not None:
+            query_params["filter_roles"] = filter_roles
```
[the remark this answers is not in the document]
m1yag1: @sirosen I updated these usage sites with `utils.commajoin()`.

tests/functional/services/flows/test_list_runs.py
```diff
@@ -68,3 +69,37 @@ def test_list_runs_filter_flow_id(flows_client, pass_as_uuids):
...
+
+@pytest.mark.parametrize(
+    "filter_roles, expected_filter_roles",
+    [
+        # empty list
+        ([], []),
+        # list with empty string
+        ([""], [""]),
+        # None
+        (None, None),
```
sirosen, line 82: The way this was written actually taught me a bit of corner-case behavior in `requests` that I didn't know about. I didn't realize that passing an empty string resulted in a parameter being _omitted_. (Remember that we're going through `utils.commajoin`)
However, I think the way the test is written gives a different impression about what's being asserted.

I tested this rewrite locally and it does the same thing (and passes):
```suggestion
        # empty list, list with empty string, and None do not send the param
        ([], None),
        ([""], None),
        (None, None),
```

I think this is much better at communicating what the test asserts -- the variable right-hand-side values in these parameters makes it look like we're testing three different outcomes. Making them all `None` clarifies that we're testing for the same outcome.

tests/functional/services/flows/test_list_runs.py
```diff
@@ -68,3 +69,37 @@ def test_list_runs_filter_flow_id(flows_client, pass_as_uuids):
...
+def test_list_runs_filter_roles(flows_client, filter_roles, expected_filter_roles):
+    load_response(flows_client.list_runs).metadata
+    res = flows_client.list_runs(filter_roles=filter_roles)
+    assert res.http_status == 200
+
+    req = get_last_request()
+    assert req.body is None
+    parsed_qs = urllib.parse.parse_qs(urllib.parse.urlparse(req.url).query)
+
+    if expected_filter_roles and any(role for role in expected_filter_roles):
```
sirosen, line 102: The above also lets us simplify this check:
```suggestion
    if expected_filter_roles:
```

The `any()` check was what initially drew my attention -- it seemed odd but passes when run.

tests/functional/services/flows/test_list_flows.py
```diff
@@ -109,3 +110,70 @@ def test_list_flows_orderby_multi(flows_client, orderby_style, orderby_value):
...
+
+    req = get_last_request()
+    assert req.body is None
+    parsed_qs = urllib.parse.parse_qs(urllib.parse.urlparse(req.url).query)
+
+    # Only check filter_roles in parsed_qs if we expect a non-empty value
+    if expected_filter_roles and any(role for role in expected_filter_roles):
+        assert parsed_qs["filter_roles"] == expected_filter_roles
+    else:
+        assert "filter_roles" not in parsed_qs
```
sirosen, line 179: I'd like to see this test updated in a way that accords with my recommendation on the `list_runs` test, for the same reasons -- declaring multiple "expected results" which are all equivalent gives a reader the wrong impression when going through the parameters.

src/globus_sdk/services/flows/client.py
```diff
@@ -240,25 +241,39 @@ def list_flows(
         List deployed flows
 
         :param filter_role: A role name specifying the minimum permissions required for
-            a flow to be included in the response.
+            a flow to be included in the response. Mutually exclusive with
+            **filter_roles**.
```
sirosen, line 245: As a very minor tweak, I think we can call this deprecated here, as a hint to users that they should use `filter_roles` instead:
```suggestion
            **filter_roles**. (deprecated)
```

src/globus_sdk/services/flows/client.py
```diff
@@ -327,8 +342,15 @@ def list_flows(
             query_params = {}
         if filter_role is not None:
             query_params["filter_role"] = filter_role
```
sirosen, line 344: We discussed adding a deprecation warning here (we have tooling for this, use the `warn_deprecated` helper), which I think we're on solid ground to include. It's not clear when this is being removed from the API, but adding the warning gives us better grounds to remove the parameter in v4 and have a simpler SDK interface.

sirosen (changes requested):

tests/functional/services/flows/test_list_flows.py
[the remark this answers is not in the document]
sirosen: Yeah, I'm very happy with the new tests! I have an adjustment I'd like us to make -- noted in a review -- but the initial parametrization put us on good footing to discuss that.

src/globus_sdk/services/flows/client.py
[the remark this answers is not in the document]
sirosen: Yep, I think the `src/` tree stuff is now all good on this account!

src/globus_sdk/services/flows/client.py
sirosen: Thanks! I tried to give what I thought was a pretty good suggestion, so I'm happy with you taking it. :grin:

src/globus_sdk/services/flows/client.py
m1yag1: I applied this and it was in the wrong place (filter_roles) but i fixed it 🙃.

PR #1176, opened 2025-05-02
Minor grammar touchup

khk-globus:
"is distinguishes", and then remove a less-formal "you'll" in favor of a simple conditioned directive.  Also fix-up example to use a naked `raise`.

----
📚 Documentation preview 📚: https://globus-sdk-python--1176.org.readthedocs.build/en/1176/


sirosen (approved):
Yep, thanks for this!

I bet this doc survives all the way from v1.x or even v0.x days. But you can teach an old doc new tricks. :wink:

PR #1177, opened 2025-05-02
Remove unnecessary inline import

khk-globus:
----
📚 Documentation preview 📚: https://globus-sdk-python--1177.org.readthedocs.build/en/1177/


sirosen (approved):

sirosen:
Kevin and I talked about this very briefly. We don't really know why that extra import was there, but probably it was part of some older deferred import work and didn't survive our rounds of refactoring.

PR #1179, opened 2025-05-02
Support `session_message` in authorization URL

ada-globus:
What it says on tin!

You can trivially test with:

```python
from globus_sdk import UserApp
from globus_sdk.gare import GlobusAuthorizationParameters

CLIENT_ID = "[YOUR CLIENT ID HERE]"

app = UserApp("sdk-session-message-test-app", client_id=CLIENT_ID)
auth_params = GlobusAuthorizationParameters(
    session_message="This is a test message",
    prompt="login",
)
app.login(auth_params=auth_params)
```

Then look for it in the generated URL. I clicked through to verify this is honored by Auth.

Screenshot 2025-05-02 at 11 19 41

----
📚 Documentation preview 📚: https://globus-sdk-python--1179.org.readthedocs.build/en/1179/

### DIFF
--- a/changelog.d/20250502_103545_ada_support_session_message_authorize_url.rst
+++ b/changelog.d/20250502_103545_ada_support_session_message_authorize_url.rst
@@ -1,0 +1,7 @@
+Added
+~~~~~
+
+- ``AuthLoginClient`` now supports a ``session_message`` when constructing an
+  OAuth2 authorization URL. (:pr:`NUMBER`)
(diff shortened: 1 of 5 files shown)


sirosen (approved):
I almost want to treat this as a bugfix (in terms of changelog labeling), but when I look at the individual parts I get why you labeled them `Added`.

Everything here looks good to me. Let's get this into the next release!

PR #1185, opened 2025-05-09
Remove base_path from all client classes

MaxTueckeGlobus:
[sc-26346](https://app.shortcut.com/globus/story/26346/sdk-remove-base-path-from-client-classes)

### Changes
- All `base_path` declarations have been removed from all client classes.
- All methods on a client class that call an API for the corresponding service have been updated to include the base path explicitly in their path string.

**Reviewer notes:**
- Much of the test suite broke when removing `base_path` since many tests were not including the `base_path` for a service when creating a `RegisteredResponse`. Rather than update each individual test, I chose to [update](https://github.com/globus/globus-sdk-python/blob/c05f5bd8484d973dbf3bc6f8c44da4672359f1e0/src/globus_sdk/_testing/models.py#L108-L114) `RegisteredResponse` to always add a particular services `base_path` to the path if it was not included. 
- To preserve the previous behavior of `BaseClient`, `BaseClient.base_url` will always have a trailing slash added if needed.

----
📚 Documentation preview 📚: https://globus-sdk-python--1185.org.readthedocs.build/en/1185/

### DIFF
--- a/changelog.d/20250512_144528_max.tuecke_sc_26346_remove_base_path_from_clients.rst
+++ b/changelog.d/20250512_144528_max.tuecke_sc_26346_remove_base_path_from_clients.rst
@@ -1,5 +1,5 @@
 Removed
 ~~~~~~~
 
-- SDK clients no longer define a ``base_path`` attribute which they prefix to paths.
+- SDK client classes no longer define nor prepend a ``base_path`` attribute which they prefix to paths.
   Make sure to use the full path now when using client methods. (:pr:`1185`)
--- a/changelog.d/20250512_144528_max.tuecke_sc_26346_remove_base_path_from_clients.rst
+++ b/changelog.d/20250512_144528_max.tuecke_sc_26346_remove_base_path_from_clients.rst
@@ -1,0 +1,5 @@
+Removed
+~~~~~~~
+
+- SDK clients no longer define a ``base_path`` attribute which they prefix to paths.
+  Make sure to use the full path now when using client methods. (:pr:`1185`)
--- a/src/globus_sdk/_testing/models.py
+++ b/src/globus_sdk/_testing/models.py
@@ -39,15 +39,11 @@ class RegisteredResponse:
         "nexus": "https://nexus.api.globusonline.org/",
         "transfer": "https://transfer.api.globus.org/",
         "search": "https://search.api.globus.org/",
-        "gcs": "https://abc.xyz.data.globus.org/api",
+        "gcs": "https://abc.xyz.data.globus.org/api/",
         "groups": "https://groups.api.globus.org/",
         "timer": "https://timer.automate.globus.org/",
         "flows": "https://flows.automate.globus.org/",
         "compute": "https://compute.api.globus.org/",
-    }
-    _path_prefix_map = {
-        "transfer": "/v0.10/",
-        "groups": "/v2/",
     }
 
     def __init__(
@@ -105,12 +101,6 @@ class RegisteredResponse:
         self.service = service
 
         if service:
-            # ensure a registered response path is prefixed by _path_prefix_map
-            # this allows a registered response to use a path like `/groups` with
-            # the GroupsClient, rather than *requiring* it use the full `/v2/groups`
-            path_prefix = self._path_prefix_map.get(service)
-            if path_prefix and not path.startswith(path_prefix):
-                path = slash_join(path_prefix, path)
             self.full_url = slash_join(self._url_map[service], path)
         else:
             self.full_url = path
--- a/src/globus_sdk/_testing/models.py
+++ b/src/globus_sdk/_testing/models.py
@@ -37,15 +37,15 @@ class RegisteredResponse:
     _url_map = {
         "auth": "https://auth.globus.org/",
         "nexus": "https://nexus.api.globusonline.org/",
-        "transfer": "https://transfer.api.globus.org/v0.10",
+        "transfer": "https://transfer.api.globus.org/",
         "search": "https://search.api.globus.org/",
         "gcs": "https://abc.xyz.data.globus.org/api",
-        "groups": "https://groups.api.globus.org/v2/",
+        "groups": "https://groups.api.globus.org/",
         "timer": "https://timer.automate.globus.org/",
         "flows": "https://flows.automate.globus.org/",
         "compute": "https://compute.api.globus.org/",
     }
-    _base_path_map = {
+    _path_prefix_map = {
         "transfer": "/v0.10/",
         "groups": "/v2/",
     }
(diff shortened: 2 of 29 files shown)


src/globus_sdk/client.py
```diff
@@ -105,8 +99,9 @@ def __init__(
 
         # resolve the base_url for the client (see docstring for resolution precedence)
         self.base_url = self._resolve_base_url(base_url, self.environment)
-        # append the base_path to the base_url if necessary
-        self.base_url = utils.slash_join(self.base_url, self.base_path)
+
+        # ensure base_url always has a trailing slash
+        self.base_url = utils.slash_join(self.base_url, "/")
```
sirosen, line 104: I think we should talk about whether or not this is desirable. There are a number of tests which check for the trailing slash incidentally, but it's not clear if any are intentionally trying to apply this constraint.

Since we always `slash_join` input paths with the base URL as part of the HTTP methods, I don't think there's anything gained by ensuring that the base_url always has a trailing slash.

I'm not very strongly against this, but I wouldn't want to keep it just as a historical artifact. I propose removing it, on the basis that it's not necessary anymore. (And this requires adjusting the few tests which this impacts.)

src/globus_sdk/_testing/models.py
```diff
@@ -37,15 +37,15 @@ class RegisteredResponse:
...
+        "groups": "https://groups.api.globus.org/",
         "timer": "https://timer.automate.globus.org/",
         "flows": "https://flows.automate.globus.org/",
         "compute": "https://compute.api.globus.org/",
     }
-    _base_path_map = {
+    _path_prefix_map = {
         "transfer": "/v0.10/",
         "groups": "/v2/",
     }
```
sirosen, line 51: I think we should, as part of getting rid of base path, also remove this mapping. This will require updating a bunch of fixtures, across a couple dozen files, but I think it fits in as part of the removal.

sirosen (changes requested):
I have a couple of minor adjustments I'd like to make. One is definitely debatable (trailing slash). Overall, excited for getting this into v4!

src/globus_sdk/client.py
MaxTueckeGlobus: I don't think there is a particularly strong argument either for or against it. I included it since that was the previous behavior and as you mentioned, there are a good amount of tests doing that check. Functionally, AFAICT, there is no reason to keep it though. If I'm already updating a number of tests to fully get rid of `_base_path_map` (as mentioned below), then I'm inclined to also remove that constraint and update any tests that are doing that check. Thoughts?

src/globus_sdk/_testing/models.py
MaxTueckeGlobus: I agree, I just wasn't sure if updating all the tests/fixtures was indented for this change. I'll go ahead and start working though the tests that need fixing.

src/globus_sdk/client.py
sirosen: Primarily, +1 to what you're saying; I'd like you to try to remove it. If it proves surprisingly difficult we can reassess.

I think you'll find that when you check which tests are broken by this, most just need to be updated to not expect a trailing slash -- they're testing other things and incidentally trip over this detail. There's one case in one GCS test which looks like it's testing this directly, which you can just remove.

sirosen:
The updated changes look good, but we need a changelog for this.
Since it's going into v4, I'd phrase this as a "removed" item. e.g.,

```rst
Removed
~~~~~~~

- SDK clients no longer define a ``base_path`` attribute which they prefix to paths.
  Make sure to use the full path when using client methods. (:pr:`1185`)
```

sirosen:
I don't think we need any special naming for the changelog files or contents -- by being in the `4.x-dev` branch they'll be isolated from any changes made in 3.x/`main`.

We should fix the changelog checking workflow, but I don't think it's a blocker right now.

changelog.d/20250512_144528_max.tuecke_sc_26346_remove_base_path_from_clients.rst
```diff
@@ -0,0 +1,5 @@
+Removed
+~~~~~~~
+
+- SDK clients no longer define a ``base_path`` attribute which they prefix to paths.
```
kurtmckee, line 4: I would rephrase this to make it clear that the `base_path` isn't just "not defined", but in fact the prefixing logic has been completely removed.

src/globus_sdk/_testing/models.py
kurtmckee: @MaxTueckeGlobus If you haven't started on that yet, I recommend making that a separate PR. It's related, but it means that this very easy-to-read PR is doing a lot more.

kurtmckee (approved):
Looks good. I left feedback regarding the changelog fragment phrasing but approve of the changes already here.

As noted in a comment, I don't think that more work needs to be done in this particular PR, which is well-targeted and sized.

changelog.d/20250512_144528_max.tuecke_sc_26346_remove_base_path_from_clients.rst
sirosen: I think that was the phrasing I suggested -- so I want to be explicit that I'm happy for us to rephrase if this seems unclear.
Would you rather see two bullet points, rather than two sentences?

```rst
Removed
~~~~~~~

- SDK clients no longer define a ``base_path`` attribute. (:pr:`1185`)
- SDK clients no longer prepend a ``base_path`` to request paths when ``get``, ``put``, ``post``, ``patch``, ``delete``, and ``request`` are called. (:pr:`1185`)
```

src/globus_sdk/_testing/models.py
sirosen: I'm unsure if there was a time-splice here, but the PR was updated per this thread a couple of hours before that last comment. I think the size of the fixture update was actually pretty modest -- several files, but almost entirely confined to `src/globus_sdk/_testing/data/`. Just flagging this in case.

sirosen (approved):
I'm happy with this, but recommend resolving the changelog thread before merging -- I think we could also merge and fix the changelog in follow-up, potentially along with an upgrading doc section.

changelog.d/20250512_144528_max.tuecke_sc_26346_remove_base_path_from_clients.rst
kurtmckee: I was thinking along the lines "SDK client classes no longer define nor prepend a ``base_path`` blah blah blah". Would that be sufficient?

Two sentences seems like more than what's necessary, but that's just how I'm currently thinking...

src/globus_sdk/_testing/models.py
kurtmckee: The time-splice was simply me not being able to read. I think this can be marked as resolved! :+1:

changelog.d/20250512_144528_max.tuecke_sc_26346_remove_base_path_from_clients.rst
sirosen: :+1: for "no longer define nor prepend"

PR #1186, opened 2025-05-12
Remove the "default requested scopes"

sirosen:
[[sc-20185]](https://app.shortcut.com/globus/story/20185/remove-default-scopes-from-the-sdk)

These were the default for login flows and client credentials. When GlobusApp-driven logins are used, these defaults are not used. They were only applied for direct use of the login flow and token request mechanisms for login clients.

One log line has been changed to produce slightly more useful debug-level logs.

Also, introduce the first v3.x -> v4 upgrading section to the upgrading doc.

----
📚 Documentation preview 📚: https://globus-sdk-python--1186.org.readthedocs.build/en/1186/

### DIFF
--- a/changelog.d/20250512_143605_sirosen_remove_default_scopes.rst
+++ b/changelog.d/20250512_143605_sirosen_remove_default_scopes.rst
@@ -1,0 +1,8 @@
+Removed
+~~~~~~~
+
+- The SDK no longer sets default scopes for direct use of client
+  credentials and auth client login flow methods. Users should either use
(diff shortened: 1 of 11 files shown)


changelog.d/20250512_143605_sirosen_remove_default_scopes.rst
```diff
@@ -0,0 +1,8 @@
+Removed
```
kurtmckee, line 1: Based on my past experiences, I recommend that we create a `Breaking changes` section and capture changes like these in that section (even if via duplication of the bullet item's text).

PR #1189, opened 2025-05-13
Updated CollectionDocument with MissingType

JasonAlt:
I'm working on clearing out code that's duplicated between the GCS CLI and the SDK. I thought I'd take a stab at updating `MappedCollectionDocument`/`GuestCollectionDocument` with `MissingType` instead of `None`. If this is useful, I can do the same with the role and gateway document classes.

This may constitute a breaking change since current callers supplying `None` now wind up with a `None` value in the document (for several fields this is a valid value).

I'm also not sure that the additional tests add any real value. It was nice as a sanity check to make sure I didn't make a typo (which I did!), but beyond that, /shrug

Justification for change:
* consistency between GCS document types
* `None` is a valid value for some fields. Passing `None` in `__init__()` (as opposed to doc[field]=None) ensures the DATA_TYPE gets set correctly
* simplifies the GCS CLI's use of different document types

----
📚 Documentation preview 📚: https://globus-sdk-python--1189.org.readthedocs.build/en/1189/

### DIFF
--- a/changelog.d/20250513_182417_jasonalt.rst
+++ b/changelog.d/20250513_182417_jasonalt.rst
@@ -1,0 +1,53 @@
+..
+.. A new scriv changelog fragment
+..
+.. Uncomment the header that is right (remove the leading dots).
+..
+.. Leave the "(:pr:`...`)" text in your change description.
+.. GitHub Actions will automatically replace it when the PR is merged.
+..
+.. Python Support
+.. ~~~~~~~~~~~~~~
+..
+.. - A bullet item for the Python Support category. (:pr:`NUMBER`)
+..
+.. Added
+.. ~~~~~
+..
+.. - A bullet item for the Added category. (:pr:`NUMBER`)
+..
+.. Removed
+.. ~~~~~~~
+..
+.. - A bullet item for the Removed category. (:pr:`NUMBER`)
+..
+Changed
+~~~~~~~
+
+- Updated MappedCollectionDoc and GuestCollectionDoc with MissingType. (:pr:`NUMBER`)
+
+.. Deprecated
+.. ~~~~~~~~~~
+..
+.. - A bullet item for the Deprecated category. (:pr:`NUMBER`)
+..
+.. Fixed
+.. ~~~~~
+..
+.. - A bullet item for the Fixed category. (:pr:`NUMBER`)
+..
+.. Documentation
+.. ~~~~~~~~~~~~~
+..
+.. - A bullet item for the Documentation category. (:pr:`NUMBER`)
+..
+.. Security
+.. ~~~~~~~~
+..
+.. - A bullet item for the Security category. (:pr:`NUMBER`)
+..
+.. Development
+.. ~~~~~~~~~~~
+..
+.. - A bullet item for the Development category. (:pr:`NUMBER`)
+..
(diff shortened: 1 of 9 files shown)


JasonAlt:
Forced-pushed with changelog

kurtmckee:
@MaxTueckeGlobus, you and Jason may be doing some similar work.

sirosen:
It's also notable that we're working on this as part of the `4.x-dev` branch, and plan to work through getting pretty much everything onto the `MISSING` paradigm as part of 4.0. This PR fits that goal like a glove, but I second the suggestion above to check in with Max and see how this could potentially dovetail so that we're all in sync.

If 4.0 were further off, I'd be more open to discussing whether or not this is an acceptable level of incompatibility for a feature release. But since we're actively working on 4.0 right now, I feel like we don't need to have that debate!

kurtmckee:
I anticipate that the base branch will need to change to the 4.x dev branch that all stories targeting 4.x changes are merging into.

MaxTueckeGlobus:
Sorry I'm a little late here, but perfect timing on this, I was about to start looking at adding `MissingType` to the GCS client today! As mentioned above, we are planning on making these changes for everything as part of SDKv4. Could you go ahead and change the base branch to `4.x-dev` and I'll work around your changes on my end?

kurtmckee:
I changed the base branch to `4.x-dev`, but that introduced some additional commits that will make the review harder. I recommend that we review and approve, _and then change the base branch before merging_.

changelog.d/20250513_182417_jasonalt.rst
```diff
@@ -0,0 +1,53 @@
...
+..
+.. - A bullet item for the Added category. (:pr:`NUMBER`)
+..
+.. Removed
+.. ~~~~~~~
+..
+.. - A bullet item for the Removed category. (:pr:`NUMBER`)
+..
+Changed
+~~~~~~~
```
kurtmckee, line 25: As we march to v4, a new header has been introduced recently: "Breaking Changes". This helps call out the changes.

```suggestion
Breaking Changes
~~~~~~~~~~~~~~~~
```

This might not need to be addressed in this PR because Max is converting EVERYTHING to use `MISSING` and could potentially remove/update your changelog fragment, but I wanted to call it out.

kurtmckee (approved):

JasonAlt:
Whatever you guys think, I can rebase on whatever works.

sirosen (approved):
Setting aside meta-issues and looking at the changes, this seems great, and I appreciate the included test matrix!
Maybe those "set optional thing" helpers will disappear as part of v4, based on what this looks like. That would be cool.

I think what Kurt was saying is that we can easily merge -- no need to rebase -- as long as we do our review before we change the base branch. And that changing the base branch made the compare view wonky, but wasn't actually doing _harm_ to the merge semantics.

Anyway, I'm leaving a clear approval here; @kurtmckee, are you okay handling the merge (and maybe rebranding the changelog item)? Since you were looking at this when trying to change the base, I think you have the best grasp of what might happen here / needs doing.

kurtmckee:
I'll handle the changelog updates.

Since we're planning to cut alpha releases I'll make changes to the final changelog when prepping the v4.0.0a1 release.

PR #1194, opened 2025-05-20
Remove the badges at the top of the README

kurtmckee:
Documentation
-------------

- Remove the badges at the top of the README.

-----

Because the README is frozen in time when published to PyPI, the badges represent a risk of aging poorly:

> image

This PR removes the badges to reduce clutter ("sorted with isort") and to reduce the risk of breaking over time ("build status").

----
📚 Documentation preview 📚: https://globus-sdk-python--1194.org.readthedocs.build/en/1194/


sirosen (approved):
Yep; we've discussed this and I still think some of the badges (namely, version related ones) offer _some_ value, but it's marginal at best.

PR #1195, opened 2025-05-20
Remove the `globus_sdk.version` module

kurtmckee:
Breaking Changes
----------------

- The SDK version is no longer available in ``globus_sdk.version.__version__``.

  Packages that want to query the SDK version must use ``importlib.metadata``:

  ```python
  import importlib.metadata

  GLOBUS_SDK_VERSION = importlib.metadata.distribution("globus_sdk").version
  ```

----

This PR aligns the SDK's versioning and documentation with current best practice, which is to use `importlib.metadata` to query for the installed package version. This technique works universally for all packages and, bonus, avoids executing package code and whatever side effects may be associated with that execution.


sirosen (approved):
♥️ Thank you for this! It's been on my to-do list for ages; let's get it done

PR #1199, opened 2025-05-28
Reduce cognitive complexity of the scope parser

sirosen:
This effort was based on analysis using `complexipy` to identify difficult-to-maintain codepaths.
The scope parser stands out as having dramatically higher complexity than the rest of the codebase, which makes it interesting to examine and update.

The updated version is at least arguably, if not definitively, more maintainable, at the cost of some micro-optimizations which dodge function-call overheads.

## Performance

Using the benchmark script, we see the following degradation in performance:

1. `depth=10`

- original: 2.9798308038152753e-05 seconds
- updated:  4.319636104628444e-05 seconds

Therefore, we're ~30% slower on small scopes, but still take fractions of a millisecond to parse.

2. `depth=100`

- original: 0.00031491286400705577 seconds
- updated:  0.00044869115098845213 seconds

This maintains the 30% perf degradation.

3. `depth=5000`

- original: 0.1826306022098288 seconds
- updated:  0.19780024648993277 seconds

Interestingly, as the scope becomes arbitrarily deeply nested, the perf loss shrinks.
Somehow, the overheads of calling more functions disappear the more times those functions are called.
Perhaps sets or strings are being interned more effectively as the program runs, or CPython is finding other ways to optimize this "hot path".

Therefore, at worst we've gotten a 30% perf loss in already *very fast* (for Python) code.

## Cognitive Complexity / Maintainability

`complexipy` is a cognitive complexity measurement tool.

It presents a nice chart of scores, but for comparison against `main` we just want the raw data.

Getting the 3 "worst" functions in the scope parser:
```
$ uvx complexipy src/globus_sdk/scopes/_parser.py -q -o > /dev/null; tail -n 3 complexipy.csv
_parser.py,_parser.py,ScopeGraph::_check_cycles,8
_parser.py,_parser.py,_tokenize,8
_parser.py,_parser.py,_parse_tokens,10
```

(scores are 8, 8, 10 above)

And getting the 3 "worst" functions in the src tree:
```
$ uvx complexipy src -q -o > /dev/null; tail -n 3 complexipy.csv
src/globus_sdk/transport/requests.py,requests.py,RequestsTransport::request,12
src/globus_sdk/_lazy_import.py,_lazy_import.py,_ParsedPYIData::_load_statement,13
src/globus_sdk/services/transfer/client.py,client.py,TransferClient::endpoint_manager_task_list,15
```

The above are taken after these changes.
Compare against scores before the change:
```
$ uvx complexipy src -q -o > /dev/null; tail -n 3 complexipy.csv
src/globus_sdk/services/transfer/client.py,client.py,TransferClient::endpoint_manager_task_list,15
src/globus_sdk/scopes/_parser.py,_parser.py,_tokenize,22
src/globus_sdk/scopes/_parser.py,_parser.py,_parse_tokens,23
```

We can see that `_tokenize()` was scored at 22, and `_parse_tokens()` at 23.

### Cognitive Complexity Scores are Only a Metric!

Obviously, a metric is only as good as what it actually tells us.

I tried cognitive complexity analysis on our codebase to see if it could find any maintenance risks by showing us outliers.

My *opinion* after examining what it showed is that the high cognitive complexity functions, `_tokenize` and `_parse_tokens` are good candidates for revision and are hard to maintain.

The metric is easy to confuse with an actual *assessment* of maintainability, but should be carefully distinguished from that.
I've made changes which lower the metric, and then reassessed -- "Is this code more readable?" -- and believe that the changes are net-good.

----
📚 Documentation preview 📚: https://globus-sdk-python--1199.org.readthedocs.build/en/1199/

### DIFF
--- a/src/globus_sdk/scopes/_parser.py
+++ b/src/globus_sdk/scopes/_parser.py
@@ -21,7 +21,7 @@ def _tokenize(scope_string: str) -> list[str]:
             tokens.append(scope_string[start:idx])
         start = idx + 1
 
-        if current_char in ("*", "[", "]"):
+        if current_char in SPECIAL_TOKENS:
             tokens.append(current_char)
         elif current_char == " ":
             pass
--- a/src/globus_sdk/scopes/_parser.py
+++ b/src/globus_sdk/scopes/_parser.py
@@ -12,36 +12,50 @@ SPECIAL_TOKENS = set("[]*")
 def _tokenize(scope_string: str) -> list[str]:
     tokens: list[str] = []
     start = 0
-    for idx, c in enumerate(scope_string):
-        try:
-            peek: str | None = scope_string[idx + 1]
-        except IndexError:
-            peek = None
+    for idx, c, peek in _peek_enumerate(scope_string):
+        if c not in SPECIAL_CHARACTERS:
+            continue
+        _reject_bad_adjacent_pairs(c, peek)
 
-        if c in SPECIAL_CHARACTERS:
-            if start != idx:
-                tokens.append(scope_string[start:idx])
+        if start != idx:
+            tokens.append(scope_string[start:idx])
 
-            start = idx + 1
-            if c == "*":
-                if peek == " ":
-                    raise ScopeParseError("'*' must not be followed by a space")
-                tokens.append(c)
-            elif c == "[":
-                tokens.append(c)
-            elif c == "]":
-                if peek is not None and peek not in (" ", "]"):
-                    raise ScopeParseError("']' may only be followed by a space or ']'")
-                tokens.append(c)
-            elif c == " ":
-                if peek == "[":
-                    raise ScopeParseError("'[' cannot have a preceding space")
-            else:
-                raise NotImplementedError
+        start = idx + 1
+        if c in ("*", "[", "]"):
+            tokens.append(c)
+        elif c == " ":
+            pass
+        else:
+            raise NotImplementedError
     remainder = scope_string[start:].strip()
     if remainder:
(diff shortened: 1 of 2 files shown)


src/globus_sdk/scopes/_parser.py
```diff
@@ -233,6 +239,24 @@ def _convert_trees(trees: list[ScopeTreeNode]) -> ScopeGraph:
...
+    """
+    if not data:
+        return
+
+    prev: str = data[0]
+    for idx, c in enumerate(data[1:]):
+        yield (idx, prev, c)
+        prev = c
+
+    yield (len(data) - 1, prev, None)
```
kurtmckee, line 257: It's gonna be so nice when Python 3.10 is the lowest supported version and we can use `itertools.pairwise()`!

src/globus_sdk/scopes/_parser.py
```diff
@@ -12,38 +12,41 @@
...
+    for idx, current_char, next_char in _peek_enumerate(scope_string):
+        if current_char not in SPECIAL_CHARACTERS:
+            continue
+        _reject_bad_adjacent_characters(current_char, next_char)
+
+        if start != idx:
+            tokens.append(scope_string[start:idx])
+        start = idx + 1
+
+        if current_char in ("*", "[", "]"):
```
kurtmckee, line 24:
```suggestion
        if current_char in SPECIAL_TOKENS:
```

kurtmckee (changes requested):

kurtmckee (approved):

PR #1200, opened 2025-05-29
Add an entry to the upgrading doc for MutableScope

sirosen:
I intentionally omitted any doc update when removing `MutableScope`, but have reconsidered.
I think a minimal stub is still worth adding, and we can enhance it later.

---

This is a small/minimal entry so that we're guaranteed to have something to build upon.

It intentionally does not go into detail on how to use `Scope` because these interfaces are being redesigned.

### DIFF
--- a/docs/upgrading.rst
+++ b/docs/upgrading.rst
@@ -45,6 +45,31 @@ Then, code can dispatch with
 From 3.x to 4.0
 ---------------
 
+``MutableScope`` is Removed, use ``Scope`` Instead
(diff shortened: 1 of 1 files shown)


sirosen:
I didn't get nice behavior from `:class:` without an explicit reference, but I agree it makes a better resulting doc, so I wrote out the full thing to make it work.

kurtmckee (approved):

kurtmckee:
Should this merge to `4.x-dev` if it's an upgrading guide that can be applied now?

PR #1206, opened 2025-06-03
Remove deprecated aliases for "Timers"

sirosen:
Several objects named `Timer*` were exposed in the SDK as deprecated
aliases for their `Timers*` renamed variants. These aliases are here
removed.

Also, delete the `globus_sdk.services.timer` module which aliased
them.

Tests and docs are updated to remove all reference to these old names.

----
📚 Documentation preview 📚: https://globus-sdk-python--1206.org.readthedocs.build/en/1206/


docs/services/timer.rst
kurtmckee, line 1: This file's name should also change.

kurtmckee (changes requested):

kurtmckee (approved):

PR #1207, opened 2025-06-03
Update search client classes to use MISSING defaults

MaxTueckeGlobus:
[sc-15807](https://app.shortcut.com/globus/story/15807/sdkv4-convert-defaults-of-none-to-globus-sdk-missing-for-payload-types-allowing-explicit-null-by-setting-none)

This PR is for SDKv4 and is a followup to #1205.

### Changes
 - Converted defaults of `None` to `globus_sdk.MISSING` for payload types in the search client.
 - Updated search unit tests to test with `MISSING` defaults instead of `None`.

[Search OpenAPI spec](https://search.api.globus.org/autodoc/redoc) for reference.

### DIFF
--- a/src/globus_sdk/services/search/client.py
+++ b/src/globus_sdk/services/search/client.py
@@ -7,6 +7,7 @@ from globus_sdk import client, paging, response, utils
 from globus_sdk._types import UUIDLike
 from globus_sdk.exc.warnings import warn_deprecated
 from globus_sdk.scopes import Scope, SearchScopes
+from globus_sdk.utils import MISSING, MissingType, filter_missing
 
 from .data import SearchQuery, SearchScrollQuery
 from .errors import SearchAPIError
@@ -278,17 +279,13 @@ class SearchClient(client.BaseClient):
 
                 .. expandtestfixture:: search.search
         """  # noqa: E501
-        if query_params is None:
-            query_params = {}
-        query_params.update(
-            {
-                "q": q,
-                "offset": offset,
-                "limit": limit,
-                "advanced": advanced,
-            }
-        )
-
+        query_params = {
+            **(query_params or {}),
+            "q": q,
+            "offset": offset,
+            "limit": limit,
+            "advanced": advanced,
+        }
         log.debug(f"SearchClient.search({index_id}, ...)")
         return self.get(f"/v1/index/{index_id}/search", query_params=query_params)
 
@@ -304,8 +301,8 @@ class SearchClient(client.BaseClient):
         index_id: UUIDLike,
         data: dict[str, t.Any] | SearchQuery,
         *,
-        offset: int | None = None,
-        limit: int | None = None,
+        offset: int | MissingType = MISSING,
+        limit: int | MissingType = MISSING,
     ) -> response.GlobusHTTPResponse:
         """
         Execute a complex Search Query, using a query document to express filters,
@@ -359,13 +356,16 @@ class SearchClient(client.BaseClient):
                     :ref: search/reference/post_query/
         """
         log.debug(f"SearchClient.post_search({index_id}, ...)")
-        add_kwargs = {}
-        if offset is not None:
-            add_kwargs["offset"] = offset
-        if limit is not None:
-            add_kwargs["limit"] = limit
-        if add_kwargs:
-            data = {**data, **add_kwargs}
+        add_kwargs = filter_missing(
+            {
+                "offset": offset,
+                "limit": limit,
+            }
+        )
+        data = {
+            **data,
+            **add_kwargs,
+        }
         return self.post(f"v1/index/{index_id}/search", data=data)
 
     @paging.has_paginator(paging.MarkerPaginator, items_key="gmeta")
@@ -374,7 +374,7 @@ class SearchClient(client.BaseClient):
         index_id: UUIDLike,
         data: dict[str, t.Any] | SearchScrollQuery,
         *,
-        marker: str | None = None,
+        marker: str | MissingType = MISSING,
     ) -> response.GlobusHTTPResponse:
         """
         Scroll all data in a Search index. The paginated version of this API should
@@ -411,11 +411,15 @@ class SearchClient(client.BaseClient):
                     :ref: search/reference/scroll_query/
         """
         log.debug(f"SearchClient.scroll({index_id}, ...)")
-        add_kwargs = {}
-        if marker is not None:
-            add_kwargs["marker"] = marker
-        if add_kwargs:
-            data = {**data, **add_kwargs}
+        add_kwargs = filter_missing(
+            {
+                "marker": marker,
+            }
+        )
+        data = {
+            **data,
+            **add_kwargs,
+        }
         return self.post(f"v1/index/{index_id}/scroll", data=data)
 
     #
@@ -579,9 +583,10 @@ class SearchClient(client.BaseClient):
         # convert the provided subjects to a list and use the "safe iter" helper to
         # ensure that a single string is *not* treated as an iterable of strings,
         # which is usually not intentional
-        body = {"subjects": list(utils.safe_strseq_iter(subjects))}
-        if additional_params:
-            body.update(additional_params)
+        body = {
+            "subjects": list(utils.safe_strseq_iter(subjects)),
+            **(additional_params or {}),
+        }
         return self.post(f"/v1/index/{index_id}/batch_delete_by_subject", data=body)
 
     #
@@ -621,10 +626,11 @@ class SearchClient(client.BaseClient):
                 .. extdoclink:: Get By Subject
                     :ref: search/reference/get_subject/
         """
-        if query_params is None:
-            query_params = {}
-        query_params["subject"] = subject
         log.debug(f"SearchClient.get_subject({index_id}, {subject}, ...)")
+        query_params = {
+            **(query_params or {}),
+            "subject": subject,
+        }
         return self.get(f"/v1/index/{index_id}/subject", query_params=query_params)
 
     def delete_subject(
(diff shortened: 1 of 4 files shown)


changelog.d/20250603_130159_max.tuecke_sc_15807_search_missing_defaults.rst
```diff
@@ -0,0 +1,4 @@
+Breaking Changes
+~~~~~~~~~~~~~~~~
+
+- All defaults of None converted to globus_sdk.MISSING for all payload types in the search client. (:pr:`1207`)
```
kurtmckee, line 4:
```suggestion
- All defaults of ``None`` converted to ``globus_sdk.MISSING`` for all payload types in the Search client. (:pr:`1207`)
```

src/globus_sdk/services/search/client.py
```diff
@@ -359,13 +356,16 @@ def post_search(
...
+        add_kwargs = filter_missing(
+            {
+                "offset": offset,
+                "limit": limit,
+            }
+        )
+        data = {
+            **data,
+            **add_kwargs,
+        }
```
kurtmckee, line 368: I was under the impression that `data` would have `MISSING` filtered already, so this would be implicitly the same as:

```suggestion
        data = {
            "offset": offset,
            "limit": limit,
            **data,
        }
```

Note that I've swapped put `data` second, so that if a user puts `offset` or `limit` into their own payload, it will overwrite the default `MISSING` value.

If that's accurate, I think other places in the code can be updated/simplified in a similar way.

src/globus_sdk/services/search/client.py
```diff
@@ -278,17 +279,13 @@ def search(
...
-            }
-        )
-
+        query_params = {
+            **(query_params or {}),
+            "q": q,
+            "offset": offset,
+            "limit": limit,
+            "advanced": advanced,
+        }
```
kurtmckee, line 288: I think that this should be reversed so that users can put all of their parameters into `query_params` without them getting overwritten.

```suggestion
        query_params = {
            "q": q,
            "offset": offset,
            "limit": limit,
            "advanced": advanced,
            **(query_params or {}),
        }
```

I'm thinking of this use case:

```python
fn(query_params={"offset": 33, "limit": 77})  # called with offset=0, limit=10
```

This feedback may apply to other areas of the code, in the specific case that we're providing default values that could overwrite a user's `query_params` values.

src/globus_sdk/services/search/data.py
```diff
@@ -215,9 +212,11 @@ def add_sort(
...
         """
         self["sort"] = self.get("sort", [])
-        sort = {"field_name": field_name, **(additional_fields or {})}
-        if order is not None:
-            sort["order"] = order
+        sort = {
+            "field_name": field_name,
+            "order": order,
+            **(additional_fields or {}),
+        }
```
kurtmckee, line 219: This is an interesting case -- `field_name` is a mandatory method parameter, but `order` is not. Therefore, `order` could be overwritten by `additional_fields`, but `field_name` probably should not be.

```suggestion
        sort = {
            "order": order,
            **(additional_fields or {}),
            "field_name": field_name,
        }
```

kurtmckee (changes requested):

src/globus_sdk/services/search/data.py
sirosen: I'd argue in favor of the current proposed version on this one (`**` expansion at the end).

`query.add_sort("foo", additional_fields={"field_name": "bar"})` is an inherently ambiguous request. I don't think we need to guarantee some well-defined behavior in such a case. I think such usage should be -- intentionally -- undefined in behavior, and at best consistent and deterministic on a given SDK version.

So I like putting `**` expansion at the end as a consistent style element for the SDK.

---

I could be convinced that the other is better, but I think I'd need to see what the usage we're trying to support is.

src/globus_sdk/services/search/client.py
MaxTueckeGlobus: You are correct, `MISSING` is filtered out at the time of sending the request. The reason I'm doing the above logic is to account for a case like this:
```
search_client.post_search(SearchQuery(), limit=100)
```
In the proposed solution above where `data` is constructed like so:
```
data = {
    "offset": offset,
    "limit": limit,
    **data,
}
```
`limit` will end up being `MISSING`, since `SearchQuery.limit` defaults to `MISSING`.

The old behavior of this method was that explicitly passed query params take precedence over query params set in `data`. The current implementation was to keep the same behavior by making an `add_kwargs` dict that does not have any `MISSING` values and use that to update `data`.

For reference, this was the old behavior of post search:
```
add_kwargs = {}
if offset is not None:
    add_kwargs["offset"] = offset
if limit is not None:
    add_kwargs["limit"] = limit
if add_kwargs:
    data = {**data, **add_kwargs}
```

Correct me if I'm wrong here, but I think the current proposed version is correct (consistent with the current SDK).

src/globus_sdk/services/search/client.py
MaxTueckeGlobus: I completely agree but I want to call out that such a change would be a different behavior from the one currently in the SDK. 

Currently `query_params` is constructed like so:
```
if query_params is None:
    query_params = {}
query_params.update(
    {
        "q": q,
        "offset": offset,
        "limit": limit,
        "advanced": advanced,
    }
)
```
which I believe is doing what you mention above of overriding the query param fields with the method defaults. I think you are correct that it is the wrong behavior, but I wanna explicitly call out that it would be different before going ahead with your proposed change. Thoughts?

src/globus_sdk/services/search/data.py
MaxTueckeGlobus: I think we need to answer the broader question here about how we want these 'escape valve' params (`additional_fields` and `query_params`) to behave. Should they always take or give precedence to the explicitly passed params or just leave the behavior undefined and up to the specific method implementation.

I think that the current behavior in the SDK of leaving it undefined is not great. For example, this is what `add_sort` is doing currently:
```
sort = {"field_name": field_name, **(additional_fields or {})}
if order is not None:
    sort["order"] = order
```
which can lead to some odd behavior where something like this:
```
SearchQuery().add_sort("field1", order="asc", additional_fields={
         "field_name": "field2",
         "order": "desc",
    }
)
```
results in `field_name="field2"` and `order="asc"`.
I think this is incredibly unintuitive even if it was a bad query to begin with.

All this too say, I agree with Stephen and like the idea of "putting ** expansions at the end as a consistent style element for the SDK". Curious to hear your thoughts though...

src/globus_sdk/services/search/client.py
sirosen: Hm; okay, I need to stare at this a little to make a good recommendation.

You're definitely correct about the order of overrides in SDK v3 and what I would consider desirable behavior for `search_client.post_search(SearchQuery(), limit=100)`. I don't like the path this puts us on for supporting it though.

The issue, IMO, is that the ordering here is pretty delicate in order to get the effect of
- we don't modify the user input object
- we merge with a parameter if one is provided
- we don't merge if no parameter is provided

I'm trying to work out how the code here should look, and then we can "backfill" the implementation.
So, one example:
```python
# _copy_with is a method private to the SDK for doing
# copy+set on payload types with missing-awareness?
data = data._copy_with(limit=limit, offset=offset)
```
Or, very similar:
```python
# merge_parameters could have a signature which takes `*source_params: dict | PayloadWrapper`
# and then merges them in precedence order
data = utils.merge_parameters(data, {"limit": limit, "offset": offset})
```

I think this is subtle enough and high enough value that we should talk through a design for a solution.

src/globus_sdk/services/search/client.py
sirosen: Part of the resolution from discussion is that the defaults should be `MISSING` for these values -- since there is a service-side default. I missed that when commenting before, but that's a good improvement to add and it resolves this case.

I'm not sure why the SDK has defaults for these; as the original author here, I can't recall / speak to what justification I might have had for doing that.

src/globus_sdk/services/search/data.py
```diff
@@ -11,6 +11,15 @@
 SearchQueryT = t.TypeVar("SearchQueryT", bound="SearchQueryBase")
 
 
+def _format_histogram_range(
+    value: tuple[t.Any, t.Any] | MissingType,
+) -> dict[str, t.Any] | MissingType:
+    if isinstance(value, MissingType):
+        return value
```
kurtmckee, line 18: Similar to `None`, I find it easier to read syntax like this when dealing with singleton values:

```python
if value is None:
    return None  # Explicitly return the singleton value
```

```suggestion
    if value is MISSING:
        return MISSING
```

sirosen: Yeah, mypy seems to accept
```python
if t.TYPE_CHECKING:
    import enum

    class MissingType(enum.Enum):
        MISSING = enum.auto()

    MISSING = MissingType.MISSING
else:
    class MissingType:
        ...
```

but _boy_ I am unsure about how we might regret doing that.

kurtmckee: Sorry, my point was that I prefer to `return MISSING`, not `return value`! :grin:

sirosen: Ah! Yes, I also prefer
```python
if x == CONSTANT:
    return CONSTANT
```
for any values of `CONSTANT` and `==` :wink:

src/globus_sdk/services/search/client.py
MaxTueckeGlobus: Per our discussion with the rest of the team, usages of `query_params` and `additional_fields` have been updated to now all take precedence over explicitly passed parameters.

PR #1216, opened 2025-06-05
Update transfer client classes to use MISSING defaults

MaxTueckeGlobus:
[sc-15807](https://app.shortcut.com/globus/story/15807/sdkv4-convert-defaults-of-none-to-globus-sdk-missing-for-payload-types-allowing-explicit-null-by-setting-none)

This PR is for SDKv4 and builds on the feedback in #1205 and #1207.

### Changes
 - Converted defaults of `None` to `globus_sdk.MISSING` for payload types in the transfer client.
 - Updated transfer and timers functional tests to test with `MISSING` defaults instead of `None`.

----
📚 Documentation preview 📚: https://globus-sdk-python--1216.org.readthedocs.build/en/1216/


changelog.d/20250605_142657_max.tuecke_sc_15807_transfer_missing_defaults.rst
```diff
@@ -0,0 +1,4 @@
+Breaking Changes
+~~~~~~~~~~~~~~~~
+
+- All defaults of ``None`` converted to ``globus_sdk.MISSING`` for all payload types in the transfer client. (:pr:`1216`)
```
kurtmckee, line 4:
```suggestion
- All defaults of ``None`` converted to ``globus_sdk.MISSING`` for all payload types in the Transfer client. (:pr:`1216`)
```

src/globus_sdk/services/transfer/client.py
```diff
@@ -1560,7 +1581,7 @@ def submit_transfer(
                     :ref: transfer/task_submit/#submit_transfer_task
         """  # noqa: E501
         log.debug("TransferClient.submit_transfer(...)")
-        if "submission_id" not in data:
+        if isinstance(data.get("submission_id", MISSING), MissingType):
```
kurtmckee, line 1584: What's changing that would require looking for `MISSING` inside of `data`? `.submit_transfer()` doesn't have a method signature change.

This question applies to other code, too.

src/globus_sdk/services/transfer/data/delete_data.py
```diff
@@ -87,25 +88,25 @@ class DeleteData(utils.PayloadWrapper):
...
-        endpoint: UUIDLike | None = None,
+        endpoint: UUIDLike | MissingType = MISSING,
         *,
-        label: str | None = None,
-        submission_id: UUIDLike | None = None,
+        label: str | MissingType = MISSING,
+        submission_id: UUIDLike | MissingType = MISSING,
         recursive: bool = False,
         ignore_missing: bool = False,
         interpret_globs: bool = False,
```
kurtmckee, line 97: These boolean defaults (and a few others below) catch my eye.

Can these be converted to `MISSING` and left out of the payload?

(And seeing this, it's possible that my question re: looking for `MISSING` in `data`, above, may be answered, but I still want to confirm.)

kurtmckee (changes requested):

src/globus_sdk/services/transfer/client.py
MaxTueckeGlobus: There is none! I had gotten confused with this, since `TransferData` and `DeleteData` had the type of `submission_id` updated, I thought this piece needed to be updated to account for the change, forgetting that `utils._set_value` was also updated to skip values of `MISSING`. The change has been reverted now.

src/globus_sdk/services/transfer/data/delete_data.py
MaxTueckeGlobus: I believe they can, but I was hesitant to do so since I could not find an openAPI spec for Transfer.

If we assume that any field in the SDK with a default also has the same service default, then yes, 100% they can all be converted to `MISSING` and left out of the payload. The bit that concerns me is that I don't know that to be the case and it has been difficult trying to figure that info out from the Transfer API docs. 

I've gone and updated all the fields to have MISSING defaults and I'll do some manual testing with actual transfers tomorrow morning. Assuming nothing breaks/behaves unexpectedly, I'll go ahead and make that change?

kurtmckee: I just hunted for this, and the answer is: the SDK is duplicating the defaults and we can switch to `MISSING` for all six of the fields that have `False` or `True` as defaults. :+1:  :partying_face: 

https://docs.globus.org/api/transfer/task_submit/#common_transfer_and_delete_fields

https://docs.globus.org/api/transfer/task_submit/#delete_specific_fields

src/globus_sdk/services/transfer/data/transfer_data.py
```diff
@@ -328,7 +328,7 @@ def add_filter_rule(
         self,
         name: str,
         *,
-        method: t.Literal["include", "exclude"] = "exclude",
+        method: t.Literal["include", "exclude"] | MissingType = MISSING,
```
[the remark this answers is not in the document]
kurtmckee: @sirosen You're correct -- the documentation lists no default value.

https://docs.globus.org/api/transfer/task_submit/#filter_rules

@MaxTueckeGlobus This may need to switch to a declaration like:

```suggestion
        method: t.Literal["include", "exclude"],
```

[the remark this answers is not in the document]
sirosen: I'm not convinced that removing the default which was set prior to this PR is our desired DX.
At the very least, I believe that changing that behavior is out of scope for this PR and that if we want to change it, we should do so in a dedicated changeset with a separate changelog entry and discussion.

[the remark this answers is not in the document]
MaxTueckeGlobus: Ok, I'll go ahead and revert that line and we can revisit removing SDK defaults like this at a later date.

sirosen (approved):

PR #1221, opened 2025-06-09
Add the `SpecificFlow.validate_run()` method

kurtmckee:
### Added

- Add the ``SpecificFlow.validate_run()`` method.

----
📚 Documentation preview 📚: https://globus-sdk-python--1221.org.readthedocs.build/en/1221/

### DIFF
--- a/src/globus_sdk/_testing/data/flows/validate_run.py
+++ b/src/globus_sdk/_testing/data/flows/validate_run.py
@@ -1,5 +1,3 @@
-from responses import matchers
-
 from globus_sdk._testing.models import RegisteredResponse, ResponseSet
 
 from ._common import TWO_HOP_TRANSFER_FLOW_ID
@@ -22,7 +20,7 @@ validate_invalid_simple_input_request = {
 RESPONSES = ResponseSet(
     metadata={
         "flow_id": TWO_HOP_TRANSFER_FLOW_ID,
-        "request_params": validate_simple_input_request,
+        "request_body": VALIDATE_RUN_SIMPLE_INPUT_BODY,
     },
     default=RegisteredResponse(
         service="flows",
@@ -30,12 +28,6 @@ RESPONSES = ResponseSet(
         method="POST",
         status=200,
         json=VALIDATE_RUN_SIMPLE_SUCCESS_RESPONSE,
-        match=[
-            matchers.json_params_matcher(
-                params={"body": VALIDATE_RUN_SIMPLE_INPUT_BODY},
-                strict_match=False,
-            )
-        ],
     ),
     invalid_input_payload=RegisteredResponse(
         service="flows",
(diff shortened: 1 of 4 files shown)


src/globus_sdk/_testing/data/flows/validate_run.py
```diff
@@ -0,0 +1,108 @@
...
+    },
+    default=RegisteredResponse(
+        service="flows",
+        path=f"/flows/{TWO_HOP_TRANSFER_FLOW_ID}/validate_run",
+        method="POST",
+        status=200,
+        json=VALIDATE_RUN_SIMPLE_SUCCESS_RESPONSE,
+        match=[
+            matchers.json_params_matcher(
+                params={"body": VALIDATE_RUN_SIMPLE_INPUT_BODY},
```
sirosen, line 35: Is there a reason not to write this as
```suggestion
                params=validate_simple_input_request,
```
?

As a related aside, I'm a little wary of our applying `match` usage to our SDK-provided testing responses (we originally didn't do it at all, then started doing it). It has proven problematic in some cases where you want "a" response but you don't care about or want to vary the contents of the request.
For example, if you want a test in a downstream application `test_frobulate_validates_run_input` with parametrized inputs to `frobulate()`, you may have dozens of _different_ calls to this API, and you would prefer that the default SDK response *always* matched so you can write your test more simply.
I'm considering shifting us to not have a match specified for `default` and allowing it only for the other responses, but that would need a broader discussion.

tests/functional/services/flows/test_validate_run.py
```diff
@@ -0,0 +1,43 @@
...
+from globus_sdk import FlowsAPIError, SpecificFlowClient
+from globus_sdk._testing import load_response
+
+
+def test_validate_run(specific_flow_client_class: type[SpecificFlowClient]):
+    metadata = load_response(SpecificFlowClient.validate_run).metadata
+
+    flow_client = specific_flow_client_class(flow_id=metadata["flow_id"])
+
+    resp = flow_client.validate_run(**metadata["request_params"])
```
sirosen, line 14: Would it be better to remove a layer of indirection here, and store `body` in `metadata` instead of the wrapper, `request_params`?
Then we'd write this line:
```suggestion
    resp = flow_client.validate_run(body=metadata["request_body"])
```

Personally, I prefer this style over expansion of a dict sourced from elsewhere in the codebase. It means that when I read the test, I know what call signature for `validate_run` is being tested. Then, if `validate_run` gains a new parameter, we'll be able to 'see" that this test doesn't use that new param.

sirosen (approved):
Optional / non-blocking feedback left, only on the testing pieces. Client changes look great! :+1:

tests/functional/services/flows/test_validate_run.py
kurtmckee: Works for me!

sirosen (approved):

PR #1222, opened 2025-06-11
Break 'globus_sdk.utils' into components and update payload types to inherit from 'dict'

sirosen:
## What

Break up the `globus_sdk.utils` monolith and improve the characteristics of the payload class.

## Why

Two primary observations motivated this work:
1. `globus_sdk.utils` is large, with unclear scope (as the name implies)
2. `PayloadWrapper._set_optstrs` and related helpers are not as important now that we have `MISSING` for defaults

Additionally, this is a good opportunity to address these additional improvements:
- `safe_strseq_iter` is handy, and hints at a family of similar useful operations for formatting data
- `PayloadWrapper` inherits from `UserDict`; inheriting from `dict` would simplify types which accept `dict | PayloadWrapper` to just accepting `dict`
- `PayloadWrapper` is a bad name -- `Payload` would be better

## How

- `classproperty` definition moved to `globus_sdk._classproperty`.
- `MISSING` definition moved to `globus_sdk._missing`.
- Rename `PayloadWrapper` to ~`Payload`~ `GlobusPayload`.
- `Payload` inherits from `dict`, not `UserDict`.
- Introduce ~`AbstractPayload`~ `AbstractGlobusPayload` to handle the metaclass conflict between `dict` and `abc.ABC`
- Move components for reshaping data from `globus_sdk.utils` to `globus_sdk._remarshal` -- specifically, `commajoin` and `safe_strseq_iter` to start.
- Rename `globus_sdk.utils` to `globus_sdk._utils`, making its private scope clearer.
- Convert payload types away from `_set_opt*` methods and `_set_value`, using `MISSING` setitem assignments instead.
  - To aid and simplify in this process, new helpers like `listify` and `strseq_listify` are added.
- Remove the `safe_` prefix from names in `_remarshal.py`, and rely on docstrings and tests to demonstrate behavior instead.
- Introduce `NullableOmittable` in `_remarshal.py` as a type alias for conversions of missing/null data.

----
📚 Documentation preview 📚: https://globus-sdk-python--1222.org.readthedocs.build/en/1222/


m1yag1:
@sirosen While I agree removing 3.8 support is a worthwhile effort, I thought it might be quicker to fix the GitHub action. I have a fix in place in https://github.com/globus/globus-sdk-python/pull/1223.

src/globus_sdk/_classproperty.py
```diff
@@ -0,0 +1,71 @@
+"""
```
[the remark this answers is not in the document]
derek-globus: Sure, if it's a pretty quick followon

src/globus_sdk/_remarshal.py
```diff
@@ -0,0 +1,159 @@
...
+
+T = t.TypeVar("T")
+R = t.TypeVar("R")
+
+
+def safe_strseq_iter(
+    value: t.Iterable[t.Any] | str | uuid.UUID,
+) -> t.Iterator[str]:
+    """
+    Given an Iterable (typically of strings), produce an iterator over it of strings.
```
derek-globus, line 26: This docstring mis-identifies what parameters are required. Not all input types are Iterable.

```suggestion
    Iterate over one or more string/string-convertible values.
```

src/globus_sdk/_remarshal.py
```diff
@@ -0,0 +1,159 @@
...
+from globus_sdk._missing import MISSING, MissingType
+from globus_sdk._types import UUIDLike
+
+T = t.TypeVar("T")
+R = t.TypeVar("R")
+
+
+def safe_strseq_iter(
+    value: t.Iterable[t.Any] | str | uuid.UUID,
+) -> t.Iterator[str]:
```
derek-globus, line 24: Why is UUID the special case for a re-marshaller that will handle any type so long as there are multiple of them?
Would it make more sense to allow an iterable of anything or anything & anything gets converted to a string?

---

Alternatively if it's really meant to be "string or UUID" and not "anything", then we should scope down `t.Iterable` to `t.Iterable[str | UUID]`

src/globus_sdk/_remarshal.py
```diff
@@ -0,0 +1,159 @@
...
+    elif isinstance(value, uuid.UUID):
+        yield str(value)
+    else:
+        for x in value:
+            yield str(x)
+
+
+def safe_stringify(value: object | MissingType | None) -> str | MissingType | None:
+    """
+    Given any object or a MISSING|None, return `str(object) | MISSING | None`.
```
derek-globus, line 50:
```suggestion
    Convert a value to a string, with handling for None and Missing.
```

src/globus_sdk/_remarshal.py
```diff
@@ -0,0 +1,159 @@
...
+    if isinstance(value, str):
+        yield value
+    elif isinstance(value, uuid.UUID):
+        yield str(value)
+    else:
+        for x in value:
+            yield str(x)
+
+
+def safe_stringify(value: object | MissingType | None) -> str | MissingType | None:
```
derek-globus, line 48: Out of curiosity, why `object` and not `t.Any` here?

src/globus_sdk/_remarshal.py
```diff
@@ -0,0 +1,159 @@
...
+
+@t.overload
+def listify(value: None) -> None: ...
+@t.overload
+def listify(value: MissingType) -> MissingType: ...
+@t.overload
+def listify(value: t.Iterable[T]) -> list[T]: ...
+
+
+def listify(value: t.Iterable[T] | MissingType | None) -> list[T] | MissingType | None:
```
derek-globus, line 97:
```suggestion
def safe_listify(value: t.Iterable[T] | MissingType | None) -> list[T] | MissingType | None:
```

Justification: This is identical to `safe_stringify` in concept but named differently.

src/globus_sdk/_payload.py
```diff
@@ -0,0 +1,59 @@
...
+# TODO: Remove this dispatch after we drop Python 3.8 support.
+#       In 3.9+ `dict.__class_getitem__` is available.
+if t.TYPE_CHECKING:
+    # pylint: disable=unsubscriptable-object
+    _PayloadBaseDict = t.Dict[str, t.Any]
+else:
+    _PayloadBaseDict = dict
+
+
+class Payload(_PayloadBaseDict):
```
derek-globus, line 21: If we're going to rename `PayloadWrapper`, I think we should push it to be more verbose.

Right now the request structure is such that a client's request method:
* returns a `GlobusHTTPResponse`
* raises a `GlobusAPIError`
* but accepts a `Payload` 

Can we call it `GlobusPayload` or `GlobusHTTPPayload` or `GlobusRequestPayload` or something that qualifies it from other usages of the term?

src/globus_sdk/_classproperty.py
```diff
@@ -0,0 +1,71 @@
...
+class _classproperty(t.Generic[T, R]):
+    """
+    This is a well-typed Generic Descriptor which can be used to wrap decorated
+    functions.
+
+    Note that this descriptor will pass an instance (self) if possible, and the
+    class (cls) only if there is no instance. This is unlike ``classmethod``.
+
+    For more guidance on how this works, see the python3 descriptor guide:
+      https://docs.python.org/3/howto/descriptor.html#properties
```
derek-globus, line 39:
```suggestion
    A hybrid class/instance property descriptor.

    On a class, the decorated method will be invoked with `cls`.
    On an instance, the decorated method will be invoked with `self`.
```

src/globus_sdk/_remarshal.py
```diff
@@ -0,0 +1,159 @@
...
+import uuid
+
+from globus_sdk._missing import MISSING, MissingType
+from globus_sdk._types import UUIDLike
+
+T = t.TypeVar("T")
+R = t.TypeVar("R")
+
+
+def safe_strseq_iter(
```
derek-globus, line 22: I don't think this method should be named `safe`.

All other functions in the module use the prefix `safe` to convey "if the value you pass is Missing or None, we'll just return that" except for this one.

Not sure if there's some history around this naming discrepency, but it leads to developer confusion when a term is almost-perfectly consistent.

derek-globus (changes requested):

src/globus_sdk/_remarshal.py
sirosen: `Any` and `object` are really similar in these cases, but I'm trying (personally) to break the habit of using `Any` where only `object`.

In this case it doesn't have an impact. Here's one where it does:
```python
def frob(obj: object) -> str:
    return obj.frobulate()  # error! `object` doesn't provide `.frobulate()`!

def frob(obj: Any) -> str:
    return obj.frobulate()  # pass! you said 'Any', so how can we know if it has `.frobulate()`?
```

src/globus_sdk/_payload.py
sirosen: Let's go with `GlobusPayload`. I feel more strongly _against_ `PayloadWrapper` than I do in favor of `Payload`.
My goal isn't to reduce verbosity, but to fix the mismatch of "it's not a wrapper tho!"

src/globus_sdk/_remarshal.py
sirosen: I've done some significant renaming in my latest commit, introducing `Omittable[T]` and `NullableOmittable[T]` to describe these types.

Let me know if you want more changes or if we can consider this sufficiently refined.

src/globus_sdk/_remarshal.py
sirosen: I decided to remove `safe_` from all of the names. The commit message notes the history in brief, but it's better to rely on docstrings to demonstrate the difference than it is to try to encode it in this way.

If we wanted to bring it back, I'd slam `safe` with `str`, so we'd have `safestr_iter` or some names like that -- indicating that the "safe" goes with "str".

derek-globus (approved):

PR #1228, opened 2025-06-18
Support use of MISSING throughout the BaseAuthClient and its variants

sirosen:
This started from a body of work done by @MaxTueckeGlobus , but I rebased it and continued to tinker on it to get things ironed out as part of the improvements I delivered in #1222.

As with other conversions, many of these are "rote", simple rewrites, and a few have special nuance.
In particular, we discussed whether or not the oauth2 flow manager objects should rewrite to use MISSING or continue to use None.
I examined the parameters and made some judgement calls there, as there are fields which at least _in theory_ a user could want to explicitly remove from an authorization URL, although there is no current practical reason to enable such usage.

Also notable: I decided not to touch the GARE components in this work. As a result, there is a new `none2missing` helper added to adapt a GARE into OAuth2 authorization inputs.

----
📚 Documentation preview 📚: https://globus-sdk-python--1228.org.readthedocs.build/en/1228/

### DIFF
--- a/src/globus_sdk/login_flows/login_flow_manager.py
+++ b/src/globus_sdk/login_flows/login_flow_manager.py
@@ -52,6 +52,8 @@ class LoginFlowManager(metaclass=abc.ABCMeta):
         """
         self._oauth2_start_flow(auth_parameters, redirect_uri)
 
+        # prompt is assigned first because its usage is type-ignored below
+        # this makes it clear that the ignore applies to that usage
         prompt = none2missing(auth_parameters.prompt)
         return self.login_client.oauth2_get_authorize_url(
             session_required_identities=none2missing(


src/globus_sdk/services/auth/client/confidential_client.py
```diff
@@ -447,6 +445,17 @@ def create_child_client(
                 "AuthClient.create_client requires either 'public_client' or "
                 "'client_type'."
             )
+        # terms_and_conditions and privacy_policy must both be set or unset
+        if bool(terms_and_conditions) ^ bool(privacy_policy):
+            raise exc.GlobusSDKUsageError(
+                "terms_and_conditions and privacy_policy must both be set or unset"
+            )
+        links: dict[str, str | MissingType] | MissingType = MISSING
+        if terms_and_conditions or privacy_policy:
```
kurtmckee, line 454: It's not incorrect to use an `or` here, but since we know from above that both or neither are set, it's sufficient to only check one.

I _definitely_ think it's more readable to show both variables on this line, rather than checking only one, but the `or` catches my eye because it will never evaluate if `terms_and_conditions` is set. Perhaps:

```suggestion
        if terms_and_conditions and privacy_policy:
```

Extra work, but it doesn't rely on readers understanding the XOR above to figure out a bare `if terms_and_conditions:`...

src/globus_sdk/services/auth/client/confidential_client.py
```diff
@@ -455,23 +464,12 @@ def create_child_client(
...
-        links: dict[str, str | MissingType] = {
-            "terms_and_conditions": terms_and_conditions,
-            "privacy_policy": privacy_policy,
-        }
-        if terms_and_conditions or privacy_policy:
-            body["links"] = links
 
-        if not isinstance(additional_fields, MissingType):
+        if additional_fields is not None:
             body.update(additional_fields)
```
kurtmckee, line 474: This is accounted for above by the `**(additional_fields or {})`.

```suggestion
```

src/globus_sdk/services/auth/client/service_client.py
```diff
@@ -208,7 +210,7 @@ def get_jwk(
         :param as_pem: Decode the JWK to an RSA PEM key, typically for JWT decoding
         :type as_pem: bool
         """
-        if openid_configuration is None:
+        if isinstance(openid_configuration, MissingType):
```
kurtmckee, line 213:
```suggestion
        if openid_configuration is MISSING:
```

src/globus_sdk/services/auth/client/service_client.py
```diff
@@ -348,35 +350,36 @@ def get_identities(
...
 
-        if "usernames" in query_params and "ids" in query_params:
+        if (
+            query_params["usernames"] is not MISSING
+            and query_params["ids"] is not MISSING
+        ):
             log.warning(
                 "get_identities call with both usernames and "
                 "identities set! Expected to result in errors"
             )
```
kurtmckee, line 370: The tenses here don't read like a warning to me. Perhaps:

```suggestion
            log.warning(
                "get_identities called with both usernames and "
                "identities set! Expecting an error."
            )
```

src/globus_sdk/services/auth/client/service_client.py
```diff
@@ -441,28 +444,23 @@ def get_identity_providers(
...
-        # letting us consume args whose `__str__` methods produce "the right
-        # thing"
-        elif domains is not None:
-            query_params["domains"] = commajoin(domains)
-        elif ids is not None:
-            query_params["ids"] = commajoin(ids)
-        else:
+        elif domains is MISSING and ids is MISSING:
             log.warning(
                 "neither 'domains' nor 'ids' provided to get_identity_providers(). "
```
kurtmckee, line 454: This is getting out-of-scope for this PR, but since we're here:

```suggestion
                "Neither 'domains' nor 'ids' provided to get_identity_providers(). "
```

src/globus_sdk/services/auth/flow_managers/native_app.py
```diff
@@ -150,7 +153,7 @@ def __init__(
             f"verifier=<REDACTED>,challenge={self.challenge}"
         )
 
-        if prefill_named_grant is not None:
+        if not isinstance(prefill_named_grant, MissingType):
```
kurtmckee, line 156:
```suggestion
        if prefill_named_grant is not MISSING:
```

src/globus_sdk/services/auth/flow_managers/native_app.py
```diff
@@ -181,11 +184,10 @@ def get_authorize_url(self, query_params: dict[str, t.Any] | None = None) -> str
             "code_challenge": self.challenge,
             "code_challenge_method": "S256",
             "access_type": (self.refresh_tokens and "offline") or "online",
+            "prefill_named_grant": self.prefill_named_grant,
```
kurtmckee, line 187: I think this could follow our "standard" process:

```suggestion
            "prefill_named_grant": self.prefill_named_grant,
            **(query_params or {}),
```

src/globus_sdk/services/auth/flow_managers/native_app.py
```diff
@@ -181,11 +184,10 @@ def get_authorize_url(self, query_params: dict[str, t.Any] | None = None) -> str
...
         }
-        if self.prefill_named_grant is not None:
-            params["prefill_named_grant"] = self.prefill_named_grant
-        if query_params:
-            params.update(query_params)
+        params.update(query_params or {})
```
kurtmckee, line 189:
```suggestion
```

tests/functional/services/auth/service_client/test_get_identities.py
```diff
@@ -55,15 +55,15 @@ def test_get_identities_success(usernames, service_client):
...
+        (None, "false"),
     ],
 )
 def test_get_identities_provision(inval, outval, service_client):
     load_response(service_client.get_identities)
-    service_client.get_identities(usernames="globus@globus.org", provision=inval)
+    if inval is not None:
+        service_client.get_identities(usernames="globus@globus.org", provision=inval)
+    else:
+        service_client.get_identities(usernames="globus@globus.org")
```
kurtmckee, line 66: The parametrized list makes it seem like `None` is being tested as an invalid value, but this condition is preventing that testing.

I recommend moving this `None` test to simply check default values instead. Otherwise, the condition needs to be removed so `None` can be passed as the `provision` argument.

tests/functional/services/auth/test_auth_client_flow.py
```diff
@@ -130,7 +131,7 @@ def test_oauth2_get_authorize_url_supports_session_params(
         "session_required_single_domain" if domain_option else None,
         "session_required_identities" if identity_option else None,
         "session_required_policies" if policy_option else None,
-        "session_required_mfa" if mfa_option is not None else None,
+        "session_required_mfa" if not isinstance(mfa_option, MissingType) else None,
```
kurtmckee, line 134:
```suggestion
        "session_required_mfa" if mfa_option is not MISSING else None,
```

kurtmckee (changes requested):

src/globus_sdk/services/auth/client/base_login_client.py
```diff
@@ -403,9 +395,7 @@ def oauth2_token(
         log.debug("Fetching new token from Globus Auth")
         # use the fact that requests implicitly encodes the `data` parameter as
         # a form POST
-        data = dict(form_data)
-        if body_params:
-            data.update(body_params)
+        data = {**dict(form_data), **(body_params or {})}
```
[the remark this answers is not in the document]
sirosen: Yes, good eye!

src/globus_sdk/services/auth/client/confidential_client.py
```diff
@@ -92,18 +92,15 @@ def get_identities(
...
-                "false" if str(provision).lower() == "false" else "true"
-            )
-        if ids is not None:
-            query_params["ids"] = commajoin(ids)
-
+        query_params = {
+            "usernames": commajoin(usernames),
+            # only specify `provision` if `usernames` is given
+            "provision": (
+                str(provision).lower() if usernames is not MISSING else MISSING
```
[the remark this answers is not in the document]
sirosen: The identity check doesn't type-narrow, so some of those `isinstance` checks are being done to make type checking happy.

src/globus_sdk/services/auth/client/service_client.py
sirosen: I'll allow it! :stuck_out_tongue_winking_eye:

tests/functional/services/auth/service_client/test_get_identities.py
sirosen: Yeah, I'll move it to a separate test. I also considered
```
_OMIT = object()
```
as a local sentinel for this test, but that felt similarly wonky.

src/globus_sdk/services/auth/client/confidential_client.py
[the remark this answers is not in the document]
sirosen: I'm happy to change this check to an `isinstance()` if your goal is that we are consistent about always checking it in the same way. I'm not happy to move in the other direction (and am really hoping that we get sentinel values in 3.15 or 3.16 to solve this), though I consider the behavior of the type checkers here to be regrettable.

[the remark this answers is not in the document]
kurtmckee: Hmm...perhaps we should have standardized on `...` instead of `MISSING` so that we have true sentinels that aren't valid values...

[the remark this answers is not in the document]
kurtmckee:
```python
MISSING = ...
MissingType = type(...)
```

:wink:

src/globus_sdk/login_flows/login_flow_manager.py
```diff
@@ -51,14 +52,20 @@ def _get_authorize_url(
...
+            ),
+            session_required_single_domain=none2missing(
+                auth_parameters.session_required_single_domain
+            ),
+            session_required_policies=none2missing(
+                auth_parameters.session_required_policies
+            ),
+            session_required_mfa=none2missing(auth_parameters.session_required_mfa),
+            session_message=none2missing(auth_parameters.session_message),
+            prompt=prompt,  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
kurtmckee: That was not clear at all to me.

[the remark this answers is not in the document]
sirosen: Would you be cool with keeping it the way I've written, but commenting on it above? I'm happy to take the feedback that it's non-obvious, but prefer to have the specificity of the ignore be crystal clear -- that helps when you look at ignores as "things to fix / cleanup" and are trying to understand why a line is ignored.

src/globus_sdk/login_flows/login_flow_manager.py
```diff
@@ -51,14 +52,20 @@ def _get_authorize_url(
         """
         self._oauth2_start_flow(auth_parameters, redirect_uri)
 
-        session_required_single_domain = auth_parameters.session_required_single_domain
+        prompt = none2missing(auth_parameters.prompt)
```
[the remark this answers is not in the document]
kurtmckee: As noted below, separating this is necessary so that the function call isn't implicated in a type-ignore comment a few lines below this, where `prompt` is actually used.

Putting this up here keeps the type-ignore isolated and specific.

PR #1235, opened 2025-06-24
Make 'is MISSING' comparison our style, and allow type checkers to narrow on it

sirosen:
_Note: includes #1234_

Add a `TYPE_CHECKING`-time definition of the `MissingType` and `MISSING` sentinel as an enum and enum member. This lets type checkers understand that an identity comparison against the `MISSING` sentinel is a valid way to type-narrow.

Update the local flake8 plugin to scan for `isinstance(..., MissingType)` and treat that as an error.

Update all of the usages flagged by this new lint.

Add a type checking test for `MISSING` and `MissingType`.
This confirms that `MISSING` and `MissingType` behave appropriately at type checking time.

----
📚 Documentation preview 📚: https://globus-sdk-python--1235.org.readthedocs.build/en/1235/

### DIFF
--- a/src/globus_sdk/_missing.py
+++ b/src/globus_sdk/_missing.py
@@ -10,20 +10,29 @@ import typing as t
 
 T = t.TypeVar("T")
 
+# type checkers don't know that MISSING is a sentinel, so we will describe it
+# differently at typing time, allowing for type narrowing on `is MISSING` and
+# similar checks
 if t.TYPE_CHECKING:
     # pretend that `MISSING: MissingType` is an enum at type-checking time
-    # this allows type checkers to use identity comparisons to type narrow
     #
-    # for example:
+    # enums are treated as `Literal[...]` values and are narrowed under simple
+    # checks, as unions and literal types are
+    # therefore, under this definition, `MissingType ~= Literal[MissingType.MISSING]`
+    #
+    # Therefore, consider this example:
     #
     #       x: int | float | MissingType
     #       if x is not MISSING:
     #           reveal_type(x)
     #
-    # should show `x: int | float`
+    # This is effectively the same as if we wrote:
     #
-    # however, because type checkers don't know that `MISSING` is a sentinel
-    # they do not narrow in this case based on the runtime data
+    #       x: int | float | Literal["a"]
+    #       if x != "a":
+    #           reveal_type(x)
+    #
+    # Both should show `x: int | float`
     import enum
 
     class MissingType(enum.Enum):


src/globus_sdk/_globus_sdk_flake8.py
```diff
@@ -124,16 +126,35 @@ def visit_Assign(self, node: ast.Assign) -> None:
...
+            # the function call is of the form 'OBJ.info(...)'
+            isinstance(func_node := node.func, ast.Attribute)
+            and func_node.attr == "info"
+            # and, more specifically, the object is named 'log', so
+            # it's `log.info(...)
+            and isinstance(func_node.value, ast.Name)
+            and func_node.value.id == "log"
+        ):
+            self._record(node, "SDK002")
+        else:
```
ada-globus, line 159: I'm tickled that it's possible to express the check this way. It's clever, but... I think we should turn these into functions for readability. I think the necessity of the comments gives away that this is rather tricky to read as it's currently expressed. I think we could express the same logic more plainly.

src/globus_sdk/_missing.py
```diff
@@ -10,40 +10,58 @@
...
+    # for example:
+    #
+    #       x: int | float | MissingType
+    #       if x is not MISSING:
+    #           reveal_type(x)
+    #
+    # should show `x: int | float`
+    #
+    # however, because type checkers don't know that `MISSING` is a sentinel
+    # they do not narrow in this case based on the runtime data
```
ada-globus, line 26: I think the comment block should lead with this. As it appears at the end, it suggests to me that this is the current situation, not the one that we've ~hoodwinked the type checker into believing~ cleverly engineered our way around.

tests/non-pytest/mypy-ignore-tests/test_consents_usage.py
```diff
@@ -11,7 +12,8 @@
 # create some variant types
 xfer_str: str = "urn:globus:auth:scope:transfer.api.globus.org:all"
 strlist: list[str] = [xfer_str]
-scopelist: list[Scope] = Scope.parse(xfer_str)
+scopelist: list[Scope] = ScopeParser.parse(xfer_str)
+scopeobj: Scope = Scope.parse(xfer_str)
```
ada-globus, line 16: This seems unrelated... 🤔

ada-globus (changes requested):
Weird way to get there, but I do like the plain english semantics this gets us. Worth the trade, IMO.

tests/non-pytest/mypy-ignore-tests/test_consents_usage.py
sirosen: Oh, that fix made it in separately already. Maybe I need a rebase...

src/globus_sdk/_missing.py
sirosen: Yeah, good note for improvement. Will rearrange the comment a bit.

PR #1238, opened 2025-06-30
Update `ComputeClientV2.get_endpoints` to list MEPs via a new `role` kwarg

yadudoc:
* Currently `get_endpoints` only queries for endpoints owned by the user.
  To list MEPs that are accessible but not owned by the user, the web-service supports a role query param that can be toggled from the
default of "owner" to "any". Refer: https://compute.api.globus.org/docs#/Endpoints/get_endpoints_v2_endpoints_get
* A new `role` kwarg is added to `ComputeClientV2.get_endpoints` that
  takes `roles="any"` to query for MEPs in addition to those owned by the user.

----
📚 Documentation preview 📚: https://globus-sdk-python--1238.org.readthedocs.build/en/1238/

### DIFF
--- a/changelog.d/20250630_144536_yadudoc1729_get_endpoints_with_role.rst
+++ b/changelog.d/20250630_144536_yadudoc1729_get_endpoints_with_role.rst
@@ -1,0 +1,4 @@
+Added
+-----
+
+- Updated ``ComputeClientV2.get_endpoints`` with a new ``role`` kwarg. (:pr:`1238`)
--- a/src/globus_sdk/services/compute/client.py
+++ b/src/globus_sdk/services/compute/client.py
@@ -3,10 +3,9 @@ from __future__ import annotations
 import logging
 import typing as t
 
-from globus_sdk import GlobusHTTPResponse, client, utils
+from globus_sdk import MISSING, GlobusHTTPResponse, MissingType, client, utils
 from globus_sdk._types import UUIDLike
 from globus_sdk.scopes import ComputeScopes, Scope
-from globus_sdk import MISSING, MissingType
 
 from .errors import ComputeAPIError
 
--- a/src/globus_sdk/services/compute/client.py
+++ b/src/globus_sdk/services/compute/client.py
@@ -6,6 +6,7 @@ import typing as t
 from globus_sdk import GlobusHTTPResponse, client, utils
 from globus_sdk._types import UUIDLike
 from globus_sdk.scopes import ComputeScopes, Scope
+from globus_sdk import MISSING, MissingType
 
 from .errors import ComputeAPIError
 
@@ -101,8 +102,10 @@ class ComputeClientV2(client.BaseClient):
         """  # noqa: E501
         return self.get(f"/v2/endpoints/{endpoint_id}/status")
 
-    def get_endpoints(self) -> GlobusHTTPResponse:
+    def get_endpoints(self, role: str | MissingType = MISSING) -> GlobusHTTPResponse:
         """Get a list of registered endpoints associated with the authenticated user.
+
+        :param role: Role of the user in relation to endpoints. (e.g.: owner, any)
 
         .. tab-set::
 
@@ -112,7 +115,8 @@ class ComputeClientV2(client.BaseClient):
                     :service: compute
                     :ref: Endpoints/operation/get_endpoints_v2_endpoints_get
         """  # noqa: E501
-        return self.get("/v2/endpoints")
+        query_params = {"role": role} if role else None
+        return self.get("/v2/endpoints", query_params=query_params)
 
     def delete_endpoint(self, endpoint_id: UUIDLike) -> GlobusHTTPResponse:
         """Delete a registered endpoint.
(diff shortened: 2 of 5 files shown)


src/globus_sdk/services/compute/client.py
```diff
@@ -101,18 +101,22 @@ def get_endpoint_status(self, endpoint_id: UUIDLike) -> GlobusHTTPResponse:
         """  # noqa: E501
         return self.get(f"/v2/endpoints/{endpoint_id}/status")
 
-    def get_endpoints(self) -> GlobusHTTPResponse:
+    def get_endpoints(self, role: str | None = None) -> GlobusHTTPResponse:
         """Get a list of registered endpoints associated with the authenticated user.
 
-        .. tab-set::
+                :param role: Role of the user in relation to endpoints. (e.g.: owner, any)
```
kurtmckee, line 107: This block looks like it was indented quite a bit. I'm not convinced this needs to be indented further.

```suggestion
        :param role: Role of the user in relation to endpoints. (e.g.: owner, any)

File: src/globus_sdk/services/compute/client.py [hunk 2]
```diff
@@ -101,18 +101,22 @@ def get_endpoint_status(self, endpoint_id: UUIDLike) -> GlobusHTTPResponse:
...
 
-            .. tab-item:: API Info
+                .. tab-set::
+        c
```
kurtmckee, line 110: This doesn't look correct. Errant typo?

```suggestion
```

src/globus_sdk/services/compute/client.py
```diff
@@ -101,18 +101,22 @@ def get_endpoint_status(self, endpoint_id: UUIDLike) -> GlobusHTTPResponse:
         """  # noqa: E501
         return self.get(f"/v2/endpoints/{endpoint_id}/status")
 
-    def get_endpoints(self) -> GlobusHTTPResponse:
+    def get_endpoints(self, role: str | None = None) -> GlobusHTTPResponse:
```
kurtmckee, line 104: As this is a new parameter, I recommend switching from `None` as the sentinel value to `globus_sdk.MISSING`, which has the type `globus_sdk.MissingType`:

```suggestion
    def get_endpoints(self, role: str | MissingType = MISSING) -> GlobusHTTPResponse:
```

We're making a major effort in the v4 branch to convert to a well-known parameter sentinel value so that there's no ambiguity. Since there's no legacy concern with this new parameter, using `MISSING` from the get-go will make this merge easily into the v4 branch.

src/globus_sdk/services/compute/client.py
```diff
@@ -101,18 +101,22 @@ def get_endpoint_status(self, endpoint_id: UUIDLike) -> GlobusHTTPResponse:
...
-                    :service: compute
-                    :ref: Endpoints/operation/get_endpoints_v2_endpoints_get
+                        .. extdoclink:: Get Endpoints
+                            :service: compute
+                            :ref: Endpoints/operation/get_endpoints_v2_endpoints_get
         """  # noqa: E501
-        return self.get("/v2/endpoints")
+
+        query_params = {"role": role} if role else None
+        return self.get("/v2/endpoints", query_params=query_params)
```
kurtmckee, line 119: QQ: Is there a v3 equivalent that needs to be added as well?

kurtmckee (changes requested):

src/globus_sdk/services/compute/client.py
yadudoc: I believe not. But @rjmello probably can confirm.

src/globus_sdk/services/compute/client.py
yadudoc: Fixed.

src/globus_sdk/services/compute/client.py
yadudoc: Thanks for the super quick review. I have these changes merged.

src/globus_sdk/services/compute/client.py
kurtmckee: Doesn't have to go into this PR, but wanted to get that loaded into our brains if it's needed. :+1:

changelog.d/20250630_144536_yadudoc1729_get_endpoints_with_role.rst
```diff
@@ -0,0 +1,4 @@
+Added
+-----
+
+- Updated `ComputeClientV2.get_endpoints` with a new `role` kwarg. (:pr:`1238`)
```
kurtmckee, line 4: pre-commit caught this and I missed it (RST vs Markdown backtick styles, which render differently):

```suggestion
- Updated ``ComputeClientV2.get_endpoints`` with a new ``role`` kwarg. (:pr:`1238`)
```

kurtmckee (changes requested):
I didn't see the MISSING change yet, but in addition, pre-commit caught a Markdown vs RestructuredText backtick issue.

kurtmckee (approved):
Looks good. When you're ready to take this out of draft, I'll be ready to merge. :partying_face:

changelog.d/20250630_144536_yadudoc1729_get_endpoints_with_role.rst
kurtmckee: This is addressed.

yadudoc:
@kurtmckee Just updated the PR.

PR #1257, opened 2025-07-14
Remove the compatibility shim from `create_policy`

sirosen:


src/globus_sdk/services/auth/client/service_client.py
```diff
@@ -803,7 +764,6 @@ def get_policies(self) -> IterableResponse:
         """
         return GetPoliciesResponse(self.get("/v2/api/policies"))
 
-    @_create_policy_compat
     def create_policy(  # pylint: disable=missing-param-doc
```
kurtmckee, line 767: Is there any chance that removing the shim means that the pylint disable comment can be removed as well? Or is that pylint issue unrelated?

kurtmckee (approved):

PR #1261, opened 2025-07-16
Add RequestCallerInfo data object to RequestsTransport.request

m1yag1:
[sc-43037](https://app.shortcut.com/globus/story/43037/)
This adds `RequestCallerInfo` as an extensible container for caller context information in `RequestsTransport.request()`.

The change breaks backward compatibility and replaces the `authorizer` parameter while enabling future extensions for additional caller context beyond authorization.

`RequestCallerInfo` takes precedence when both parameters are provided, with a warning logged for conflicting usage.

----
📚 Documentation preview 📚: https://globus-sdk-python--1261.org.readthedocs.build/en/1261/

### DIFF
--- a/tests/unit/transport/test_transport_authz_handling.py
+++ b/tests/unit/transport/test_transport_authz_handling.py
@@ -55,6 +55,9 @@ def test_requests_transport_accepts_caller_info():
 
         assert response.status_code == 200
 
+        sent_request = mock_session.send.call_args[0][0]
+        assert sent_request.headers["Authorization"] == "Bearer token"
+
 
 def test_requests_transport_caller_info_required():
     transport = RequestsTransport()


src/globus_sdk/transport/_clientinfo.py
```diff
@@ -124,3 +127,14 @@ def _check_reserved_chars(key: str, value: str) -> None:
...
+
+class RequestCallerInfo:
+    """
+    Data object that holds contextual information about the caller of a request.
+
+    :param authorizer: The authorizer object from the client making the request
+    """
+
+    def __init__(self, *, authorizer: GlobusAuthorizer | None = None) -> None:
+        self.authorizer = authorizer
```
sirosen, line 140: Let's put this somewhere else. Per the module docstring, this implements the `X-Globus-Client-Info` header spec. We shouldn't mix in other functionality.

src/globus_sdk/transport/requests.py
```diff
@@ -307,6 +307,7 @@ def request(
         headers: dict[str, str] | None = None,
         encoding: str | None = None,
         authorizer: GlobusAuthorizer | None = None,
+        caller_info: RequestCallerInfo | None = None,
```
sirosen, line 310: I think we should make this parameter non-nullable -- callers will be required to pass it.

I'd be open to updating the signature to
```
method: str,
url: str,
*,
caller_info: RequestCallerInfo,
...
```

so that it's keyword-only and we slightly update the requirement to pass keyword arguments.

Making it required removes a level of conditionality from the checks against it, and I don't think it does us any harm in the base client usage which is responsible for passing this.

sirosen (commented):
The intent with that card, although I now see that I did not write it unambiguously, was that we replace `authorizer` with the new caller info object. Could you update to do so? The lack of a need for backwards compatibility in v4 is why we benefit from doing this now, in particular.

changelog.d/20250716_105609_8730430+m1yag1_sc_43037_caller_info.rst
```diff
@@ -0,0 +1,4 @@
+Added
+-----
+
+- Add ``RequestCallerInfo`` data object to ``RequestsTransport.request`` for passing caller context information. (:pr:`NUMBER`)
```
MaxTueckeGlobus, line 4:
```suggestion
- Add ``RequestCallerInfo`` data object to ``RequestsTransport.request`` for passing caller context information. (:pr:`1261`)
```

MaxTueckeGlobus (commented):
Just want to quick make a note of this for now, I'll hold off on the main body of my review until Stephen's feedback has been resolved.

src/globus_sdk/transport/requests.py
```diff
@@ -335,15 +340,27 @@ def request(
...
+        # Always ensure we have caller_info
+        if caller_info is None:
+            caller_info = RequestCallerInfo(authorizer=authorizer)
+        elif authorizer is not None and caller_info.authorizer is not authorizer:
+            # Warn if both parameters are provided but differ
+            log.warning(
+                "Both 'caller_info' and 'authorizer' parameters provided to "
+                "request(). Using authorizer from caller_info and ignoring "
+                "standalone authorizer parameter."
+            )
```
kurtmckee, line 353: I don't know if this was a TAC item but I would prefer for this to be an exception, not a warning.

My reasoning is that no existing code will be affected by this new `caller_info` parameter. However, developers that make a deliberate choice to use the `caller_info` parameter are opting into a new contract. Therefore, if they pass both in, I think that we're safe to raise an exception and explicitly say "don't pass both".

It's a clearer contract than a warning, I think.

changelog.d/20250716_105609_8730430+m1yag1_sc_43037_caller_info.rst
m1yag1: @MaxTueckeGlobus I've left this same comment on a SDK PR as well! There is automation for adding the PR number into the docs upon merge so this isn't necessary. https://github.com/globus/globus-sdk-python/blob/main/.github/workflows/readthedocs-pr-links.yaml

tests/unit/transport/test_transport_authz_handling.py
```diff
@@ -28,3 +28,27 @@ def test_will_null_authz_header_with_null_authorizer():
...
+
+    with mock.patch.object(transport, "session") as mock_session:
+        mock_response = mock.Mock(status_code=200)
+        mock_session.send.return_value = mock_response
+
+        response = transport.request(
+            "GET", "https://example.com", caller_info=caller_info
+        )
+
+        assert response.status_code == 200
```
sirosen, line 56: I think I'd like to see the headers on the sent request here, to confirm that `headers={"Authorization": "Bearer token"}` -- meaning that the "unwrapping" of this data was correct.

src/globus_sdk/transport/retry.py
```diff
@@ -24,23 +25,23 @@ class RetryContext:
...
     :param exception: The error raised when trying to send the request
-    :param authorizer: The authorizer object from the client making the request
     """
 
     def __init__(
         self,
         attempt: int,
         *,
-        authorizer: GlobusAuthorizer | None = None,
+        caller_info: RequestCallerInfo | None = None,
```
sirosen, line 37: I don't think the caller info should be nullable. The inner layer (`authorizer`) is, but I don't think there's utility/value in allowing the wrapper itself to be omitted.

`RetryContext` is only constructed inside of the transport layer.

sirosen (changes requested):

src/globus_sdk/transport/requests.py
m1yag1: I've made updates that remove this code, so this comment no longer applies.

src/globus_sdk/transport/retry.py
m1yag1: This is no longer nullable and all the tests have been updated that broke (~13 tests)

sirosen (approved):

changelog.d/20250716_105609_8730430+m1yag1_sc_43037_caller_info.rst
MaxTueckeGlobus: I had no idea, good to know for the future!

PR #1265, opened 2025-07-17
Make 'v1' token storage less accessible

sirosen:
NB: This isn't something we've discussed as a team, but I think it aligns with our general naming and deprecation schedule without adding scope-creep to our SDK v4 milestone.

I noted an example which is still showing the 'v1' interface, which I intend to update in a separate follow-up.

---

This is a small internal change which reduces the size of the ``globus_sdk.token_storage`` top-level interface.

It has impact on code completions and the general developer experience when using these legacy tools.

A new section is added to the upgrading guide on the changes in names (not previously covered).

----
📚 Documentation preview 📚: https://globus-sdk-python--1265.org.readthedocs.build/en/1265/

### DIFF
--- a/changelog.d/20250717_154854_sirosen_bury_v1_tokenstorage.rst
+++ b/changelog.d/20250717_154854_sirosen_bury_v1_tokenstorage.rst
@@ -1,0 +1,8 @@
+Changed
+-------
+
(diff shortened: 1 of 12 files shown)


kurtmckee:
Introducing a `v1` module on the march to `v4` doesn't strike me as desirable.

PR #1268, opened 2025-07-18
Rename 'function_data' to 'data'

sirosen:
This was desired shortly after the first release containing
`function_data`, but we held the change back at the time to wait for
SDK v4 to make a technically breaking change.

See also: #1092


kurtmckee (approved):
Tagging @rjmello for review as well.

rjmello (approved):

PR #1271, opened 2025-07-18
Remove 'client_id' from 'AuthClient'

sirosen:
----
📚 Documentation preview 📚: https://globus-sdk-python--1271.org.readthedocs.build/en/1271/

### DIFF
--- a/tests/unit/test_auth_clients.py
+++ b/tests/unit/test_auth_clients.py
@@ -9,39 +9,6 @@ CLIENT_ID_UUID = uuid.uuid4()
 CLIENT_ID_STR = str(CLIENT_ID_UUID)
 
 
-def test_service_client_does_not_require_client_id():
-    client = globus_sdk.AuthClient()
-    # accessing the attribute warns, but provides None
-    with pytest.warns(globus_sdk.RemovedInV4Warning):
-        assert client.client_id is None
-
-
-pass_value_params = pytest.mark.parametrize(
-    "pass_value", (CLIENT_ID_STR, CLIENT_ID_UUID), ids=("str", "uuid")
-)
-
-
-@pass_value_params
-def test_service_client_allows_client_id_but_warns(pass_value):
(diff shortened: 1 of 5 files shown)


tests/unit/test_auth_clients.py
```diff
@@ -74,3 +43,9 @@ def test_confidential_app_auth_client_rejects_authorizer():
         globus_sdk.ConfidentialAppAuthClient(
             CLIENT_ID_UUID, "foo-secret", authorizer=authorizer
         )
+
+
+def test_service_client_rejects_client_id():
+    # init will warn because a value is being passed
+    with pytest.raises(TypeError):
+        globus_sdk.AuthClient(client_id=CLIENT_ID_STR)
```
kurtmckee, line 51: This test isn't helpful. It's testing the Python language itself, not SDK code.

```suggestion
```

sirosen: > I suspect that its utility is not measured in years, and possibly not even in weeks.

Yeah, that I agree with.

In fact, I think it's more an artifact of "demonstrate that this change did the right thing and none of the past wrong things", and that it has already outlived its useful life.

The comment is plain wrong, so I have to touch this to fix anyway. Why don't I remove it, and we'll move on? My objection was more one of principle than a measure of the importance of this specific test.

I may try adding a typing test that says "mypy will reject it if you try to instantiate AuthClient with a client_id", which is more in line with the kind of contractual testing we have in the typing tests.

sirosen: Having just written https://github.com/globus/globus-sdk-python/pull/1271/commits/8760539420b1b051303f93b4ad32ac02e6da47ac , I now have a new guideline for myself:
"If you're checking `TypeError`, maybe you actually want a type-checking test?"

Curious if that reframing of "what the test is about" seems as good to you as it does to me.

kurtmckee (approved):

PR #1281, opened 2025-07-23
Merge v3.61.0 back to 4.x-dev

sirosen:
This is a merge-back of v3.61.0 into v4.
Notably, this last v3 release includes deprecations for things which are being removed in v4. So it may seem strange to see the "introduction" of more code around these (most notably, `ComputeClient`), but these additions will be removed in upcoming changes.

----
📚 Documentation preview 📚: https://globus-sdk-python--1281.org.readthedocs.build/en/1281/


m1yag1 (approved):
Only nitpick is what appears to be a fixup commit [Apply suggestions from code review](https://github.com/globus/globus-sdk-python/pull/1281/commits/7dde798fbd0562b235309344c9757a747aab0ca6).

sirosen:
That commit is in `main`, so I think it would be categorically wrong to do anything about it here.

PR #1227, opened 2025-06-18
[sc-38573] implement GCP subscription admin verification

pjhinton-globus:
**Shortcut Story**: [sc-38573 / SDK/CLI support for verified data collections](https://app.shortcut.com/globus/story/38573/sdk-cli-support-for-verified-data-collections)

This pull request implements support for setting the subscription administrator verification status on a Globus Connect Personal collection.

* Add method `set_subscription_admin_verified()` to `TransferClient` class.
* Include reference guide in method docstring.
* Add mock responses for success and  failure paths on API.
* Add functional tests to exercise happy path and all failure modes.
* Add changelog fragment.

----
📚 Documentation preview 📚: https://globus-sdk-python--1227.org.readthedocs.build/en/1227/

PR #1287, opened 2025-07-25
Support setting a group's `subscription_admin_verified_id`

cwhead-globus:
This change is intended to support setting a group's `subscription_admin_verified_id` through the Globus SDK. This leverages the groups route added as part of [sc-38858](https://app.shortcut.com/globus/story/38858/add-support-for-marking-groups-as-subscription-admin-verified).

- New `GroupsClient` method
  - Required Arguments:
    - `group_id`
    - `subscription_id` (can be `None` to unverify a group)
- Test to confirm behavior

----
📚 Documentation preview 📚: https://globus-sdk-python--1287.org.readthedocs.build/en/1287/

### DIFF
--- a/tests/functional/services/groups/test_set_subscription_admin_verified_id.py
+++ b/tests/functional/services/groups/test_set_subscription_admin_verified_id.py
@@ -1,21 +1,19 @@
-from tests.common import register_api_route_fixture_file
+import json
+
+from globus_sdk._testing import get_last_request, load_response
 
 
 def test_set_subscription_admin_verified_id(groups_client):
-    register_api_route_fixture_file(
-        "groups",
-        "/v2/groups/d3974728-6458-11e4-b72d-123139141556/subscription_admin_verified",
-        "set_subscription_admin_verified_id.json",
-        method="PUT",
-    )
+    meta = load_response(groups_client.set_subscription_admin_verified_id).metadata
 
     res = groups_client.set_subscription_admin_verified_id(
-        group_id="d3974728-6458-11e4-b72d-123139141556",
-        subscription_id="43db2f9f-601b-4c36-bbaa-26a97555935c",
+        group_id=meta["group_id"],
+        subscription_id=meta["subscription_id"],
     )
     assert res.http_status == 200
-    assert res.data["group_id"] == "d3974728-6458-11e4-b72d-123139141556"
-    assert (
-        res.data["subscription_admin_verified_id"]
-        == "43db2f9f-601b-4c36-bbaa-26a97555935c"
-    )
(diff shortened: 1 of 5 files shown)


tests/functional/services/groups/test_set_subscription_admin_verified_id.py
```diff
@@ -0,0 +1,21 @@
+from tests.common import register_api_route_fixture_file
+
+
+def test_set_subscription_admin_verified_id(groups_client):
+    register_api_route_fixture_file(
+        "groups",
+        "/v2/groups/d3974728-6458-11e4-b72d-123139141556/subscription_admin_verified",
+        "set_subscription_admin_verified_id.json",
+        method="PUT",
+    )
```
sirosen, line 10: I need to think about how we can do better in-repo contributor/developer messaging. I'd like to deprecate and eventually remove `register_api_route_fixture_file`, and we have an alternative method that it would be better to use.

Sorry about this! I'll give full info on the newer path here:

We now have testing data provided to the SDK's own testsuite and to users via `globus_sdk._testing`:
https://github.com/globus/globus-sdk-python/tree/main/src/globus_sdk/_testing/data

This doc on usage shows what we're trying to provide: https://globus-sdk-python.readthedocs.io/en/stable/testing/getting_started.html
And here's an example of it being used downstream of the SDK, in the CLI:
https://github.com/globus/globus-cli/blob/3d8034036eb5963a684dfe4119b7462902563680/tests/functional/timer/test_job_operations.py#L102-L108

If we define the test response data in `src/globus_sdk/_testing/data/groups/set_subscription_admin_verified_id.py`, it will be possible to register the mock by calling `load_response`. 
And the object we get back from that `load_response()` call can contain the `group_id` and `subscription_admin_verified_id` values, for use in the test.
Something like:
```python
meta = load_response(groups_client.set_subscription_admin_verified_id).metadata
group_id = meta["group_id"]
sub_id = meta["subscription_id"]
```

One of the neat things about passing data back to a test from the fixture is that the test no longer needs to have a hardcoded `group_id` which _must_ match the fixture data.

I'll point at two examples, which are sort of at far ends of the spectrum in terms of what you could/should do.
First, looking at recent PRs, I think #1227 is pretty much exemplary. It defines several test response cases, not just the `default` one, and then tests the success case + parsing of errors.
However, I didn't have to go too far back to find a PR of my own which doesn't do as good of a job! #1155 shows what I'd do for the typical case, where we only want to declare a default response and not worry about the failure modes, etc.

sirosen (approved):

kurtmckee (approved):

PR #1293, opened 2025-07-29
Remove SearchClient.create_entry

sirosen:
----
📚 Documentation preview 📚: https://globus-sdk-python--1293.org.readthedocs.build/en/1293/


ada-globus:
Yeah, I think it costs us little and is a nice affordance to the user. I don't think much copy is needed—effectively just an english language redirect (since nothing else in v4 will tell them where to go... the signpost is removed with this change!)

sirosen:
Since this needed a rebase anyway, I've included the doc here rather than splitting it to a separate PR.

PR #1294, opened 2025-07-29
Remove the "SearchQuery" type

sirosen:
Additionally, remove the base type which it depended upon and the `set_marker` helper on the scroll query type.
This provides better uniformity between our query helper type and the scrolling query helper type.

If users indicate that they found the setter methods useful, we can reintroduce them.

----
📚 Documentation preview 📚: https://globus-sdk-python--1294.org.readthedocs.build/en/1294/


ada-globus (approved):
Do we need a note in the upgrading guide to direct people to `SearchQueryV1`?

PR #1299, opened 2025-08-05
Put error text in `__notes__` where supported

sirosen:
For Python 3.11+, `add_note()` and `__notes__` provide a mechanism for us to greatly enhance the representation of an error.

---

This enhancement is motivated by the frequency with which users building scripts and applications on top of the SDK handle a traceback (e.g., in logs) well removed from the actual error context.
In some cases, we are asking users to retrieve `err.text` in order to get detailed information on what went wrong.

Instead of reactively looking for this data, the library can proactively push it into tracebacks without impacting programmatic usage.
With large errors, this can make tracebacks very long, but traces are already diagnostic info and contain exception chains, so I'm not worried about the downsides here.

----
📚 Documentation preview 📚: https://globus-sdk-python--1299.org.readthedocs.build/en/1299/

### DIFF
--- a/changelog.d/20250805_143650_sirosen_exception_add_note.rst
+++ b/changelog.d/20250805_143650_sirosen_exception_add_note.rst
@@ -1,0 +1,6 @@
+Added
+-----
+
+- On Python 3.11+, the SDK will populate the ``__notes__`` of API errors with a
+  message containing the full body of the error response.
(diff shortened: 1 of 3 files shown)


kurtmckee:
It's the explanatory text. I see that it's separated by double newlines, so it's parseable.

kurtmckee (approved):

PR #1300, opened 2025-08-09
Raise TypeError on invalid Scope constructions

sirosen:
Type hints declare that `Scope.with_dependency` and
`Scope.with_dependencies` require `Scope` objects as their input.
This new requirement makes construction simpler, with fewer cases, but
is easy to accidentally violate when manipulating Scopes and strings
together.

Some existing tests were actually constructing invalid scopes by
passing strings and only incidentally passing.

A Scope containing dependencies which are strings is invalid -- we
offer no guarantees around how such an object might behave.

To fix, TypeErrors are added to `with_dependency()` and
`with_dependencies()` in order to catch string inputs.

----
📚 Documentation preview 📚: https://globus-sdk-python--1300.org.readthedocs.build/en/1300/

### DIFF
--- a/src/globus_sdk/scopes/representation.py
+++ b/src/globus_sdk/scopes/representation.py
@@ -77,6 +77,11 @@ class Scope:
 
         :param other_scope: The scope upon which the current scope depends.
         """
+        if not isinstance(other_scope, Scope):
+            raise TypeError(
+                "Scope.with_dependency() takes a Scope as its input. "
+                f"Got: '{type(other_scope).__qualname__}'"
+            )
         return dataclasses.replace(
             self, dependencies=self.dependencies + (other_scope,)
         )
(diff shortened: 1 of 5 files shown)


src/globus_sdk/scopes/representation.py
```diff
@@ -89,8 +94,16 @@ def with_dependencies(self, other_scopes: t.Iterable[Scope]) -> Scope:
 
         :param other_scopes: The scopes upon which the current scope depends.
         """
+        other_scopes_tuple = tuple(other_scopes)
+        for i, item in enumerate(other_scopes_tuple):
+            if not isinstance(item, Scope):
+                raise TypeError(
+                    "Scope.with_dependencies() takes "
+                    "an Iterable of Scopes as its input. "
```
kurtmckee, line 102: I don't think that Python class capitalization is needed here. Suggesting:

```suggestion
                    "an iterable of Scopes as its input. "
```

kurtmckee (approved):

PR #1302, opened 2025-08-26
Rename 'set_subscription_admin_verified_id'

sirosen:
Remove the '_id' suffix, per discussion with the Groups team.
The new name is 'set_subscription_admin_verified'.

The old function is still available as an alias, but emits a
`RemovedInV4Warning`.

The test fixture data in `globus_sdk._testing` is slightly adjusted, but
not considered a breaking change due to the looser compatibility policy
around the provided fixtures.

---

When this merges back to 4.x, I will need to apply a change to remove the
deprecated method and test for it.


sirosen:
I'm going to merge in spite of the CI failure because I think that RTD action is now broken. It's marked deprecated and we're supposed to migrate to the RTD app integration.

PR #1311, opened 2025-09-17
Add a `FlowTimer` class to aid in creating timers that run flows

kurtmckee:
Added
-----

-   Add a ``FlowTimer`` class to aid in creating timers that run flows.


sirosen (approved):
Two quick notes as I approve:

- I'm not sure if we should be releasing this in the SDK ahead of service support? I think maybe we approve and hold this PR? I'll raise the question with the team.

- A few of these changes will need adjustment in the 4.x-dev branch -- we'll want to make sure that `_testing` -> `testing` is handled, along with `PayloadWrapper` -> `Payload`

PR #1320, opened 2025-09-29
Fix a Poetry deprecation warning in the test suite

kurtmckee:
The test suite warning is:

```
The "poetry.dev-dependencies" section is deprecated
and will be removed in a future version.
Use "poetry.group.dev.dependencies" instead.
```


sirosen (approved):
I saw this earlier too! I forgot to go back and fix it. Nice! 🚀

PR #1321, opened 2025-09-29
Merge #1317 to `4.x-dev`

kurtmckee:


changelog.d/20250929_092604_kurtmckee_update_auth_get_my_groups.rst
```diff
@@ -0,0 +1,4 @@
+Added
+-----
+
+- Add the ``statuses`` parameter to ``GroupsClient.get_my_groups()``. (:pr:`NUMBER`)
```
sirosen, line 4: Wait a tick, I just noticed that this is still marked with `NUMBER`, before the auto-fixer gets to it. In `main` it should be updated. Should we manually fix here? I think so.

sirosen: Okay! Worst case, we fix it later. :smile:

PR #1324, opened 2025-10-03
Implement a model of resource ownership which allows apps to be used as context managers, triggering resource teardown on close

sirosen:
Introduce the '_resource_owner' internal protocol

Several resources in the SDK now define a tracking protocol to allow
resources to participate in a cascading `close()` via a GlobusApp.

The design breaks down into two major sections.

## Ownership Tracking

- `_resource_owner: BaseClient | None` is a new property of transports
- `_resource_owner: GlobusApp | None` is a new property of token storage
- `_resource_owner: GlobusApp | None` is a new property of clients
- `_owned_clients: MutableSet[BaseClient]` is a new property of apps

These are the storage mechanisms being used. The `_owned_clients` set is
defined as a `weakref.WeakSet`, meaning that membership in the set will
not prevent garbage collection of the members (which implicitly removes
them).

Each resource "may know its owner" by having the `_resource_owner`
assigned, and apps keep tabs on "owned clients" (which are otherwise
not associated).

The rules for when these are set are as follows:
- When an app is attached to a client (as through init) it is stored as
  `_resource_owner`
- When an app is attached to a client, the client is added to the app's
  `_owned_clients`
- When a client creates a transport, it sets itself as the transport's
  `_resource_owner`
- When an app creates a token storage, it sets itself as the token
  storage's `_resource_owner`

## Close Semantics

- All token storages now define `close()`, but it is still only
  meaningful for some -- the base version is a no-op
- `BaseClient._close()` is a new (private) method which closes the
  underlying transport IFF the client is the owner
- `GlobusApp.close()` is a new (public) method which closes (`_close()`)
  all apps owned by the client and any token storage owned by the client
- `GlobusApp` supports use as a context manager, with `close()` on exit

-----

- Introduce the '_resource_owner' internal protocol
- Add unit tests for resource ownership + close
- Make User Guide use GlobusApp context managers
- Add explicit narrative doc for GlobusApp.close()

### DIFF
--- a/changelog.d/20251002_135923_sirosen_resource_ownership.rst
+++ b/changelog.d/20251002_135923_sirosen_resource_ownership.rst
@@ -1,0 +1,21 @@
+Added
+-----
+
+- ``GlobusApp`` usages now support resource tracking in which they will close
+  the transports used by registered clients and the underlying token storage
+  when via a ``close()`` method. (:pr:`NUMBER`)
+
+  - ``GlobusApp`` now features a ``close()`` method for doing imperative
+    resource cleanup.
+
+  - ``GlobusApp`` can be used as a context manager, and calls ``close()`` on
+    exit.
+
+  - Transports which are created by clients, and token storages which
+    are created by apps, are subject to this cleanup behavior. Imperatively
+    created ones which are passed in to clients and apps are not.
+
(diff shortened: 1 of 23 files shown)


sirosen:
This branch of work was continued, literally (commits preserved) in #1326

PR #1338, opened 2025-11-12
Added install make/tox recipes; modified contributing doc

derek-globus:
## What?
* Added a new tox environment: `install-venv` which installs a tox environment as though it were a virtual environment at `./.venv`
* Added a new make recipe: `make install` which installs a virtualenv and pre-commit
* Update the contributing docs

## Why?
The steps to get an IDE set up to function with the SDK is not apparent from the contributing docs.

Even if IDEs are not the exclusive development environment used by sdk contributors they are a core one. We should have a simpler path to encourage contribution from new contributors.

## Testing
`make clean && make install`

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -44,9 +44,6 @@ And linting with
 Optional, but recommended, linting setup
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If you didn't use the pre-canned `make install` above, we recommend setting up linting
-and autofixing integrations with your editor and git.
-
 The Globus SDK uses  https://pre-commit.com/[`pre-commit`] to automatically run linters
 and fixers. Install `pre-commit` and then run
 


tox.ini
```diff
@@ -14,6 +14,11 @@ minversion = 4.22.0
 labels =
     freezedeps = freezedeps-print,freezedeps-py{3.13,3.12,3.11,3.10,3.9,3.8}
 
+[testenv:install-venv]
+envdir = {toxinidir}/.venv
+allowlist_externals = echo
+commands = echo "Virtualenv installed (via tox) at {envdir}"
```
kurtmckee: Nice!

I _think_ this is inheriting something from the bare `testenv` environment that I want to double-check if you haven't already: I think it's installing a packaged version of the SDK, rather than an editable version.

If it's editable, you should see a `.pth` file in `.venv/lib/python3.13/site-packages/` that contains a path. If it isn't, you'll just see `globus_sdk/` in that directory, and that will lead to unexpected results.

PR #1339, opened 2025-11-14
Auto-retry gares in a GlobusApp

derek-globus:
## What?
* Added automatic GARE-retry to GlobusApps.
* Updated the GARE docs to explain how to use GlobusApp redrive ([here](https://globus-sdk-python--1339.org.readthedocs.build/en/1339/authorization/gare.html#globusapp-integration))

## How?
* When enabled on a GlobusApp, any attached client receive a new transport-level retry check. This check (once per http request handling) will intercept HTTP 403 code errors & attempt to interpret them as a GARE. If successful, the check will drive a login flow using the GlobusApp-registered LoginFlowManager then resubmit the original request with updated tokens.

## Notable Decisions

* GARE-initiated logins do persist new scope requirements, just use them for the login.
   *  All authorization parameters, "required scopes" included, are treated as a "dirty bit," i.e., requested on exclusively the next login.
* Auto-redrive is disabled by default.
   * This is (hopefully) a super useful feature, but could also be quite surprising if you aren't expecting it. For now, at least, we can start with it off & re-evaluate down the line based on adoption whether we want to keep it that way.
* resource_server is only included in a request context when an app is present
  *  The current `GCSClient` uses its own `request` method to load the `resource_server` property when requested (queries it from the collection info endpoint). This results in an infinite recursion of the form `get_collection(...) -> resource_server -> request(...) -> resource server -> request(...)`
  * Ultimately, we probably need a to decouple request from our resource server call but that issue already exists in the SDK so I opted to not fix it as a part of this scope of work.

## Why?

* OAuth2 is designed to accommodate for an interaction mode wherein resource servers may dynamically return an error to "you need to log in more or better". This response is intended to inform the client that they need to request some combination of new externally-approved consents, logins, authentications, etc. before attempting the request again or else it will fail once more.
* This response pattern is sufficiently standard it can be replicated inside of the SDK, thanks largely to the fact that these errors have a well-defined schema (Globus Auth Requirements Objects, AKA GAREs).

This change introduces that execution pattern as a built-in feature to GlobusApp, reducing even further the amount of boiler-plate code required by certain services to have a successful api interaction.

## Testing

Repeatedly calling "show" on a flow in sandbox with an aggressive (20 second) authentication timeout policy.

Full script (please reach out to me if you'd like to test this locally & I can grant access to the flow).
```python
from pprint import pprint

from globus_sdk import UserApp, GlobusAppConfig, FlowsClient

HA_FLOW_ID = "05ad3012-a2f2-46d6-8f11-1fd924cd1d59"

app = UserApp(
    client_id="<get-your-own-native-client-id!>",
    config=GlobusAppConfig(
        environment="sandbox",
        login_flow_manager="local-server",
        auto_redrive_gares=True,
    )
)

flows = FlowsClient(app=app)
pprint(flows.get_flow(HA_FLOW_ID).data)
```

### DIFF
--- a/tests/unit/test_base_client.py
+++ b/tests/unit/test_base_client.py
@@ -1,5 +1,3 @@
-from __future__ import annotations
-
 import json
 import logging
 import os
@@ -9,12 +7,7 @@ from unittest import mock
 import pytest
 
 import globus_sdk
-from globus_sdk import (
-    GlobusApp,
-    GlobusAppConfig,
-    GlobusSDKUsageError,
-    UserApp,
-)
+from globus_sdk import GlobusApp, GlobusAppConfig, GlobusSDKUsageError, UserApp
 from globus_sdk.authorizers import NullAuthorizer
 from globus_sdk.scopes import Scope, TransferScopes
 from globus_sdk.testing import RegisteredResponse, get_last_request
--- a/tests/unit/globus_app/test_auto_gare_redrive.py
+++ b/tests/unit/globus_app/test_auto_gare_redrive.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import typing as t
 import uuid
 from unittest.mock import MagicMock, Mock


docs/authorization/gare.rst
```diff
@@ -74,6 +74,44 @@ classes in an ``extra`` attribute.
...
+executing user or client is automatically prompted to remediate manually-required
+steps (e.g., consent, MFA) without any additional code.
+
+
+.. code-block:: python
+
+    from globus_sdk import GlobusApp, GlobusAppConfig, TransferClient
+
+    config = GlobusAppConfig(auto_redrive_gares=True)
+    with GlobusApp("my-gare-demo", "<client-id>", config=config) as app:
```
sirosen: Should the example use `UserApp`?

src/globus_sdk/client.py
```diff
@@ -78,6 +78,9 @@ class BaseClient:
     #: the scopes for this client may be present as a ``ScopeCollection``
     scopes: ScopeCollection | None = None
 
+    # internal flag to avoid recursion when loading resource_server
+    _is_loading_resource_server: bool = False
+
```
sirosen: I think this is unused? It was something that concerned me a bit, but maybe purely vestigial from an earlier draft.

sirosen (commented):
Does it expand scope here too much to also update the basic transfer example doc?
( https://globus-sdk-python--1339.org.readthedocs.build/en/1339/user_guide/usage_patterns/data_transfer/submit_transfer/index.html#transferring-data-where-at-least-one-collection-is-unknown )

I wouldn't want to make this PR responsible for all doc updates, but I think it's especially pertinent to that section of the user guide.

docs/authorization/gare.rst
derek-globus: Yeah it should, good catch.

tests/unit/test_base_client.py
```diff
@@ -240,6 +258,161 @@ def _reraise_token_error(_: GlobusApp, error: TokenValidationError):
     assert str(ex.value) == "No token data for transfer.api.globus.org"
 
 
+class GlobusAppConfigurator:
```
sirosen: 🤩 What a name!

derek-globus:
> Does it expand scope here too much to also update the basic transfer example doc?

Sure, I can update that doc as well to just turn on GARE redrive.

tests/unit/test_base_client.py
```diff
@@ -240,6 +258,161 @@ def _reraise_token_error(_: GlobusApp, error: TokenValidationError):
...
+    app = UserApp(client_id="client_id", config=config)
+    transfer = TransferClient(app=app)
+
+    # Set up exclusively a GARE response which will retry indefinitely.
+    _GET_TASK_GARE_RESPONSE.add()
+    with pytest.raises(globus_sdk.GlobusAPIError):
+        transfer.get_task("foobar")
+    assert configurator.login_flow_manager.login_count == 1
+
+
```
sirosen: I like all of these tests, but I feel that they're in the wrong test module. (This module is already a little bit of a hodgepodge, so that's not a new issue...)
If I were _looking for_ these tests, let's say to add another one, I wouldn't think to look in `test_base_client.py`, even though we are testing the integration between these components.

Could we maybe create a new module in `tests/functional/globus_app/` for these? Or maybe `tests/unit/globus_app/`?

`functional` vs `unit` tests are a bit blurry for the SDK, so I'm less worried about that distinction than I am about "tests of the client" vs "tests of the app".

sirosen (changes requested):

tests/unit/test_base_client.py
derek-globus: GlobusAppConfigConfig just didn't have the right ring to it.

PR #1346, opened 2025-11-26
Update the Client Credentials Example for v4

sirosen:
This example still omitted the requested scopes for client credentials and relied on the default (which was removed in v4).

Also swap in our symbolic constants from the TransferClient class, rather than the exact strings needed.

---

Hat tip to @jweirich-globus for spotting this! :tada:


jweirich-globus (approved):
Thanks for the update! 🙏

m1yag1 (approved):
I copied the examples down verbatim and ran them as a user would. It all worked!

PR #1350, opened 2025-12-04
Very marginal cleanup: startswith usage

sirosen:
A single startswith call, rather than an `or` (which will usually *not*
short-circuit) will make every call to send a request via SDK clients a
little faster.

### DIFF
--- a/src/globus_sdk/client.py
+++ b/src/globus_sdk/client.py
@@ -540,7 +540,7 @@ class BaseClient:
 
         # if a client is asked to make a request against a full URL, not just the path
         # component, then do not resolve the path, simply pass it through as the URL
-        if path.startswith("https://") or path.startswith("http://"):
+        if path.startswith(("https://", "http://")):
             url = path
         else:
             url = slash_join(self.base_url, urllib.parse.quote(path))


kurtmckee:
@sirosen If I'm interpreting your narrative text correctly, I need to point out that this is not a speed consideration because `.startswith("http://")` is not pre-evaluated and then fallen back to as a boolean value. That is, Python doesn't call `.startswith()` twice and then interpret the `or`:

```python
return s.startswith("https://") or s.startswith("http://")
     # ^^^^^^^^^^^^^^^^^^^^^^^^    ^^^^^^^^^^^^^^^^^^^^^^^
     # True                        False

return True                     or False
```

Short-circuiting in Python completely prevents evaluation until it's needed. Thus Python interprets the above in this way:

```python
return s.startswith("https://") or s.startswith("http://")
     # ^^^^^^^^^^^^^^^^^^^^^^^^
     # True

return True
```

You can see that evaluation of later values is completely suspended until preceding evaluations require it:

```python
print(True  or  (1/0))  # prints 'True'
print(False and (1/0))  # prints 'False'

print(False or  (1/0))  # ZeroDivisionError
print(True  and (1/0))  # ZeroDivisionError
```

PR #1351, opened 2025-12-04
Add support for Globus Tunnels

buzztroll:
It includes the API calls needed by programs like the globus-cli.

### DIFF
--- a/src/globus_sdk/services/transfer/client.py
+++ b/src/globus_sdk/services/transfer/client.py
@@ -13,7 +13,7 @@ from globus_sdk._missing import MISSING, MissingType
 from globus_sdk.scopes import GCSCollectionScopes, Scope, TransferScopes
 from globus_sdk.transport import RetryConfig
 
-from .data import DeleteData, TransferData
+from .data import CreateTunnelData, DeleteData, TransferData
 from .errors import TransferAPIError
 from .response import IterableTransferResponse
 from .transport import TRANSFER_DEFAULT_RETRY_CHECKS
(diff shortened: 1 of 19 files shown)


src/globus_sdk/services/tunnels/client.py
```diff
@@ -0,0 +1,330 @@
...
+    >>>     print(item["name"], item["type"])
+    """
+
+    default_iter_key = "data"
+
+
+class TunnelClient(client.BaseClient):
+    r"""
+    Client for the
+    `Globus Transfer API <https://docs.globus.org/api/v2/tunnels/>`_.
```
kurtmckee: This URL doesn't exist (at least, not currently).

src/globus_sdk/services/tunnels/client.py
```diff
@@ -0,0 +1,330 @@
...
+
+    .. sdk-sphinx-copy-params:: BaseClient
+
+    This class provides helper methods for most common resources in the
+    REST API, and basic ``get``, ``put``, ``post``, and ``delete`` methods
+    from the base rest client that can be used to access any REST resource.
```
kurtmckee:
```suggestion
    from the base REST client that can be used to access any REST resource.
```

src/globus_sdk/services/tunnels/client.py
```diff
@@ -0,0 +1,330 @@
...
+        lifetime_mins: int | None = None,
+        *,
+        query_params: dict[str, t.Any] | None = None,
+    ) -> response.GlobusHTTPResponse:
+        """
+        :param listener_stream_ap_id: ID of the Stream Access Point on the
+            listening side of the connection
+        :param initiator_stream_ap_id: ID of the Stream Access Point on the
+            initiating side of the connection
+        :param submission_id: And optional parameter used to track this request
```
kurtmckee:
```suggestion
        :param submission_id: An optional parameter used to track this request
```

src/globus_sdk/services/tunnels/client.py
```diff
@@ -0,0 +1,330 @@
...
+
+    service_name = "transfer"
+    scopes = TransferScopes
+    default_scope_requirements = [TransferScopes.all]
+
+    def create_tunnel(
+        self,
+        listener_stream_ap_id: str,
+        initiator_stream_ap_id: str,
+        submission_id: uuid.UUID | None = None,
```
kurtmckee: Could this be a `str` containing a UUID?


```suggestion
        submission_id: uuid.UUID | str | None = None,
```

src/globus_sdk/services/tunnels/client.py
```diff
@@ -0,0 +1,330 @@
...
+    def set_listener(
+        self,
+        tunnel_id: str,
+        listener_cs: str,
+        *,
+        query_params: dict[str, t.Any] | None = None,
+    ) -> response.GlobusHTTPResponse:
+        """
+        :param tunnel_id: The ID of the Tunnel which the application listener's
+            contact string will be set.
```
kurtmckee: This param string doesn't look correct, but double-check me on that.

```suggestion
        :param tunnel_id: The ID of the Tunnel.
```

src/globus_sdk/services/tunnels/client.py
```diff
@@ -0,0 +1,330 @@
...
+                    tc = globus_sdk.TunnelClient(...)
+                    tc.set_listener(tunnel_id, "192.168.0.23:8080")
+
+            .. tab-item:: API Info
+
+                ``PATCH /tunnels/<tunnel id>``
+        """
+        cs = listener_cs.split(":")
+        host = cs[0]
+        port = cs[1]
```
kurtmckee: I recommend this form to avoid `KeyError` in the case that someone leaves out the port, or perhaps uses IPv6 (if that's supported). After that, it may be that the host needs to be stripped of square brackets (though I'm not an expert on IPv6; I simply saw in a cursory search that IPv6 addresses with ports have square brackets around the host, like `[2008:2007::1]:8080`).

```suggestion
        host, _, port = listener_cs.rpartition(":")
```

src/globus_sdk/services/tunnels/client.py
buzztroll: 👍

src/globus_sdk/services/tunnels/client.py
```diff
@@ -0,0 +1,330 @@
...
+        *,
+        query_params: dict[str, t.Any] | None = None,
+    ) -> response.GlobusHTTPResponse:
+        """
+        :param tunnel_id: The ID of the Tunnel to be deleted.
+        :param query_params: Any additional parameters will be passed through
+            as query params.
+
+        This will clean up all data associated with a Tunnel. In order to delete a
+        Tunnel it must first be stopped.
```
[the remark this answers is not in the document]
buzztroll: 👍

src/globus_sdk/services/transfer/client.py
```diff
@@ -2699,3 +2699,199 @@ def endpoint_manager_delete_pause_rule(
...
+        """
+        log.debug("TransferClient.create_tunnel(...)")
+        if (
+            "submission_id" not in data["data"]["attributes"]
+            or data["data"]["attributes"]["submission_id"] is MISSING
+        ):
+            log.debug("create_tunnel autofetching submission_id")
+            data["data"]["attributes"]["submission_id"] = self.get_submission_id()[
+                "value"
+            ]
```
sirosen: I think the SDK should emit a UsageError, not a bare KeyError, if someone calls `create_tunnel({})`.

cc @derek-globus , @kurtmckee: I trust both of you to check me on this. Should we add new helpers for this? Or handle it inline here? I'm inclined towards inline handling until we see this occur again, even though we expect JSON:API to produce more of these nested data shapes.

src/globus_sdk/services/transfer/client.py
```diff
@@ -2699,3 +2699,199 @@ def endpoint_manager_delete_pause_rule(
...
+
+                .. code-block:: python
+
+                    tc = globus_sdk.TunnelClient(...)
+                    result = tc.create_tunnel(data)
+                    print(result["data"]["id"])
+
+            .. tab-item:: API Info
+
+                ``POST /tunnels``
```
sirosen: I see that in all of our other, preexisting method listings, we don't show the full route. That dates from SDK v3 and earlier, when the base path was implicit. We should fix those. I think the new ones should be written in full, like so:

```suggestion
                ``POST /v2/tunnels``
```

src/globus_sdk/services/transfer/client.py
```diff
@@ -2699,3 +2699,199 @@ def endpoint_manager_delete_pause_rule(
...
             query_params=query_params,
         )
+
+    # Tunnel methods
+
+    def create_tunnel(
+        self,
+        data: dict[str, t.Any] | CreateTunnelData,
+        *,
+        query_params: dict[str, t.Any] | None = None,
```
sirosen: I think we should remove `query_params` support from this call, `update_tunnel()`, and `delete_tunnel()`.

The `query_params` parameter is a useful escape-valve when users on an older SDK version want to use a newer parameter, before SDK support is implemented.[^1] But it's a convenience which we don't provide uniformly, across all API bindings, and we generally only have it on methods where it is, by typical API designs will make it useful. You can always work around its absence by calling the underlying `get`, `put`, `post`, etc methods. The norm is that `POST` and `PATCH` APIs don't encode significant data in their query params, but use the body instead.  `DELETE` is more variable, but `DELETE` with query params or a body is relatively rare.

[^1]: I could write at length about how I want to reshape the SDK's API such that this is no longer necessary. But I'd rather find some time to actually work on that refactor. 😉

sirosen (changes requested):
The main thing we're missing here is tests. We need dummy responses for each of these APIs in `src/globus_sdk/testing/data/` and unit tests which run the various methods.
We aren't at perfect coverage today, but we are aiming for 100% on any new work in the SDK.

My overall request is "an initial crack at testing, across the various new methods and new object", not "perfect and exhaustive testing".

I went back through the PR history to find useful reference PRs, and I've got two which should provide decent guidance.

#### Testing a New API

The most recent new-API was "Update Index" for Search, and looks like a pretty good example: https://github.com/globus/globus-sdk-python/pull/1310/files

Importantly! That PR demonstrates that you don't need to go through a lot of effort to build up some kind of fancy modeling of the responses. Some fixed mocks which have the right shape are totally fine. There are more sophisticated tests with more dynamic responses and whatnot, but we're literally just looking in the base case to make the testsuite run the lines added in `src/globus_sdk/services/transfer/client.py`.

#### Testing a New Payload Type

The most recent new-Payload-type was "FlowTimer" for Timers. It has some more complexity in modeling responses, but I really want to point at the test, and the way it "pulls back" the payload after sending it, to examine what was sent:
https://github.com/globus/globus-sdk-python/pull/1311/files

Using
```python
    req = get_last_request()
    sent = json.loads(req.body)
```
lets us call the relevant method and then assert on what we actually would send to the API.

That might also be a nice way to test the handling of `submission_id`.

pyproject.toml
```diff
@@ -107,7 +107,7 @@ globus_sdk = [
 # non-packaging tool configs follow
 
 [tool.pytest.ini_options]
-addopts = "--no-success-flaky-report --color=yes"
+#addopts = "--no-success-flaky-report --color=yes"
```
sirosen: :point_up: oops

src/globus_sdk/services/transfer/data/tunnel_data.py
```diff
@@ -0,0 +1,55 @@
...
+            "lifetime_mins": lifetime_mins,
+        }
+        if additional_fields is not None:
+            attributes.update(additional_fields)
+
+        self["data"] = {
+            "type": "Tunnel",
+            "relationships": relationships,
+            "attributes": attributes,
+        }
```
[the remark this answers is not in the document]
derek-globus: Seems like we're trying to expose an initialization interface which obfuscates each attribute from the type of facet it is; e.g., to instantiate this class, I don't need to know that an "initiator_stream_access_point" is a relationship or that a "submission_id" is an attribute.

I think if we want to design this class to be more explicitly mutable as you're suggesting, stephen, we should keep the interface consistent opaque in that mutation. In the example you provided, users would need to know the explicit foo/bar mappings which the initializer conveniently hides away from them.

Something like:
```python
myobj = CreateTunnelData(access_point_1, access_point2)

# Modify a direct user-facing facet which adjusts the associated data representation (attribute or relationship-nested field).
myobj.submission_id = submission_id
```

src/globus_sdk/services/transfer/client.py
```diff
@@ -2699,3 +2699,199 @@ def endpoint_manager_delete_pause_rule(
...
+        log.debug("TransferClient.list_tunnels(...)")
+        r = self.get("/v2/tunnels", query_params=query_params)
+        return IterableTransferResponse(r)
+
+    def get_stream_ap(
+        self,
+        stream_ap_id: str,
+        *,
+        query_params: dict[str, t.Any] | None = None,
+    ) -> response.GlobusHTTPResponse:
```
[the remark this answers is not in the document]
derek-globus: +1 to your proposed change, Stephen.

It's also unfortunate that access point abbreviates to AP; since in the current globus ecosystem, AP already means something: Action Provider.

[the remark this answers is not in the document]
buzztroll: I removed the abbreviation

src/globus_sdk/services/transfer/client.py
```diff
@@ -2699,3 +2699,192 @@ def endpoint_manager_delete_pause_rule(
...
+        try:
+            attributes = data["data"]["attributes"]
+            if (
+                attributes.get("submission_id", MISSING) is MISSING
+                or data["data"]["attributes"]["submission_id"] is MISSING
+            ):
+                log.debug("create_tunnel auto-fetching submission_id")
+                data["data"]["attributes"]["submission_id"] = self.get_submission_id()[
+                    "value"
+                ]
```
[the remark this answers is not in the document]
buzztroll: I refactored this

src/globus_sdk/services/transfer/data/tunnel_data.py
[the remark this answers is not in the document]
buzztroll: I am not sure what was decided here.

src/globus_sdk/services/transfer/client.py
```diff
@@ -2699,3 +2699,195 @@ def endpoint_manager_delete_pause_rule(
...
+                ``POST /v2/tunnels``
+        """
+        log.debug("TransferClient.create_tunnel(...)")
+        try:
+            data_element = data["data"]
+        except KeyError:
+            raise exc.GlobusSDKUsageError(
+                "create_tunnel() body was malformed (missing the 'data' key). "
+                "Use CreateTunnelData to easily create correct documents."
+            )
```
sirosen: pylint is mad about this right now, since it's implicit exception chaining. It wants us to be explicit:

```suggestion
        except KeyError as e:
            raise exc.GlobusSDKUsageError(
                "create_tunnel() body was malformed (missing the 'data' key). "
                "Use CreateTunnelData to easily create correct documents."
            ) from e
```

(Aside: we don't run pylint via pre-commit because it's so gosh-darn slow.)

buzztroll:
🎄 🎅 🤶

PR #1379, opened 2026-04-08
feat: [sc-47448] implement bookmark methods in Transfer v2 client

pjhinton-globus:
**Shortcut Story**: https://app.shortcut.com/globus/story/47448/

Added the following methods to the experimental TransferClientV2:

* create_bookmark()
* update_bookmark()
* get_bookmark()
* delete_bookmark()
* list_bookmarks()

This provides access to the new Transfer v2 API operations for boomkark management, which includes support for high assurance collections.

Functional unit tests are provided for all methods. Change log entry is present.

### DIFF
--- a/src/globus_sdk/experimental/transfer_v2/client.py
+++ b/src/globus_sdk/experimental/transfer_v2/client.py
@@ -10,7 +10,12 @@ from globus_sdk.scopes import TransferScopes
 from globus_sdk.services.transfer.errors import TransferAPIError
 from globus_sdk.transport import RetryConfig
 
-from .data import TunnelCreateDocument, TunnelUpdateDocument
+from .data import (
+    BookmarkCreateDocument,
+    BookmarkUpdateDocument,
+    TunnelCreateDocument,
+    TunnelUpdateDocument,
+)
 from .transport import TRANSFER_V2_DEFAULT_RETRY_CHECKS
 
 log = logging.getLogger(__name__)
@@ -131,7 +136,7 @@ class TransferClientV2(client.BaseClient):
                 .. code-block:: python
 
                     tc = globus_sdk.experimental.TrasferClientV2(...)
-                    result = tc.show_tunnel(tunnel_id)
+                    result = tc.get_tunnel(tunnel_id)
                     print(result["data"])
 
             .. tab-item:: API Info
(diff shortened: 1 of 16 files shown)


src/globus_sdk/experimental/transfer_v2/data/bookmark_documents.py
```diff
@@ -0,0 +1,87 @@
...
+from globus_sdk._payload import GlobusPayload
+
+log = logging.getLogger(__name__)
+
+
+class BookmarkCreateDocument(GlobusPayload):
+    """
+    Convenience class for constructing a bookmark document to use as the
+    ``data`` parameter to
+    :meth:`create_bookmark <globus_sdk.TransferClientV2.create_bookmark>`.
```
kurtmckee: I think that this works but I recommend this instead, which I think is more explicit and eliminates one of the duplicated method names:


```suggestion
    :meth:`~globus_sdk.TransferClientV2.create_bookmark`.
```

This is documented in the Sphinx Python domain's [cross-referencing modifiers bullet list](https://www.sphinx-doc.org/en/master/usage/domains/python.html#target-specification).

_Similar feedback applies to all other places where ``:meth:`c <a.b.c>`_`` appears._

src/globus_sdk/experimental/transfer_v2/data/bookmark_documents.py
```diff
@@ -0,0 +1,87 @@
...
+        self,
+        collection: uuid.UUID | str,
+        name: str,
+        path: str,
+        *,
+        pinned: bool | MissingType = MISSING,
+        additional_fields: dict[str, t.Any] | None = None,
+    ) -> None:
+        super().__init__()
+        log.debug("Creating a new BookmarkCreateDocument object")
```
kurtmckee, line 30: I'm surprised to see a debug statement attached to an instantiation that doesn't trigger I/O. Is this common in the SDK?

src/globus_sdk/experimental/transfer_v2/client.py
```diff
@@ -290,3 +295,140 @@ def list_stream_access_points(
...
+        """
+        :param data: Parameters for bookmark creation
+
+        .. tab-set::
+
+            .. tab-item:: Example Usage
+
+                .. code-block:: python
+
+                    tc = globus_sdk.experimental.TrasferClientV2(...)
```
kurtmckee:
```suggestion
                    tc = globus_sdk.experimental.TransferClientV2(...)
```

_Similar feedback applies throughout the code examples in this file. I recommend doing a deep search-and-replace to make sure this typo doesn't appear outside of the code I'm seeing in this one file._

src/globus_sdk/experimental/transfer_v2/client.py
```diff
@@ -290,3 +295,140 @@ def list_stream_access_points(
...
+                    data = globus_sdk.experimental.BookmarkCreateDocument(...)
+                    result = tc.create_bookmark(data)
+                    print(result["data"]["id"])
+
+            .. tab-item:: API Info
+
+                ``POST /v2/bookmarks``
+        """
+        log.debug("TransferClientV2.create_bookmark(...)")
```
kurtmckee: _Not a blocker, just a thought._

Would it make sense for these debug statements to look up the class name, rather than hard-coding it?

For example:


```suggestion
        log.debug(f"{self.__class__.__name__}.create_bookmark(...)")
```

src/globus_sdk/experimental/transfer_v2/client.py
```diff
@@ -290,3 +295,140 @@ def list_stream_access_points(
...
+
+                    tc = globus_sdk.experimental.TrasferClientV2(...)
+                    query_params = {"include": "collection"}
+                    tc.list_bookmarks(query_params)
+
+            .. tab-item:: API Info
+
+                ``GET /v2/bookmarks``
+        """
+        log.debug("TransferClientV2.list_bookmarks({query_params})")
```
kurtmckee:
```suggestion
        log.debug(f"TransferClientV2.list_bookmarks({query_params})")
```

src/globus_sdk/testing/data/transfer/v2/create_bookmark.py
```diff
@@ -0,0 +1,45 @@
...
+
+BOOKMARK_ID = str(uuid.uuid4())
+
+_collection_id = str(uuid.uuid4())
+_name = "public datasets"
+_path = "/data_repository/public"
+
+
+RESPONSES = ResponseSet(
+    default=RegisteredResponse(
```
kurtmckee, line 13: _Not a blocking request._

Would it be an option to include error responses as well?

src/globus_sdk/testing/data/transfer/v2/list_bookmarks.py
```diff
@@ -0,0 +1,65 @@
...
+            ],
+            "meta": {
+                "request_id": "xj8AWyQr8",
+            },
+            "included": [
+                {
+                    "type": "Collection",
+                    "id": b["collection_id"],
+                    "attributes": {
+                        "display_name": "Wotsomatta U Dataset Collection",
```
kurtmckee, line 56: ❤️ Rah, rah, siss-boom-bah!

kurtmckee (changes requested):

src/globus_sdk/experimental/transfer_v2/data/bookmark_documents.py
pjhinton-globus: @kurtmckee wrote:

> I'm surprised to see a debug statement attached to an instantiation that doesn't trigger I/O. Is this common in the SDK?

I modeled most of my changes on the work @aaschaer had in defining the `TransferV2Client` methods for tunnel management and their document helper classes, which logs instantiation of `TunnelCreationDocument` objects.

https://github.com/globus/globus-sdk-python/blob/f8a4a27439c680a7baf396a61c7f0ecb1253ba7e/src/globus_sdk/experimental/transfer_v2/data/tunnel_documents.py#L20-L34

src/globus_sdk/testing/data/transfer/v2/create_bookmark.py
pjhinton-globus: @kurtmckee  wrote:

> _Not a blocking request._
> 
> Would it be an option to include error responses as well?

I double-checked the API spec for these operations:

https://transfer.api.sandbox.globuscs.info/v2/docs

The only ones we have published are 422s which are raised on schema validation error. Looking under the hood at implementation, I see there are places where we can return 409s, 403s, and another case where we should return 404 (but currently 500).  What sort of completeness on error responses would the expectation be?

src/globus_sdk/experimental/transfer_v2/data/bookmark_documents.py
pjhinton-globus: @kurtmckee wrote:

> I think that this works but I recommend this instead, which I think is more explicit and eliminates one of the duplicated method names:
> 
> This is documented in the Sphinx Python domain's [cross-referencing modifiers bullet list](https://www.sphinx-doc.org/en/master/usage/domains/python.html#target-specification).
> 
> _Similar feedback applies to all other places where `` :meth:`c <a.b.c>`_ `` appears._

This is implemented in commit da44e0b16ee2b05c1ee358b96e0adcfe1135070d.

src/globus_sdk/experimental/transfer_v2/client.py
pjhinton-globus: @kurtmckee wrote:

> _Similar feedback applies throughout the code examples in this file. I recommend doing a deep search-and-replace to make sure this typo doesn't appear outside of the code I'm seeing in this one file._

This is implemented in commit da44e0b16ee2b05c1ee358b96e0adcfe1135070d.

src/globus_sdk/experimental/transfer_v2/client.py
pjhinton-globus: @kurtmckee wrote:

> _Not a blocker, just a thought._
> 
> Would it make sense for these debug statements to look up the class name, rather than hard-coding it?
> 
> For example:

This is implemented in commit da44e0b16ee2b05c1ee358b96e0adcfe1135070d.

src/globus_sdk/experimental/transfer_v2/client.py
pjhinton-globus: (adding missing f-prefix to interpolated string)

This is implemented in commit da44e0b16ee2b05c1ee358b96e0adcfe1135070d.

src/globus_sdk/experimental/transfer_v2/data/bookmark_documents.py
```diff
@@ -0,0 +1,87 @@
...
+            "attributes": attributes,
+            "relationships": relationships,
+        }
+
+
+class BookmarkUpdateDocument(GlobusPayload):
+    """
+    Convenience class for constructing a bookmark document to use as the
+    ``data`` parameter to
+    :meth:`~globus_sdk.TransferClientV2.update_bookmark>`.
```
kurtmckee:
```suggestion
    :meth:`~globus_sdk.TransferClientV2.update_bookmark`.
```

src/globus_sdk/testing/data/transfer/v2/create_bookmark.py
kurtmckee: I think that I've encountered some issues with testing Auth interactions in the past in various services and have thought "it would be nice if error responses could be mocked as easily as happy-path responses".

It's not necessary to include here, but I'm always on the lookout for opportunities to make the SDK's test framework easier for services to test their behavior with. :+1:

kurtmckee (changes requested):

src/globus_sdk/experimental/transfer_v2/data/bookmark_documents.py
pjhinton-globus: This change has been implemented in commit 2bed61839443334b08af2fdb20f69a0bf23ff6c0.

kurtmckee (approved):

sirosen (approved):
Some notes worth sharing, but nothing negative and therefore nothing to block approval!

---

I see there were some changes to logging style in other methods. I'm 👍 on that change itself, but I think that needs to be codified as a standard for us (even if only by policy and code uniformity). I would not have asked for those changes in review here, as our other classes don't follow the `self.__class__.__name__` pattern.

Without further action, things will drift. Now that we have started down this road, I think we have to see it through. I'll try to prep a PR when I have a less busy day. I'm curious to see how we can lint for this (static-analysis + runtime inspection?).

---

@pjhinton-globus, this PR started a couple of threads of conversation among the SDK maintainers, but we don't have a clear-cut plan yet for changes. I think it's important to share back what we're thinking about.

These were the things which were noted as interesting:
- It seems that the JSON:API documents benefit greatly from specialized payload types. Do we want to create a new type for each API with a body? (Across all services, this would be a large number of types.) We're not sure how to handle this.
- `additional_fields` is a name that originated from adding data to a subdocument (see: `TransferData.add_item`); not sure it's the right name for a top-level JSON:API payload. Maybe `additional_attributes`?
- Should we even _have_ `additional_fields`, or an equivalent thereof, for a payload type? You could always `doc = BookmarkCreateDocument(...); doc["data"]["attributes"]["x"] = 1`.

Regarding that last item, I think that if we create a base `JsonApiPayload(GlobusPayload)`, we could provide properties to directly access `attributes` and other important keys. So this would become possible:
```
doc = BookmarkCreateDocument(...)
doc.attributes["x"] = 1
```

Issue #1380, opened 2026-04-15
Streaming Responses Don't Stream


AK2000:
In `globus_sdk.BaseClient.request` streaming is given [as an a option,](https://github.com/globus/globus-sdk-python/blob/b957ba48d639b2bd99e3e1050bad7ae6b4adda48/src/globus_sdk/client.py#L511) but when the GlobusHTTPResponse is created, the `__init__` method by default calls `response.json()` which will wait till a streaming response closes before returning. If the json is not fetched until `_parsed_json` is called or `data` is called, this would fix the problem and allow a user to use the streaming of the `_raw_response`.

sirosen:
I think we want the `response.json()` call to be eager most of the time, since it drives the error handling semantics for the SDK. That is, when you call `foo_client.bar()`, if the "Foo API" doesn't give back JSON, that should be an immediate error (typically).

However, given that this totally breaks streaming, I think we can justify making that call conditional on whether or not `stream=True` was passed.

sirosen:
Oh, I just realized, looking at #1381 and thinking about it for a moment, that we don't propagate the error anyway. This might actually be fine to make always-lazy.

AK2000:
Closed by #1381

PR #1381, opened 2026-04-15
Lazy evaluating parsed json in GlobusHttpResponse

AK2000:
Fix for issue #1380 that works with my event streaming client. Lazy evaluating parse json in globus responses to fix streaming. This does not affect any other error handling in the test suite but it also does not use @sirosen's suggestion that we only move to lazy parsing when streaming is enabled


src/globus_sdk/response.py
```diff
@@ -67,18 +65,24 @@ def __init__(
...
+        # always use `text`
+        if self._wrapped is not None:
+            return self._wrapped._parsed_json
+
+        assert self._response is not None
+        try:
+            return self._response.json()
+        except ValueError:
+            log.warning("response data did not parse as JSON, data=None")
+            return None
```
sirosen: I'd like a very minor refactor here, to avoid using `assert` (which is stripped out if someone runs with `python -O`).
I also think we can make it clearer to ourselves, as maintainers, by leaving a `NotImplementedError`  with a nice message.

My thought is to do

```python
if self._response is not None:
    ...  # current code after the assert goes here

raise NotImplementedError(
    "Cannot read JSON data from a response which did not "
    "wrap another response or contain a transport-layer response."
)
```

sirosen (commented):
Thanks for this patch, it looks good! (And I was wrong about needing to make it conditional, so 👍 to not doing that. 😅 )

I think we need a regression unit test, but I'm happy for the SDK maintainers to take on that responsibility (unless you want to do it). I plan to add a test and ask for someone else to act as a co-reviewer.

sirosen:
I added the requisite unit test and fixed that `assert` usage. I've asked for someone else on the team to double-check here -- both to check my work and to confirm that I haven't missed anything subtle that this change might imply.

EDIT: oh, and I noticed that the changelog needed some fixing, so that's included too.

ada-globus (approved):

PR #1382, opened 2026-05-01
Simplify env var config lookups

sirosen:
Our env var loading code has not aged well with the advent of type checking -- it is very dynamically typed, and attempting to annotate it precisely is overly complex. Even with type overloads, the resulting helper is not properly typed in all cases, and cannot extend to new callback types gracefully.

Simplify env var loading to follow some common patterns without using complex helpers.

No behaviors were intentionally changed, although a TODO note was added to an odd decision to convert `""` to a default in one case.

### DIFF
--- a/src/globus_sdk/config/env_vars.py
+++ b/src/globus_sdk/config/env_vars.py
@@ -13,56 +13,50 @@ import pathlib
 import typing as t
 
 log = logging.getLogger(__name__)
-T = t.TypeVar("T")
-
 
 ENVNAME_VAR = "GLOBUS_SDK_ENVIRONMENT"
 HTTP_TIMEOUT_VAR = "GLOBUS_SDK_HTTP_TIMEOUT"
 SSL_VERIFY_VAR = "GLOBUS_SDK_VERIFY_SSL"
 
 
-@t.overload
-def _load_var(
-    varname: str,
-    default: t.Any,
-    explicit_value: t.Any | None,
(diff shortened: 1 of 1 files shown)


src/globus_sdk/config/env_vars.py
```diff
@@ -13,56 +13,50 @@
...
-    # only info log on non-default *values*
-    # meaning that if we define the default as 'foo' and someone explicitly sets 'foo',
-    # no info log gets emitted
-    if value != default:
-        log.debug(f"on lookup, non-default setting: {varname}={value}")
+def get_environment_name(inputenv: str | None = None) -> str:
+    if inputenv is None:
+        value = os.getenv(ENVNAME_VAR, "production")
     else:
-        log.debug(f"on lookup, default setting: {varname}={value}")
```
m1yag1, line 59: In the old code, the logs explicitly flagged whether the value was "default" or "non-default". I wanted to point this out in case losing that context wasn't intentional.

m1yag1 (approved):
Only one question about logging differences between the old and new code. This does not block approval.

src/globus_sdk/config/env_vars.py
sirosen: I considered it acceptable, since we know what the default values are anyway.

That conditional logging would have made each of these getters more complex (either they do it themselves or there's a shared logging helper), so absent a strong motivation to keep the old logging, I simplified it.

PR #1383, opened 2026-05-04
Support `orjson` as an alternative encoder/decoder

sirosen:
*Rebased onto #1384 to solve tox/ci config*

Add support for user selection of `orjson` as a faster and more correct
JSON encoder/decoder. Request encoding is done by sending encoded bytes
through `requests` (which is supported), and request decoding is done
via a helper which reads `requests.Response.content` (bytes).

This is primarily applied in
- the JSON request encoder
- the Response class
- the APIError base class

Additionally, GlobusApp and Transfer retry policies decode response
contents, and are here updated to load from bytes.

`mypy` runs are expanded to include `orjson` as a dependency, and
test runs include it via a factor.

New tests can directly target the encoder class, but as there is no
central point of control for decoding, there is no clearly testable
element for that purpose.

tox configuration and testing requirements are updated to pull in
`orjson` when the `-orjson` factor is used.
GitHub Actions testing is now expanded to include `-orjson` tests.


sirosen:
Although #1385 is larger in size, I think it's superior because it provides a much clearer point of control and makes `use_orjson=True/False` work for encoding and decoding. Closing in favor.

PR #1385, opened 2026-05-05
Add support for 'orjson' and introduce a 'RequestsRepresentationProvider' API for handling both serialization and deserialization

sirosen:
**Update 2026-06-01**: We discussed and adjusted the goals of this changeset. The comment which follows is therefore outdated. See [our course adjustment in this comment](https://github.com/globus/globus-sdk-python/pull/1385#issuecomment-4608889235).

## Goal

*Why?*

Provide support for ``orjson`` as an alternative way of encoding and decoding HTTP body data.
This should be _opt in_ because it could be breaking for someone, but in a future major version it should become the default.

It should be possible to set this passively to change "the whole SDK" via an env var, but for application authors, there should be explicit `True`/`False` programmatic control somewhere.

Extensibility (e.g., to other JSON implementations or other formats) is a non-goal, but the design should not be hostile to future needs.

## What Changed

Add a new `OrjsonRequestEncoder`, which mirrors the `JSONRequestEncoder` type.

Add a new `decoders` module, containing `ResponseDecoder` and `OrjsonResponseDecoder`.
Both provide a method, `get_body_json(response)` for decoding a response body as JSON.

The transport can select which encoder and decoder to use based on an init-time flag, `use_orjson=True` or an env var, `GLOBUS_SDK_USE_ORJSON=1`.
Setting these values when `orjson` is not installed raises a `RuntimeError` on transport init.

The objects doing decoding are various: responses, errors, and retry hooks.
Each of these must be customized in some way to do decoding with a configured/desired decoder.
To support this, the "currently active" transport is now available through a contextvar and a narrow getter interface, `RequestsTransport.get_current_transport()`.

Tests are extended to cover `orjson` testing, including several new tests, an `orjson` dependency group, new frozen requirements, tox config, and CI config.
Because responses read the client's decoder, a large number of test tweaks are needed to provide a "better" client mock which satisfies this requirement.

In CI and the default tox env list, we only test `orjson` on a small selection of Python versions, to moderate the expansion of our test matrix.

## Tradeoffs & Open Questions

There aren't perfect solutions in the current state of the SDK when accounting for backwards compatibility.

### `requests.Response` in public APIs

Anywhere that the SDK is passing a `requests.Response` through a public interface[^1], we have lost control of that interface.
We would now like to be passing a pair `(response, decoder)`, but changing these interfaces would be breaking.
Making changes in a future major version to always pass around our own response objects is one option to solve this.

[^1]: Some of these interfaces were originally conceived as public and some were not. So that also adds complexity.

All three of the contexts in which the decoder is passed (retry, response, api errors) are tricky.
Only one of those three (retry hooks) operates at the abstraction layer of the transport, and even that has open questions.

In all of these cases, the "get current transport" contextvar is used to sidestep our inability to fully control these APIs as we would like.
(A previous draft updated the `RetryContext`, so that one has a better solution than the others.)

### `RequestsTransport.encoders`

Currently, encoding behaviors are defined at the class level, via a mapping `RequestsTransport.encoders`.
Users can access, customize, and replace the JSON encoder today by reading or writing `RequestsTransport.encoders["json"]`.

Unfortunately, a class var acts as both the class-level and instance-level control, since instances all reference that same dict.
Any instance-level customization will be paradigm breaking here, since class-level customization will stop being effective.

In reality, SDK users are probably not touching this at all.
GitHub code searches in-org show no results, and our own usages are some of the most elaborate.
Even my own work where I have enabled `orjson` did not use this path -- not based on foresight, but because it was simpler to encode things in custom client methods.

I have chosen a moderate path, by which we
- deprecate `RequestsTransport.encoders` (users are now being told, in the docs, not to modify or use it)
- copy it at runtime into `encoder_map`, an instance attribute
- when `orjson` use is enabled, `encoder_map` is modified during init
  - this modification explicitly checks to see if the user replaced the `json` encoder (as someone might do today)

This is actually mildly breaking because transports will no longer transparently pick up on modifications to the class var, even with `orjson` disabled.
I think that tradeoff is worth it, but we could modify this to be even more considerate.

### `requests` import deferral and import paths

Because API errors now need access to `globus_sdk.transport`, these import paths are more heavily implicated in other parts of the SDK.
This breaks tests which help enforce deferred imports of `requests`, due to its very slow import time.
To rectify, a number of `requests` imports, specifically, are now `TYPE_CHECKING` flagged and deferred.

In general, the whole SDK is trying to lazily load `requests`, so this is all directionally aligned.
It does, however, highlight the new dependency (`globus_sdk.exc` -> `globus_sdk.transport`).

Our imports are getting a little bit tangled -- lots of parts of the SDK pull in `exc` for `GlobusSDKUsageError`, but `exc.api` needs `transport`.
We should think about rearranging to clean this up. Many of the deferred import arrangements could be seen as punting on this issue.

## Diff Info

GitHub's diff looks massive because it's counting all those lines of lockfile/frozen-deps content. True diff is much more modest, but still nontrivial:
```
$ git diff main --shortstat -- src tests
 26 files changed, 599 insertions(+), 102 deletions(-)
```

### Added

-   The SDK now supports use of `orjson` as an alternative JSON encoder and decoder.
    When `GLOBUS_SDK_USE_ORJSON=1` is set, request sending and response decoding will use `orjson`.

    -   Use of `orjson` is optional, but if the variable is set and `orjson` is not installed, errors will be emitted.

    -   The setting can also be configured on transport objects with the init option, `use_orjson=True`.

    -   In a future major version of the SDK, use of `orjson` will default to true when it is available.

-   `RequestsTransport` objects are now visible via `RequestsTransport.get_current_transport()`, a staticmethod, while the transport is sending a request or being used to handle a response.
    This method raises a `LookupError` if there is no currently active transport.

Deprecated
----------

-   The `RequestsTransport` class supports configuration of request encoding via a class-variable mapping, `encoders`. This limits the ability of the SDK to apply per-object customizations, as in the case of `orjson` support. The class variable `encoders` is deprecated, and users should leverage the new `encoder_map` instance variable instead.


tests/functional/base_client/test_encodings.py
```diff
@@ -45,3 +55,86 @@ def test_text_encoding_can_send_non_ascii_utf8_bytes(client):
...
+    my_transport.encoder_map["myjson"] = MyRequestEncoder()
+    client = client_class(transport=my_transport)
+
+    responses.add(responses.POST, "https://foo.api.globus.org/bar", body="hi")
+    client.post("/bar", data={"baz": 1}, encoding="myjson")
+
+    my_transport.close()
+
+    last_req = responses.calls[-1].request
+    assert json.loads(last_req.body) == {"foo": {"baz": 1}}
```
kurtmckee: If encoders and decoders can be swapped in and out, I'd prefer that everything migrate to that system. I'm specifically hopeful that it can look similar to the methodology chosen by Flask for managing JSON encode/decode.

src/globus_sdk/transport/caller_info.py
```diff
@@ -1,9 +1,12 @@
 from __future__ import annotations
 
-from globus_sdk.authorizers import GlobusAuthorizer
+import typing as t
 
 from .retry_config import RetryConfig
 
+if t.TYPE_CHECKING:
+    from globus_sdk.authorizers import GlobusAuthorizer
```
kurtmckee, line 8: I don't understand these type annotation changes.

tests/common/response_mock.py
```diff
@@ -48,6 +48,10 @@ def json(self):
         else:
             raise ValueError("globus sdk mock value error")
 
+    @property
+    def content(self):
+        return json.dumps(self.json()).encode()
+
```
kurtmckee: I would really prefer that there be a single location from which JSON encoding and decoding is defined, and that the code use that. Here, whether orjson is available or not, the builtin JSON package is used.

kurtmckee (commented):
I'm stuck on this PR at the level of "this doesn't currently look like the right change". Some of it is attributable to me not having deep knowledge of the SDK's transport layer.

* I'm not sold on a new -- and arguably temporary -- environment variable that has to be set to use orjson even after it's installed. I argue strongly for orjson to always be the default if it's available full stop. The way that surfaces as a breaking change, as far as I can tell, is that if users are submitting e.g. `float('nan')` as JSON values, then this won't get all the way to the surfaces and cause HTTP 4XX or HTTP 5XX level failure responses -- it'll get stopped going out the door at the client side, and rightly so IMHO.

* I'm not convinced that breaking apart the encoding/decoding into separate classes is helpful. I'd strongly prefer to have a single class _a la_ Flask's implementation...though this preference is informed by deeper knowledge of Flask's implementation than I have of the SDK's transport layer.

In addition to these concerns, I kept getting tripped up by imports getting swapped around in files that weren't otherwise getting modified. That seems to be unrelated to this change. They're here now, so I'm not asking for it to get ripped out.

In short, I don't think that this is the direction to go for orjson support. I think it needs to be on by default if orjson is installed, that there should not be an environment variable to control this, and that this should be customizable via an attribute that gets set with an instance of a class that handles both encoding and decoding, as is done when setting the JSON handler for Flask apps.

It will probably be helpful to meet out-of-band to discuss this change so I can understand the technical requirements better, since I'm not well-versed in the SDK's existing mechanisms here.

src/globus_sdk/transport/caller_info.py
sirosen: Per discussion, this is part of the careful work to ensure that `requests` is appropriately deferred as an import.

sirosen:
After discussion, we've made some changes to the direction of this PR.

1. `orjson` is now opt-out, not opt-in. On-by-default is the right behavior for this 99% of the time, and we want users to easily get the perf boost from `orjson`.
2. The opt-out pathway is no longer an env var or flag. You now have to provide a substitute implementation to switch away from `orjson`. This simplifies the implementation for us, and is unlikely to cost users anything.
3. "encoders" and "decoders" are combined into `RequestsRepresentationProvider`s. These providers handle media types or other representational decisions, and are scoped specifically to `requests` objects.

tests/functional/base_client/test_encodings.py
sirosen: I hope you like the new `RequestsRepresentationProvider`!

tests/common/response_mock.py
sirosen: I think this is the only un-addressed comment from review. Note that this is testsuite mocking. i.e. It is providing data to the code which chooses between `json` and `orjson`.

I think it's valuable to keep the test data stubbing on `json`. That way we're feeding the same inputs to both `json` and `orjson`, and testing that we don't get different results. I do note that `text` is defined in `__init__` though, and will fix to replace this property with `self.content = self.text.encode()`.

sirosen: Done in 87b73ec54491f80cab32aa75c2d8feb104882b69 ; please mark this resolved if my thoughts about intentionally testing `json` against `orjson` are satisfactory.

src/globus_sdk/config/__init__.py
kurtmckee, line 1: Changes to this file appear to be remnants.

Not harmful, and not necessary to revert, but there's another file I'm going to flag so I'm flagging this as well.

src/globus_sdk/config/env_vars.py
kurtmckee, line 1: Changes in the file appear to be remnants. I recommend eliminating this change.

src/globus_sdk/transport/representation_providers.py
```diff
@@ -0,0 +1,238 @@
...
+        method: str,
+        url: str,
+        params: dict[str, t.Any] | None,
+        data: t.Any,
+        headers: dict[str, str],
+    ) -> requests.Request:
+        import requests
+
+        if data is not None:
+            headers = {"Content-Type": "application/json", **headers}
```
kurtmckee, line 171: Just confirming: the goal is for `headers` to override the `Content-Type` if it's specified in `headers`, correct?

src/globus_sdk/transport/representation_providers.py
```diff
@@ -0,0 +1,238 @@
...
+    def decode_body(self, response: requests.Response) -> t.Any:
+        if orjson_compat.ORJSON_AVAILABLE:
+            return orjson_compat.loads(response.content)
+        else:
+            return response.json()
+
+
+class RequestsHttpFormProvider(RequestsRepresentationProvider):
+    """
+    This provider prepares requestdata as a form-encoded body.
```
kurtmckee:
```suggestion
    This provider prepares request data as a form-encoded body.
```

src/globus_sdk/transport/requests.py
```diff
@@ -250,6 +327,34 @@ def request(
...
+                method,
+                url,
+                caller_info=caller_info,
+                query_params=query_params,
+                data=data,
+                headers=headers,
+                encoding=encoding,
+                allow_redirects=allow_redirects,
+                stream=stream,
+            )
```
kurtmckee: I'd prefer that `self._as_current_transport()` be a decorator for this method. The call stack will be affected, but we won't have to maintain a call stack visually, with `self._send_request()` being the only thing that this method does. 🤞 🙏

src/globus_sdk/client.py
```diff
@@ -571,7 +571,10 @@ def request(
 
         if 200 <= r.status_code < 400:
             log.debug(f"request completed with response code: {r.status_code}")
-            return GlobusHTTPResponse(r, self)
+            with self.transport._as_current_transport():
```
kurtmckee, line 574: Ah, I see that `._as_current_transport()` isn't used only in the `.request()` method. I still carry a preference for a decorator implementation so that the code in `._send_request()` can collapse back into `.request()`...thoughts?

tests/common/response_mock.py
```diff
@@ -41,6 +41,7 @@ def __init__(
         self._json_body = json_body
 
         self.text = text or (json.dumps(json_body) if json_body else "")
```
kurtmckee: Is there a way to leverage orjson (when available) for dumping at this location?

tests/unit/transport/test_transfer_transport.py
```diff
@@ -42,6 +43,7 @@ def test_transfer_does_not_retry_external():
 
     dummy_response = mock.Mock()
     dummy_response.json = lambda: body
+    dummy_response.content = json.dumps(body).encode()
```
kurtmckee: Similar feedback throughout -- I'd love for the test suite to prefer orjson where possible!

tests/conftest.py
```diff
@@ -27,7 +28,20 @@ def mocked_responses():
 
 
 @pytest.fixture
-def make_response():
+def mock_client_factory():
+    def build():
+        client = mock.Mock()
+        # because responses are deserialized using the decoder attached to the client
+        # we need to ensure it is populated if we want to be able to pass the mock
+        # clients to response classes and get good behavior
```
kurtmckee: The phrase "get good behavior" might be rephrased/expanded for future readers' benefit!

kurtmckee (changes requested):

src/globus_sdk/config/__init__.py
sirosen: Yes, good call! My intention was to fully revert the changes to `config`; but it seems I didn't quite achieve that because I did it manually. (Perhaps I should have checkout out the files from `main`).

src/globus_sdk/transport/representation_providers.py
sirosen: Yes; or at least, that's the logic we have today. I think we should strongly consider changing (I would say fixing) it, but in a separate changeset. In theory, someone may want to set `Content-Type: application/custom+json` and we should (IMO) respect that.

But since we today overwrite whatever they send, I don't want to include such a change in behavior in this PR.

src/globus_sdk/transport/representation_providers.py
sirosen: :scream:

src/globus_sdk/transport/requests.py
sirosen: Oh, very good idea; I like it! I hadn't thought of turning it into a decorator because the decorator needs to touch `self`.

I'll add a separate decorator to keep the code simple.

src/globus_sdk/client.py
sirosen: Yep, per above I like the decorator idea.
The only real question IMO is what the decorator name should be -- I'll try out a few names and see what feels natural.

tests/common/response_mock.py
sirosen: I initially was thinking that testing `json.dumps()` against `orjson.loads()` (intentionally) was a good thing. Then I tried to write up my reasoning and realized that the case for it is rather weak. I'll take a look.

tests/conftest.py
sirosen: Wait, is this even true anymore? It was strictly necessary before the `current_transport` work, but now I think this comment (and that attribute) is a relic! I'll test removal.

sirosen: Confirmed, this is safe to remove.

kurtmckee (approved):

PR #1392, opened 2026-06-05
removed pinned from v2 bookmarks

aaschaer:
We made a somewhat late decision to pull `pinned` from bookmarks v2 as it was unused in v1 since 2024.

I'm marking this no-news-is-good-news since its removing a nonexistant field, but lmk if I should actually add a changelog

### DIFF
--- a/changelog.d/20260605_135304_aaschaer_remove_pinned.rst
+++ b/changelog.d/20260605_135304_aaschaer_remove_pinned.rst
@@ -1,0 +1,4 @@
+Removed
+-------
+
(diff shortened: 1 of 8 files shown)


kurtmckee (changes requested):
I think it'll be appropriate to have a changelog fragment for this, though it sounds like it's provable that it hasn't been in use.

aaschaer:
@kurtmckee I've added a changelog fragment

kurtmckee (approved):

pjhinton-globus (approved):

PR #1396, opened 2026-06-16
Fix a crash bug in the orjson compatibility module

### LINKED ISSUE #1114: issues with globus-sdk 4.8.0
[globus-sdk 4.8.0](https://globus-sdk-python.readthedocs.io/en/stable/changelog.html) causes some issues with unit-tests on [iblrig](https://github.com/int-brain-lab/iblrig) related to their [added support for orjson](https://github.com/globus/globus-sdk-python/pull/1385):

```
FAILED iblrig/test/test_misc.py::TestPortSettings::test_port_settings - RuntimeError: 'orjson' is not available but globus-sdk was configured to use it. This is not valid. Please ensure that 'orjson' is installed.
FAILED iblrig/test/test_video.py::TestCameraSession::test_run_video_session - RuntimeError: 'orjson' is not available but globus-sdk was configured to use it. This is not valid. Please ensure that 'orjson' is installed.
```

This seems to be an upstream issue with globus-sdk. Options:

- fix upstream issue,
- add orjson dependency (might be nice to have for performance reasons), or
- pin globus-sdk to < 4.8.0

bimac:
`globus_sdk._internal.orjson_compat.__getattr__` violates Python's attribute protocol. The contract for `getattr(obj, name, default)` and `hasattr(obj, name)` is that a missing attribute raises `AttributeError`, which those builtins catch to return the default / `False`. The module breaks this: when `orjson` is not installed, it raises `RuntimeError` for *every* unknown name, because `__getattr__` calls `require()` before checking which name was requested.

As a result, any code that defensively probes the module crashes:

``` python
from globus_sdk._internal import orjson_compat

hasattr(orjson_compat, 'anything')   # should be False; instead: RuntimeError
```

This is how it surfaced for us - unrelated tests failed because `unittest.assertWarns` / `pytest.warns` scan `sys.modules` and do `getattr(mod, "__warningregistry__", None)` on every module:

``` python
import warnings
from unittest import TestCase
from globus_sdk import TransferAPIError  # loads transport -> orjson_compat

with TestCase().assertWarns(UserWarning):
    warnings.warn("boom")
```

The fix is to move the call to `require()` inside the if-clause of `__get_attr__`.

### DIFF
--- a/changelog.d/20260616_111455_kurtmckee_fix_orjson_compat_getattr.rst
+++ b/changelog.d/20260616_111455_kurtmckee_fix_orjson_compat_getattr.rst
@@ -1,0 +1,7 @@
+Fixed
+-----
+
+- Fix a crash bug in the orjson compatibility module. (:pr:`NUMBER`)
+
+  This manifested as a ``RuntimeError``, and could occur when a test framework
+  accessed the module to check for warning registries in the module.
--- a/src/globus_sdk/_internal/orjson_compat.py
+++ b/src/globus_sdk/_internal/orjson_compat.py
@@ -35,10 +35,10 @@ else:
         return ["__all__", "__file__", "__path__"] + list(__all__)
 
     def __getattr__(name: str) -> t.Any:
-        require()
-
-        mod = sys.modules[__name__]
         if name in ("loads", "dumps"):
+            require()
+
+            mod = sys.modules[__name__]
             value = getattr(orjson, name)
             setattr(mod, name, value)
             return value


kurtmckee:
Hello @bimac, thanks for reporting this. I'll take a crack at addressing this.

kurtmckee (approved):

sirosen:
We've just released this in v4.8.1 , so you should be all set! :rocket: 

Thanks not only for the report but the fix as well! (We're so used to bug-reports only, we didn't realize it was a PR at first. :sweat_smile: )

bimac:
Thanks for the quick release turnaround 🙂

PR #1399, opened 2026-06-16
Set `persist-credentials` on GHA checkout

sirosen:
As flagged by zizmor, we should at least be explicit and ideally always
set this to `false`, so that a workflow cannot push commits.

Initially fixed with

    zizmor .github/workflows/ --fix=all

(applies "unsafe" fixes)

The only additional change is to set it to `true` for the PR update job,
which does push commits.

### DIFF
--- a/.github/workflows/update_pr_references.yaml
+++ b/.github/workflows/update_pr_references.yaml
@@ -4,6 +4,10 @@ on:
     branches:
       - main
 
+permissions:
+  contents: write
+  pull-requests: read
+
 jobs:
   update_pr_numbers_in_change_fragments:
     runs-on: ubuntu-latest
--- a/.github/workflows/update_pr_references.yaml
+++ b/.github/workflows/update_pr_references.yaml
@@ -9,6 +9,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: true
       - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
           python-version: '3.x'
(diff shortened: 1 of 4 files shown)


.github/workflows/update_pr_references.yaml
```diff
@@ -9,6 +9,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: true
```
kurtmckee, line 17: In a situation like this, I'd prefer to see permissions specified as well.

I _think_ it would be something like this, but please double-check that!

```yaml
permissions:
  contents: write
```

kurtmckee (changes requested):

.github/workflows/update_pr_references.yaml
sirosen: I agree, but was intending to handle `permissions` in a separate PR.

I think `contents: write` on its own breaks the workflow, and that we need
```yaml
permissions:
  contents: write
  pull-requests: read
```

Unless I've misunderstood what the `pull-requests` permission does, which is very possible! :sweat_smile: 

I'll add this in, and worst case we fix it later.

sirosen: Confirmed (I think!) that we need `pull-requests: read`:
https://docs.github.com/en/rest/commits/commits?apiVersion=2026-03-10#list-pull-requests-associated-with-a-commit
says
> The fine-grained token must have the following permission set:
> - "Pull requests" repository permissions (read)

kurtmckee: Ooh, good catch! This gets me the outcome that I want (explicit declarations of permissions) and the outcomes that we both want (the workflow continues to function). :laughing:

kurtmckee (approved):

PR #1402, opened 2026-06-16
Enable zizmor linting on pushes and PRs

sirosen:
This is a new workflow which has a single job, calling the zizmor
reusable workflow.
It publishes findings as SARIF data to GitHub security findings APIs.

The workflow will not fail on pull requests or pushes, but will flag new
and existing usages which fail zizmor checking.

### DIFF
--- a/.github/workflows/zizmor.yaml
+++ b/.github/workflows/zizmor.yaml
@@ -1,0 +1,12 @@
+name: zizmor
+
+on:
+  pull_request:
+  push:
+
+jobs:
+  zizmor:
+    permissions:
+      security-events: write
+
+    uses: zizmorcore/workflow/.github/workflows/reusable-zizmor.yml@465600191822382e76d26abf77772f21262ecdb6  # main
--- a/changelog.d/20260616_152921_sirosen_add_zizmor.rst
+++ b/changelog.d/20260616_152921_sirosen_add_zizmor.rst
@@ -1,0 +1,5 @@
+Security
+--------
+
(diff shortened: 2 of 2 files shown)


.github/workflows/zizmor.yaml
```diff
@@ -0,0 +1,12 @@
...
+on:
+  pull_request:
+  push:
+
+jobs:
+  zizmor:
+    permissions:
+      security-events: write
+
+    uses: zizmorcore/workflow/.github/workflows/reusable-zizmor.yml@1e20adb0862e932363a4d85d68c92e5cc6fcb5d4
```
kurtmckee: This commit SHA is not the latest (the latest change isn't strictly needed, but it catches my eye).

sirosen: I copied from another repo where I have this! Whoops! I'll update.

sirosen: Fixed!

Neither `pinact` nor `gha-update` like that there are no version tags on the workflow. (It's marked as a "preview".)

kurtmckee: Completely believe they don't like the lack of version tags!

That shared workflow is a bit odd from my perspective. They have the `setup-uv` action pinned, but nothing else. It's a very interesting setup right now (IMHO).

kurtmckee: Also of note, they're not locking the uv version that's set up -- they explicitly comment that they want the latest version of uv, which surprises me.

kurtmckee (approved):

PR #1408, opened 2026-07-30
Remove flow run label length constraint documentation

kurtmckee:
It's not documented consistently across all methods and is subject to change in the service.


sirosen (approved):
I wouldn't mind such details being documented if we can extract the information from OpenAPI specs in the future (to guarantee freshness driven by the services).

PR #1410, opened 2026-08-03
Add Web Input-related methods to the `FlowsClient` class

kurtmckee:
Added
-----

- Add Web Input-related methods to the ``FlowsClient`` class.

  The new methods are: ``list_web_inputs``, ``get_web_input``,  and ``respond_to_web_input``.


src/globus_sdk/testing/data/flows/get_web_input.py
```diff
@@ -0,0 +1,162 @@
...
+        status=404,
+        json=NOT_FOUND_RESPONSE,
+    ),
+    auth_policy_required=RegisteredResponse(
+        service="flows",
+        path=f"/web_inputs/{WEB_INPUT_ID}",
+        status=403,
+        json=AUTH_POLICY_REQUIRED_RESPONSE,
+    ),
+)
```
sirosen, line 180: I don't think we're testing several of these cases that you've declared here. And likewise for some of the other responses declared in this PR.
I see the test for paginated listing, but I think that's the only non-default case? 

I *think* we should run through all of them? That's weakly held; let me know if you disagree.

kurtmckee: These test cases exist so downstream services can leverage these tests instead of writing their own.

Some of these are getting used in the CLI test suite, for example.

kurtmckee: Oh, to your point -- we could perhaps run through all of them, but that didn't strike me as necessary. They just need to exist for downstream usage.

sirosen: Okay, let's not block on this! I'll look through the rest of the review, which should be fast -- it looked good on my first pass.

There's an argument for having the SDK verify that each response parses as expected: the error ones are errors, the GAREs are detected as GAREs, and so forth. That way if a response mock is malformed (e.g., typo) we'll catch it here. But also, if we're consuming these same mock responses in CLI tests, we'd catch issues there too.

src/globus_sdk/services/flows/client.py
```diff
@@ -983,6 +984,169 @@ def list_registered_apis(
...
         )
 
+    @paging.has_paginator(
+        paging.NullableMarkerPaginator, items_key="web_input_summaries"
+    )
+    def list_web_inputs(
+        self,
+        *,
+        filter_roles: str | t.Iterable[str] | MissingType = MISSING,
+        filter_states: str | t.Iterable[str] | MissingType = MISSING,
```
sirosen: Can the roles and states be (optionally type-aliased) `Literal`s?

For example, in `TransferClient.endpoint_search`, we have:
```python
        filter_entity_type: (
            t.Literal[
                "GCP_mapped_collection",
                "GCP_guest_collection",
                "GCSv5_endpoint",
                "GCSv5_mapped_collection",
                "GCSv5_guest_collection",
            ]
            | MissingType
        ) = MISSING,
```

This allows an unexpected value at runtime (any `str`), but offers a hint for proper usage.

In `SearchClient`, the same thing is handled via a type alias (`_VALID_ROLE_NAMES_T`).
I think you'd want that here because of the `<value> | Iterable[<value>]` types.

src/globus_sdk/services/flows/client.py
```diff
@@ -983,6 +984,169 @@ def list_registered_apis(
...
+        :param filter_roles:
+            Filter web inputs to only include those the user has the given role for.
+        :param filter_states:
+            Filter web inputs to only include those in the given state(s).
+        :param filter_flow_ids:
+            Filter web inputs to only include those associated with the given flow IDs.
+        :param filter_run_ids:
+            Filter web inputs to only include those associated with the given run IDs.
+        :param orderby:
+            A criterion for ordering web inputs in the listing.
```
sirosen: I think `orderby` could use more explanation. If you look at `FlowsClient.list_flows` you'll see that we break down the supported fields and ordering options.

It would be cool to have something like `tuple[_FieldName, Literal["ASC", "DESC"]]` as a supported input type, but I don't think we have prior art for that. I think the types could get awkward to handle.

For now, I'd like to just see the extended doc.

sirosen (changes requested):
Minor doc request before we merge!

src/globus_sdk/services/flows/client.py
kurtmckee: Updated and force-pushed. [See the diff.](https://github.com/globus/globus-sdk-python/compare/be20322e1fab59a772103a7454c827f7520bde31..b40df8ad40428814de0dd6eaa07ca9be6958fbb0)

src/globus_sdk/services/flows/client.py
kurtmckee: Updated and force-pushed. [See the diff.](https://github.com/globus/globus-sdk-python/compare/be20322e1fab59a772103a7454c827f7520bde31..b40df8ad40428814de0dd6eaa07ca9be6958fbb0)

sirosen (approved):
Image

PR #1414, opened 2026-08-05
Annotate `resource_server` as `str` for some cases

sirosen:
Several classes set `scopes` as a class attribute, meaning that their `resource_server` class property will evaluate to `str`, rather than `str | None`.
_Technically_, this can be violated (if `scopes` is deleted from the class), but in terms of a closed system of type-safe code, the annotation is correct.

An annotation without an assignment does not rebind the name, so the subclasses' `resource_server` attribute still points at the classproperty descriptor.

This corrects several type-checking errors in the docs without requiring improper checks or type-ignores to provide type-safety.

### DIFF
--- a/changelog.d/20260805_101317_sirosen_improve_resource_server_annotations.rst
+++ b/changelog.d/20260805_101317_sirosen_improve_resource_server_annotations.rst
@@ -1,0 +1,6 @@
+Changed
+-------
+
+-   The ``resource_server`` attribute of the clients for Auth, Compute, Flows,
+    Groups, Search, Timers, and Transfer is now annotated as ``str``, rather than
+    the inherited type of ``str | None``. (:pr:`NUMBER`)
--- a/src/globus_sdk/services/auth/client/base_login_client.py
+++ b/src/globus_sdk/services/auth/client/base_login_client.py
@@ -46,6 +46,10 @@ class AuthLoginClient(client.BaseClient):
     error_class = AuthAPIError
     scopes = AuthScopes
 
+    # annotate but do not assign 'resource_server'
+    # because we know that the classproperty of this name will evaluate to a string
+    resource_server: str
+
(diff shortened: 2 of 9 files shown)


derek-globus (approved):
I'm surprised that this narrowing is type-safe, but our CI linting demonstrates that it is.

sirosen:
I was also surprised to see it work when I tried it! :sweat_smile: 
I expected to need type-ignore comments.

A simplified form also works:
```python
class A:
    x: int | str

class B(A):
    x: int

reveal_type(B.x)  # int!
```

It's unquestionably a soundness hole in typing, but there are lots of those, so I'm not that surprised.

<details><summary>other bad soundness hole tangent</summary>

There are worse things! For example, early on `float` was treated as meaning `float | int`. This still holds today!

```python
def f(x: float) -> None:
    pass

y: int = 0

# type-checks okay! :,(
f(y)
```

It's wrong! Many things which take floats also take ints, but not all. There are C APIs which need a `float`. :skull: 

And any protocol or superclass annotation which attempts to restrict the methods you use can be circumvented by the narrowing in an `isinstance` check. For example:
```python
class P:
    def plus1(self, x: int) -> int:
        return x + 1

class C(P):
    def __init__(self) -> None:
        self.counter = 0

    def incr(self, x: int) -> int:
        self.counter += x
        return self.counter

def only_use_mutation_free_methods(obj: P, x: int) -> int:
    ...  # 70 lines of code go here
    if isinstance(obj, C):
        y = obj.incr(x)
    else:
        y = obj.plus1(x)
    ...  # 30 more lines of code
    return y
```

That means that even if you annotate it `Iterable`, I can make it `list`, and if it's `Mapping` (rather than `MutableMapping`), I can make it `dict`. :skull: 

But it still works well enough to be useful! Mostly.

</details>

PR #1415, opened 2026-08-06
Introduce `GlobusApp.logout(sweep=True)`

sirosen:
This new flag codifies how to clear all tokens for services currently
registered with the app (typically via clients).

### DIFF
--- a/src/globus_sdk/globus_app/app.py
+++ b/src/globus_sdk/globus_app/app.py
@@ -25,6 +25,7 @@ from globus_sdk.scopes import AuthScopes, Scope, ScopeParser
 from globus_sdk.token_storage import (
     ScopeRequirementsValidator,
     TokenStorage,
+    TokenStorageData,
     TokenValidationError,
     ValidatingTokenStorage,
 )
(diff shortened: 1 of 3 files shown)


src/globus_sdk/globus_app/app.py
```diff
@@ -385,21 +386,37 @@ def login_required(self) -> bool:
                 return True
         return False
 
-    def logout(self) -> None:
+    # TODO: sweep should default to True in globus-sdk v5+
```
derek-globus: This isn't the right way to document that intention imo.

There's an epic [here](https://app.shortcut.com/globus/epic/40780?group_by=none&hide_subtasks=false&ct_workflow=all&cf_workflow=500015354) where we're accumulating v5 plans.

---

As I mentioned in slack, I separately don't actually think we should do this. But if you want to discuss the idea, could you create a story there instead of a TODO here?

derek-globus (changes requested):

src/globus_sdk/globus_app/app.py
sirosen: Sorry, yes, that comment was written contemporaneously with our conversation. Let me remove it.

I'm not going to create other tracking, since I think it needs some amount of further discussion first.

derek-globus (approved):

PR #1416, opened 2026-08-08
Update the create-and-run-flow example scripts

sirosen:
The create-and-run-flow example was written under globus-sdk v3 , and used tokenstorage to handle logins. They have drifted out of date.
The minimal fix would be to correct the import paths used, but these changes refit the scripts more dramatically to use `GlobusApp`.

Not only are the scripts shorter and more to the point, with modernized usage, they are also fully type annotated, such that `tox r -e mypy-docs` passes on this part of the docs tree.

-----

@derek-globus, for your particular interest, note the use of `logout(sweep=True)` in these examples. :grin:

### DIFF
--- a/docs/examples/create_and_run_flow/manage_flow.py
+++ b/docs/examples/create_and_run_flow/manage_flow.py
@@ -8,14 +8,8 @@ import globus_sdk
 CLIENT_ID = "61338d24-54d5-408f-a10d-66c06b59f6d2"
 
 
-def get_flows_client(app: globus_sdk.GlobusApp) -> globus_sdk.FlowsClient:
-    return globus_sdk.FlowsClient(
-        app=app, app_scopes=[globus_sdk.FlowsClient.scopes.manage_flows]
-    )
-
-
 def create_flow(app: globus_sdk.GlobusApp, args: argparse.Namespace) -> None:
-    with get_flows_client(app) as flows_client:
+    with globus_sdk.FlowsClient(app=app) as flows_client:
         print(
             flows_client.create_flow(
                 title=args.title,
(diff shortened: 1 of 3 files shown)


docs/examples/create_and_run_flow/manage_flow.py
```diff
@@ -1,117 +1,59 @@
...
-                    "DoIt": {
-                        "Type": "Action",
-                        "ActionUrl": "https://actions.globus.org/hello_world",
-                        "Parameters": {
-                            "echo_string": "Hello, Asynchronous World!",
-                        },
-                        "End": True,
-                    }
+def create_flow(app: globus_sdk.GlobusApp, args: argparse.Namespace) -> None:
+    with get_flows_client(app) as flows_client:
```
derek-globus: What do you think of removing the `get_flows_client` call entirely?

There's not too much benefit in scoping down the request space from `flows:all` -> `flows:manage_flows`, they're both very expansive in terms of what they allow.

At that point, `with globus_sdk.FlowsClient(app=app)` is  basically the same length as `with get_flows_client(app)` without hiding the call behind a function.

derek-globus (approved):
One minor suggestion but otherwise looks good; good improvement to use GlobusApp & the use of `sweep` makes sense here.

docs/examples/create_and_run_flow/manage_flow.py
sirosen: Oh, is our default `all`?

I tried it with the call inlined, but then I found that it was harder to point out "here's where and how we construct the client, with configured scopes". If we don't need to configure scopes I'm in favor of inlining it. Let me check/confirm and apply.

derek-globus: > Oh, is our default all?

[Yep](https://github.com/globus/globus-sdk-python/blob/a3aed17abf8870417578b8602224a084fbb3667a/src/globus_sdk/services/flows/client.py#L53)

derek-globus (approved):

PR #1417, opened 2026-08-08
Update the auth management example scripts

sirosen:
These were written under globus-sdk v3 and had outdated imports.

Updating to use GlobusApp also allowed a dramatic simplification, as this example explained automatic login redriving without the use of `auto_redrive_gares` (because such a feature did not exist at time of writing).
Using GlobusApp for the login requirements of these scripts therefore allows us to reduce the example from 3 scripts to only 2.


docs/examples/auth_manage_projects/manage_projects.py
```diff
@@ -1,144 +1,72 @@
...
-
-def get_auth_client():
-    tokens = get_tokens()
+def get_auth_client(app: globus_sdk.GlobusApp) -> globus_sdk.AuthClient:
     return globus_sdk.AuthClient(
-        authorizer=globus_sdk.AccessTokenAuthorizer(tokens["access_token"])
+        app_scopes=[
+            globus_sdk.AuthClient.scopes.manage_projects,
+            globus_sdk.AuthClient.scopes.openid,
+            globus_sdk.AuthClient.scopes.email,
```
derek-globus, line 15: nit: you don't need `openid` when you're using a `GlobusApp`, that gets included in every auth scope requirement list.

derek-globus (approved):
One nit

docs/examples/auth_manage_projects/manage_projects.py
sirosen: True, but I think I prefer that we keep it?
We actually use the `openid` scope capabilities in this script to access `userinfo`. So if we drop it from this list, then we're relying on an implementation detail of `GlobusApp`. (One which will probably never change, but still.)

derek-globus: Yeah that's fine 👍

PR #1421, opened 2026-08-14
Update DynamoDB TokenStorage example + mypy config

sirosen:
In order to make the DynamoDB token storage example pass mypy linting and
use the current paradigm for TokenStorage, it needs:

- missing annotations (easy)
- retrofitting of existing functions (easy)
- mypy overrides to ignore `boto3` types (easy, but note below)
- an implementation of `get_token_data_by_resource_server` (see note!)

For mypy, I have decided to move to a dedicated `mypy.ini` config in the
`docs/` directory. With the config growing, it didn't make sense to keep
it embedded in the repo root pyproject.toml . That also lets us set
different top-level behavioral flags in the config.

For `get_token_data_by_resource_server`, I decided to implement it, in
the simplest possible way using a sequential scan, but no pagination.
It also prints a very annoying warning (printing, not emitting a warning,
so that warning filters are not applicable), to make sure that anyone
who picks it up and runs it verbatim gets some signal that they ought to
think about it and possibly make some changes.
The associated doc notes that you _could_ make this method raise an error
and it would mostly work, but that some SDK features would fail if you
do so.

### DIFF
--- a/docs/examples/token_storage/dynamodb_token_storage.py
+++ b/docs/examples/token_storage/dynamodb_token_storage.py
@@ -7,7 +7,7 @@ import typing as t
 import boto3
 
 import globus_sdk
-from globus_sdk.token_storage import StorageAdapter
+from globus_sdk.token_storage import TokenStorage, TokenStorageData
 
 CLIENT_ID = "61338d24-54d5-408f-a10d-66c06b59f6d2"
 tablename = "example-globus-tokenstorage"
@@ -19,11 +19,13 @@ parser.add_argument(
 )
 
 boto_client = boto3.client("dynamodb")
-auth_client = globus_sdk.NativeAppAuthClient(CLIENT_ID)
 
 
-class DynamoDBStorageAdapter(StorageAdapter):
-    def __init__(self, client, tablename: str, namespace: str = "DEFAULT") -> None:
+class DynamoDBStorageAdapter(TokenStorage):
+
+    def __init__(
+        self, client: t.Any, tablename: str, namespace: str = "DEFAULT"
+    ) -> None:
         """
         :param client: A boto3 DyanmoDB client to use
         :param tablename: The name of the dynamodb table to use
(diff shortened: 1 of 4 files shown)


docs/examples/token_storage/dynamodb_token_storage.py
```diff
@@ -19,11 +19,13 @@
 )
 
 boto_client = boto3.client("dynamodb")
-auth_client = globus_sdk.NativeAppAuthClient(CLIENT_ID)
 
 
-class DynamoDBStorageAdapter(StorageAdapter):
-    def __init__(self, client, tablename: str, namespace: str = "DEFAULT") -> None:
+class DynamoDBStorageAdapter(TokenStorage):
```
derek-globus:
```suggestion
class DynamoDBTokenStorage(TokenStorage):
```

The term `Adapter` is associated with the legacy version of `StorageAdapters`.

All the new token storages use `TokenStorage` as their suffix now (`MemoryTokenStorage`, `SQLiteTokenStorage`, `JSONTokenStorage`, etc.) so our example extension doc should too.

docs/examples/token_storage/index.rst
```diff
@@ -9,12 +9,23 @@ DynamoDB Token Storage
...
 DynamoDB as the backend storage mechanism.
 
-Unlike the builtin adapters for JSON and sqlite, there is no capability here
-for an enumeration of all of the tokens in storage. This is because DynamoDB
-functions as a key-value store, and can efficiently map keys, but features slow
-sequential scans for enumeration. The example therefore demonstrates that
-key-value stores with limited or no capabilities for table scans can be used to
-implement the token storage interface.
+Unlike the builtin adapters for JSON and sqlite, enumerating tokens in storage
+is not a desirable operation.
```
derek-globus:
```suggestion
Unlike the builtin ones, enumerating tokens in a DynamoDB table-backed storage
is not desirable.
```

Mostly a recommendation to avoid the word "adapter" here & in the rest of the document.

derek-globus (changes requested):
My only real note is around the term `Adapter`.

The old interface was called `StorageAdapater`, the new one is called `TokenStorage`. In my opinion, child classes shouldn't be referred to as "adapaters" in class name or text references to avoid confusion between the two versions.

docs/examples/token_storage/dynamodb_token_storage.py
sirosen: Oh! I missed this in my rewrites! Yes, let's update the name.

docs/examples/token_storage/index.rst
sirosen: I made a slight alteration in this when I applied, but spiritually it's the same. LMK if you want it tweaked more. :+1:

derek-globus (approved):

PR #1423, opened 2026-08-31
Fix a `RecursionError` that occurs with deeply-nested request bodies

kurtmckee:
> [!NOTE]
>
> This issue was discovered while testing a service release.
>
> The first commit, which introduces a test that demonstrates the problem, takes over 7 seconds locally to eventually throw `RecursionError`.
>
> The second commit, which eliminates recursion entirely, makes the test succeed in 0.04 seconds.

Fixed
-----

- Fix a ``RecursionError`` that occurs with deeply-nested request bodies.

### DIFF
--- a/src/globus_sdk/transport/representation_providers.py
+++ b/src/globus_sdk/transport/representation_providers.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import enum
 import typing as t
 import uuid
+from collections import deque
 
 from globus_sdk._internal import orjson_compat
 from globus_sdk._missing import MISSING, filter_missing
@@ -106,17 +107,65 @@ class RequestsRepresentationProvider:
         """
         Prepare the data (body) for a request.
 
-        If the body is a dict, list, or tuple, it will be recursively processed to
-        filter out MISSING and format primitives.
-
-        Otherwise, it is returned as-is.
+        MISSING values are filtered out of dicts, lists, and tuples.
+        Scalar elements, like UUIDs and enums, and converted to strings.
         """
-        if isinstance(data, dict):
-            return filter_missing({k: self._prepare_data(v) for k, v in data.items()})
-        elif isinstance(data, (list, tuple)):
-            return [self._prepare_data(x) for x in data if x is not MISSING]
-        else:
+
+        if not isinstance(data, (dict, list, tuple)):
             return self._format_primitive(data)
+
+        # Strategy: Rely on object mutability to iteratively prepare data
+        # after it has already been appended to a list / assigned in a dict.
+        data = self._create_mutable_copy(data)
+        unprocessed_data: deque[dict[str, t.Any] | list[t.Any]] = deque((data,))
+
+        while unprocessed_data:
+            current_data = unprocessed_data.popleft()
+
+            if isinstance(current_data, dict):
+                prepared_dict: dict[str, t.Any] = {
+                    k: self._prepare_element(v, unprocessed_data)
+                    for k, v in current_data.items()
+                    if v is not MISSING
+                }
+                current_data.clear()
+                current_data.update(prepared_dict)
+
+            else:  # isinstance(current_data, list)
+                prepared_list: list[t.Any] = [
+                    self._prepare_element(v, unprocessed_data)
+                    for v in current_data
+                    if v is not MISSING
+                ]
+                current_data.clear()
+                current_data.extend(prepared_list)
+
+        return data
+
+    def _prepare_element(
+        self, element: t.Any, unprocessed_data: deque[dict[str, t.Any] | list[t.Any]]
+    ) -> t.Any:
+        if not isinstance(element, (dict, list, tuple)):
+            return self._format_primitive(element)
+        element = self._create_mutable_copy(element)
+        unprocessed_data.append(element)
+        return element
+
+    @staticmethod
+    def _create_mutable_copy(
+        data: dict[str, t.Any] | list[t.Any] | tuple[t.Any, ...],
+    ) -> dict[str, t.Any] | list[t.Any]:
+        """
+        Make a mutable copy of a non-scalar value.
+
+        Tuples will be converted to lists.
+        """
+
+        if isinstance(data, tuple):
+            return list(data)
+        if isinstance(data, list):
+            return data.copy()
+        return data.copy()
 
 
 class RequestsPlainTextProvider(RequestsRepresentationProvider):
--- a/changelog.d/20260831_155548_kurtmckee_fix_recursion_errors.rst
+++ b/changelog.d/20260831_155548_kurtmckee_fix_recursion_errors.rst
@@ -1,0 +1,4 @@
+Fixed
+-----
+
+- Fix a ``RecursionError`` that occurs with deeply-nested request bodies. (:pr:`NUMBER`)
--- a/tests/unit/transport/test_representation_providers.py
+++ b/tests/unit/transport/test_representation_providers.py
@@ -1,0 +1,32 @@
+from globus_sdk import MISSING
+from globus_sdk.transport.representation_providers import RequestsRepresentationProvider
+
+
+def test_deeply_nested_body():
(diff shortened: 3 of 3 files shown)


src/globus_sdk/transport/representation_providers.py
```diff
@@ -106,18 +106,70 @@ def _prepare_data(self, data: t.Any) -> t.Any:
...
-        elif isinstance(data, (list, tuple)):
-            return [self._prepare_data(x) for x in data if x is not MISSING]
-        else:
+
+        if not isinstance(data, (dict, list, tuple)):
             return self._format_primitive(data)
 
+        data = self._create_mutable_copy(data)
+
+        unprocessed: list[dict[str, t.Any] | list[t.Any]] = [data]
```
sirosen: Because this is solely used with `pop(0)`, let's switch to a `deque` with `popleft()`.

When I benchmarked this, it shows no improvement for narrow, deep objects. But it becomes pronounced on wide objects, showing as much as a 10x perf improvement over an object 5 levels deep with 10 sub-dicts per level (read: 100,000 nested dicts, all told).

src/globus_sdk/transport/representation_providers.py
```diff
@@ -106,18 +106,70 @@ def _prepare_data(self, data: t.Any) -> t.Any:
...
+        Make a mutable copy of a non-scalar value.
+
+        Tuples will be converted to lists.
+        """
+
+        if isinstance(data, tuple):
+            return list(data)
+        if isinstance(data, list):
+            return data.copy()
+        return data.copy()
```
sirosen, line 168: The use of this method results in us very often copying data which is _immediately_ rebuilt into a fresh list or dict. I'm trying to find a way to avoid the excess copying because it seems like a large amount of extra work.

It's tricky because we want to control and scope the depth of copying we do here. It's _almost_ a deepcopy, but not quite.

sirosen: I've been trying to collapse this copying together with the other copying and still have not succeeded. I like this formulation a little better, but still find it unsatisfactory:
```python
    def _prepare_data(self, data: t.Any) -> t.Any:
        """
        Prepare the data (body) for a request.

        If the body is a dict, list, or tuple, it will be processed to
        filter out MISSING and format primitives.

        Otherwise, it is returned as-is.
        """

        if not isinstance(data, (dict, list, tuple)):
            return self._format_primitive(data)

        if isinstance(data, (list, tuple)):
            data = list(data)
        elif isinstance(data, dict):
            data = data.copy()

        unprocessed: deque[dict[str, t.Any] | list[t.Any]] = deque((data,))

        def _handle_value(v: t.Any) -> t.Any:
            if not isinstance(v, (dict, list, tuple)):
                return self._format_primitive(v)
            elif isinstance(v, dict):
                v = v.copy()
                unprocessed.append(v)
                return v
            else:
                v = list(v)
                unprocessed.append(v)
                return v

        while unprocessed:
            current_data = unprocessed.popleft()

            if isinstance(current_data, dict):
                updated_dict: dict[str, t.Any] = {
                    k: _handle_value(v)
                    for k, v in current_data.items()
                    if v is not MISSING
                }
                current_data.clear()
                current_data.update(updated_dict)

            if isinstance(current_data, list):
                updated_list: list[t.Any] = [
                    _handle_value(v) for v in current_data if v is not MISSING
                ]
                current_data.clear()
                current_data.extend(updated_list)

        return data
```

I found that trying to improve upon this further resulted in noticeably worse performance.
For example:
<details><summary>worse version which reduces the amount of copying, using a "setitem" queue</summary>

```python
    def _prepare_data(self, data: t.Any) -> t.Any:
        """
        Prepare the data (body) for a request.

        If the body is a dict, list, or tuple, it will be processed to
        filter out MISSING and format primitives.

        Otherwise, it is returned as-is.
        """

        if not isinstance(data, (dict, list, tuple)):
            return self._format_primitive(data)

        # wrap the top-level item in a list in order to make everything look uniform
        data_wrapper: list[t.Any] = [data]

        unprocessed: deque[
            tuple[
                dict[str, t.Any] | list[t.Any] | tuple[t.Any, ...],
                int | str,
                dict[str, t.Any] | list[t.Any] | tuple[t.Any, ...],
            ]
        ] = deque([(data_wrapper, 0, data)])

        unprocessed_items_in_current_scope: list[tuple[int | str, t.Any]] = []

        def _handle_value(key: int | str, v: t.Any) -> t.Any:
            if not isinstance(v, (dict, list, tuple)):
                return self._format_primitive(v)
            elif isinstance(v, dict):
                unprocessed_items_in_current_scope.append((key, v))
                return v
            else:
                unprocessed_items_in_current_scope.append((key, v))
                return v

        while unprocessed:
            container, container_key, current_data = unprocessed.popleft()

            if isinstance(current_data, dict):
                updated_dict: dict[str, t.Any] = {
                    k: _handle_value(k, v)
                    for k, v in current_data.items()
                    if v is not MISSING
                }
                container[container_key] = updated_dict

                for key, value in unprocessed_items_in_current_scope:
                    unprocessed.append((updated_dict, key, value))

            if isinstance(current_data, (list, tuple)):
                # in the list case, this has to be two-pass to handle missing record
                # removals
                updated_list: list[t.Any] = [
                    v for v in current_data if v is not MISSING
                ]
                updated_list = [
                    _handle_value(i, v) for (i, v) in enumerate(updated_list)
                ]
                container[container_key] = updated_list

                for key, value in unprocessed_items_in_current_scope:
                    unprocessed.append((updated_list, key, value))

            unprocessed_items_in_current_scope.clear()

        return data_wrapper[0]
```

</details>

For now, I'm content to acknowledge the copying as necessary. I do not see a way to avoid it which doesn't go back to iteratively building containers, which is a surprisingly big perf hit.

src/globus_sdk/transport/representation_providers.py
```diff
@@ -106,18 +106,70 @@ def _prepare_data(self, data: t.Any) -> t.Any:
...
+                updated_list: list[t.Any] = []
+                for v in current_data:
+                    if v is MISSING:
+                        continue
+                    if not isinstance(v, (dict, list, tuple)):
+                        updated_list.append(self._format_primitive(v))
+                    else:
+                        v = self._create_mutable_copy(v)
+                        updated_list.append(v)
+                        unprocessed.append(v)
```
sirosen: I think that iteratively building the `updated_dict` and `updated_list` leaves both performance and readability on the table. Formulating these as comprehensions makes it clearer what the scope of the synthesis is (a full reconsititution of `current_data`), and performs better.

It's slightly unfortunate that `unprocessed.append()` needs to be buried into a helper to make it work, but the result is still clearer overall.

sirosen (changes requested):
Because this is exchanging a very small and simple bit of recursive code for a much larger body of work to "unroll the loop", I think it's appropriate that we make sure we're reshaping this into something which is
1. Maximally readable
2. Maximally performant

We're reducing readability (net-net), so we should put in the effort to get as much benefit from that trade as possible.

I opened #1424 with a benchmark that I used to validate my suggestions. I encourage its use here -- also feel free to merge or cherry-pick it into here and make further changes! -- to verify these changes:
- use a `deque`
- use comprehensions rather than iteratively building dicts and lists (even though this requires the addition of a new helper function)

When I measured on Python 3.12 and 3.14, I found rather dramatic improvements, especially in the worst-case type of testing.
I only explore a very small fraction of the possible space with the proposed microbenchmark, so it may be worth our time to enhance it further?

src/globus_sdk/transport/representation_providers.py
kurtmckee: Updated and force-pushed. Good suggestion to use a deque!

src/globus_sdk/transport/representation_providers.py
kurtmckee: I've updated along these lines and force-pushed.

sirosen (approved):

PR #1424, opened 2026-09-01
Add a benchmark for filter_missing

sirosen:
Add a benchmark for filtering out MISSING values from a wide and deep request body.

> [!NOTE]
> This is a benchmark developed during review of #1423, and is not safe to add to `main` until that fix merges.

@kurtmckee, I'm especially interested in what you think of this benchmark as the author of the original fix PR. I'm using a cherry-pick of this work to evaluate potential improvements.
(I think some measure of perf testing is important to validate my assumptions. I had some initial feedback that I scrapped once I actually measured the effects.)

EDIT: I added a small second benchmark which grabs testing registered response data as "typical" JSON data. This doesn't demonstrate any perf wins or losses in my testing across branches, but serves as a good "do no harm" signal.

### DIFF
--- a/tests/benchmark/test_filter_missing.py
+++ b/tests/benchmark/test_filter_missing.py
@@ -1,0 +1,38 @@
+import copy
+
+import pytest
+
+from globus_sdk.transport.representation_providers import RequestsRepresentationProvider
+
+
+@pytest.mark.parametrize(
+    ("width", "depth"),
+    (
+        # note that the expected number of objects is roughly "WIDTH raised to DEPTH"
+        # so relatively modest numbers like 5x10 = 5 to the 10th = 9.7 million
+        # be careful to build wide-but-shallow or deep-but-narrow trees
+        pytest.param(1, 1, id="1w-1d"),
+        pytest.param(5, 1, id="5w-1d"),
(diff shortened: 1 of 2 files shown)


tests/benchmark/test_filter_missing.py
```diff
@@ -0,0 +1,63 @@
...
+        pytest.param(1, 1000, id="1w-1000d"),
+        pytest.param(5, 5, id="5w-5d"),
+        pytest.param(10, 5, id="10w-5d"),
+    ),
+)
+def test_deeply_nested_object_encoding(benchmark, depth, width):
+    data = {}
+    for _ in range(depth):
+        data = {
+            # significant optimizaion: if we're building something really deep with
```
kurtmckee:
```suggestion
            # significant optimization: if we're building something really deep with
```

kurtmckee (changes requested):

tests/benchmark/test_filter_missing.py
sirosen: Rebased after application, and turned into a fixup. :+1:

kurtmckee (approved):

PR #1426, opened 2026-09-04
Add codspeed benchmarking in a dedicated CI job

sirosen:
Also add a new tox config to support running codspeed locally and in CI.


sirosen:
I think the codspeed setup somehow doesn't work cross-fork. Moving to an in-repo branch to see if I can get it up and running.

PR #1427, opened 2026-09-04
Add codspeed benchmarking in a dedicated CI job

sirosen:
Also add a new tox config to support running codspeed locally and in CI.

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -1,0 +1,45 @@
+name: CodSpeed Benchmarks
+
+on:
+  push:
+    branches:
+      - "main"
+  pull_request:
+  # `workflow_dispatch` allows CodSpeed to trigger backtest
+  # performance analysis in order to generate initial data.
+  workflow_dispatch:
+
+jobs:
+  benchmark:
+    name: run benchmarks
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      # required for OIDC authentication with CodSpeed
+      id-token: write
+
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
+
+      # Only run codspeed benchmarking on the latest stable CPython version --
+      # not on a matrix of Pythons -- to maximize value while minimizing our
+      # resource usage.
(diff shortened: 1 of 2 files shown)


sirosen:
I think this will not run cleanly until merged, because codspeed is looking for baseline data from `main`. I can do some hacky workaround stuff, but will first raise the possibility, with the team, of merging as-is so that I can simply trigger on `main`.

.github/workflows/codspeed.yml
```diff
@@ -0,0 +1,45 @@
+name: CodSpeed Benchmarks
```
derek-globus, line 1: Does this just generate an artifact that we can inspect to more easily identify where a performance impact occurred? Or will this do something more aggressive like blocking PRs that hurt performance against prior benchmarks?

derek-globus (commented):
First blush with CodSpeed, I'd like to understand better how you expect this tool should fit into the sdk ecosystem.

sirosen:
FWIW, I'm also new to codspeed. I'd like us to explore it, as a harness for running and tracking benchmark history, and improving our org-wide discipline. If it doesn't work out, we can remove it.

.github/workflows/codspeed.yml
sirosen: It doesn't generate profiles or traces, but it does some sophisticated emulation work to produce stable benchmark results, which it then tracks in a project history.

It can comment on PRs which have significant performance impacts. And it defaults to blocking PRs which show a regression but I have explicitly turned that off for us. Here's how I configured it for now:

image

So if we have a perf regression it will tell us.

---

For a sample of what the PR comment looks like, I was experimenting with this in `slyp` over the weekend a little:
https://github.com/sirosen/slyp/pull/34

sirosen:
I fixed the CI failure from codspeed; I was previously wrong about the cause. The `tox` env needs `passenv = CODSPEED_*` so that codspeed can pass info via those environment variables.

.github/workflows/codspeed.yml
derek-globus: Okay, thanks; that's helpful

derek-globus (approved):
I'm onboard to give it a try. If it becomes onerous, I may advocate that we rethink our usage, but I can see the benefit to catching performance impacts before they make it to source code.

sirosen:
FWIW, I see the value more as giving us a better way of talking about performance impacts. If I want to refactor something to be 3x larger but 10x faster, I should be able to prove that to my reviewers with a new benchmark. And we want to encode the idea that "this code is longer and more complex than other solutions because it's faster" into our testing process.

PR #1428, opened 2026-09-08
Speed up scope parsing, especially for deep scopes

sirosen:
There are three distinct optimizations here, applied in series:

- **Make scope parsing faster on deep scope strings**
- **Remove unused scope graph 'nodes' set**
- **Prefer 'reversed' over reversing slice**

By way of explanation, the first optimization, which does some graph-coloring with an enum to search for back-edges.
This replaces the prior implementation, which made sets of "all ancestors" for each node, which results in a large number of new sets being built, especially for deeper scopes.

The micro-optimizations are very minor cleanup (the commits should be self-explanatory).

### DIFF
--- a/src/globus_sdk/scopes/_graph_parser.py
+++ b/src/globus_sdk/scopes/_graph_parser.py
@@ -270,12 +270,6 @@ class _CycleDetector:
         # The use of `iter()` copies the set of edges into a consumable iterator.
         self.visit_stack.append((node, iter(self.graph.adjacency_matrix[node])))
 
-    def __getitem__(self, key: str) -> _DetectorStates:
-        return self.node_states.get(key, _DetectorStates.UNVISITED)
-
-    def __setitem__(self, key: str, value: _DetectorStates) -> None:
-        self.node_states[key] = value
-
     def run(self) -> None:
         """
         Check the graph for cycles, and if one is detected immediately error.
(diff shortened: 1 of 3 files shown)


src/globus_sdk/scopes/_graph_parser.py
```diff
@@ -276,3 +243,84 @@ def _peek_enumerate(data: str | list[str]) -> t.Iterator[tuple[int, str, str | N
...
+
+    def _stack_push(self, node: str) -> None:
+        # The use of `iter()` copies the set of edges into a consumable iterator.
+        self.visit_stack.append((node, iter(self.graph.adjacency_matrix[node])))
+
+    def __getitem__(self, key: str) -> _DetectorStates:
+        return self.node_states.get(key, _DetectorStates.UNVISITED)
+
+    def __setitem__(self, key: str, value: _DetectorStates) -> None:
+        self.node_states[key] = value
```
derek-globus: I don't think we're buying anything by overriding getter & setters on this class. The interface is "make a detector and run, potentially raising an error", not be "a data object where callers can access node states". Using the get/set interface feels like you're making this class into something that it's not (and doesn't need to be).

I'd suggest pulling these out of the external interface, removing the setter (`self.node_states[key] = value` is pretty straightforward for call sites to write) and renaming the getter to something internal if you want to preserve the default (`def _get_node_state(key: str)`)

derek-globus (changes requested):
One note on interface design, otherwise this looks correct to me.

src/globus_sdk/scopes/_graph_parser.py
sirosen: Yeah, I can do that. I added these while futzing around to try to make it all readable. Based on your feedback, I'll just stick with `dict.get` usage.

derek-globus (approved):

PR #1432, opened 2026-09-14
Update to pass under the latest flake8-bugbear

sirosen:
Primarily, flake8-bugbear flags several of our exceptions as being
unsafe to stringify, copy, or pickle because they don't pass args to
`super()` in the mandated way. (This is the new B042 rule.)

In order to pass, most of these cases are simple to update to define and
test explicit `str()` and `copy.copy()` calls. Copy is a simpler
operation than pickling and exercises the same reducer. For the target
error types, defining custom `__reduce__` methods is sufficient.

For `GlobusAPIError`, because it's so key to SDK interfaces, no action
is taken for now, other than the addition of a TODO comment.

New unit tests exercise str and copy.copy on all of the updated errors.


sirosen:
I believe the codspeed "improvement" is not real, and is an artifact of some other change (e.g., new machine image from GitHub). I can't see any way that making this change would improve performance of that part of the code.

My current working theory is that changing the imports in the exception modules is enough to rejigger the layout of some data in memory and this benchmark is so hypersensitive that it detects the difference. That might point at a potential improvement to benchmarks for the future, but I'm not sure what exactly we should be doing to make this more stable.

tests/unit/errors/test_error_copy.py
```diff
@@ -0,0 +1,104 @@
+"""
+Several errors define `__reduce__` to ensure that they are pickleable.
+
+These tests ensure that we can copy without errors and that the results compare equal
+under some definition of "equal" (which may be specific top the error type).
```
kurtmckee:
```suggestion
under some definition of "equal" (which may be specific to the error type).
```

tests/unit/errors/test_error_copy.py
```diff
@@ -0,0 +1,104 @@
...
+    ExpiredTokenError,
+    IdentityMismatchError,
+    MissingTokenError,
+    UnmetScopeRequirementsError,
+)
+
+
+def test_copy_of_network_error():
+    err = NetworkError("bad cxn", Exception("kaboom"))
+    err_copy = copy.copy(err)
```
kurtmckee, line 22: I don't recall from above if `NetworkError` implemented `.__reduce__()`. Should this check for pickle-ability?

kurtmckee (changes requested):

tests/unit/errors/test_error_copy.py
sirosen: `copy` also uses `__reduce__`, which is not totally clear in the docs. There's a note about `copyreg` for objects whose types are not classes, but that's about it. But you can verify it experimentally:
```
>>> class A:
...     def __reduce__(self):
...         print('hi')
...         return (A, ())
...
>>> x = A()
>>> import copy
>>> copy.copy(x)
hi
<__main__.A object at 0x763dc252fed0>
```

We could fully pickle-and-unpickle these objects, but I think it makes the tests more expensive for little benefit.

kurtmckee: I didn't know that! :brain:

kurtmckee (approved):

tests/unit/errors/test_error_copy.py
sirosen: Neither did I! I learned it while looking into this issue -- B042 documentation in flake8-bugbear mentions `copy`!

PR #1433, opened 2026-09-14
Update pre-commit hook versions

sirosen:
Of primary interest, the latest version of `slyp` should be significantly faster.


sirosen:
@codspeedbot explain why performance improved

I'm sure I will live to regret this, but I'm curious what it will do. I will hide the comments it puts here.

kurtmckee (approved):

PR #1434, opened 2026-09-17
Make scope parsing a little bit faster + add "normal case" benchmark

sirosen:
I noticed that we took a string slice (`data[1:]`) in a place where it was avoidable.
A new benchmark showed me in local runs that avoiding the slice could save a little bit of time in realistic cases, although it's not _much_. I'm curious if codspeed benchmarks will show a clearer win for this change.

- **Add a benchmark for "ordinary" scope parsing**
- **Avoid unnecessary str slice (copy) in scope parse**


src/globus_sdk/scopes/_graph_parser.py
```diff
@@ -237,8 +237,10 @@ def _peek_enumerate(data: str | list[str]) -> t.Iterator[tuple[int, str, str | N
     if not data:
         return
 
-    prev: str = data[0]
-    for idx, c in enumerate(data[1:]):
+    iterator = enumerate(data, start=-1)
+    prev: str
+    _, prev = next(iterator)
+    for idx, c in iterator:
```
kurtmckee, line 243: For this kind of micro-optimization, I recommend an explanation like:


```suggestion
    # Equivalent to `for idx, c in enumerate(data[1:])`, but slightly faster.
    iterator = enumerate(data, start=-1)
    prev: str
    _, prev = next(iterator)
    for idx, c in iterator:
```

Also, is it necessary to state that `prev` is a string? Its type seems clear, just as `c` is clear in the `for` loop.

sirosen:
I agree pretty well with both bits of feedback. A comment does seem worthwhile, and maybe we don't need to write down the type of the string we get from iteration.

However, I'm self-closing because I think the CI benchmark run shows that I was not accurately measuring things when I tried this locally. I saw a speedup that looked like it was in the 5-10% range on some benchmarks, and it just doesn't reproduce.

Kurt and I had a brief conversation about this and I plan to pull out separate, dedicated benchmark changes. Not only do I like my new benchmark, but I also think the one which shows change here is too fragile.

I still feel instinctively like this bit of parsing code could be made faster, but short of busting out `re` (which I really don't want for this), I'm not finding it.

